PR #33, opened 2022-11-13
Adding qryn.dev to destinations

lmangani:
Thanks for this Great Tool! 

I'm adding our project [qryn](https://github.com/metrico/qryn) to the supported destinations as it can handle the same APIs and protocols as Grafana Cloud does. We'll sure add Odigos to our use-case examples next.


edeNFed:
Thanks @lmangani looking forward for this 😄

lmangani:
The [Grafana Cloud](https://github.com/keyval-dev/odigos/blob/main/ui/vendors/grafana.tsx) vendor file works out of the box, the only exception being we have a single URL rather than multiple ones. Shall we clone the Grafana vendor into a qryn vendor even if pretty much identical, or shall we promote using the same for simplicity?

lmangani:
> Thanks @lmangani looking forward for this 😄

Here's our "adapted" version of the Odigos tutorial https://blog.qryn.dev/odigos-qryn

edeNFed:
Hi @lmangani  why not add qryn as a standalone destination so you will have your logo displayed and it will be cleared to the user which backend they are using? It will also allow you to customize the OTel collector configuration as you wish.

See this commit for an example of adding a new destination (this one adds 3 new dests, but you can simply follow all the files related to Jaeger for instance): https://github.com/keyval-dev/odigos/commit/13d7b30797ee8d501306863e80cf44463f3466a3

lmangani:
Thanks @edeNFed we can consider this down the road but actually our goal is to be compatible so this is a great usecase. Even with a new destination, I'm afraid the same blockers in #34 would apply short of duplicating a lot of code. Any feedback on that?

lmangani:
@edeNFed 

> Hi @lmangani why not add qryn as a standalone destination so you will have your logo displayed and it will be cleared to the user which backend they are using? It will also allow you to customize the OTel collector configuration as you wish.
> 

After some consideration we'll go for this route and add a dedicated qryn destination block

lmangani:
Closing to reopen a dedicated PR with the qryn destination

PR #34, opened 2022-11-14
WIP: Do not hardcode service ports

lmangani:
The Loki and Tempo port and path configs are very hardcoded. This is not very flexible and it does not necessarily apply globally, as many service entry points are proxied or behind other services (ie: cloudflare) or even served by different platforms on the same port (ie: qryn)

The PR proposal allows (and requires) users to configure a full path to the API instead of just the host, and removes the port hardcoding much like the prometheus configuration does. This would need a good example (or form placeholder) explaining the desired full path format, ie:

`https://hostname:port/path/to/api`

Some Input sanitization for paths seems to be already done [here](https://github.com/keyval-dev/odigos/blob/main/autoscaler/controllers/gateway/config/grafana.go#L61)


lmangani:
@edeNFed Thanks for considering this. I agree if the port and path are provided they should be used as-is, leaving space for flexibility and customisation for those using TLS/SSL and serving everything on 443. The fields could have a `placeholder` attribute to suggest the supported path without enforcing it. This change would allow us to use qryn without any code duplication.

edeNFed:
Sounds good. I'll be happy to merge this after the change

PR #59, opened 2023-01-25
Add ignore namespace option as a flag

eugercek:
Roadmap:

- [x] Add namespace ignoring for controller
- [x] Add namespace option in `values.yaml` on https://github.com/keyval-dev/odigos-charts
- [x] Test Cases
  - [x] Default namespaces from the code (moved to helm)
  - [x] Single namespace is given
  - [x] No namespace is given
  - [x] Multiple namespace is given
  - [x] Single non-existing namespace is given
  - [x] Multiple namespaces given but one is not existing
  - [x] Multiple namespaces given but more than one are not existing

### DIFF
--- a/cli/cmd/resources/instrumentor.go
+++ b/cli/cmd/resources/instrumentor.go
@@ -357,6 +357,11 @@ func NewInstrumentorDeployment(version string, telemetryEnabled bool) *appsv1.De
 		"--health-probe-bind-address=:8081",
 		"--metrics-bind-address=127.0.0.1:8080",
 		"--leader-elect",
+		"--ignore-namespace=odigos-system",
+		"--ignore-namespace=kube-system",
+		"--ignore-namespace=local-path-storage",
+		"--ignore-namespace=istio-system",
+		"--ignore-namespace=linkerd",
 		fmt.Sprintf("--lang-detector-tag=%s", version),
 		fmt.Sprintf("--lang-detector-image=%s", langDetectorImage),
 	}
--- a/instrumentor/controllers/common.go
+++ b/instrumentor/controllers/common.go
@@ -17,6 +17,7 @@ import (
 )
 
 var (
+	// IgnoredNamespaces have [odigos-system kube-system local-path-storage istio-system linkerd] by default
 	IgnoredNamespaces []string
 	SkipAnnotation    = "odigos.io/skip"
 )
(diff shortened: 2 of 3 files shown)


eugercek:
Manually edited helm chart and install odigos.

```bash
# Pushed this branch's instrumentor as eugercek/ns-ignore
docker build -t eugercek/ns-ignore  . --build-arg SERVICE_NAME=instrumentor
k create deploy foo --image stefanprodan/podinfo
k create ns other
k create -n other deploy fooother --image stefanprodan/podinfo
helm install my-odigos ./odigos-charts/charts/odigos --namespace odigos-system --create-namespace
```

Couldn't do `yq` magic 🙇🏻  But changes are these
`values.yaml`
```
instrumentor:
  deleteLangDetectionPods: true
  image:
    repository: eugercek/ns-ignore
    tag: latest
```

`instrumentor/deployment.yaml`
```
          - --ignore-namespace=other
```

---
Results

```
> k get instrumentedapplications -A                             
NAMESPACE   NAME   AGE
default     foo    3m29s
```

image

---

Updated First comment to give test cases, after your review and final approve I'll open helm-chart PR and test the test cases given in first comment.

edeNFed:
Looks like a good plan!

eugercek:
I'm waiting for your review btw 👀 Also I have a question on helm, should we get default ignored namespaces from helm chart or we should continue to hardcode them in code.

edeNFed (approved):

edeNFed:
Getting ignored namespaces from helm sounds good

eugercek:
Finished. the feature and tested the scenarios, it's alive 😁 It's ready for your review.

Also maybe we can discuss hash map usage for `shouldSkip`function.

edeNFed:
Cool! Regarding `shouldSkip`, I think using a hash map makes a lot of sense. 
A small comment:
We also have CLI for installing Odigos in addition to the helm chart. You may want to add the default ignored namespaces to the CLI also, you can see a similar code here:
https://github.com/keyval-dev/odigos/blob/9b6eb588a69d1f0677d94f8713a358681379e2a0/cli/cmd/resources/instrumentor.go#L355

eugercek:
Added both, second is a kinda hard coded, can discuss a better way if you wish.

edeNFed (approved):

edeNFed:
Looks great, merging

PR #70, opened 2023-02-13
Add missing option in CLI instruction

eugercek:
I noticed that I forgot to add cli option in odigos installation. It's totally OK to ignore this PR, with time Odigos will have lot of options and you may want to carefully choose how to do it.

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -19,10 +19,12 @@ const (
 )
 
 var (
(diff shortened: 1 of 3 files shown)


edeNFed (approved):

edeNFed:
Looks good
I totally agree that we need to be mindful of the number of CLI options as more and more options are added.

PR #46, opened 2023-01-09
Adding OpsVerse as a destination

navinpai:
[OpsVerse](https://opsverse.io) offers a fully managed, open-source based Observability platform that provides capabilities to collect, manage and analyze logs, traces, metrics, profiles, user-heatmaps etc. Since we offer endpoints that are compatible destinations for Loki _(for logs)_, Jaeger OTLP _(for traces)_ and Prometheus _(for metrics)_ with minor changes like authentication headers, adding OpsVerse a managed vendor destination for odigos

Screenshot 2023-01-10 at 12 20 28 AM

Screenshot 2023-01-10 at 12 21 21 AM

Screenshot 2023-01-10 at 12 24 22 AM

----
PS: We're big fans of fellow projects leveraging open-telemetry, and hope odigos grows into a great project. Looking forward to collaborating more! ♥️

PR #135, opened 2023-04-21
[MINOR] Move OpsVerse destination into Managed

navinpai:
Followup to #46 

----

Noticed that despite mentioning that [OpsVerse](https://opsverse.io) is a fully managed oss-based o11y solution, I had added it into OSS tools _(While the tools that OpsVerse offers are made up of OSS, the platform itself is a managed platform)_

Hence moving OpsVerse into the list of Managed Destinations to avoid any confusion. Alongside, also updated `assets/dests.png` to show all supported destinations in `README`


edeNFed (approved):
lgtm

Issue #72, opened 2023-02-13
Add VictoriaMetrics destination 


edeNFed:
As VictoriaMetrics gains popularity, we would like to add it as a new destination in Odigos.

eugercek:
May I pick this up. I've been curious about VictoriaMetrics since I saw its benchmarks 😄

edeNFed:
Sure sounds great. Let me know if you need help 🙂

josephlim75:
@eugercek is there any progress for this feature sending to Victoriametrics destination ?

PR #215, opened 2023-05-30
feat(vm): adding support for vm as an observability destination

arhamj:
Closes issue #72


arhamj:
I've used the `/api/v1/import/prometheus` API exposed by VM for this integration. @edeNFed Can you please review this when you get time?

Few questions:
- Are there unit tests which I need to add for this change?
- What if a self-hosted VM is not used and a hosted solution is used, do we handle auth for such cases?

arhamj:
Although the change to be made here was pretty well documented, is there a separate process which consumes config like the `endpoint`, `Exporters`, etc. to do the actual forwarding of metrics?

Issue #487, opened 2023-09-14
destination: logz.io region field should be a dropdown


edeNFed:
In logz.io destination, the region field should be a dropdown that contains the following values: https://docs.logz.io/user-guide/accounts/account-region.html

Similar to the site field in Datadog destination.

PR #534, opened 2023-09-22
Destination: logz.io region field dropdown

saputkin:
#487 
Created a dropdown similar to what was done in datadog's case.
image


edeNFed (approved):

edeNFed:
Looks great, thanks for contributing

PR #558, opened 2023-10-04
refactor: remove add.go file

### LINKED ISSUE #556: Remove odigos add command
It seems to `odigos add` cli command [is not doing anything](https://github.com/keyval-dev/odigos/blob/main/cli/cmd/add.go).

I think it can be safely removed

ftonato:
Closes #556


blumamir (approved):
Thanks!

Issue #543, opened 2023-10-03
Pressing enter in the create destination form does not invoke submit button


edeNFed:
In the create destination page (for example at `/overview/destinations/create/form?dest=splunk`)
Pressing enter should be the same as clicking the create destination button.

Krishnakalani111:
I want to try this.

prathameshgaikwad:
Hey can I solve this issue?

rauno56:
Weirdly(since there's no form elements in the DOM) enough enter works for me in Chrome and Firefox, but required inputs are not well-handled.

PR #585, opened 2023-10-04
on clicking enter functionality added

Krishnakalani111:
resolves : #543

### DIFF
--- a/frontend/webapp/components/setup/connection/create.connection.form/create.connection.form.tsx
+++ b/frontend/webapp/components/setup/connection/create.connection.form/create.connection.form.tsx
@@ -146,7 +146,7 @@ export function CreateConnectionForm({
   }
 
   return (
-    <form onClick={onCreateClick}>
+    <form onSubmit={onCreateClick}>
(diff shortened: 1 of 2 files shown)


frontend/webapp/components/setup/connection/create.connection.form/create.connection.form.tsx
```diff
@@ -179,14 +179,14 @@ export function CreateConnectionForm({
         {renderFields(fields, dynamicFields, handleDynamicFieldChange)}
       </DynamicFieldsWrapper>
       <CreateDestinationButtonWrapper>
-        <KeyvalButton disabled={isCreateButtonDisabled} onClick={onCreateClick}>
+        <KeyvalButton disabled={isCreateButtonDisabled} type='submit'>
```
alonkeyval, line 182: Thank you for your contribution, @Krishnakalani111. While the code seems to be functioning, there appears to be an issue that arises after form submission. Please double-check your code and resubmit it once the correction is made.

Issue #578, opened 2023-10-04
remove unused dependencies from nodejs instrumentation agent


blumamir:
[Odigos nodejs agent](https://github.com/keyval-dev/odigos/blob/main/odiglet/agents/nodejs/package.json) includes some unused dependencies in `package.json`.

Since the agent does not collect metrics at the moment, I think these are redundant:
- "@opentelemetry/sdk-metrics"
- "@opentelemetry/exporter-metrics-otlp-grpc"
- "@opentelemetry/exporter-prometheus"

@edeNFed - WDYT?

edeNFed:
👍🏻 sounds like a great idea

Krishnakalani111:
i want to contribute, please assign this to me

PR #587, opened 2023-10-04
Update package.json

Krishnakalani111:
resolves : #578

### DIFF
--- a/odiglet/agents/nodejs/package.json
+++ b/odiglet/agents/nodejs/package.json
@@ -16,15 +16,12 @@
   "dependencies": {
     "@opentelemetry/api": "1.6.0",
     "@opentelemetry/auto-instrumentations-node": "0.39.2",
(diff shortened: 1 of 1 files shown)


Krishnakalani111:
It would be very helpful if I had the docs to setup the required things to check the effect of changes in package.json of node js file.
@blumamir

blumamir:
> It would be very helpful if I had the docs to setup the required things to check the effect of changes in package.json of node js file. @blumamir

First step is to setup a cluster and install odigos as described in the [quickstart](https://docs.odigos.io/intro). Once you have a kind cluster that runs odigos, you can build the images from your source code with `TAG=v0.1.79 make build-images` and then push it to the kind cluster with `TAG=v0.1.79 make load-to-kind`. This pushes the new code, after which you can restart `odiglet` so it loads the updated build, and restart a nodejs service in the target application to have it instrumented with the new version.

Please let me know if you need further details. I can also verify it myself if you don't have time to setup everything 🙏

Issue #574, opened 2023-10-04
Use markdown linter for higher quality readmes


blumamir:
[Similar to opentelemetry repos](https://github.com/open-telemetry/opentelemetry-go/blob/main/.github/workflows/markdown.yml), we can add markdown linter steps to the CI to make sure the READMEs apply best practices.

I think we can use opentelemetry workflow as is:
https://github.com/open-telemetry/opentelemetry-go/blob/main/.github/workflows/markdown.yml

ankur12-1610:
Hi, can work on it?

PR #594, opened 2023-10-04
Add markdown linter for high quality READMEs

ankur12-1610:
Fixes: #574


ankur12-1610:
@blumamir can you please take a look? It should be the same as of the resource provided in the issue right?

blumamir (commented):
So sorry, I forgot to hit the "submit" button after reviewing.

It seems I misled you by giving reference to the wrong file 😅 sorry about that. After looking again, I think the CI linter we need should run 
```yaml
on:
  push:
  pull_request:
```

And we probably also need to add a `.markdownlint.yaml` file referenced in the yaml

Issue #565, opened 2023-10-04
odigos ui prints unuseful info to the console


blumamir:
When running `odigos ui` command and using the UI in `localhost:3000`, I get some prints to the console which are not helpful:

```
$ odigos ui
2023/10/04 11:01:58 Starting Odigos UI...
2023/10/04 11:01:58 Odigos UI is available at: http://localhost:3000
2023/10/04 11:02:09 Namespace kube-node-lease not found in request, skipping
2023/10/04 11:02:09 Namespace kube-public not found in request, skipping
```

We should identify the code paths that lead to this message being printed and make sure it is not outputted to console in `odigos ui` (or removed entirely)

prady0t:
I'm not really sure about the origin of `Namespace kube-node-lease not found in request, skipping`.  `namespace.go` doesn't even have a checker for namespaces in request.

PR #546, opened 2023-10-03
fix : Removed unnecessary logs

Rhythm-08:
issue: #542 
removed logs and code refactoring

PR #596, opened 2023-10-04
Fixed unuseful log messages

prady0t:
Fixes: #565 

Few log messages were removed.

### DIFF
--- a/frontend/main.go
+++ b/frontend/main.go
@@ -143,8 +143,8 @@ func main() {
 		log.Fatalf("Error starting server: %s", err)
 	}
 
-	log.Println("Starting Odigos UI...")
-	log.Printf("Odigos UI is available at: http://%s:%d", flags.Address, flags.Port)
+	// log.Println("Starting Odigos UI...")
+	// log.Printf("Odigos UI is available at: http://%s:%d", flags.Address, flags.Port)
 	err = r.Run(fmt.Sprintf("%s:%d", flags.Address, flags.Port))
 	if err != nil {
 		log.Fatalf("Error starting server: %s", err)


frontend/main.go
```diff
@@ -143,8 +143,8 @@ func main() {
 		log.Fatalf("Error starting server: %s", err)
 	}
 
-	log.Println("Starting Odigos UI...")
-	log.Printf("Odigos UI is available at: http://%s:%d", flags.Address, flags.Port)
+	// log.Println("Starting Odigos UI...")
+	// log.Printf("Odigos UI is available at: http://%s:%d", flags.Address, flags.Port)
```
blumamir, line 147: Thank you for working on a fix.

In the issue, I was actually referring to these lines:
```
2023/10/04 11:02:09 Namespace kube-node-lease not found in request, skipping
2023/10/04 11:02:09 Namespace kube-public not found in request, skipping
```

Which I find not helpful.

The URL itself is very helpful. Without it, the user will need to search docs or guides to understand what the next step is, and by outputting the URL to the console the user can immediately go there and use the UI.

To clean up the odigos UI console outputs, we need:
1. Browser the code in the `frontend` module to identify where such prints occur (the one in the example and maybe more).
2. Understand the logical intention of these messages and review if they might contain useful information in some situations or not.
3. Follow the code paths that might trigger such prints and make sure they do not carry any meaning in other contexts (such as running not from the UI?).

Sorry about not being clear in the original issue. Please let me know if you need more context and help regarding this fix.

prady0t: That actually makes much more sense. Can you provide steps to reproduce this locally as I can't see these lines when running `odigos ui`

blumamir: I started by creating the kind cluster and demo application as described in [odigos docs](https://docs.odigos.io/installation).

Run `odigos install` on my playground cluster. My cli version is:
```
➜  ~ odigos version
version.Info{Version:'v0.1.75', GitCommit:'d20861b', BuildDate:'2023-10-04T19:15:51Z'}
```

When I run `odigos ui`:
```sh
➜  ~ odigos ui
2023/10/08 12:02:32 Starting Odigos UI...
2023/10/08 12:02:32 Odigos UI is available at: http://localhost:3000
```

The use browser to load odigos UI at `http://localhost:3000` and upon reaching the destinations page in the setup, I get these in my console:
```sh
2023/10/08 12:02:37 Namespace default not found in request, skipping
2023/10/08 12:02:37 Namespace kube-node-lease not found in request, skipping
```

To be honset, I can find few places in code that [prints something similar](https://github.com/keyval-dev/odigos/blob/4901838f4dad9f3ab78c76e175b175aecef2c58d/frontend/endpoints/namespaces.go#L147):
```go
		if !exists {
			log.Printf("Deployment %s not found in request, skipping\n", dep.Name)
```

But not the Namespace message itself. I am a bit confused 🤔 perhaps something went messy in my local setup? 

Can you please share if you are able to reproduce by browsing to the destinations screen in setup? Thank you so much 🙏

prady0t: This was fixed in #546 , those log messages were removed. You might be using a previous version.

blumamir: Yeah I see what you mean. Thank you so much for taking a look 🙏

So, I think the log prints are already as required and we can close the PR?

prady0t: Yes we are done here.

Issue #569, opened 2023-10-04
Upgrade go version to newer 1.21


blumamir:
The project currently uses go version `1.20`.

This can be found in the `go.mod` and `go.work` files as
```
module github.com/keyval-dev/odigos/...

go 1.20
```

And in docker files:
```
RUN curl -LO https://go.dev/dl/go1.20.linux-${TARGETARCH}.tar.gz && tar -C /usr/local -xzf go*.linux-${TARGETARCH}.tar.gz
```

```
FROM golang:1.20 as builder
```

Since go latest version is currently 1.21, we should review the changes and update the repo to use the newer version

PR #597, opened 2023-10-05
Upgraded Go 1.20 -> 1.21

prady0t:
Fixes: #569


blumamir:
Thanks for the contribution @prady0t .

I think that bumping the go version should be OK for the project.
Did you build and run it on a test cluster after the change to verify everything still works as expected?

I would also suggest quickly browsing [the release notes](https://tip.golang.org/doc/go1.21) to make sure that we are not affected by anything introduced in the new version.

Thanks!

prady0t:
Ah yes I did! It was working fine with all it's cli commands working. The reason the above build failed is because starting from go 1.21 the way versions were numbered has slightly changed (as per the release note https://tip.golang.org/doc/go1.21#:~:text=Go%201.21%20introduces,new%20version%20numbering.) 

So in this line : 
https://github.com/keyval-dev/odigos/blob/d20861bb1429e6160b27200a1278bcc4f61899d4/odiglet/Dockerfile#L20
go1.20 should be replaced by something like 1.21.0 (or maybe higher). But I'm not sure if I should keep 1.21 elsewhere as it is or change there too.

blumamir (approved):
Thanks again 🥇

Issue #552, opened 2023-10-04
Add --version flag to odigos ui binary


edeNFed:
The Odigos UI binary should have a flag that will print the current version of the UI.

blumamir:
This was added in #600

PR #600, opened 2023-10-05
Inject ui version and add option to print it

### LINKED ISSUE #551: Inject version to Odigos UI binary
Odigos UI binary should include the git tag version.
This can be done using `ldflags` in goreleaser, similar to the way that we do it for the CLI:
https://github.com/keyval-dev/odigos/blob/02639256a750cfa8e0e7555954af046fc8140c23/.goreleaser.yaml#L19-L22

kousikmitra:
- Inject version info to `odigos-ui` binary
- Add flag to print version in `odigos-ui` binary

Fixed #551 #552

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -31,36 +31,40 @@ builds:
       - arm64
     dir: ./frontend
     id: odigos-ui
+    ldflags:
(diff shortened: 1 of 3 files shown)


edeNFed (approved):
lgtm
thanks for contributing!

PR #605, opened 2023-10-06
Fix #599

### LINKED ISSUE #599: Odigos UI bind to different port if port 3000 already taken
when launching Odigos UI via`odigos ui` we are trying to listen on port 3000, if this port is taken by another application we should try other ports.

I think trying 3001,3002,3003 before giving up sounds reasonable.

rahul0x00:
Fixes issue #599

### DIFF
--- a/cli/cmd/ui.go
+++ b/cli/cmd/ui.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


rahul0x00:
The code attempts to bind to the specified port (default: 3000) for serving the UI. If port 3000 is already in use, it automatically increments the port number and tries the next one until an available port is found.

edeNFed:
Hi @rahul0x00 
The code for the UI http server already exists at:
https://github.com/keyval-dev/odigos/blob/main/frontend/main.go

I think this is the file you should change.

rahul0x00:
okay, i'll create a new PR

PR #611, opened 2023-10-08
Fix odigos ui bind port #599

rahul0x00:
Fixes issue https://github.com/keyval-dev/odigos/issues/599

The code attempts to bind to the specified port (default: 3000) for serving the UI. If port 3000 is already in use, it automatically increments the port number and tries the next one until an available port is found.


frontend/main.go
```diff
@@ -145,16 +146,24 @@ func main() {
...
-	log.Println("Starting Odigos UI...")
-	log.Printf("Odigos UI is available at: http://%s:%d", flags.Address, flags.Port)
-	err = r.Run(fmt.Sprintf("%s:%d", flags.Address, flags.Port))
-	if err != nil {
-		log.Fatalf("Error starting server: %s", err)
+	// Attempt to start the server on port 3000 and increment the port number if it's already in use
+	for {
+		r, err := startHTTPServer(&flags)
+		if err != nil {
+			log.Printf("Error starting server on port %d: %s", flags.Port, err)
```
blumamir, line 153: Can you share an example of how to line will be printed to the user console? will the %s expand to long text or is it a short one?

frontend/main.go
```diff
@@ -145,16 +146,24 @@ func main() {
...
-		log.Fatalf("Error starting server: %s", err)
-	}
-
-	log.Println("Starting Odigos UI...")
-	log.Printf("Odigos UI is available at: http://%s:%d", flags.Address, flags.Port)
-	err = r.Run(fmt.Sprintf("%s:%d", flags.Address, flags.Port))
-	if err != nil {
-		log.Fatalf("Error starting server: %s", err)
+	// Attempt to start the server on port 3000 and increment the port number if it's already in use
+	for {
```
blumamir, line 150: If the err we encounter is not related to the port being bound already, then we might run this loop many times before something will break us out.

Perhaps we can limit the number of tested ports to something like 5 or 10? this way, if there is an issue - we do not iterate dozens of ports.

frontend/main.go
```diff
@@ -145,16 +146,24 @@ func main() {
...
-	if err != nil {
-		log.Fatalf("Error starting server: %s", err)
-	}
-
-	log.Println("Starting Odigos UI...")
-	log.Printf("Odigos UI is available at: http://%s:%d", flags.Address, flags.Port)
-	err = r.Run(fmt.Sprintf("%s:%d", flags.Address, flags.Port))
-	if err != nil {
-		log.Fatalf("Error starting server: %s", err)
+	// Attempt to start the server on port 3000 and increment the port number if it's already in use
```
blumamir, line 149: 3000 or what the user supplied in flags.

frontend/main.go
```diff
@@ -145,16 +146,24 @@ func main() {
...
+		if err != nil {
+			log.Printf("Error starting server on port %d: %s", flags.Port, err)
+			flags.Port++ // Try the next port
+		} else {
+			log.Printf("Starting Odigos UI...")
+			log.Printf("Odigos UI is available at: http://%s:%d", flags.Address, flags.Port)
+			err = r.Run(fmt.Sprintf("%s:%d", flags.Address, flags.Port))
+			if err != nil {
+				log.Printf("Error starting server on port %d: %s", flags.Port, err)
+				flags.Port++ // Try the next port
```
blumamir, line 161: I see we check for errors in 2 places: `startHTTPServer(&flags)` and `r.Run(fmt.Sprintf("%s:%d", flags.Address, flags.Port))`, but I guess only one of them returns the err when the port is already in use. We might simplify the flow and attempt a new port only in the path where it is expected, and fail fast if the error is unrelated to the port.

frontend/main.go
```diff
@@ -145,16 +146,24 @@ func main() {
...
+			} else {
+				// Server started successfully				
+				break
+			}
```
blumamir, line 165: nit: The `r.Run` will actually block until the server is stopped:
```
// Note: this method will block the calling goroutine indefinitely unless an error happens.
```

So the comment is not accurate in this case.

frontend/main.go
blumamir, line 150: nit: matter of styling, so completely optional - this for loop has 3 levels of nesting: `for`, `if-else`, `if-else`.

An alternative could be to use just one level of nesting and use `continue` and `break` to structure the code.

So this:
```go
		if err != nil {
			log.Printf("Error starting server on port %d: %s", flags.Port, err)
			flags.Port++ // Try the next port
		} else {
			log.Printf("Starting Odigos UI...")
```

Can become this:
```go
		if err != nil {
			log.Printf("Error starting server on port %d: %s", flags.Port, err)
			flags.Port++ // Try the next port
                         continue
		}

		log.Printf("Starting Odigos UI...")
```

blumamir (commented):
Thank you so much for your contribution.
Added few minor comments :)

frontend/main.go
rahul0x00: The line that's printed to the user console will look something like this if an error occurs:
` Error starting server on port 3000: <error_message_here> `

%d format specifier is for `flags.Port` integer variable and %s formatter is for the `err` string

frontend/main.go
rahul0x00: You are correct; only the r.Run call is likely to return an error related to the port being already in use.
To simplify the flow and fail fast when encountering errors unrelated to the port, i moved the port increment logic to the part of the code where r.Run is called. This way, it will only attempt to increment the port when it know it's necessary.

frontend/main.go
rahul0x00: limited the number of testing ports to 10

nicolascb:
hello

we could use ":0" to get a randomly open port, does that make sense here?

https://stackoverflow.com/a/43425461

rahul0x00:
> hello
> 
> we could use ":0" to get a randomly open port, does that make sense here?
> 
> https://stackoverflow.com/a/43425461

Using a random port is certainly an option, and it can help avoid port conflicts. However, in some cases, it's desirable to use a specific range of ports for better control and predictability.

rahul0x00:
Hi, @blumamir 
Sorry for pinging you,i've made the necessary changes, can you review the changes?

blumamir (commented):
Thanks and sorry for the time to review.

The code LGTM, but I wonder if we need to attempt a new port if `startHTTPServer(&flags)` returns err. WDYT?

frontend/main.go
```diff
@@ -144,17 +144,30 @@ func main() {
 	if err != nil {
 		log.Fatalf("Error creating Kubernetes client: %s", err)
 	}
+	
+	// Attempt to start the server on default port(3000) or user-specified port and increment the port number if it's already in use
+	maxPortAttempts := 10 // Maximum number of port attempts
+	for attempt := 1; attempt <= maxPortAttempts; attempt++ {
+		r, err := startHTTPServer(&flags)
+		if err != nil {
```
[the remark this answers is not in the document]
rahul0x00: You're correct, and I appreciate your observation. If there's a port conflict, r.Run() will indeed return an error, and it doesn't make sense to increment the port before starting the server using r.Run(). In this case, the error handling for port conflicts in the loop isn't necessary.

frontend/main.go
```diff
@@ -144,17 +144,30 @@ func main() {
...
+		err = r.Run(fmt.Sprintf("%s:%d", flags.Address, flags.Port))
 
-	log.Println("Starting Odigos UI...")
-	log.Printf("Odigos UI is available at: http://%s:%d", flags.Address, flags.Port)
-	err = r.Run(fmt.Sprintf("%s:%d", flags.Address, flags.Port))
-	if err != nil {
-		log.Fatalf("Error starting server: %s", err)
+		if err != nil {
+			log.Printf("Error starting server on port %d: %s", flags.Port, err)
+			if attempt < maxPortAttempts {
```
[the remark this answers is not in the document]
rahul0x00: You're correct; it's important to make both cases consistent.

blumamir (approved):
LGTM, Thanks again! 🙏
Can I kindly ask you to fix the lint on the file?
It seems to change indentations. It might be related to some plugins in your IDE maybe?

blumamir:
```
#24 73.12 ./main.go:90:62: syntax error: unexpected newline in argument list; possibly missing comma or )
```

rahul0x00:
> ```
> #24 73.12 ./main.go:90:62: syntax error: unexpected newline in argument list; possibly missing comma or )
> ```
There was a missing ) at line number 90, i fixed it , should work now

rahul0x00:
> LGTM, Thanks again! 🙏 Can I kindly ask you to fix the lint on the file? It seems to change indentations. It might be related to some plugins in your IDE maybe?

Hi, can you tell me how much spaces do you use as indent as i'm not able to figure it out?

rahul0x00:
Hi, @blumamir,
I've pushed the latest changes to the code, and I believe it's working better than ever now.

frontend/main.go
```diff
@@ -125,36 +129,63 @@ func httpFileServerWith404(fs http.FileSystem) http.Handler {
...
+    }()
+
+    // Attempt to start the server on the specified or default port
+    maxPortAttempts := 10
+    for {
+        r, err := startHTTPServer(&flags)
+        if err != nil {
+            log.Printf("Error starting server on port %d: %s", flags.Port, err)
+
+            if flags.Port < defaultPort+maxPortAttempts {
```
blumamir, line 167: This code assumes `flags.Port` is initialized to `defaultPort`. 
But the user can run `odigos ui --port 4000` or `odigos ui --port 2000` and then you would run the loop once or many many times.

frontend/main.go
```diff
@@ -125,36 +129,63 @@ func httpFileServerWith404(fs http.FileSystem) http.Handler {
...
+    signal.Notify(interruptChan, os.Interrupt, syscall.SIGINT)
+
+    go func() {
+        for range interruptChan {
+            log.Println("Received Ctrl+C. Shutting down gracefully...")
+            // Implement any cleanup or shutdown logic here
+
+            os.Exit(0)
+        }
+    }()
```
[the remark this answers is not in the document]
rahul0x00: There was a bug in the code, when running the ui if the default port was taken and if it took other ports then this was the output:

```
2023/10/15 13:38:30 Starting Odigos UI...
2023/10/15 13:38:30 Odigos UI is available at: http://localhost:3000
2023/10/15 13:38:30 Error starting server on port 3000: listen tcp 127.0.0.1:3000: bind: address already in use
2023/10/15 13:38:30 Attempting to start the server on the next available port...
2023/10/15 13:38:30 Starting Odigos UI...
2023/10/15 13:38:30 Odigos UI is available at: http://localhost:3001```

The first 2 lines were not supposed to be run when default port was taken, i did a workaround to fix it and it worked but there was no option to cancel to process, so i used os to regsiter for the cancel signal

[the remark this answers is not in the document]
Comment (rahul0x00) [reply frontend/main.go:158]: if you don't like this approach maybe we should keep the previous code with output like this:
```
2023/10/15 17:46:50 Starting Odigos UI on port 3000...
2023/10/15 17:46:50 Error starting server on port 3000: listen tcp 127.0.0.1:3000: bind: address already in use
2023/10/15 17:46:50 Attempting to start the server on the next available port...
2023/10/15 17:46:50 Starting Odigos UI on port 3001...
```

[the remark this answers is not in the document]
Comment (rahul0x00) [reply frontend/main.go:158]: > There was a bug in the code, when running the ui if the default port was taken and if it took other port then this was the output:
> 
> ```
> 2023/10/15 13:38:30 Starting Odigos UI...
> 2023/10/15 13:38:30 Odigos UI is available at: http://localhost:3000
> 2023/10/15 13:38:30 Error starting server on port 3000: listen tcp 127.0.0.1:3000: bind: address already in use
> 2023/10/15 13:38:30 Attempting to start the server on the next available port...
> 2023/10/15 13:38:30 Starting Odigos UI...
> 2023/10/15 13:38:30 Odigos UI is available at: http://localhost:3001
> ```
> 
> The 2nd line was not supposed to be printed when default port was taken i was not able to fix it ,so i did a workaround to fix it and it worked but now there was no option to cancel the process, so i explicitly used th interrupt signal

or we could show this output?:
```
2023/10/15 17:53:36 Starting Odigos UI at http://localhost:3000
2023/10/15 17:53:36 Error starting server on port 3000: listen tcp 127.0.0.1:3000: bind: address already in use
2023/10/15 17:53:36 Attempting to start the server on the next available port...
2023/10/15 17:53:36 Starting Odigos UI at http://localhost:3001
```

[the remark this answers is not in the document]
rahul0x00: what's your opinion on this?

PR #626, opened 2023-10-10
Adds linters to Go modules

### LINKED ISSUE #572: Add linters to Go modules
We should add linters that validate our Go code on pull request checks.
Relevant for the following modules: Odiglet, Autoscaler, Scheduler, Instrumentor, Frontend
I had a good experience using: https://github.com/golangci/golangci-lint

ankur0904:
Fix #572

Added the workflow for the linters.


ankur0904:
@edeNFed Please review the PR.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.19.0
+
+    - name: Install golangci-lint
+      run: go get -u github.com/golangci/golangci-lint/cmd/golangci-lint
+
+    - name: Run golangci-lint
+      run: golangci-lint run --allow-serial-runners
```
blumamir, line 25: I suggest moving this to `Makefile`, so the user can execute it locally to get lint output before submitting a PR and running the CI

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,28 @@
...
+
+    - name: Fix Go code with golangci-lint (optional)
+      run: golangci-lint run --fix --allow-serial-runners
```
blumamir, line 28: I think this step will not run if the previous step failed.
If CI step is making changes to files (like lint fix), they have to be committed and pushed to take effect, otherwise, they just go down with the CI container.

My suggestion is to just verify in CI that `make lint` is successful, and if not, the user will have to run locally `make golangci-lint-fix` and push a fixed version  after linting

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.19.0
+
+    - name: Install golangci-lint
+      run: go get -u github.com/golangci/golangci-lint/cmd/golangci-lint
```
blumamir, line 22: I see there is a [github action](https://github.com/golangci/golangci-lint-action) for that which does [many awesome things](https://golangci-lint.run/usage/install/#ci-installation):

> We recommend using [our GitHub Action](https://github.com/golangci/golangci-lint-action) for running golangci-lint in CI for GitHub projects. It's [fast and uses smart caching](https://github.com/golangci/golangci-lint-action#performance) inside and it can be much faster than the simple binary installation.
> 
> Also, the action creates GitHub annotations for found issues: you don't need to dig into build log to see found by golangci-lint issues

Perhaps we can use it instead?

ankur0904:
@blumamir I have made the required changes.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,24 @@
...
+    paths:
+      - '**/*.go'
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
```
blumamir, line 14: There is already v4 for this action:
https://github.com/actions/checkout

I suggest using the latest version

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,24 @@
...
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
```
blumamir, line 17: also here, there is already v4 which is preferable: https://github.com/actions/setup-go

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,24 @@
...
+      with:
+        go-version: 1.19.0
```
blumamir, line 19: The project uses go version `1.21`. You can check this for example [here](https://github.com/keyval-dev/odigos/blob/78631271fd3123f01a78f5585cd503b3ea438a1a/.github/workflows/main.yml#L23) and [here](https://github.com/keyval-dev/odigos/blob/78631271fd3123f01a78f5585cd503b3ea438a1a/go.work#L1)

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,24 @@
...
+
+    - name: Run golangci-lint
+      uses: golangci/golangci-lint-action@v2
```
blumamir, line 22: Use [latest version v3](https://github.com/golangci/golangci-lint-action/releases)

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,24 @@
...
+      with:
+        version: latest
```
blumamir, line 24: By using the latest version, we can bring in breaking (or non-breaking) changes that will make the CI fail one day because of changes in the dependency.

I suggest pinning the version so the CI runs are stable and would not introduce any unexpected changes.

blumamir (commented):
Thanks

Add few comments

ankur0904:
@blumamir Thanks for the feedback. Made the required changes.

blumamir (commented):
Sorry for the delay.
I reviewed it again, and added 2 minor comments.

ankur0904:
@blumamir Hi, I changed the version of the go from `1.21` to `1.55.0`.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Set up Go
+      uses: actions/setup-go@v4
+      with:
+        go-version: 1.21
+
+    - name: Run golangci-lint
```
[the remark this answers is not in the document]
ankur0904: @blumamir 
Sorry for this but I think the file is contained in [this](https://github.com/open-telemetry/opentelemetry-go) repo and the PR is for [this](https://github.com/keyval-dev/odigos) repo.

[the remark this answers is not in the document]
blumamir: > @blumamir Sorry for this but I am unable to understand.

@ankur0904 I think that you also need to add a `.golangci.yml` file to the project, which this github actions will look for.

ankur0904:
hey @blumamir any other changes required?

ankur0904:
@blumamir Can we merge the PR or does any addition issue exists or changes necessary?

blumamir (approved):

PR #631, opened 2023-10-10
Update instrumenting.mdx

creator0131:
I do some Minor Fixes. I hope, you like my Changes.

### DIFF
--- a/docs/instrumenting.mdx
+++ b/docs/instrumenting.mdx
@@ -1,12 +1,12 @@
 ---
-title: "Generating Traces"
-sidebarTitle: "Generating Traces"
+Title:- "Generating Traces"
+SidebarTitle:- "Generating Traces"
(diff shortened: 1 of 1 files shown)


docs/instrumenting.mdx
```diff
@@ -1,12 +1,12 @@
 ---
-title: "Generating Traces"
-sidebarTitle: "Generating Traces"
+Title:- "Generating Traces"
```
blumamir, line 2: I see many other places in the repo where this pattern is used, for example [here](https://github.com/keyval-dev/odigos/blob/eaec27a5db5cad0b241450c17dfec6a3b8e69601/docs/installation.mdx#L2). but there are many other such files.

Can you share more about the motivation to make this change, and if so, should we apply it everwhere in the repo?

PR #637, opened 2023-10-12
feat: setup odigos own telemetry logs pipeline for odigos cloud

blumamir:


cli/cmd/install.go
```diff
@@ -48,8 +49,21 @@ var installCmd = &cobra.Command{
...
 			client, cmd, ns, createNamespace)
+		if isOdigosCloud {
+			createKubeResourceWithLogging(ctx, "Creating Odigos Cloud Secret",
+				client, cmd, ns, createOdigosCloudSecret)
+			createKubeResourceWithLogging(ctx, "Creating Own Telemetry Pipeline",
+				client, cmd, ns, createOwnTelemetryPipeline)
+			createKubeResourceWithLogging(ctx, "Deploying Odigos Cloud Proxy",
+				client, cmd, ns, createKeyvalProxy)
+		} else {
+			createKubeResourceWithLogging(ctx, "Own Telemetry Pipeline disabled",
```
edeNFed, line 64: Not sure we should print it to the screen, may sounds to user like there is some kind of error. Maybe just call the function without logging this text

edeNFed (approved):

cli/cmd/install.go
blumamir: good point. fixed 👍

PR #657, opened 2023-10-17
feat: persist and print odigos version in cli

blumamir:


cli/cmd/version.go
```diff
@@ -20,7 +24,26 @@ var versionCmd = &cobra.Command{
 	Use:   "version",
 	Short: "Print odigos version.",
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Printf("version.Info{Version:'%s', GitCommit:'%s', BuildDate:'%s'}\n", OdigosVersion, OdigosCommit, OdigosDate)
+		fmt.Printf("Odigos Cli Version: version.Info{Version:'%s', GitCommit:'%s', BuildDate:'%s'}\n", OdigosVersion, OdigosCommit, OdigosDate)
+
```
edeNFed, line 28: Users can run `odigos version` even before installing Odigos on their cluster or even connecting to Kubernetes. In that case, we probably only want to show the version of the CLI.
I think `kube.CreateClient` panics or print an error

edeNFed (approved):

PR #664, opened 2023-10-18
feat(ci): macos-latest-xlarge arm64 runners

CyberHippo:
Building the ARM images on ARM runners.

Closes https://github.com/keyval-dev/odigos/issues/555

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,8 +8,15 @@ permissions:
   packages: write
 jobs:
   release:
-    runs-on:
-      labels: large-runner
+    strategy:
+      matrix:
+        include:
+          - platforms: "linux/amd64"
+            runner:
+              labels: large-runner
+          - platforms: "linux/arm64"
+            runner: "macos-latest-xlarge"
+    runs-on: ${{ matrix.runner }}
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


edeNFed:
Hi @CyberHippo 
Did you run it? I suspect it will not work, I think executing `docker push` to two tags with the same name will overwrite each other.
So for the first run (for example amd64) the workflow will execute `docker push keyval/odigos-autoscaler:v1.0`
and in the second run (for example arm64) the workflow will be executing again `docker push keyval/odigos-autoscaler:v1.0` and will overwrite the image pushed earlier so we are left with only arm64 image.
I think the solution to that is to push two images with two separate tags and then merge them. This blog post describes similar method: https://dev.to/aws-builders/using-docker-manifest-to-create-multi-arch-images-on-aws-graviton-processors-1320

CyberHippo:
Hi @edeNFed,
I did not run it because I neither have access to hosted runner, nor a paid subscription to run macos-latest-xlarge runners. If you think of a way to test it please let me know.
You're right about docker tag conflict. I'll take a look at the blog post and figure something out.

edeNFed:
@CyberHippo interesting, this looks like it may work

CyberHippo:
I'll close this PR for now as I don't have time to work on it anymore. Good luck

Issue #658, opened 2023-10-17
Odigos uninstall output is not helpful when odigos in not installed


blumamir:
When running the odigos cli `odigos uninstall`, there can be 2 options.

1. odigos is install, in which case the following output is printed to console which is **good**
```
➜  cli git:(odigos-deployment) odigos uninstall
About to uninstall Odigos from namespace odigos-system
Are you sure? [Y/n]: Y
Uninstalling Odigos Deployments                   ✔
Uninstalling Odigos DaemonSets                    ✔
Uninstalling Odigos ConfigMaps                    ✔
Uninstalling Odigos CRDs                          ✔
Uninstalling Odigos RBAC                          ✔
Uninstalling Odigos Secrets                       ✔
Uninstalling Namespace odigos-system              ✔
Waiting for namespace to be deleted               ✔
Rolling back odigos changes to pods
Rolling back odigos changes to namespaces
        ✔

SUCCESS: Odigos uninstalled.
```

2. If Odigos is not installed to begin with, the output is not very helpful:
```
➜  cli git:(odigos-deployment) odigos uninstall
About to uninstall Odigos from namespace odigos-system
Are you sure? [Y/n]: Y
Uninstalling Odigos Deployments                   ✔
Uninstalling Odigos DaemonSets                    ✔
Uninstalling Odigos ConfigMaps                    ✔
Uninstalling Odigos CRDs                          ✔
Uninstalling Odigos RBAC                          ✔
Uninstalling Odigos Secrets                       ✔
Uninstalling Namespace odigos-system              X
ERROR namespaces "odigos-system" not found
```

I think it is more useful to print something like "Odigos is not installed in the cluster" if the namespace is not found.

To check your fix, you can set up an odigos environment with [Quickstart](https://docs.odigos.io/intro). Then, make necessary changes in the `cli` module code, and run in terminal:
```sh
➜  odigos git:(main) cd cli
➜  cli git:(main) go run . uninstall
```

To run the cli uninstall command locally from code and verify your fix.

Thanks!

HeetVekariya:
- @blumamir I understand the issue as to just change the print statement of the second case.
- If you find my solution correct, can you please assign me the issue.
- Also, are you using the older version, because i have already corrected the position of Tic marks(v) in this issue: https://github.com/keyval-dev/odigos/issues/614

blumamir:
Yes I am using v0.1.75. 
Thanks for the help @HeetVekariya

[...]

PR #673, opened 2023-10-19
Fix/uninstall log

HeetVekariya:
# Fixes issue:
- Issue: #658 
- Update the error message for uninstall of odigos if it is not installed

<br>

# Changes made:
- Updated the code such that if no **Namepace** is found then print the updated error message and exit
- Other changes:
  - Previously the code continues ```defaultnamespace``` but the updated code print the error and exit

<br>

### Files changed:
- **uninstall.go** (cli/cmd/uninstall.go)

### DIFF
--- a/cli/cmd/uninstall.go
+++ b/cli/cmd/uninstall.go
@@ -39,17 +39,10 @@ var uninstallCmd = &cobra.Command{
 
 		ns, err := resources.GetOdigosNamespace(client, ctx)
 		if err != nil {
-			errString := err.Error()
-
-			if errString[len(errString) - 1] == '0' {
-				errMessage := "\033[31mERROR\033[0m Odigos is not currently installed in the cluster, so there is nothing to uninstall"
-				fmt.Println(errMessage)
-			} else {
-				fmt.Println("\033[31mERROR\033[0m", err)
-			}
+			fmt.Println("\033[31mERROR\033[0m", err)
 			os.Exit(1)
 		}
-		
+
 		fmt.Printf("About to uninstall Odigos from namespace %s\n", ns)
 		confirmed, err := confirm.Ask("Are you sure?")
 		if err != nil || !confirmed {
--- a/cli/cmd/uninstall.go
+++ b/cli/cmd/uninstall.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
@@ -13,6 +14,7 @@ import (
 	"github.com/keyval-dev/odigos/cli/pkg/log"
 	"github.com/keyval-dev/odigos/common/consts"
 	v1 "k8s.io/api/core/v1"
+	// "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 
@@ -36,18 +38,19 @@ var uninstallCmd = &cobra.Command{
 		}
 		ctx := cmd.Context()
 
-		errMessage := fmt.Errorf("Odigos is not installed in the cluster")
 		ns, err := resources.GetOdigosNamespace(client, ctx)
 		if err != nil {
-			if errors.IsNotFound(err) {
+			errString := err.Error()
+
+			if errString[len(errString) - 1] == '0' {
+				errMessage := "\033[31mERROR\033[0m Odigos is not currently installed in the cluster, so there is nothing to uninstall"
 				fmt.Println(errMessage)
 			} else {
-				ns = consts.DefaultNamespace
-				fmt.Println(err)
+				fmt.Println("\033[31mERROR\033[0m", err)
 			}
 			os.Exit(1)
 		}
-
+		
 		fmt.Printf("About to uninstall Odigos from namespace %s\n", ns)
 		confirmed, err := confirm.Ask("Are you sure?")
 		if err != nil || !confirmed {
--- a/cli/cmd/uninstall.go
+++ b/cli/cmd/uninstall.go
@@ -38,8 +38,11 @@ var uninstallCmd = &cobra.Command{
 		ctx := cmd.Context()
 
 		ns, err := resources.GetOdigosNamespace(client, ctx)
-		if err != nil {
-			fmt.Println("\033[31mERROR\033[0m", err)
+		if resources.IsErrNoOdigosNamespaceFound(err) {
+			fmt.Println("\033[31mERROR\033[0m odigos is not currently installed in the cluster, so there is nothing to uninstall")
+			os.Exit(1)
+		} else if !resources.IsErrNoOdigosNamespaceFound(err) && err != nil{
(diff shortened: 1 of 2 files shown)


HeetVekariya:
I have accidentally updated the branch, so one additional commit is added in the PR

cli/cmd/uninstall.go
```diff
@@ -36,9 +36,16 @@ var uninstallCmd = &cobra.Command{
 		}
 		ctx := cmd.Context()
 
+		errMessage := fmt.Errorf("Odigos is not installed in the cluster")
```
blumamir, line 39: This variable should be declared in the internal `if` next to where it is being used.

blumamir, line 39: Should we use `fmt.Errorf`? Can we use plain string instead?

cli/cmd/uninstall.go
```diff
@@ -36,9 +36,16 @@ var uninstallCmd = &cobra.Command{
...
 		ns, err := resources.GetOdigosNamespace(client, ctx)
 		if err != nil {
-			ns = consts.DefaultNamespace
+			if errors.IsNotFound(err) {
+				fmt.Println(errMessage)
+			} else {
+				ns = consts.DefaultNamespace
```
blumamir, line 45: this is redundant since we `os.Exit` right after

cli/cmd/uninstall.go
```diff
@@ -36,9 +36,16 @@ var uninstallCmd = &cobra.Command{
...
+				fmt.Println(err)
```
blumamir, line 46: What will be printed in this case? will it include a full-blown stack trace?
can you please provide an example for the output?

I suggest at least adding some text that target the user with instructions, like "An error occurred during odigos uninstall. Please check your kubeconfig and try again. If the problem persists or you need assistance, don't hesitate to open an issue or reach out in odigos slack channel"

cli/cmd/uninstall.go
blumamir, line 39: The current message lacks context.

How about? 
```
Odigos is not currently installed in the cluster, so there is nothing to uninstall
```

blumamir (commented):
Thanks again @HeetVekariya .

Left few comments.

Did you run a local test for this new logic? Can you please provide examples of how the output will look?

blumamir:
[I have added a section to the `CONTRIBUTING.md` file](https://github.com/keyval-dev/odigos/pull/672/files) that you can use to test you changes. Hope it will help

cli/cmd/uninstall.go
HeetVekariya: I have updated the error message with your suggested message

cli/cmd/uninstall.go
HeetVekariya: - I just found that my if condition never triggers, so i removed it

cli/cmd/uninstall.go
HeetVekariya: Hey @blumamir, I want to use something like l.Error(errMessage) but i am not able to do it
- Initially i want to proceed with ```l.Error(errMessage)``` but when i use that it says ```l not defined```
- So i left it there.
- I have changed the code you can see at the end of this thread.

cli/cmd/uninstall.go
HeetVekariya: - We can do that, but my concern is that, is this message ```Please check your kubeconfig and try again.```,  generic all types of problem ?

HeetVekariya:
> [I have added a section to the `CONTRIBUTING.md` file](https://github.com/keyval-dev/odigos/pull/672/files) that you can use to test you changes. Hope it will help

- It helped me to test the code, you can check the example outputs of the errors in above comment :+1:

cli/cmd/uninstall.go
```diff
@@ -38,9 +39,17 @@ var uninstallCmd = &cobra.Command{
 
 		ns, err := resources.GetOdigosNamespace(client, ctx)
 		if err != nil {
-			ns = consts.DefaultNamespace
-		}
+			errString := err.Error()
 
+			if errString[len(errString) - 1] == '0' {
```
blumamir, line 44: Can you please describe what this code checks in practice?

blumamir (commented):
Thanks again, great work.
Added one last question regarding the PR

cli/cmd/uninstall.go
HeetVekariya: This code checks for the condition if the namespace is not found

HeetVekariya:
@blumamir can you please check why one of the checks is failing, is the problem is from my side ?

cli/cmd/uninstall.go
blumamir: Is this because you count on the error message to contain the specific text: `"expected to get 1 namespace got 0"`?

I have a few concerns regarding this practice:
1. What if someone in the future decides to change this error text? then the feature will break and it will be tough for anyone to figure it out.
2. There is also the case that [some general `err` is returned](https://github.com/keyval-dev/odigos/blob/bdf9a967b8960b7e55e06e4d5daa4074290a6ffc/cli/cmd/resources/namespace.go#L29). For this error, we cannot know for sure if it ends with the character `0` or not. Thus, making the current logic less robust and inviting bugs.
3. The code line `if errString[len(errString) - 1] == '0' {` is not trivial to understand for future maintainers and readers of this repo. It does not describe what is checked here and why it is expected to work.

You can consider using [`errors.Is`](https://pkg.go.dev/errors#Is) to test if an error is of specific type. There are numerous blog posts in google and even some examples in this repo on how to use it.

HeetVekariya: Your concern is absolutely right,
I am thinking about something like:
```go

ns, err := resources.GetOdigosNamespace(client, ctx)
if err != nil {
  var ErrOdigosNotInstalled = errors.New("expected to get 1 namespace got 0")
  if errors.Is(err, ErrOdigosNotInstalled) {
    fmt.Println("\033[31mERROR\033[0m", Odigos is not currently installed in the cluster, so there is nothing to uninstall )
    } else {
    fmt.Println("\033[31mERROR\033[0m", err)
    }
  }
 os.Exit(1)
}
```

HeetVekariya: Correct me if i am wrong

HeetVekariya:
@blumamir I have made the changes as suggested,
- handle the condition (error generation) of no namespace in the ```namespace.go``` file instead of ```uninstall.go```

Testing:
1) no namespace:
image

2) other error:
image

blumamir (approved):

HeetVekariya:
- @blumamir i have made changes, but i haven't handled the error for condition if namespace != 1.
- If we need to handle that then we have to create another function for it same as ```IsErrNoOdigosNamespaceFound```

blumamir:
> * @blumamir i have made changes, but i haven't handled the error for condition if namespace != 1.
> * If we need to handle that then we have to create another function for it same as `IsErrNoOdigosNamespaceFound`

I think it's fine, as we do not expect it to happen in any normal operation. LGTM

HeetVekariya:
Thank you for merging

Issue #660, opened 2023-10-18
Print nice message when odigos install is called when odigos is already installed


blumamir:
Currently, when odigos is already installed, and one calls `odigos install`, the following output is ejected to console:
```
➜ odigos install
Installing Odigos version v0.1.75 in namespace odigos-system ...
Creating namespace odigos-system                  X
ERROR namespaces "odigos-system" already exists
```

I think it would be nicer to print some useful friendly message, like:
```
Odigos is already installed in namespace "odigos-system".
If you wish to re-install, run "odigos uninstall" first.
```

blumamir:
@HeetVekariya Since you are already working on the `odigos uninstall` fix, would you be interested in fixing this one as well?

blumamir:
To check if odigos is already installed, one can use [GetOdigosNamespace](https://github.com/keyval-dev/odigos/blob/609985701c758032f58a030b7ed35b765dfd61ce/cli/cmd/resources/namespace.go#L21) function to check if there is an existing `odigos` installation in the cluster.

HeetVekariya:
- @blumamir Sorry for the silence, as i was in college.
- I see that you have assigned me two issue, thank you for that.

HeetVekariya:
closed this as completed

PR #676, opened 2023-10-20
fix: exit if odigos is already installed

HeetVekariya:
# Fixes issue:
- Issue: #660 
- Update the error message for install of odigos if it is already installed

<br>

# Changes made:
- Updated the code such that if  ```Namepace``` is found then print the updated error message and exit

<br>

# Files changed:
- **install.go** (cli/cmd/install.go)

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/keyval-dev/odigos/cli/pkg/containers"
@@ -54,6 +55,12 @@ This command will install k8s components that will auto-instrument your applicat
 		cmd.Flags().StringSliceVar(&ignoredNamespaces, "ignore-namespace", DefaultIgnoredNamespaces, "--ignore-namespace foo logging")
 		fmt.Printf("Installing Odigos version %s in namespace %s ...\n", versionFlag, ns)
 
+		nsflag, err := resources.GetOdigosNamespace(client, ctx)
+		if  err == nil {
+			fmt.Printf("\033[31mERROR\033[0m Odigos is already installed in namespace \"%s\". If you wish to re-install, run \"odigos uninstall\" first.\n", nsflag)
+			os.Exit(1)
+		}
+		
 		isOdigosCloud := odigosCloudApiKeyFlag != ""
 		createKubeResourceWithLogging(ctx, fmt.Sprintf("Creating namespace %s", ns),
 			client, cmd, ns, createNamespace)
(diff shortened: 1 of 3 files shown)


HeetVekariya:
- @blumamir Test cases:

1) If already installed:
Screenshot from 2023-10-20 10-49-49

2)  if not installed:
Screenshot from 2023-10-20 10-50-40

- I found that the code goes in the infinite loop due to this code [here](https://github.com/keyval-dev/odigos/blob/3319cab6c2ea6034d3aa25ca073a1dd8602339a3/cli/cmd/install.go#L89):
```go
if !skipWait {
  l := log.Print("Waiting for Odigos pods to be ready ...")
  err := wait.PollImmediate(1*time.Second, 3*time.Minute, arePodsReady(ctx, client, ns))
  if err != nil {
     l.Error(err)
  }
  
  l.Success()
}
```

- This calls the function ```arePodsReady(ctx, client, ns```, whose function definition is:

```go
func arePodsReady(ctx context.Context, client *kube.Client, ns string) func() (bool, error) {
  return func() (bool, error) {
  pods, err := client.CoreV1().Pods(ns).List(ctx, metav1.ListOptions{})
  if err != nil {
	  return false, err
  }
  runningPods := 0
  for _, p := range pods.Items {
	  if p.Status.Phase == corev1.PodFailed {
		  return false, fmt.Errorf("pod %s failed", p.Name)
	  }
  
	  if p.Status.Phase == corev1.PodRunning {
		  runningPods++
	  }
  }
  
  return runningPods == len(pods.Items), nil
  }
}
```
- Here the for loop does not terminates.
- Is this problem already exists or my code has created this problem ?

cli/cmd/install.go
```diff
@@ -54,6 +55,12 @@ This command will install k8s components that will auto-instrument your applicat
 		cmd.Flags().StringSliceVar(&ignoredNamespaces, "ignore-namespace", DefaultIgnoredNamespaces, "--ignore-namespace foo logging")
 		fmt.Printf("Installing Odigos version %s in namespace %s ...\n", versionFlag, ns)
 
+		nsflag, err := resources.GetOdigosNamespace(client, ctx)
```
blumamir, line 58: nit: this ns has nothing to do with any flag.
```suggestion
		existingOdigosNs, err := resources.GetOdigosNamespace(client, ctx)
```

blumamir (approved):

cli/cmd/install.go
HeetVekariya: Changed name to `existingOdigosNs`

HeetVekariya:
@blumamir I have made the suggested changes

HeetVekariya:
Thank you for merging

PR #706, opened 2023-10-28
Created run_odigos_ui.sh

### LINKED ISSUE #703: Do not run Odigos UI if Odigos is not installed on the cluster
When invoking the command `odigos ui` we should test if Odigos is installed on the Kubernetes cluster.
If Odigos is not installed we should abort the command (not running the UI) and show an error message.
We can detect if Odigos is installed via a similar logic to what is implemented in `odigos version`.

Ritish134:
Added run_odigos_ui.sh file to test if odigos is installed or not . fixes #703


edeNFed:
Hi, @Ritish134 this is not exactly what I meant in the issue.
There is a Go file named `ui.go` which invokes Odigos user interface.
In this file, we need to make sure Odigos is installed on the current Kubernetes cluster before invoking the UI.
You can find an example of such check here: 
https://github.com/keyval-dev/odigos/blob/763606620b93f9b68c74ba1dcc56692b69012a6d/cli/cmd/install.go#L61-L68

Ritish134:
@edeNFed can you check once now

PR #723, opened 2023-10-31
feat: checking the presence of odigos in k8 cluster before starting odigos ui

iamshreeram:
Checking the presence of odigos in k8 cluster before starting odigos ui;

This is for addressing #703


cli/cmd/ui.go
```diff
@@ -29,43 +31,70 @@ var uiCmd = &cobra.Command{
...
+	if err != nil {
+		kube.PrintClientErrorAndExit(err)
+		return false
+	}
+	ctx := cmd.Context()
+
+	// check if odigos is installed
+	existingOdigosNs, err := resources.GetOdigosNamespace(client, ctx)
+	if err == nil {
+		fmt.Printf("Odigos is installed in namespace \"%s\".", existingOdigosNs)
```
edeNFed, line 89: I think this info is not important when running the `odigos ui` command, I guess we can remove this

iamshreeram:
@edeNFed , Sure, I have addressed the review comments in [a7a6d9e](https://github.com/keyval-dev/odigos/pull/723/commits/a7a6d9e3aa21487dc5729dea01a17b6a6d324bc9). thanks!

cli/cmd/ui.go
iamshreeram: Sure, I have addressed the review comments in [a7a6d9e](https://github.com/keyval-dev/odigos/pull/723/commits/a7a6d9e3aa21487dc5729dea01a17b6a6d324bc9). thanks!

cli/cmd/ui.go
```diff
@@ -29,43 +31,69 @@ var uiCmd = &cobra.Command{
...
+			process := exec.Command(binaryPath, flags...)
+			process.Stdout = os.Stdout
+			process.Stderr = os.Stderr
+			err = process.Run()
+			if err != nil {
+				fmt.Printf("Error starting UI: %v\n", err)
+				os.Exit(1)
+			}
+		} else {
+			fmt.Printf("\033[31mERROR\033[0m Unable to find Odigos in kubernetes cluster.\n")
```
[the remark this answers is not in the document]
iamshreeram: Yes, That is right. We are already printing the error message. Do you want me to remove the duplicate? 

Below error will be printed in console 

```
ERROR Cannot install/start UI. Failed to check if Odigos is already installed: Odigos installation is not found in any namespace
ERROR Unable to find Odigos in kubernetes cluster.  
```

cli/cmd/ui.go
```diff
@@ -63,9 +70,29 @@ var uiCmd = &cobra.Command{
...
+
+	// check if odigos is installed
+	_, err = resources.GetOdigosNamespace(client, ctx)
+	if err == nil {
+		return true
+	} else if !resources.IsErrNoOdigosNamespaceFound(err) {
+		fmt.Printf("\033[31mERROR\033[0m Cannot install/start UI. Failed to check if Odigos is already installed: %s\n", err)
+		return false
+	}
+	return true
```
blumamir, line 93: shouldn't we return `false` in this case?
this line will hit when the namespace is not found (which means odigos is not installed).

If this function returns `true` and the caller above checks for `if !checkOdigosInstallation(cmd)`, it means we will not `os.Exit` and will attempt to run the UI

PR #800, opened 2023-12-02
feat: checking the presence of odigos in k8 cluster before starting odigos ui - duplicate of #723

iamshreeram:
Same as #723 ;


cli/cmd/ui.go
```diff
@@ -29,6 +31,11 @@ var uiCmd = &cobra.Command{
 	Short: "Start the Odigos UI",
 	Long:  `Start the Odigos UI. This will start a web server that will serve the UI`,
 	Run: func(cmd *cobra.Command, args []string) {
+
+		if !checkOdigosInstallation(cmd) {
+			fmt.Printf("\033[31mERROR\033[0m Unable to find Odigos in kubernetes cluster. Aborting odigos ui.\n")
```
blumamir, line 36: It is possible that we print here after we already printed a similar message inside `checkOdigosInstallation`, right?
I suggest printing a message to the screen only once

blumamir (approved):
LGTM.
Thank you for the contribution 🙏

Added one minor suggestion

cli/cmd/ui.go
iamshreeram: @blumamir , 
I wanted to display the error trace when an exception occurs at [L#90](https://github.com/keyval-dev/odigos/blob/909432b4dffb2f2cc47ddcaa60a358f909eabd83/cli/cmd/ui.go#L90C2-L90C2). 

If you think this would be a duplicate, I can modify the error message to the following:
```
ERROR Cannot install/start UI. Failed to find Odigos in the Kubernetes cluster. Aborting Odigos UI.
```
By doing this, the stack trace of the error will be suppressed, and only the custom error message will be printed.

iamshreeram:
This PR is to address #703; 

@edeNFed/ @blumamir  , Anything else needed on this issue?

iamshreeram:
This PR is to address https://github.com/keyval-dev/odigos/issues/703;

@edeNFed / @blumamir , Anything else needed on this issue?

blumamir:
@iamshreeram  apologies for missing this PR on time.

The fix was merged at #886 so closing this one. Thank you again for working on this 🙏

PR #809, opened 2023-12-03
fix: allow non-standard port for Jaeger destination

edeNFed:


edeNFed:
@blumamir @RonFed please take a look

RonFed:
@edeNFed shouldn't the traces.json file be on git ?

blumamir (approved):

PR #841, opened 2023-12-18
new destination: coralogix

ryantanjunming:


autoscaler/controllers/gateway/config/coralogix.go
```diff
@@ -0,0 +1,59 @@
...
+
+type Coralogix struct{}
+
+func (c *Coralogix) DestType() common.DestinationType {
+	return common.CoralogixDestinationType
+}
+
+func (c *Coralogix) ModifyConfig(dest *odigosv1.Destination, currentConfig *commonconf.Config) {
+	if isTracingEnabled(dest) || isLoggingEnabled(dest) || isMetricsEnabled(dest) {
+		privateKey, exists := dest.Spec.Data[coralogixPrivateKey]
```
edeNFed, line 22: Is this correct? I think secrets should be consumed by `"${CORALOGIX_PRIVATE_KEY}" and regular fields should be consumed by `dest.Spec.Data["CORALOGIX_DOMAIN"]`.
The reason is that Odigos automatically injects secret fields as env variables not to expose them in the regular Destination object.

ryantanjunming: resolved this in later commits, thanks @edeNFed

edeNFed (approved):

edeNFed:
lgtm
thanks for contributing @ryantanjunming!

PR #886, opened 2024-01-16
pass namespace to odigos ui command

### LINKED ISSUE #662: "odigos ui" does not work with custom namespace
`odigos install` command can be run with custom namespace with the `-n` flag. 

`odigos ui` can then be used to launch a UI to control odigos, but it also needs to know the namespace to use. The UI binary already accept a `-n` flag [here](https://github.com/keyval-dev/odigos/blob/609985701c758032f58a030b7ed35b765dfd61ce/frontend/main.go#L55), but the cli does not propagate the flag to the UI when executing it, thus the UI opens with the wrong namespace and shows empty results. 

To fix this, we need to add the following changes to the UI command:
- find the namespace of the odigos installation [with this function](https://github.com/keyval-dev/odigos/blob/609985701c758032f58a030b7ed35b765dfd61ce/cli/cmd/resources/namespace.go#L21).
- Add the relevant `-n` flag [here](https://github.com/keyval-dev/odigos/blob/609985701c758032f58a030b7ed35b765dfd61ce/cli/cmd/ui.go#L58C5-L58C5) so the UI binary will be `exec.Command(...)` with the proper namespace of odigos.
- Verify the fix by setting up odigos installation [with the instructions here](https://docs.odigos.io/intro) but use `odigos install -n foo` and then run `odigos ui` and make sure we can see sources in odigos

tomasmota:
fixes #662 and includes the changes needed for #703


blumamir (approved):
Thanks! 🥇

tomasmota:
> Thanks! 🥇 

No problem, remember to close #703 if you see fit

Issue #541, opened 2023-10-03
CLI: odigos install return success even if DaemonSet pods are not up


edeNFed:
When installing odigos via the CLI by running `odigos install` the cli reports installation successful even if Odiglet (the deployed DaemonSet) is not running. Kubernetes may refuse to start the pods of Odiglet (for example: if there is a PodSecurityPolicy that blocks `hostPID`).

`arePodsReady` function should also check that for any DaemonSet in the `ns` namespace the number of running pods equal to the number of desired and is bigger than 1.
https://github.com/keyval-dev/odigos/blob/2f7d45c9a2deb1751bab3e120fe86a1ea1fdc7f5/cli/cmd/install.go#L86

blumamir:
> I can take it

Thanks @SzymonSt

PR #890, opened 2024-01-19
Deamonset pod check

tomasmota:
fixes: #541


tomasmota:
not sure why it thinks the changes to ui.go are new. Those are already in [keyval-dev:main](https://github.com/keyval-dev/odigos/tree/main)...

cli/cmd/install.go
```diff
@@ -110,6 +110,21 @@ This command will install k8s components that will auto-instrument your applicat
...
+			return false, err
+		}
+
+		for _, ds := range daemonSets.Items {
+			desiredPods := ds.Status.DesiredNumberScheduled
+			readyPods := ds.Status.NumberReady
+			if readyPods == 0 || readyPods != desiredPods {
+				return false, nil
+			}
+		}
```
blumamir, line 125: I think the daemonsets will ends up creating pods that will show up in the `client.CoreV1().Pods(ns).List(ctx, metav1.ListOptions{})` list below, no?

blumamir: right :) I missed that.
Thanks for the contribution 💯

blumamir (approved):

PR #939, opened 2024-02-01
feat: add --short flag to version command

WallysFerreira:


cli/cmd/version.go
```diff
@@ -61,4 +68,6 @@ func printOdigosClusterVersion(cmd *cobra.Command) {
 
 func init() {
 	rootCmd.AddCommand(versionCmd)
+
+	versionCmd.PersistentFlags().Bool("short", false, "prints only the CLI version")
```
edeNFed, line 72: Why use PersistentFlags? I see that in all the other commands we are using Flags

edeNFed (approved):

blumamir (approved):
Thanks @WallysFerreira

PR #941, opened 2024-02-02
feat(elasticsearch): remove hard-coded port 9200, validate URL input

### LINKED ISSUE #940: Remove Hardcoded Elasticsearch Port
**Is your feature request related to a problem? Please describe.**
The problem is, my elasticsearch is not running on port 9200, I want to have flexibility on the port configuration
https://github.com/keyval-dev/odigos/blob/206ef74eccff945a582571186365d2410d76347d/autoscaler/controllers/gateway/config/elasticsearch.go#L39
https://github.com/keyval-dev/odigos/blob/206ef74eccff945a582571186365d2410d76347d/autoscaler/controllers/gateway/config/elasticsearch.go#L59

**Describe the solution you'd like**
We can remove the hardcoded port, and also make sure that user put a valid URL by validating it using `url.Parse`

**Describe alternatives you've considered**
-

**Additional context**
I can work on this issue

clavinjune:
This PR enables a flexible ES endpoints by removing hard-coded ES port in the endpoints section. This PR also sanitize/validate whether user input a valid URL or not by utilizing `url.Parse`. I'm wondering whether we do need to have a test or not since `url.Parse` function already have their own [test](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/net/url/url_test.go)

Please advise if we need more test cases

This PR will close #940

### DIFF
--- a/autoscaler/controllers/gateway/config/elasticsearch.go
+++ b/autoscaler/controllers/gateway/config/elasticsearch.go
@@ -1,7 +1,8 @@
 package config
 
 import (
-	"fmt"
+	"context"
+	"net/url"
 
 	odigosv1 "github.com/keyval-dev/odigos/api/odigos/v1alpha1"
 	commonconf "github.com/keyval-dev/odigos/autoscaler/controllers/common"
@@ -15,6 +16,8 @@ const (
 	esLogsIndexKey      = "ES_LOGS_INDEX"
 )
 
+var _ Configer = (*Elasticsearch)(nil)
+
 type Elasticsearch struct{}
 
 func (e *Elasticsearch) DestType() common.DestinationType {
@@ -22,9 +25,15 @@ func (e *Elasticsearch) DestType() common.DestinationType {
 }
 
 func (e *Elasticsearch) ModifyConfig(dest *odigosv1.Destination, currentConfig *commonconf.Config) {
-	url, exists := dest.Spec.Data[elasticsearchUrlKey]
+	rawURL, exists := dest.Spec.Data[elasticsearchUrlKey]
 	if !exists {
 		log.Log.V(0).Info("ElasticSearch url not specified, gateway will not be configured for ElasticSearch")
+		return
+	}
+
+	parsedURL, err := e.SanitizeURL(context.TODO(), rawURL)
+	if err != nil {
+		log.Log.V(0).Error(err, "failed to sanitize URL", "elasticsearch-url", rawURL)
 		return
 	}
 
@@ -36,7 +45,7 @@ func (e *Elasticsearch) ModifyConfig(dest *odigosv1.Destination, currentConfig *
 		}
 
 		currentConfig.Exporters[esTraceExporterName] = commonconf.GenericMap{
-			"endpoints":    []string{fmt.Sprintf("%s:9200", url)},
+			"endpoints":    []string{parsedURL},
 			"traces_index": traceIndexVal,
 		}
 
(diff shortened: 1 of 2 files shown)


autoscaler/controllers/gateway/config/elasticsearch.go
```diff
@@ -36,7 +45,7 @@ func (e *Elasticsearch) ModifyConfig(dest *odigosv1.Destination, currentConfig *
 		}
 
 		currentConfig.Exporters[esTraceExporterName] = commonconf.GenericMap{
-			"endpoints":    []string{fmt.Sprintf("%s:9200", url)},
+			"endpoints":    []string{parsedURL},
```
blumamir, line 48: Since the current odigos version always appends the port, I assume that people configure the endpoint with no port (something like `https://elastic.example.com`). Thus, if we now stop appending the port they might break.

What do you think about checking if the port exists in the `SanitizeURL` function -
- It the user supplied a port in the URL use it.
- If there is no port in the URL, we can fallback to using `9200` which will keep existing users configuration working.

You can consider using the [urlHostContainsPort](https://github.com/keyval-dev/odigos/blob/206ef74eccff945a582571186365d2410d76347d/autoscaler/controllers/gateway/config/utils.go#L73) function to do that.

autoscaler/controllers/gateway/config/elasticsearch_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/keyval-dev/odigos/autoscaler/controllers/gateway/config"
+	"github.com/stretchr/testify/require"
+)
+
+func TestElasticsearch_SanitizeURL(t *testing.T) {
```
blumamir, line 11: I suggest adding a test for failure. e.g. that the `SanitizeURL` function returns an error for invalid URL endpoint

blumamir (commented):
Thank you for adding support for it and adding tests 🙏 Odigos should support any port and not only the default.

Add few minor comments

autoscaler/controllers/gateway/config/elasticsearch.go
```diff
@@ -15,19 +16,27 @@ const (
...
 
 func (e *Elasticsearch) ModifyConfig(dest *odigosv1.Destination, currentConfig *commonconf.Config) {
-	url, exists := dest.Spec.Data[elasticsearchUrlKey]
+	rawURL, exists := dest.Spec.Data[elasticsearchUrlKey]
 	if !exists {
 		log.Log.V(0).Info("ElasticSearch url not specified, gateway will not be configured for ElasticSearch")
 		return
 	}
 
+	parsedURL, err := e.SanitizeURL(context.TODO(), rawURL)
```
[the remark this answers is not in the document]
clavinjune: this is done ✅

autoscaler/controllers/gateway/config/elasticsearch.go
clavinjune: this is done ✅

autoscaler/controllers/gateway/config/elasticsearch_test.go
clavinjune: this is done ✅

blumamir (approved):
Looks great.

Thank you for adding support for this and addressing the comments 🙏

clavinjune:
please help to merge if all's good since I don't have any write access :bow:

PR #945, opened 2024-02-04
Ignore already existing namespace

edeNFed:


cli/cmd/install.go
```diff
@@ -146,6 +138,10 @@ func arePodsReady(ctx context.Context, client *kube.Client, ns string) func() (b
 
 func createNamespace(ctx context.Context, cmd *cobra.Command, client *kube.Client, ns string) error {
 	_, err := client.CoreV1().Namespaces().Create(ctx, resources.NewNamespace(ns), metav1.CreateOptions{})
+	if apierrors.IsAlreadyExists(err) {
```
blumamir, line 141: should we check that the namespace contains the `odigos.io/system-object` label?
If not, it will be broken downstream

cli/cmd/install.go
```diff
@@ -54,16 +56,6 @@ This command will install k8s components that will auto-instrument your applicat
 		ns := cmd.Flag("namespace").Value.String()
 		cmd.Flags().StringSliceVar(&ignoredNamespaces, "ignore-namespace", DefaultIgnoredNamespaces, "--ignore-namespace foo logging")
 
-		// check if odigos is already installed
-		existingOdigosNs, err := resources.GetOdigosNamespace(client, ctx)
-		if err == nil {
-			fmt.Printf("\033[31mERROR\033[0m Odigos is already installed in namespace \"%s\". If you wish to re-install, run \"odigos uninstall\" first.\n", existingOdigosNs)
```
blumamir, line 60: If the user will run this command and odigos itself is already installed (not just the namespace), it will now attempt to override the current installation. 

WDYT about checking for the presence of the `odigos-deployment` config map in the namespace to distinguish between namespace exists (with no active installation) vs namespace exists + odigos is installed thus we should not re-install?

blumamir (approved):

PR #964, opened 2024-02-10
add npm install before npm build

tomasmota:
if npm run build is run before npm install, next is not available:
image


alonkeyval (approved):
LGTM

Issue #904, opened 2024-01-23
honeycomb destination: add EU instance endpoint


blumamir:
Currently, odigos honeycomb destination [exports it's data to `api.honeycomb.io:443`](https://github.com/keyval-dev/odigos/blob/b8ffcfa3c4234ad1319a25dcc540629969b5dc44/autoscaler/controllers/gateway/config/honeycomb.go#L18).

According to [honeycomb docs](https://docs.honeycomb.io/getting-data-in/otel-collector/), there exists an EU endpoint at "api.eu1.honeycomb.io:443".

This issue is to add support for the user to configure the endpoint in the UI from a dropdown. As a reference, one can, for example, use the [CORALOGIX_DOMAIN](https://github.com/keyval-dev/odigos/blob/b8ffcfa3c4234ad1319a25dcc540629969b5dc44/destinations/data/coralogix.yaml#L23).

To test this, one would need to open a free account in honeycomb, and verify that the UI allows choosing the endpoint as a dropdown, and that data is shown in honeycomb APM correctly.

The [docs](https://github.com/keyval-dev/odigos/blob/main/docs/backends/honeycomb.mdx) would also need to be updated.

PR #965, opened 2024-02-10
add honeycomb eu endpoint

tomasmota:
fixes: #904


autoscaler/controllers/gateway/config/honeycomb.go
```diff
@@ -13,20 +23,31 @@ func (h *Honeycomb) DestType() common.DestinationType {
...
+		return
+	}
+
+	log.Log.V(0).Info("Honeycomb tracing is enabled, configuring Honeycomb destination")
+
+	domain, exists := dest.Spec.Data[honeycombDomain]
+	if !exists {
+		log.Log.V(0).Info("Honeycomb domain not specified, gateway will not be configured")
+		return
+	}
```
blumamir, line 37: existing users will have the destination already configured, but will lack this new key. In this case, in order not to break anyone, I recommend to fallback to `api.honeycomb.io`.

```suggestion
	if !exists {
	        domain = "api.honeycomb.io"
	}
```

destinations/data/honeycomb.yaml
```diff
@@ -20,3 +20,10 @@ spec:
       componentProps:
         type: password
       secret: true
+    - name: HONEYCOMB_DOMAIN
+      displayName: Domain
```
blumamir, line 24: Consider calling it `endpoint`.

I think that domain in this context refers to `honeycomb.io` and in this domain there are 2 endpoints

docs/backends/honeycomb.mdx
```diff
@@ -16,4 +16,9 @@ Make sure the API Key is permitted to **Send Events** to Honeycomb.
 
 ## Configuring Honeycomb Backend
 
-Insert the API Key from the previous step in the **API Key** field and name your backend in the **Name** field.
+- **Name**: Choose a name for this backend (e.g. Honeycomb).
+- **API Key**: Insert the API Key from the previous step in the **API Key** field.
+- **Domain** - Chose one of the domains in the dropdown, as documented in the [Honeycomb documentation](https://docs.honeycomb.io/getting-data-in/otel-collector/).
```
blumamir, line 21: We can also consider adding here that `api.honeycomb.io` is the US instance and `api.eu1.honeycomb.io` is for EU instance. This can a bit more convenient for users that are not sure what to choose and like to skip reading the long docs

blumamir (commented):
Thank you @tomasmota great work again 🥇 

Added a few minor comments

blumamir (approved):
Thanks for addressing everything so quickly 🙏

PR #969, opened 2024-02-11
feat(autoscaler): Sync Data-Collection Affinity, Tolerations, and NodeSelector Attributes with Odiglet Daemonset

### LINKED ISSUE #927: Add Affinity and Tolerations to odiglet and odigos-data-collection Daemonset
**Is your feature request related to a problem? Please describe.**
We need to be able to set Affinity and Tolerations attributes to odiglet and odigos-data-collection

**Describe the solution you'd like**
Need advice on this one

**Describe alternatives you've considered**
-

**Additional context**
-

clavinjune:
This PR enables data-collection Daemonset to have the same Affinity, Tolerations, and NodeSelector configuration with Odiglet Daemonset

this PR may close #927


clavinjune:
please help to review @blumamir :bow:

autoscaler/controllers/datacollection/daemonset.go
```diff
@@ -37,10 +38,27 @@ var (
 	}
 )
 
+func getOdigletDaemonsetPodSpec(ctx context.Context, c client.Client, namespace string) (*corev1.PodSpec, error) {
+	odigletDaemonset := &appsv1.DaemonSet{}
+
+	if err := c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: resources.OdigletDaemonSetName}, odigletDaemonset); err != nil {
```
blumamir, line 44: We are polling the entire cli module as a dependency just to get the `resources.OdigletDaemonSetName` string which is not likely to ever change.

I wonder if we can just define this const here and skip adding a new dependency. WDYT?

blumamir (commented):
Thanks again @clavinjune 

Great work as usual

added one nit comment about the new dependency of autoscaler on the cli. let me know what you think

autoscaler/controllers/datacollection/daemonset.go
clavinjune: done

clavinjune: a better approach would be create 1 module that will be shared over all the rest of modules in order to keep things standard I guess?

blumamir: > a better approach would be create 1 module that will be shared over all the rest of modules in order to keep things standard I guess?

Yes I was thinking the same thing. Perhaps we will do it one day and cleanup a bit

autoscaler/go.mod
```diff
@@ -1,12 +1,14 @@
...
-go 1.21
+go 1.21.0
+
 toolchain go1.22.0
 
 require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-logr/zapr v1.3.0
 	github.com/keyval-dev/odigos/api v0.0.0
+	github.com/keyval-dev/odigos/cli v0.0.0-20240212071645-f34b1db17b0c
```
blumamir, line 11: we should revert this change, with `git checkout` or `go mod tidy`

blumamir:
You are so fast 🤩 thanks again

can you please revert the change in `go.mod` and `go.sum` before merging?

clavinjune:
`go.mod` and `go/sum` changes reverted

blumamir (approved):

blumamir:
merged. I will release a verion of Odigos later today 🙏

PR #1011, opened 2024-02-25
[TASK-766] config cluster attr action

alonkeyval:


frontend/webapp/assets/icons/actions/index.ts
```diff
@@ -1,6 +1,6 @@
 import ClusterAttributes from './cluster-attr.svg';
 
 export const ACTION_ICONS = {
-  'cluster-attributes': ClusterAttributes,
+  'insert-cluster-attributes': ClusterAttributes,
```
blumamir, line 4: change to add cluster info

PR #1023, opened 2024-02-28
Update qryn docs

lmangani:
Updating the qryn documentation to use native examples


blumamir:
Thanks @lmangani 

Does these updated instruction deprecate the current `qryn` destination type?
image

lmangani:
@blumamir good question - not necessarily. I just noticed the docs on the web were outdated so this is a safe set of placeholder instructions for the OSS qryn version until we complete a full revalidation. I'll post further updates as we progress and/or update the integration fields to support more authentication methods.

docs/backends/qryn.mdx
```diff
@@ -1,11 +1,30 @@
...
+To send metrics to **qryn**, use the native Prometheus `remote_write` integration in the Odigos UI. 
+
+#### Configuration
+The URL should be in the format `https://<qryn-host>:<qryn-port>/api/prom/remote/write`.
+
+### Logs
+To send logs to **qryn**, use the native Loki integration in the Odigos UI. 
+
+#### Configuration
+The URL should be in the format `https://<qryn-host>:<qryn-port>/loki/api/v1/push`.
```
blumamir, line 26: do you think most users of the OSS qryn would use `https` or `http`?

blumamir (commented):
Thank you @lmangani , added few nit thoughts.

Odigos currently has a `qryn` destination which I think should be renamed to `qryn cloud` to reduce confusion, as the "our docs" link for this destination will lead to the current page which will not be consistent:

image

We can do it in the future when revisiting the qryn cloud integration.

lmangani:
@blumamir thank you so much for the points. I will rethink and refactor and open a new PR.
The focus will be on qryn OSS as that's our primary target and we will try to unify both since the difference is only in the authentication method (HTTP vs. Token Based) and Tenant IDs. So hopefully we can compact it all in one without confusion for end users, since all APIs are the same. We'll be back 😄

PR #1040, opened 2024-03-06
Fix: #543 Implemented useKeyDown hook to handle Enter key press

prathameshgaikwad:
Passed in the handleKeyPress function into the KeyvalButton component that is triggered when "Enter" key is pressed.


prathameshgaikwad:
Sorry for the repeated open and close actions.

I use an extension that sorts the imports in a file. This caused irrelevant lines to be changed in the code. So, I disabled the extension and made a fresh commit. Now the commit has changes only in the relevant sections.

alonkeyval:
The useKeyDown hooks already implement this functionality. Thanks @prathameshgaikwad

prathameshgaikwad:
You're right.

PR #1064, opened 2024-03-13
fix: respect `KUBECONFIG` env var

rauno56:
Many linux users force config location to `XDG_CONFIG_HOME` instead of home folder and have `KUBECONFIG` env variable set. Current install script did not respect that and failed if that was the case.

Fix checks the environment for mentioned env var and uses that if set. If `KUBECONFIG` is not set, the behavior is unchanged.


blumamir (approved):

rauno56:
I see there's another location that assumes the config is always in the default location. It's in the frontend code.

blumamir:
> I see there's another location that assumes the config is always in the default location. It's in the frontend code.

Yeah that makes sense, good catch

I think the `common` module might be a good place for it

blumamir:
perhaps [here](https://github.com/keyval-dev/odigos/blob/main/common/utils/env.go)?

rauno56:
Actually... now when I see that the used function comes from k8s client package, it makes me think the bug is there, but regardless... I'll fix it here first.

rauno56:
Made the move and reused it in frontend. Should I do anything about changed deps in modules as well?

blumamir (approved):

blumamir:
> Should I do anything about changed deps in modules as well?

Amazing! thanks

Do you mean about the `go.mod` and `go.sum` files?

rauno56:
> Do you mean about the go.mod and go.sum files?

That's right.

PR #1080, opened 2024-03-18
fix: configure filelog with only instrumented apps' logs

rauno56:
Approach we were taking was to select only relevant files from `/var/log/pods` for `filelog` receiver.

The files are specific to **pods**, but we don't want change the configuration every time there are changes on the pod level, so we wildcard the pod-specific part of the files out.

However, since there are easy collisions between statefulsets and deployments/daemonsets we have to glob the file pattern for each of those types **and** be specific per the kind of the instrumented app(deployment/statefulset/daemonset).

Next-best alternative to this would be to fork `filelog` altogether and make it very aware of those rules, keep the whole list of instrumented app in memory and watch the `/var/log/pods` for changes.

It's not pretty, but I think it's better than what we have now - totally broken logs pipeline. I'd continue with this general approach(happy to take suggestions about how to implement it better), and work on a long-term solution later. Latter probably comes with a considerably more work.


autoscaler/controllers/datacollection/testdata/logs_included.yaml
```diff
@@ -0,0 +1,109 @@
...
+    - azure
+receivers:
+  filelog:
+    exclude:
+    - /var/log/pods/kube-system_*/**/*
+    - /var/log/pods/odigos-system_*/**/*
+    include:
+    - /var/log/pods/default_test-deployment-[0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z]-[0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z]_*/*/*.log
+    - /var/log/pods/default_test-daemonset-[0-9a-z][0-9a-z][0-9a-z][0-9a-z][0-9a-z]_*/*/*.log
+    - /var/log/pods/other-namespace_test-statefulset-+([0-9])_*/*/*.log
```
rauno56, line 30: These are examples for resulting globs for each of the three kinds.

blumamir (approved):
LGTM 💯

rauno56:
Turns out that listing the instrumented applications is buggy. Need to fix it before the merge.

rauno56:
Turns out, the manager was configured to watch `odigos-system` namespace only, but the instrumented applications are in their deployed namespaces.

Discussed it with @blumamir, and decided to broaden permissions on `odigos-system-specific` resources. Alternative would've been to split the operators per scope - one for cluster scope, the other for custom resources in `odigos-system` - which would've been tons more work. Rollback on the roles is trivial if we want to do that in future.

blumamir (approved):
Thank you for this effort, excited to have this fixed

edeNFed (approved):

PR #1086, opened 2024-03-24
feat: add memory limiter to gateway collector

blumamir:


autoscaler/controllers/gateway/deployment.go
```diff
@@ -25,6 +26,20 @@ const (
...
+	requestMemory        = "500Mi"
+
+	// this configures the processor limit_mib, which is the hard limit in MiB, afterwhich garbage collection will be forced.
+	// as recommended by the processor docs, this is set to 50MiB less than the memory limit of the collector
+	memoryLimiterLimitMib = 450
+
+	// confgiures the processor spike_limit_mib. When memory usage exceeds the hard limit minus this amount,
+	// the processor will enter memory limited mode and will start refusing data.
+	// as recommended by the processor docs, this is set to 20% of the hard limit
+	memoryLimiterSpikeLimitMiB = 90
```
RonFed, line 38: Maybe worth considering defining these constants in a relative way like `0.2 * memoryLimiterLimitMib`. This way we can define a single default memory size of `requestMemory` and calculate all the others based on that value.
That way we could add code that retries with a smaller `requestMemory` in case of failure to deploy (and all the other sizes will adjust)

blumamir: Thanks @RonFed ,

I refactored the code so now it's configurable and relative. Can you please take a look again?

rauno56 (commented):
Do I understand correctly that the defaults are as percentages calculated, but when user wants to specify memory limits, they can only use absolute values?

The user cannot set the actual request limit on the gateway, because that will be overwritten by a value on `OdigosConfiguration`?

autoscaler/controllers/gateway/memory.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// the memory limiter hard limit is set as 50 MiB less than the memory request
+	memoryLimiterLimitMib := memoryRequestMib - defaultMemoryLimiterLimitDiffMib
+	if odigosConfig.Spec.CollectorGatewayMemoryLimiterLimitMib != 0 {
+		memoryLimiterLimitMib = odigosConfig.Spec.CollectorGatewayMemoryLimiterLimitMib
+	}
+
+	memoryLimiterSpikeLimitMiB := memoryLimiterLimitMib * defaultMemoryLimiterSpikePercentage / 100.0
+	if odigosConfig.Spec.CollectorGatewayMemoryLimiterSpikeLimitMiB != 0 {
+		memoryLimiterSpikeLimitMiB = odigosConfig.Spec.CollectorGatewayMemoryLimiterSpikeLimitMiB
```
RonFed, line 43: Maybe worth adding a check here that `memoryLimiterSpikeLimitMiB` < `memoryLimiterLimitMib`.
Why are the configuration options used as absolute values and our defaults relative? I think we should be consistent with the relative approach

autoscaler/controllers/gateway/deployment.go
```diff
@@ -28,9 +29,9 @@ const (
 )
 
 func syncDeployment(dests *odigosv1.DestinationList, gateway *odigosv1.CollectorsGroup, configData string,
-	ctx context.Context, c client.Client, scheme *runtime.Scheme, imagePullSecrets []string, odigosVersion string) (*appsv1.Deployment, error) {
+	ctx context.Context, c client.Client, scheme *runtime.Scheme, imagePullSecrets []string, odigosVersion string, memConfig *memoryConfigurations) (*appsv1.Deployment, error) {
 	logger := log.FromContext(ctx)
-	desiredDeployment, err := getDesiredDeployment(dests, configData, gateway, scheme, imagePullSecrets, odigosVersion)
+	desiredDeployment, err := getDesiredDeployment(dests, configData, gateway, scheme, imagePullSecrets, odigosVersion, memConfig)
```
RonFed, line 34: I'm not sure how often it might happen, but if our default is to request 500MB, and there is some resource limit preventing us from doing so, we should probably have some handling for it.

blumamir:
> Do I understand correctly that the defaults are as percentages calculated, but when user wants to specify memory limits, they can only use absolute values?

The relative defaults allow a user to set just one value (the memory request) and have all the other values automatically derived from it. 

If a user wants to set specific values, I thought it would be more straight forward to set the values as they would be written into the k8s manifest, instead of specifying relative amounts which requires more mental effort to calculate and understand. but I am open for change if you think it will be easier for users to use relative values @rauno56 @RonFed

rauno56:
I get where you are coming from, but I was also surprised a bit with the difference there.

Both of you have piles more experience in this specific case so I'll let you to thumb wrestle this out, but as a unsuspecting user-developer I'd expect defaults to have the same "power" as user-set values.

PR #1089, opened 2024-03-25
fix: fix autoscaler reconciliation loop with patching collector groups instead of updating them

rauno56:
- **feat: implement status patch and use it instead of updating the whole data collection**
- **feat: generate client for colletorsgroup crd**
- **style: capitalize error messages**

Not exactly idiomatic to this codebase, but works™. Also the `genclient` is not required, but I saw that annotation on every other CRD during my investigation, so added it there for symmetry. Maybe it'll be useful in the future.


rauno56:
We can take it to common package and use it in `gateway/root.go` as well.

blumamir (approved):

blumamir (approved):

Issue #1083, opened 2024-03-22
PYPATH is overridden during instrumentation


esara:
**Describe the bug**
We are using odigos to autoinstrument a python application, but the app is written/containerized with a predefined PYPATH and the autoinstrumention overrides it and breaks the application startup.

**To Reproduce**
Create a containerized python app with a PYPATH set in the container build.
```
PYTHONPATH=/pylib:/src
```
but odigos changes it to
```
PYTHONPATH=/var/odigos/python/opentelemetry/instrumentation/auto_instrumentation:/var/odigos/python
```
at which point the app does not find its own python libraries and fails

**Expected behavior**
It would great to append the PYPATH instead of overwriting it, similar to
https://github.com/keyval-dev/odigos/pull/1031 for node

RonFed:
Hi @esara, thanks for opening this issue.
We are aware of this problem, and we could add `PYTHONPATH` to the env vars we append  (similar to what was added to node in #1031), however, that approach is only valid for env vars that are declared as part of the k8s deployment. Unfortunately, we can't use that approach for env vars which are part of the container build, for that we will need a more involved solution.

esara:
Thanks @RonFed , I understand, env vars that are spelled out in the deployment/template spec can be reliably parsed

It is probably not as reliable, but it may be possible to parse out the runtime PYPATH of the running process (to be instrumented) as an option? something like this:
```
$ xargs --null --max-args=1 echo < /proc/1/environ|grep PYTHON
PYTHONPATH=/pylib:/src
```

PR #1100, opened 2024-04-01
Add env vars to runtime details

RonFed:
This PR addresses #1083 

The main goal of this PR is to avoid overriding existing environment variables with the values used for instrumentation purposes.

* Adding instrumentation-related environment variables to runtime inspection (PYTHONPATH and NODE_OPTIONS currently).
* In case such env var is defined both in the container build and in a k8s template the one in the template will take precedence.
* Support reverts once Odigos is uninstalled or instrumentation is removed - we should return to the original state. For that purpose, we added an annotation stating the original instrumentation-related env var values from the k8s manifest if there were any. This annotation is used to undo any env var changes we might have done, and the annotation itself is deleted as part of uninstall/un-instrument.
* Added unit tests for the modified `Revert` and `ApplyInstrumentationDevicesToPodTemplate` functions.

TODO: update helm charts to be consistent with this change. PR in https://github.com/keyval-dev/odigos-charts/pull/38


api/odigos/v1alpha1/instrumentedapplication_types.go
```diff
@@ -42,7 +54,7 @@ type OptionByContainer struct {
 
 // InstrumentedApplicationSpec defines the desired state of InstrumentedApplication
 type InstrumentedApplicationSpec struct {
-	Languages []common.LanguageByContainer `json:"languages,omitempty"`
+	RuntimeDetails []RuntimeDetailsByContainer `json:"languages,omitempty"`
```
blumamir, line 57:
```suggestion
	RuntimeDetails []RuntimeDetailsByContainer `json:"runtimeDetails,omitempty"`
```

cli/cmd/uninstall.go
```diff
@@ -196,31 +196,58 @@ func getWorkloadRolloutJsonPatch(obj client.Object, pts *v1.PodTemplateSpec) ([]
 		}
 	}
 
+	// read the original env vars (of the manifest) from the annotation
+	var origManifestEnv map[string]map[string]string
+	if obj.GetAnnotations() != nil {
+		manifestEnvAnnotation := obj.GetAnnotations()[consts.ManifestEnvOriginalValAnnotation]
+		json.Unmarshal([]byte(manifestEnvAnnotation), &origManifestEnv)
```
blumamir, line 203: handle err of `json.Unmarshal`?

common/envOverwrite/overwriter.go
```diff
@@ -2,33 +2,58 @@ package envOverwrite
...
+func ShouldRevert(envName string, value string) bool {
+	_, ok := EnvValues[envName]
+	if !ok {
+		// We don't care about this environment variable
+		return false
 	}
 
-	if currentVal == "" {
-		return val
+	if !strings.Contains(value, EnvValues[envName].Value) {
```
blumamir, line 35: nit: instead of `EnvValues[envName]` again you can use the value returned 4 lines above

odiglet/pkg/kube/runtime_details/shared.go
```diff
@@ -65,32 +64,37 @@ func runtimeInspection(pods []corev1.Pod) ([]common.LanguageByContainer, error)
...
-				log.Logger.V(0).Info("multiple languages detected for pod container processes, selecting first one", "processes", processes, "pod", pod.Name, "container", container.Name, "namespace", pod.Namespace)
-			}
-
-			resultsMap[container.Name] = common.LanguageByContainer{
-				ContainerName: container.Name,
-				Language:      detectionResults[0].Language,
+			for _, process := range processes {
+				lang, err := inspectors.DetectLanguage(process)
+				if err != nil {
+					log.Logger.V(0).Info("no supported language detected for container in pod", "process", process, "pod", pod.Name, "container", container.Name, "namespace", pod.Namespace)
```
blumamir, line 70:
```suggestion
					log.Logger.V(0).Info("no supported language detected for process in a container in pod", "process", process, "pod", pod.Name, "container", container.Name, "namespace", pod.Namespace)
```

This message will now be printed per process where it used to be printed per container

odiglet/pkg/kube/runtime_details/shared.go
```diff
@@ -65,32 +64,37 @@ func runtimeInspection(pods []corev1.Pod) ([]common.LanguageByContainer, error)
...
+					continue
+				}
+
+				// Convert map to slice for k8s format
+				envs := make([]odigosv1.EnvVar, 0, len(process.Envs))
```
blumamir, line 75: I think requesting a capacity of `len(process.Envs)` is a bit wasteful, as the process can have dozens of env variables but we will at most take just one currently

odiglet/pkg/kube/runtime_details/shared.go
```diff
@@ -65,32 +64,37 @@ func runtimeInspection(pods []corev1.Pod) ([]common.LanguageByContainer, error)
...
+				for envName, envValue := range process.Envs {
+					envs = append(envs, odigosv1.EnvVar{Name: envName, Value: envValue})
```
blumamir, line 77: shouldn't we filter out just those that we know we need?

procdiscovery/pkg/inspectors/langdetect.go
```diff
@@ -24,25 +26,16 @@ var inspectorsList = []inspector{
...
-			language, detected := i.Inspect(&p)
-			if detected {
-				detectionResult := LanguageDetectionResult{
-					Language: language,
-				}
-				result = append(result, detectionResult)
-				break
-			}
+// DetectLanguage returns the detected language for the process
+func DetectLanguage(process process.Details) (common.ProgrammingLanguage, error) {
```
blumamir, line 32: It's not really an error if the language is not detected. I suggest returning something else like `found` or simply `common.UnknownProgrammingLanguage` to communicate that no language is detected but no err occured

cli/cmd/uninstall.go
RonFed: Added a print if there is an error

common/envOverwrite/overwriter.go
RonFed: Fixed

procdiscovery/pkg/inspectors/langdetect.go
RonFed: Changed to return `common.UnknownProgrammingLanguage`

odiglet/pkg/kube/runtime_details/shared.go
```diff
@@ -65,32 +64,37 @@ func runtimeInspection(pods []corev1.Pod) ([]common.LanguageByContainer, error)
...
+				}
+
+				resultsMap[container.Name] = odigosv1.RuntimeDetailsByContainer{
+					ContainerName: container.Name,
+					Language:      lang,
+					EnvVars:       envs,
+				}
```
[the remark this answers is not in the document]
RonFed: Since we don't have CRD supporting multiple processes inside the same container, we don't have a good way of handling it currently. I think the the behavior before this PR is to take into account the first process and here it is the last process, both options are not ideal. Do you think we should consider the first process only?

odiglet/pkg/kube/runtime_details/shared.go
RonFed: The process.Envs do not contain all the env vars defined for the process, only those we defined in `envOverwrite.EnvValues`
The actual filtering is done in `getEnvVars` in `process.go`.
* Do you think it worth renaming Envs to somethin like InstrumentationEnvs ?
* Is it better for the filtering to happen in the lower function (as it is currrently) or here?

odiglet/pkg/kube/runtime_details/shared.go
RonFed: The filtering is done in the lower-level function `getEnvVars`

odiglet/pkg/kube/runtime_details/shared.go
RonFed: Updated to only take the first process into account, which should be the same behavior as we have today

cli/cmd/uninstall.go
```diff
@@ -196,31 +196,61 @@ func getWorkloadRolloutJsonPatch(obj client.Object, pts *v1.PodTemplateSpec) ([]
 		}
 	}
 
+	// read the original env vars (of the manifest) from the annotation
+	var origManifestEnv map[string]map[string]string
+	if obj.GetAnnotations() != nil {
+		manifestEnvAnnotation := obj.GetAnnotations()[consts.ManifestEnvOriginalValAnnotation]
+		err := json.Unmarshal([]byte(manifestEnvAnnotation), &origManifestEnv)
+		if err != nil {
+			fmt.Printf("Failed to unmarshal original env vars: %s\n", err)
```
blumamir, line 205: This message will be printed to the user console.
perhaps we can rephrase it to something more friendly to the users:

```suggestion
			fmt.Printf("Failed to rollout odigos environment variables for %s %s: %s\n", err, obj.GetName(), obj.GetKind)
```

Didn't test it, just as an example

cli/cmd/uninstall.go
```diff
@@ -196,31 +196,61 @@ func getWorkloadRolloutJsonPatch(obj client.Object, pts *v1.PodTemplateSpec) ([]
 		}
 	}
 
+	// read the original env vars (of the manifest) from the annotation
+	var origManifestEnv map[string]map[string]string
+	if obj.GetAnnotations() != nil {
+		manifestEnvAnnotation := obj.GetAnnotations()[consts.ManifestEnvOriginalValAnnotation]
+		err := json.Unmarshal([]byte(manifestEnvAnnotation), &origManifestEnv)
```
blumamir, line 203: will it err if the `manifestEnvAnnotation` is an empty string? which is the case if we did not change any env

odiglet/pkg/kube/runtime_details/shared.go
```diff
@@ -64,33 +64,42 @@ func runtimeInspection(pods []corev1.Pod) ([]common.LanguageByContainer, error)
...
+				// We only take the first process into account, when we'll support multiple processes we'll need to change this.
+				log.Logger.V(0).Info("multiple processes found in pod container, only taking the first one into account", "pod", pod.Name, "container", container.Name, "namespace", pod.Namespace)
+			}
+			process := processes[0]
 
-			detectionResults := inspectors.DetectLanguage(processes)
-			if len(detectionResults) == 0 {
-				log.Logger.V(0).Info("no supported language detected for container in pod", "processes", processes, "pod", pod.Name, "container", container.Name, "namespace", pod.Namespace)
+			lang := inspectors.DetectLanguage(process)
+			if lang == common.UnknownProgrammingLanguage {
```
blumamir, line 75: perhaps instead of returning unknown, we can return nil? I suspect this unknown can have some side effects

PR #1103, opened 2024-04-02
docs: be accurate on `--image-prefix` option

rauno56:
The docs used to say that the prefix is used for all images, but in fact it's actually only used for the collector. Updated the docs for now, but we can opt to change how it actually works.

Also... I personally feel that being able to specify the image with the version instead of providing a prefix is better interface for setting collector images.


blumamir:
I think it's not straightforward to use just any arbitrary collector image:

- to achieve the mapping of device id to service name and k8s attributes, the autoscaler includes the `odigosresourcename` processor in the datacollection collector configuration which is only found in the odigos collector distribution.
- there are custom exporters in odigos for azure blob and google cloud storage which can only work when used with the odigos collector. It is only relevant if the user actually uses them. 
- the collector components are still not stable, and introduces breaking changes here and there. the autoscaler destinations configuration logic is aligned to a specific version of the collector which we need to review and update before bumping the version.

docs/cli/odigos_install.mdx
```diff
@@ -40,7 +40,7 @@ odigos install [flags]
         --odiglet-image             odiglet container image name
         --instrumentor-image        instrumentor container image name
         --autoscaler-image          autoscaler container image name
-        --image-prefix              prefix for all container images. used when your cluster doesn't have access to docker hub
+        --image-prefix              prefix for collector images. Used when your cluster doesn't have access to docker hub or when a different base collector image is needed.
```
blumamir, line 43:
```suggestion
        --image-prefix              prefix for collector images. Used when your cluster doesn't have access to docker hub. 
```

I think it's not feasible at the moment to support other base images, due to the coupling of the autoscaler with the custom processors/exporters and the version of components

rauno56:
> I think it's not straightforward to use just any arbitrary collector image

I didn't try to suggest that actually. So I agree with all the points you made.

Perhaps I should just remove the second sentence then. I put it there more to reflect the reality than to suggest usages.

blumamir (approved):

PR #1105, opened 2024-04-03
feat: update Java Agent to v2.3.0

rauno56:
- **chore: update Java Agent to the latest 1.x version**
- **chore: update Java Agent to 2.0.0**

Currently based on top of 1.x upgrade. Would rather have 3 separate commits in the history after the upgrade:

1. `1-latest`
2. `2.0.0`
3. `2-latest`


rauno56:
JS instrumentation still has bunch of deprecated attributes in use. I'll try to upgrade that as well to match everything on that level.

RonFed (approved):
🎉

.github/workflows/e2e.yaml
```diff
@@ -92,6 +92,7 @@ jobs:
           echo "Sleeping for 10 seconds to allow traces to be collected"
           sleep 10
           kubectl cp -c filecp traces/test-opentelemetry-collector-0:tmp/trace.json ./.github/workflows/e2e/bats/traces-orig.json
+          cat ./.github/workflows/e2e/bats/traces-orig.json
```
blumamir, line 95: is it intentional?

blumamir:
> JS instrumentation still has bunch of deprecated attributes in use. I'll try to upgrade that as well to match everything on that level.

Great thank you 🙏

.github/workflows/e2e.yaml
rauno56: Yes. Otherwise, it's hard to get the traces out to examine and fix tests without running the whole CI job step-by-step locally(which I did, but figured to add it there for the next time).

rauno56:
I turned off OTel logs exporter since we read them from disk(/stdout). That matches the other agents' behavior. @RonFed

.github/workflows/e2e.yaml
RonFed: @rauno56 FYI you can use the `make e2e-test` to run the test locally.

PR #1126, opened 2024-04-15
feat(procdiscovery): make GetPidDetails public

blumamir:


rauno56:
They broke the helm chart.
Fix here: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1143

Until then you can add `ref: e0c6b9c` to the `action/checkout` to check out the last working commit.

RonFed (approved):

PR #1131, opened 2024-04-21
install odigos in openshift

esara:
address https://github.com/keyval-dev/odigos/issues/1128

### DIFF
--- a/cli/cmd/resources/odiglet.go
+++ b/cli/cmd/resources/odiglet.go
@@ -120,6 +120,15 @@ func NewOdigletClusterRole(psp bool) *rbacv1.ClusterRole {
 			{
 				Verbs: []string{
 					"get",
+				},
+				APIGroups: []string{"apps"},
+				Resources: []string{
+					"deployments/finalizers",
+				},
+			},
+			{
+				Verbs: []string{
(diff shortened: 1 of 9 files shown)


cli/cmd/resources/odiglet.go
```diff
@@ -411,6 +445,45 @@ func NewOdigletDaemonSet(ns string, version string, imagePrefix string, imageNam
...
 }
+
+// used to inject the host volumes into odigos components for selinux update
+func selinuxHostVolumes() []corev1.Volume {
+	return []corev1.Volume{
+		{
+			Name: "host",
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: "/",
```
edeNFed, line 456: Can we mount `/sbin` instead of `/`?

esara: that is what I tried first, but semanage is a python script and python is installed outside of /sbin (and I am sure if we want to manage all the python dependencies in the odiglet container), other then making this readonly, I am not too happy about mounting it... what else can you think of?

edeNFed (approved):

PR #1176, opened 2024-05-07
fix: fix file paths for odigos-otelcol package

rauno56:


collector/distribution/odigos-otelcol/postinstall.sh
```diff
@@ -0,0 +1,4 @@
+if command -v systemctl >/dev/null 2>&1; then
+    systemctl enable odigos-otelcol.service
+    systemctl start odigos-otelcol.service
```
blumamir, line 3: since we enable it by default, I suggest that the default config will have sampling so that we don't blow up the disk with otelcol logging in case the config is not changed right away:

https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/loggingexporter

blumamir (approved):

PR #1179, opened 2024-05-09
fix: race condition in instrumented application creation and deletion

blumamir:


odiglet/pkg/kube/runtime_details/workload_enabled_predicated.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	if e.ObjectNew == nil {
+		return false
+	}
+
+	// only run runtime inspection if the workload was not instrumented before
+	// and now it is.
+	oldEnabled := isInstrumentationEnabled(e.ObjectOld)
+	newEnabled := isInstrumentationEnabled(e.ObjectNew)
+	becameEnabled := !oldEnabled && newEnabled
```
RonFed, line 40: Maybe we can add here
`if becameEnabled {
       return true
}`
and avoid checking it in the following special cases for deployments, statfulsets and daemonsets

odiglet/pkg/kube/runtime_details/workload_enabled_predicated.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// special case for deployments, where we also want to pass events if the
+	// deployment pods become available so language detection can run
+	oldDeployment, oldOk := e.ObjectOld.(*appsv1.Deployment)
+	newDeployment, newOk := e.ObjectNew.(*appsv1.Deployment)
```
RonFed, line 45: Just a suggestion:
* get kind of old object
* get kind of new object
* if different kinds return false
* else: switch case on kind
  *for each case cast to appropriate typ
check the condition after casting

instrumentor/controllers/deleteinstrumentedapplication/instrumentedapplication_controller.go
```diff
@@ -0,0 +1,98 @@
...
+// DeploymentReconciler reconciles a Deployment object
+type InstrumentedApplicationReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+}
+
+func (r *InstrumentedApplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	var runtimeDetails odigosv1.InstrumentedApplication
```
RonFed, line 67: Why to call it runtimeDetails if it is of type InstrumentedApplication?

RonFed (approved):

PR #1195, opened 2024-05-13
refactor: split config creation to common module

rauno56:
Believe it or not, it's the least amount of changes I could make to have a meaningful impact.

1. Replace configuration creation function parameters with something generic(`ExporterConfigurer` and `ProcessorConfiguerer` interfaces).
2. Change all the required down-stream locations that used to expect k8s objects.
3. Move config-specific stuff under common module and fix imports.

This does not include anything towards using it in external modules yet.

- **fix: fix file paths for odigos-otelcol package**
- **refactor: abstract Destination away from exporter configurators**
- **refactor: filter processors before passing them to k8s agnostic functions**
- **test: add processor to data collector config test**
- **refactor: abstract Processor away from processor configuration interfaces**
- **refactor: move common config primitives under common module**
- **fix: fix typoed var names**
- **fix: remove unused imports in processors.go**
- **fix: fix tests for elasticsearch**


common/config/processor.go
```diff
@@ -0,0 +1,66 @@
...
+	return false
+}
+
+func GetCrdProcessorsConfigMap(processors []ProcessorConfigurer) (cfg GenericMap, tracesProcessors []string, metricsProcessors []string, logsProcessors []string) {
+	cfg = GenericMap{}
+	for _, processor := range processors {
+		fmt.Printf("processors: %+v\n", processor)
+		processorKey := fmt.Sprintf("%s/%s", processor.GetType(), processor.GetName())
+		processorsConfig, err := processor.GetConfig()
+		fmt.Printf("err: %+v\n", err)
```
blumamir, line 43: also here forgotten print to console

common/config/processor.go
```diff
@@ -0,0 +1,66 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func GetCrdProcessorsConfigMap(processors []ProcessorConfigurer) (cfg GenericMap, tracesProcessors []string, metricsProcessors []string, logsProcessors []string) {
+	cfg = GenericMap{}
+	for _, processor := range processors {
+		fmt.Printf("processors: %+v\n", processor)
```
blumamir, line 40: remove debug print

common/config/processor.go
```diff
@@ -0,0 +1,66 @@
+package config
+
+import (
+	"fmt"
+
+	"github.com/odigos-io/odigos/common"
+	"sigs.k8s.io/controller-runtime/pkg/log"
```
blumamir, line 7: I am not sure how this logger will work outside of a controller runtime k8s environment. we should consider returning the error/s back to the caller to log or write into the CRD as status conditions.

common/config/root.go
```diff
@@ -135,15 +133,15 @@ func isSignalExists(dest *odigosv1.Destination, signal common.ObservabilitySigna
 	return false
 }
 
-func isTracingEnabled(dest *odigosv1.Destination) bool {
+func isTracingEnabled(dest ExporterConfigurer) bool {
```
blumamir, line 136: nit: now that we have a common interface `SignalSpecific` we can potentially use it and share this code between destinations and processors.

blumamir (commented):
Thank you @rauno56 , added few nits 🙏

autoscaler/controllers/common/processors.go
```diff
@@ -1,45 +1,12 @@
...
-func IsProcessorLogsEnabled(processor *odigosv1.Processor) bool {
-	for _, signal := range processor.Spec.Signals {
-		if signal == common.LogsObservabilitySignal {
-			return true
-		}
-	}
-	return false
-}
-
 func FilterAndSortProcessorsByOrderHint(processors *odigosv1.ProcessorList, collectorRole odigosv1.CollectorsGroupRole) []*odigosv1.Processor {
```
blumamir, line 9: I am thinking, should this function be moved to common as well? we might need to reuse this functionality with other platforms - filtering out disabled processors, sorting by order hint and choosing the right role to use. 

We can also address it later on if we run into the situation

rauno56: I am not sure. We could go that route as well, I guess, but I don't necessary think we need it nor do I think it's better experience:

1. Order hints are useful only if we cannot control the order,
2. In a local file one could argue that removing items from yaml or commenting them out is simpler than having an additional attribute,
3. we have one type of collector for now so filtering by role is not necessary.

So I think the question end up being whether we want to maximally mirror every aspect of the environments or not. The config files will be different from k8s resources anyway so I've been working on an assumption that it didn't matter. What do you think?

common/config/processor.go
rauno56: Check 196c64ff2b0efb3d9c425abccdb4269e8a22d149.

common/config/root.go
rauno56: I actually had it in the next dev branch. Pushed it here.

autoscaler/controllers/common/processors.go
blumamir: > 1. Order hints are useful only if we cannot control the order,

Not sure I understood. how will the user hint the order of the processors otherwise?

> * 
> * In a local file one could argue that removing items from yaml or commenting them out is simpler than having an additional attribute,

While yaml files are one way to configure the destinations, there can be other ways like via odigos cloud, where the user has no control over files, and the changes are applied via UI. we can fork to use different scheme and build custom configurations per use-case, but it might be hard to maintain and implement vs aligning everywhere to use the same scheme.

Anyway, we can move forward and address this issue if we run into it.

api/odigos/v1alpha1/destination_types.go
```diff
@@ -62,3 +62,17 @@ type DestinationList struct {
 func init() {
 	SchemeBuilder.Register(&Destination{}, &DestinationList{})
 }
+
+/* Implement common.ExporterConfigurer */
+func (dest Destination) GetName() string {
+	return dest.Name
+}
```
blumamir, line 69: Destinations has 2 names: one for the CRD object itself in k8s (which is what we use here) and one free-style name that the user can fill which is found in `Spec`. 

I suggest stating which name is used so that it's not confusing in the future.

blumamir (approved):

autoscaler/controllers/common/processors.go
rauno56: > Not sure I understood. how will the user hint the order of the processors otherwise?

The order is clear whenever we're working with arrays. But I agree, that when the items are separated from one another(like in k8s or perhaps a DB schema where every processor is it's own row), we do need a way to order and change the order.

> there can be other ways like via odigos cloud, where the user has no control over files

I see what you mean now. My default way to design this would've been different from the ground, I guess.

I'll leave it as it is for now. But thanks for very valuable feedback. :pray:

PR #1203, opened 2024-05-16
Add condition for instrumentedApplication

RonFed:


blumamir (approved):
don't forget adding it also to odigos-charts

RonFed:
> don't forget adding it also to odigos-charts

Done in https://github.com/odigos-io/odigos-charts/pull/56

PR #1204, opened 2024-05-16
feature: add Causely destination

br-kearns:
Adds support for Causely as destination with Traces/Metrics options.


esara (approved):

autoscaler/controllers/gateway/config/causely.go
```diff
@@ -0,0 +1,44 @@
+package config
```
blumamir, line 1: #1195  moved the destination configurators into `common` from `autoscaler`.

To fix the compilation issues, this file should be moved there as well

autoscaler/controllers/gateway/config/causely.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+type Causely struct{}
+
+func (e *Causely) DestType() common.DestinationType {
+	return common.CauselyDestinationType
+}
+
+func (e *Causely) ModifyConfig(dest *odigosv1.Destination, currentConfig *commonconf.Config) {
+	if url, exists := dest.Spec.Data[causelyUrl]; exists {
```
blumamir, line 20: as of #1168 , each configurator now returns an error if it spots anything wrong in the config. If the required `causelyUrl` is `!exists`, we should return `err` which is written into the CRD and will soon be shown in the UI.

You can check other destinations to copy how it's done there.

docs/quickstart/next-steps.mdx
```diff
@@ -36,10 +36,11 @@ Select the relevant backend for your use case below to connect it to Odigos.
   <Card title="OpenTelemetry" href="/backends/opentelemetry" />
   <Card title="OpsVerse" href="/backends/opsverse" />
   <Card title="Prometheus" href="/backends/prometheus" />
-  <Card title="Qryn" href="/backends/qryn" />
+  <Card title="Qryn" href="/backends/qryn" />∂
```
blumamir, line 39:
```suggestion
  <Card title="Qryn" href="/backends/qryn" />
```

docs/backends/causely.mdx
```diff
@@ -0,0 +1,7 @@
+---
+title: "Causely"
+---
+
+## Configuring the Causely Backend
+
+In order to add a new Causely backend, input the Causely Mediator endpoint URL into Odigos UI.
```
blumamir, line 7: Sometimes users get confused and are unsure if they need to add protocol `http://`, URL path, port etc. Since this endpoint is `grpc`, I guess the endpoint should be in the format `host:port` which can be nice to state here.
Is the `Mediator endpoint URL` something that users just need to copy and paste from somewhere in causely UI? or do they need to construct it themselves?

blumamir (commented):
Thanks @br-kearns ,
Very excited to have Causely as a first-class destination in odigos.

Added a few minor comments. Please feel free to write me if anything is not clear 🙏

blumamir (approved):
Awesome @br-kearns that looks so neat

Happy to have you as an official destination 🎉

PR #1234, opened 2024-05-27
feat: rename GetName() to GetID() and add debug, uptrace configurers

rauno56:
Sorry about the huge rename, but I made a mistake of mirroring the previous method name instead of thinking critical about what that did. `GetID` is much better name because all that has to do is to provide a reference for a destination if two of the same types are used. On the VM Agent side we use the array index for that. On k8s we create a random hash.

Other than that, this PR
- adds two new destination types: debug and Uptrace
- skips failing destinations instead of erroring out of compiling the Collector config altogether.
- emits warnings when no signals are specified for a destination.

Commits:

- **feat: add debug destination type**
- **feat: add Uptrace destination type**
- **feat: continue compiling collector config if one of the exporters is missing**
- **docs: add a comment about assumptions made for GetName**
- **feat: make debug and uptrace configers available**
- **feat: warn when no signals configured**
- **reafactor: rename GetName to GetID**


common/config/uptrace.go
```diff
@@ -0,0 +1,66 @@
...
+	if !exists {
+		endpoint = "https://otlp.uptrace.dev:4317"
+	}
+
+	exporterName := "otlp/uptrace-" + dest.GetID()
+
+	currentConfig.Exporters[exporterName] = GenericMap{
+		"endpoint": endpoint,
+		"tls": GenericMap{
+			"insecure": false,
```
blumamir, line 37: If it's self hosted, wouldn't it most likely be used with insecure connection?

blumamir (commented):
Added one note about the uptrace destination being used as tls only.

we need to add the new destination also to the docs:

- https://github.com/odigos-io/odigos/tree/main/docs/backends
- https://github.com/odigos-io/odigos/blob/4dfe879fee1dceeca704de0e5815e4298fafa8d1/docs/mint.json#L156
- https://github.com/odigos-io/odigos/blob/4dfe879fee1dceeca704de0e5815e4298fafa8d1/docs/quickstart/next-steps.mdx?plain=1#L45

common/config/uptrace.go
rauno56: I think Uptrace is mostly used as a service. That's what I've always done at least.

I will add a check for endpoint - if it's http, will make this insecure.

blumamir (approved):

PR #1235, opened 2024-05-27
docs: add DNF instructions to VM Agent docs

rauno56:
Adds DNF instructions for RHEL and other distros that use DNF by default.
Organizes some of the quickstart to use the main doc.
Fixes some inconsistencies with capitalization("systemd" is always lowercase, etc) and typos.


blumamir (approved):
Looks so much better. Thanks 🥇

PR #1242, opened 2024-05-29
docs: update add-new-dest.mdx

rauno56:
I noticed that add-new-dest config page was severely out of date. It's now much easier to add new destinations.


blumamir (approved):
Thanks! 💯

PR #1246, opened 2024-05-31
Instrumentation Instance

RonFed:
* Adding new CRD `InstrumentationInstance` which will hold status for each active SDK. currently only created for eBPF SDKs.
* Support notifications for errors in the new CR.
* Add condition for healthy instances for specific source in the source endpoint.


alonkeyval:
The current design looks good, but my main concern is a potential flood of alerts on the frontend if there are many instrumented applications. I suggest adding batch-processing logic to mitigate this.

RonFed:
> The current design looks good, but my main concern is a potential flood of alerts on the frontend if there are many instrumented applications. I suggest adding batch-processing logic to mitigate this.

Created a common batcher and modified the instrumented app and instrumentation instance to use it.

blumamir (approved):

alonkeyval (approved):

PR #1250, opened 2024-06-03
Autoscaler v2.0

edeNFed:


autoscaler/main.go
```diff
@@ -164,6 +167,22 @@ func main() {
 		os.Exit(1)
 	}
 
+	ctx := ctrl.SetupSignalHandler()
+	gatewayAutoscaler := collectormetrics.NewAutoscaler(mgr.GetClient(),
+		collectormetrics.WithCollectorsGroup(odigosv1.CollectorsGroupRoleClusterGateway),
+		collectormetrics.WithInterval(15*time.Second),
```
RonFed, line 173: The 15 seconds interval should be a const or probably something we to be able to set in a config map, right?

autoscaler/collectormetrics/algorithm.go
```diff
@@ -0,0 +1,124 @@
...
+		if e.isExporterTooSlow(queueSize, queueCapacity) {
+			result++
+		}
+	}
+
+	return result
+}
+
+func (e *exporterQueueAndBatchQueue) isExporterTooSlow(queueSize float64, queueCapacity float64) bool {
+	return queueSize/queueCapacity >= 0.7
```
RonFed, line 85: The 0.7 should also be defined somewhere and we might want to be able to control it in runtime like the interval?

autoscaler/collectormetrics/run.go
```diff
@@ -0,0 +1,181 @@
...
+	"github.com/odigos-io/odigos/autoscaler/controllers/gateway"
+	"github.com/odigos-io/odigos/k8sutils/pkg/env"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	odigosv1 "github.com/odigos-io/odigos/api/odigos/v1alpha1"
+	appsv1 "k8s.io/api/apps/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	dto "github.com/prometheus/client_model/go"
```
RonFed, line 20: maybe change `dto` to something nicer like `prometheus`?

autoscaler/collectormetrics/autoscaler.go
```diff
@@ -0,0 +1,147 @@
...
+		o.algorithm = algorithm
+	}
+}
+
+type Autoscaler struct {
+	kubeClient    client.Client
+	options       AutoscalerOptions
+	ticker        *time.Ticker
+	notifications chan Notification
+	podIPs        map[string]string
```
RonFed, line 61: Why this map needs to have the pod names as values? Can we just use a `map[string]struct{}` which will only save the IPs?

autoscaler/collectormetrics/autoscaler.go
```diff
@@ -0,0 +1,147 @@
...
+	return func(o *AutoscalerOptions) {
+		o.algorithm = algorithm
+	}
+}
+
+type Autoscaler struct {
+	kubeClient    client.Client
+	options       AutoscalerOptions
+	ticker        *time.Ticker
+	notifications chan Notification
```
RonFed, line 60: This should probably be a buffered channel to avoid blocking the reconcile.

autoscaler/collectormetrics/run.go
```diff
@@ -0,0 +1,181 @@
...
+		a.podIPs[notification.PodName] = notification.IP
+	} else if notification.Reason == IPRemoved {
+		delete(a.podIPs, notification.PodName)
+	}
+}
+
+func (a *Autoscaler) getCollectorsMetrics(ctx context.Context) []MetricFetchResult {
+	logger := log.FromContext(ctx)
+	results := make(chan MetricFetchResult, len(a.podIPs))
+	timeoutCtx, cancel := context.WithTimeout(ctx, 1*time.Second)
```
RonFed, line 130: Maybe the timeout should be a function of the interval duration. Intuitively it should be a small fraction of the interval since we want our measurement to represent the interval and not take a significant time of  what it measures if that makes sense.

autoscaler/collectormetrics/autoscaler.go
edeNFed: I am using the pod name just in case of an error.
This allows us to print an informative error message that contains the pod name and not just the IP which will be harder to debug.

PR #1263, opened 2024-06-10
Bugfix disable sending metrics without destination

tamirdavid1:


odiglet/pkg/instrumentation/plugin.go
```diff
@@ -70,7 +75,26 @@ func (p *plugin) Allocate(ctx context.Context, request *v1beta1.AllocateRequest)
 		}
 
 		deviceId := req.DevicesIDs[0]
-		res.ContainerResponses = append(res.ContainerResponses, p.LangSpecificFunc(deviceId))
+
+		cfg, err := rest.InClusterConfig()
```
edeNFed, line 79: Maybe worth initializing kubernetes client in `NewPlugin` instead of in `Allocate` which is called every time we do instrumentation

edeNFed (approved):

edeNFed (approved):

PR #1265, opened 2024-06-12
Improve E2E tests

edeNFed:


tests/e2e/multi-apps/00-assert.yaml
RonFed, line 1: I think we need to change the names of these files to better specify what they are asserting. like `assert_tempo_dest.yaml`

tests/e2e/multi-apps/tracesql/resource-attributes.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: TraceTest
+description: |
+  This test check the following resource attributes:
+  A. odigos.version attribute exists on all spans and has the correct value
+  B. Kubernetes attributes are correctly set on all spans
+  At the time of writing this test, TraceQL api does not support not equal to nil so we use regex instead.
+query: |
+    { resource.odigos.version != "e2e-test" ||
+      resource.k8s.deployment.name !~ ".*" ||
+      resource.k8s.node.name !~ ".*" ||
```
RonFed, line 11: We know the exact values for node name in kind and deployment name of the test apps, so we can compare to those expected values. right?

edeNFed:
@RonFed @blumamir after 103 commits I think this is finally ready for review

tests/common/wait_for_dest.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+# Ensure the script fails if any command fails
+set -e
+
+# function to verify tempo is ready
+function wait_for_ready() {
```
RonFed, line 7: Add comment about why this is needed

blumamir (commented):
Still reviewing, trying to run locally.
Few nit comments

tests/common/wait_for_dest.sh
```diff
@@ -0,0 +1,24 @@
...
+
+# function to verify tempo is ready
+# This is needed due to bug in tempo - It reports Ready before it is actually ready
+# So we manually hit the health check endpoint to verify it is ready
+function wait_for_ready() {
+  local dest_namespace="traces"
+  local dest_service="e2e-tests-tempo"
+  local dest_port="tempo-prom-metrics"
+  local response=$(kubectl get --raw /api/v1/namespaces/$dest_namespace/services/$dest_service:$dest_port/proxy/ready)
+  if [ "$response" != "ready" ]; then
```
yodigos, line 14: Is there some kind of timeout if it's not ready ever?

tests/e2e/helm-chart/chainsaw-test.yaml
```diff
@@ -0,0 +1,128 @@
...
+    try:
+      - script:
+          timeout: 60s
+          content: ../../common/wait_for_dest.sh
+  - name: Install Odigos
+    try:
+    - script:
+        content: |
+          git clone https://github.com/odigos-io/odigos-charts.git /tmp/odigos-charts
+          # Retry 3 times to avoid flakiness
```
yodigos, line 27: Just change the comment  from 3 retries to 60 sec timout

tests/e2e/helm-chart/chainsaw-test.yaml
```diff
@@ -0,0 +1,128 @@
...
+            exit 1
+          fi
+    - assert:
+        file: assert-odigos-installed.yaml
+  - name: Install Demo App
+    try:
+    - script:
+        timeout: 100s
+        content: |
+          docker pull keyval/odigos-demo-inventory:v0.1
```
yodigos, line 51: I think this sure work:
`kubectl apply -f https://raw.githubusercontent.com/odigos-io/simple-demo/main/kubernetes/deployment.yaml`
Then we don't need to remember to update versions.

tests/e2e/helm-chart/chainsaw-test.yaml
```diff
@@ -0,0 +1,128 @@
...
+    - assert:
+        file: assert-instrumented-and-pipeline.yaml
+  - name: Generate Traffic
+    try:
+      - script:
+          timeout: 60s
+          content: |
+            while true; do
+              # Apply the job
+              kubectl apply -f 05-generate-traffic.yaml
```
yodigos, line 84: Why do we flush traces?
And why todo this process multiple times? 
Don't we want to generate traffic and then just wait until the trace is ready?
Is it flaky or something?

tests/e2e/helm-chart/02-install-simple-demo.yaml
```diff
@@ -1,31 +1,37 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: membership
+  name: coupon
```
yodigos, line 4: I see most of the yaml step files, like this, Are the same at helm-chart and multi-apps? 
Why do we need them both?

yodigos: I think we can do some "Common" folder where all the same yaml files will be there.

tests/e2e/README.md
```diff
@@ -0,0 +1,87 @@
...
+- kubectl configured to a fresh Kubernetes cluster. For local development, you can use KinD but also managed clusters like EKS should work.
+- Odigos cli compiled at `cli` folder. Compile via:
+```bash
+cd cli && go build -tags=embed_manifests -o odigos
+```
+- Odigos images tagged with `e2e-test` preloaded to the cluster. If you are using KinD you can run:
+```bash
+TAG=e2e-test make build-images load-to-kind 
+```
+- Chainsaw binary, can be installed via: `brew install kyverno/chainsaw/chainsaw`
```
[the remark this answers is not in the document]
yodigos: For me the tap did problems, but this worked:
`go install github.com/kyverno/chainsaw@latest`

tests/common/traceql_runner.sh
```diff
@@ -0,0 +1,74 @@
...
+  echo "Usage: $0 <traceql-test-file>"
+  exit 1
+fi
+
+# Test file path
+TEST_FILE=$1
+
+# Check if yq is installed
+if ! command -v yq &> /dev/null; then
+  echo "yq command not found. Please install yq."
```
blumamir, line 69: I got this error hidden in lots of text.
Should we add `brew install yq` to the README under tools?

yodigos: Also `brew install jq`

tests/e2e/helm-chart/chainsaw-test.yaml
edeNFed: Applying the remote YAML file does two things: pull all the images (because it's a fresh cluster) and run the deployments.
Because the GitHub action runners sometimes have bad internet connection it's hard to predict how much timeout we should put for both pulling images + running apps so I decided to split it into two steps for stability.

tests/e2e/helm-chart/chainsaw-test.yaml
edeNFed: Good question: Tempo is sort of eventually consistent, meaning you don't immediately get traces you just inserted in a query.
I force flush it so they are shown in the query without having to sleep until tempo persists the data.

tests/e2e/helm-chart/02-install-simple-demo.yaml
edeNFed: I plan to change one of the scenarios to install a different app (probably [kv-mall](https://github.com/keyval-dev/kv-mall)). So they are going to be different (will happen in a future PR)

tests/e2e/multi-apps/chainsaw-test.yaml
```diff
@@ -0,0 +1,113 @@
...
+          docker pull keyval/odigos-demo-inventory:v0.1
+          docker pull keyval/odigos-demo-membership:v0.1
+          docker pull keyval/odigos-demo-coupon:v0.1
+          docker pull keyval/odigos-demo-inventory:v0.1
+          docker pull keyval/odigos-demo-frontend:v0.2
+          kind load docker-image keyval/odigos-demo-inventory:v0.1
+          kind load docker-image keyval/odigos-demo-membership:v0.1
+          kind load docker-image keyval/odigos-demo-coupon:v0.1
+          kind load docker-image keyval/odigos-demo-inventory:v0.1
+          kind load docker-image keyval/odigos-demo-frontend:v0.2
```
blumamir, line 45: the image tag is not consistent (0.1 vs 0.2).
I think we should align with https://github.com/odigos-io/simple-demo/blob/main/kubernetes/deployment.yaml which is using v0.1.11. Can we simply `kubectl apply` the simple demo deployment which we use for the getting started?

yodigos: @blumamir You can look in the other chainsaw file - I raised the same thing and Eden answered there.

tests/e2e/multi-apps/chainsaw-test.yaml
```diff
@@ -0,0 +1,113 @@
...
+    - apply:
+        file: 04-add-destination.yaml
+    - assert:
+        file: assert-instrumented-and-pipeline.yaml
+  - name: Generate Traffic
+    try:
+    - script:
+        timeout: 60s
+        content: |
+          while true; do
```
yodigos, line 67: IMO, let's copy this to a new file: generate_traffic.sh in the common directory.
So, we don't need to duplicate it with the other chainsaw-test.
And this file will be cleaner - only steps description without logic.

blumamir (approved):

PR #1294, opened 2024-06-25
inspect first process with detected language in container

joeyleeeeeee97:
inspect the first process with detected language. 

Many applications are running with a startup.sh, so just inject the first process is not expected at those case.


odiglet/pkg/kube/runtime_details/shared.go
```diff
@@ -78,25 +79,30 @@ func runtimeInspection(pods []corev1.Pod, ignoredContainers []string) ([]odigosv
...
+			var inspectProc *procdiscovery.Details
+			var detectErr error
+
+			for _, proc := range processes {
+				lang, detectErr = inspectors.DetectLanguage(proc)
+				if detectErr == nil && lang != common.UnknownProgrammingLanguage {
+					inspectProc = &proc
+					break
+				}
+				log.Logger.V(0).Info("error detecting language", "error", detectErr, "process", proc)
```
RonFed, line 93: Please remove this logline, which might get called a lot.

joeyleeeeeee97: yes.

blumamir (approved):
Makes sense, thank you for the contribution 🙏

PR #1295, opened 2024-06-25
Movie UI into Kubernetes cluster

edeNFed:


Makefile
```diff
@@ -71,13 +76,18 @@ load-to-kind-collector:
 load-to-kind-instrumentor:
 	kind load docker-image $(ORG)/odigos-instrumentor:$(TAG)
 
+.PHONY: load-to-kind-ui
+load-to-kind-ui:
```
blumamir, line 80: could be nice to also add the restart and deploy targets

frontend/kube/client.go
```diff
@@ -30,8 +33,30 @@ type Client struct {
...
+			return nil, err
+		}
+	} else {
+		kubeConfig, err = clientcmd.BuildConfigFromFlags("", kc)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return kubeConfig, nil
+}
```
blumamir, line 56: nit: we can move this code to k8sutils instead of having it duplicated in cli and ui

cli/cmd/resources/ui.go
```diff
@@ -0,0 +1,234 @@
...
+							Env: []corev1.EnvVar{
+								{
+									Name: "CURRENT_NS",
+									ValueFrom: &corev1.EnvVarSource{
+										FieldRef: &corev1.ObjectFieldSelector{
+											FieldPath: "metadata.namespace",
+										},
+									},
+								},
+							},
```
blumamir, line 85: nit: do we need to pass CURRENT_NS into the container twice?

cli/cmd/ui.go
```diff
@@ -33,12 +34,6 @@ var uiCmd = &cobra.Command{
 	Short: "Start the Odigos UI",
 	Long:  `Start the Odigos UI. This will start a web server that will serve the UI`,
```
blumamir, line 35: nit: update the text here

cli/cmd/ui.go
```diff
@@ -55,222 +50,104 @@ var uiCmd = &cobra.Command{
...
 
-	ver, _ := strings.CutPrefix(release.TagName, "v")
-	return ver, nil
+	return &pods.Items[0], nil
 }
 
 func init() {
 	rootCmd.AddCommand(uiCmd)
-	uiCmd.Flags().String("address", "localhost", "address to listen on")
-	uiCmd.Flags().Int("port", defaultPort, "port to listen on")
```
blumamir, line 274: what if port 3000 is already taken on the host? don't we want the user to allow binding to other port?

blumamir (approved):

PR #1297, opened 2024-06-25
refactor: env overwriter

blumamir:


common/envOverwrite/originalenv.go
```diff
@@ -0,0 +1,14 @@
...
+// In k8s we need to store such map for each container in the pod.
+// In systemd, this can store the original values of the systemd service.
+//
+// When we want to rollback the changes we made to the environment variables, we can fetch the original
+// values from this map and set them back to the manifest.
+//
+// The key is the environment variable name.
+// The value is either the original value of the environment variable or nil if the environment variable
+// was not set by the user.
+type OriginalEnv map[string]*string
```
RonFed, line 14: If the env var was not set by the user I think we should not create an entry for it.

common/envOverwrite/overwriter.go
```diff
@@ -22,67 +23,93 @@ var EnvValuesMap = map[string]envValues{
...
 		}
 	}
 
-	return true
+	// Scenario 3: both odigos and user defined values are present
+	// happens: when the user set some values to this env (either via manifest or dockerfile) and odigos instrumentation is applied.
+	// action: we want to keep the user defined values and upsert the odigos value.
+	for _, sdkEnvValue := range envMetadata.values {
+		if strings.Contains(observedValue, sdkEnvValue) {
+			if sdkEnvValue == desiredOdigosPart {
```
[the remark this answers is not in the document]
blumamir: This part is in `if(strings.Contains(observedValue, sdkEnvValue))` so I think it's ok.

This if will be untrue when:
- `desiredOdigosPart` contains the new value for new SDK
- `observedValue` contains an old value, which is hit in the iteration on all possible values for any SDK
- `sdkEnvValue` which is still in `observedValue` is the old value
- then `sdkEnvValue != desiredOdigosPart`

common/envOverwrite/originalenv.go
blumamir: We create an entry with `nil` so that we know to delete it when reverting. Otherwise it's not straight forward to know if an arbitrary environment variable should be deleted from manifest (injected by odigos) or if it was there to begin with and should be untouched.

RonFed: We used the annotation for that, right?

instrumentor/controllers/instrumentationdevice/manager.go
```diff
@@ -3,10 +3,62 @@ package instrumentationdevice
...
+	if len(oldPodSpec.Spec.Containers) != len(newPodSpec.Spec.Containers) {
+		return true
+	}
+	for i := range oldPodSpec.Spec.Containers {
+		if len(oldPodSpec.Spec.Containers[i].Env) != len(newPodSpec.Spec.Containers[i].Env) {
+			return true
+		}
+		for j := range oldPodSpec.Spec.Containers[i].Env {
+			if oldPodSpec.Spec.Containers[i].Env[j].Value != newPodSpec.Spec.Containers[i].Env[j].Value {
+				return true
```
RonFed, line 46: Do we need to check the env name before checking the value?

blumamir: good point, thanks! updated

RonFed: I still think this can be improve to only check if the env vars we care about have changed/added/deleted

instrumentor/controllers/instrumentationdevice/workload_controllers.go
```diff
@@ -0,0 +1,50 @@
...
+func (r *StatefulSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	instrumentedAppName := workload.GetRuntimeObjectName(req.Name, "StatefulSet")
+	err := reconcileSingleInstrumentedApplicationByName(ctx, r.Client, instrumentedAppName, req.Namespace)
+	return ctrl.Result{}, err
+}
+
+func reconcileSingleInstrumentedApplicationByName(ctx context.Context, client client.Client, instrumentedAppName string, namespace string) error {
+	var instrumentedApplication odigosv1.InstrumentedApplication
+	err := client.Get(ctx, types.NamespacedName{Name: instrumentedAppName, Namespace: namespace}, &instrumentedApplication)
+	if err != nil {
```
RonFed, line 46: Ignore not found error

RonFed (approved):

PR #1298, opened 2024-06-26
Added an option to do odigos version --cli/--cluster to print only th…

yodigos:
…e version.

Removed the need of adding the TAG in the building commands.


Makefile
```diff
@@ -1,4 +1,11 @@
+NAMESPACE := odigos-system
+CONFIGMAP_NAME := odigos-deployment
+
 ORG := keyval
+
+# Cluster Version
+TAG := $(shell kubectl get configmap $(CONFIGMAP_NAME) -n $(NAMESPACE) -o jsonpath='{.data.ODIGOS_VERSION}')
```
RonFed, line 7: Also possible is `TAG ?= $(shell odigos version --short)`

Eden has a similar change in the Makefile at: https://github.com/odigos-io/odigos/pull/1250/files

yodigos: You are right!
In this PR I also added a flag for cli version and cluster version.
So, I'll use it :)

cli/cmd/version.go
```diff
@@ -13,26 +13,44 @@ import (
...
+	OdigosDate           string
 )
 
 // versionCmd represents the version command
 var versionCmd = &cobra.Command{
 	Use:   "version",
 	Short: "Print odigos version.",
 	Run: func(cmd *cobra.Command, args []string) {
-		short_flag, _ := cmd.Flags().GetBool("short")
+		short_flag, _ := cmd.Flags().GetBool(cliFlag)
```
blumamir, line 33: The conventions for this project are camelCase 
```suggestion
		shortFlag, _ := cmd.Flags().GetBool(cliFlag)
```

Makefile
```diff
@@ -1,5 +1,12 @@
-TAG ?= $(shell odigos version --short)
+NAMESPACE := odigos-system
+CONFIGMAP_NAME := odigos-deployment
+
+TAG ?= $(shell odigos version --cluster)
 ORG := keyval
+
+# Cluster Version
+#TAG ?= $(shell kubectl get configmap $(CONFIGMAP_NAME) -n $(NAMESPACE) -o jsonpath='{.data.ODIGOS_VERSION}')
```
blumamir, line 8: Do we still need it here?

Makefile
```diff
@@ -1,5 +1,12 @@
-TAG ?= $(shell odigos version --short)
+NAMESPACE := odigos-system
+CONFIGMAP_NAME := odigos-deployment
```
blumamir, line 2: these can be removed, right?

cli/cmd/version.go
```diff
@@ -13,30 +13,65 @@ import (
...
 	Short: "Print odigos version.",
 	Run: func(cmd *cobra.Command, args []string) {
-		short_flag, _ := cmd.Flags().GetBool("short")
+		cliFlag, _ := cmd.Flags().GetBool(cliFlag)
+		clusterFlag, _ := cmd.Flags().GetBool(clusterFlag)
 
-		if short_flag {
+		if cliFlag {
 			fmt.Printf("%s\n", OdigosVersion)
+		}
```
yodigos, line 38: @blumamir I changed it so you can do multiple arguments (cli + cluster).. I'm not sure there is or will be a use case for it. But, usually CLIs allow multiple arguments. So even in the future if we will want to add

blumamir (approved):

PR #1307, opened 2024-06-30
Reduce k8s cache memory usage

RonFed:
The following changes are made to reduce memory consumption in odiglet:
* Only include pods within the same node in the cache. This will make sure that we are only reconciling pods that are in the same node as odiglet (instead of filtering them after querying).
* Avoid saving the managed fields for cached objects https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/cache#TransformStripManagedFields
* Avoid querying ReplicaSet (which increased the memory consumption).

### DIFF
--- a/odiglet/pkg/kube/instrumentation_ebpf/pods.go
+++ b/odiglet/pkg/kube/instrumentation_ebpf/pods.go
@@ -3,12 +3,13 @@ package instrumentation_ebpf
 import (
 	"context"
 	"errors"
+	"strings"
 
 	"github.com/odigos-io/odigos/common"
+	"github.com/odigos-io/odigos/common/consts"
 	"github.com/odigos-io/odigos/odiglet/pkg/ebpf"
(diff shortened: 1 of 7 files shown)


odiglet/pkg/kube/instrumentation_ebpf/pods.go
```diff
@@ -25,6 +26,11 @@ type PodsReconciler struct {
 
 func (p *PodsReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
+
+	if request.Namespace == consts.DefaultOdigosNamespace {
```
edeNFed, line 30: better to use `env.GetCurrentNamespace()` to support non `odigos-system` namespaces

edeNFed (approved):

blumamir (approved):

PR #1308, opened 2024-06-30
Improve memory consumption of autoscaler and instrumentor

edeNFed:


autoscaler/main.go
```diff
@@ -105,15 +118,35 @@ func main() {
...
+				},
+				&corev1.Service{}: {
+					Label: labels.Set(gateway.CommonLabels).AsSelector(),
+					Field: nsSelector,
+				},
+				&appsv1.DaemonSet{}: {
+					Field: nsSelector,
+				},
+				&corev1.ConfigMap{}: {
+					Field: nsSelector,
```
RonFed, line 143: If we are only interested in the odigos ns here, we can use `DefaultNamespaces` field in the top cache options.

RonFed (approved):

PR #1317, opened 2024-07-02
Head sampling CRD

RonFed:
Introduce head sampling configuration to the instrumentation config CRD.
This PR only includes the definitions without implementations.


blumamir:
replaced by #1445

PR #1326, opened 2024-07-05
Verify sync to odigos-chart repo

edeNFed:


blumamir:
charts moved to this repo, so the PR is no longer required

PR #1327, opened 2024-07-05
Test connection for destinations

RonFed:


frontend/endpoints/test_connection/otlphttp_test_connection.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func NewOTLPHTTPTester() *otlphttpExporterConnectionTester {
+	return &otlphttpExporterConnectionTester{
+		f: otlphttpexporter.NewFactory(),
+	}
+}
+
+func (t *otlphttpExporterConnectionTester) Factory() exporter.Factory {
+	return t.f
```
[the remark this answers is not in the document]
RonFed: The idea was to call `NewFactory` once at the init phase and re-use it on each connection test. Each test connection test calls this function multiple times to get the factory.

frontend/endpoints/destinations.go
```diff
@@ -56,6 +58,21 @@ type Destination struct {
...
+}
+func (dest Destination) GetType() common.DestinationType {
+	return dest.Type
+}
+func (dest Destination) GetConfig() map[string]string {
+	return dest.Fields
+}
+func (dest Destination) GetSignals() []common.ObservabilitySignal {
+	return exportedSignalsObjectToSlice(dest.ExportedSignals)
+}
```
blumamir, line 74: This already exists [here](https://github.com/odigos-io/odigos/blob/b2f196915907bf5f7460fd02d22df9e3ab6bc2bd/api/odigos/v1alpha1/destination_types.go#L67).

Is defining it again intentional?

blumamir (approved):

frontend/endpoints/destinations.go
RonFed: Yes, these are for the ui struct and those you linked are for the k8s one - they both implement this interface, and it is used when we call `ModifyCOnfig` at the `TestConnection` function

blumamir: Right!
so confusing 😅

LGTM

PR #1350, opened 2024-07-13
feat: nodejs opamp instrumentation libraries enabled remote config

blumamir:


opampserver/pkg/sdkconfig/configsections/types.go
```diff
@@ -0,0 +1,41 @@
+package configsections
+
+import "github.com/odigos-io/odigos/opampserver/pkg/sdkconfig/configresolvers"
+
+type ConfigSectionName string
+
+// This interface is the message sent to opamp client to configure aspects of the SDK
```
RonFed, line 7: Where is the interface?

blumamir:
helm chart: https://github.com/odigos-io/odigos-charts/pull/99

opampserver/pkg/connection/conncache.go
```diff
@@ -0,0 +1,144 @@
...
+		return &connCopy, ok
+	}
+}
+
+func (c *ConnectionsCache) AddConnection(deviceId string, conn *ConnectionInfo) {
+	// copy the conn object to avoid it being accessed concurrently
+	connCopy := *conn
+	c.mux.Lock()
+	defer c.mux.Unlock()
+	c.liveConnections[deviceId] = &connCopy
```
tamirdavid1, line 55: I remember we discussed about an option to save {deviceId}-{InstanceId} to handle fork cases?

opampserver/pkg/server/handlers.go
```diff
@@ -50,56 +55,73 @@ func (c *ConnectionHandlers) OnNewConnection(ctx context.Context, deviceId strin
...
 	instrumentedAppName := workload.GetRuntimeObjectName(k8sAttributes.WorkloadName, k8sAttributes.WorkloadKind)
-	c.logger.Info("new OpAMP client connected", "deviceId", deviceId, "namespace", k8sAttributes.Namespace, "podName", k8sAttributes.PodName, "instrumentedAppName", instrumentedAppName, "workloadKind", k8sAttributes.WorkloadKind, "workloadName", k8sAttributes.WorkloadName, "containerName", k8sAttributes.ContainerName, "otelServiceName", k8sAttributes.OtelServiceName)
+	remoteResourceAttributes, err := configresolvers.CalculateServerAttributes(k8sAttributes, c.nodeName)
+	if err != nil {
+		c.logger.Error(err, "failed to calculate server attributes", "k8sAttributes", k8sAttributes)
+		return nil, nil, err
+	}
 
-	sdkConfig, err := json.Marshal(RemoteConfigSdk{RemoteResourceAttributes: serverOfferedResourceAttributes})
+	fullRemoteConfig, err := c.sdkConfig.GetFullConfig(ctx, remoteResourceAttributes, &podWorkload)
```
tamirdavid1, line 82: We can maybe send the instrumentedAppName to the GetFullConfig so we wont need to call GetRuntimeObjectName again in it?

opampserver/pkg/server/handlers.go
```diff
@@ -50,56 +55,73 @@ func (c *ConnectionHandlers) OnNewConnection(ctx context.Context, deviceId strin
...
+	c.logger.Info("new OpAMP client connected", "deviceId", deviceId, "namespace", k8sAttributes.Namespace, "podName", k8sAttributes.PodName, "instrumentedAppName", instrumentedAppName, "workloadKind", k8sAttributes.WorkloadKind, "workloadName", k8sAttributes.WorkloadName, "containerName", k8sAttributes.ContainerName, "otelServiceName", k8sAttributes.OtelServiceName)
+
+	connectionInfo := &connection.ConnectionInfo{
+		DeviceId:                 deviceId,
+		Workload:                 podWorkload,
+		Pod:                      pod,
+		Pid:                      pid,
+		InstrumentedAppName:      instrumentedAppName,
+		AgentRemoteConfig:        fullRemoteConfig,
+		RemoteResourceAttributes: remoteResourceAttributes,
```
tamirdavid1, line 97: as i understood it, the `remoteResourceAttributes` is part of the fullRemoteConfig, why we need it in the ConnectionInfo as well?
I saw that we save it in the connectionCache and never use it from the cache

tamirdavid1:
Amir, what do you think about aligning the opamp server response/request with the terminology of
`agentToServer` instead of `opampRequest`
and `serverToAgent` instead of `opampResponse`?

opampserver/pkg/server/handlers.go
blumamir: great catch. fixed 👍

opampserver/pkg/server/handlers.go
blumamir: We use it when one of the other reconcilers updates and we need to calculate the entire full config from scratch. If we don't store it anywhere, we will need to calculate on each new config update.
I agree it's not that clear right now, would love to address it oneday and improvre

blumamir:
> Amir, what do you think about aligning the opamp server response/request with the terminology of `agentToServer` instead of `opampRequest` and `serverToAgent` instead of `opampResponse`?

good point. updated 👍

tamirdavid1 (approved):

PR #1351, opened 2024-07-14
Filelog receiver update

yodigos:


common/config/debug.go
```diff
@@ -6,14 +6,25 @@ import (
...
 	exporterName := "debug/" + dest.GetID()
 
+	verbosity, exists := dest.GetConfig()[VERBOSITY]
+	if !exists {
+		// Default verbosity
+		verbosity = "basic"
+	}
+
 	currentConfig.Exporters[exporterName] = GenericMap{
+		VERBOSITY: verbosity,
```
blumamir, line 27: nit: the `const` above is used to reference the field name in the CRD.
for the otelcol config map, the convention in all other destinations is to use the name directly:
```suggestion
	"verbosity": verbosity,
```

blumamir (approved):

PR #1357, opened 2024-07-16
fix: rate limit node collector restarts on config change

yodigos:


autoscaler/controllers/datacollection/configmap.go
```diff
@@ -38,11 +38,11 @@ func syncConfigMap(apps *odigosv1.InstrumentedApplicationList, dests *odigosv1.D
 	setTracesLoadBalancer := SamplingExists != nil
 
 	desired, err := getDesiredConfigMap(apps, dests, processors, datacollection, scheme, setTracesLoadBalancer)
-	desiredData := desired.Data[configKey]
 	if err != nil {
 		logger.Error(err, "failed to get desired config map")
 		return "", err
 	}
+	desiredData := desired.Data[configKey]
```
yodigos, line 45: possible failure if desired is nil -> just changed order

autoscaler/controllers/datacollection/daemonset.go
```diff
@@ -289,3 +294,38 @@ func patchDaemonSet(existing *appsv1.DaemonSet, desired *appsv1.DaemonSet, ctx c
...
+		return
+	}
+
+	dm.inProgress = true
+
+	time.AfterFunc(delay, func() {
+		dm.mu.Lock()
+		defer dm.mu.Unlock()
+
+		logger.Info("Sync DaemonSet function execution started...")
```
yodigos, line 320: I can also create a generic Retry function / use external library like: https://github.com/avast/retry-go
Or leave it as is..
WDYT?

autoscaler/controllers/datacollection/root.go
```diff
@@ -57,18 +66,17 @@ func syncDataCollection(instApps *odigosv1.InstrumentedApplicationList, dests *o
 	logger := log.FromContext(ctx)
 	logger.V(0).Info("Syncing data collection")
 
-	configData, err := syncConfigMap(instApps, dests, processors, dataCollection, ctx, c, scheme)
+	_, err := syncConfigMap(instApps, dests, processors, dataCollection, ctx, c, scheme)
```
yodigos, line 69: I calculate this later -> when that patch triggers

autoscaler/controllers/datacollection/daemonset.go
```diff
@@ -289,3 +294,35 @@ func patchDaemonSet(existing *appsv1.DaemonSet, desired *appsv1.DaemonSet, ctx c
...
 	return &b
 }
+
+type DelayManager struct {
+	mu         sync.Mutex
+	inProgress bool
+}
+
+// runFunctionWithDelayAndSkipNewCalls runs the function with the specified delay and skips new calls until the function execution is finished
+func (dm *DelayManager) runFunctionWithDelayAndSkipNewCalls(delay time.Duration, fn func(args ...interface{}) (*appsv1.DaemonSet, error), fnArgs ...interface{}) {
```
RonFed, line 304: I'd prefer to use a library function for this if possible.
Looks like https://pkg.go.dev/k8s.io/apimachinery@v0.30.2/pkg/util/wait#PollUntilContextTimeout might be relevant or other functions from that lib

autoscaler/controllers/datacollection/root.go
```diff
@@ -2,13 +2,22 @@ package datacollection
...
 	odigosv1 "github.com/odigos-io/odigos/api/odigos/v1alpha1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+var dm = &DelayManager{}
+
+const (
+	DAEMONSET_PATCH_DELAY = 20 * time.Second
```
RonFed, line 17: The convention for consts in camelCase
`daemonSetPatchDelay` etc`

autoscaler/controllers/datacollection/daemonset.go
```diff
@@ -38,17 +37,43 @@ var (
...
+		dm.mu.Lock()
+		defer dm.mu.Unlock()
+		defer dm.finishProgress()
+
+		for i := 0; i < retries; i++ {
+			_, err := syncDaemonSet(ctx, dests, collection, c, scheme, secrets, version)
+			if err == nil {
+				return
+			}
+		}
```
blumamir, line 68: should we log something if we run out of retries and give up?
otherwise, the error is silently dropped, right?

and should we also log something on success to aid debugging in case there are issues?

blumamir (commented):
Looks good! added 2 nit comments

blumamir (approved):

PR #1370, opened 2024-07-18
Fix: Delete Node Collector Group when deleting all Sources

yodigos:
… active.

And Create a new Node Collector Group  when Apps are added


cli/cmd/resources/scheduler.go
```diff
@@ -154,6 +154,35 @@ func NewSchedulerClusterRole() *rbacv1.ClusterRole {
 					"destinations/status",
 				},
 			},
+			{
+				Verbs: []string{
```
yodigos, line 158: Yoav: fix this to only relevant priorities

common/utils/collectorgroup_util.go
```diff
@@ -0,0 +1,39 @@
...
+func GetCollectorGroup(ctx context.Context, c client.Client, namespace string, collectorGroupName string) (*odigosv1.CollectorsGroup, error) {
+	var collectorGroup odigosv1.CollectorsGroup
+	err := c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: collectorGroupName}, &collectorGroup)
+
+	return &collectorGroup, err
+}
+
+func DeleteCollectorGroup(ctx context.Context, c client.Client, namespace string, collectorGroupName string) error {
+	logger := log.FromContext(ctx)
+	logger.Info("Deleting collector group", "collectorGroupName", collectorGroupName)
```
yodigos, line 25: Yoav TODO: I assume I can do something like logger.extra addition instead of adding to each log.

cli/cmd/resources/scheduler.go
RonFed, line 1: These changes will also need to be made in the charts repo

common/utils/collectorgroup_util.go
RonFed, line 1: This file should be under `k8sutils`. The common module shouldn't have k8s dependencies.

RonFed (approved):

PR #1372, opened 2024-07-20
Fix env vars uninstall

edeNFed:


odiglet/pkg/kube/runtime_details/inspection.go
```diff
@@ -109,8 +109,10 @@ func runtimeInspection(pods []corev1.Pod, ignoredContainers []string) ([]odigosv
 			if inspectProc == nil {
 				log.Logger.V(0).Info("unable to detect language for any process", "pod", pod.Name, "container", container.Name, "namespace", pod.Namespace)
 				lang = common.UnknownProgrammingLanguage
-			} else if len(processes) > 1 {
-				log.Logger.V(0).Info("multiple processes found in pod container, only taking the first one with detected language into account", "pod", pod.Name, "container", container.Name, "namespace", pod.Namespace)
+			} else if len(inspectProc.Envs) > 0 {
```
blumamir, line 112: why do we need to check `inspectProc.Envs > 0`?

blumamir (approved):

odiglet/pkg/kube/runtime_details/inspection.go
edeNFed: Good catch, fixed

PR #1385, opened 2024-07-22
Avoid app to crash python agent

tamirdavid1:


agents/python/configurator/lib_handling.py
```diff
@@ -0,0 +1,43 @@
...
+        'asgiref'
+        'idna',
+        'deprecated',
+        'importlib_metadata',
+        'packaging',
+        'psutil',
+        'zipp',
+        'urllib3',
+        'uuid_extensions.uuid7',
+        'typing_extensions',
```
RonFed, line 30: Are these taken from `requirements.txt`? If so, maybe it is better to read them from the file or pass them as an argument

tamirdavid1: Hey Ron,
This file is not exists in the runtime, also it is containing packages names that sometimes are different from the module name that i'm reloading.
**e.g** -> protobuf (`pip install protobuf`) but the module name is `google.protobuf`.
I thought about reloading all modules available in sys.modules but it can potentially take time

agents/python/configurator/lib_handling.py
```diff
@@ -0,0 +1,43 @@
...
+    ]
+
+    for module_name in needed_modules:
+        if module_name in sys.modules:
+            module = sys.modules[module_name]
+            importlib.reload(module)
+            
+
+def get_module_version(module_name: str):
```
blumamir, line 39: is it used anywhere?

blumamir (approved):

agents/python/configurator/lib_handling.py
tamirdavid1: removed, used for debugging.

PR #1396, opened 2024-07-27
Add additional ClickHouse fields

edeNFed:


common/config/clickhouse.go
```diff
@@ -24,20 +30,57 @@ func (c *Clickhouse) ModifyConfig(dest ExporterConfigurer, currentConfig *Config
...
+
 	exporterName := "clickhouse/clickhouse-" + dest.GetID()
 	exporterConfig := GenericMap{
 		"endpoint": endpoint,
 	}
 	if userExists {
 		exporterConfig["username"] = username
-		exporterConfig["password"] = password
+		exporterConfig["password"] = "${CLICKHOUSE_PASSWORD}"
+	}
```
RonFed, line 55: It looks like CLICKHOUSE_PASSWORD is removed from the consts

common/config/clickhouse.go
```diff
@@ -24,20 +30,57 @@ func (c *Clickhouse) ModifyConfig(dest ExporterConfigurer, currentConfig *Config
...
+
+	dbName, exists := dest.GetConfig()[clickhouseDatabaseName]
+	if !exists {
+		return errors.New("clickhouse database name not specified, gateway will not be configured for Clickhouse")
+	}
+	exporterConfig["database"] = dbName
+
+	tracesTable, exists := dest.GetConfig()[clickhouseTracesTable]
+	if !exists {
+		return errors.New("clickhouse traces table not specified, gateway will not be configured for Clickhouse")
```
RonFed, line 69: I don't think we should return an error here. It is legitimate to have a table only for metrics or logs. right?

RonFed (approved):

PR #1405, opened 2024-07-29
feat: remove OdigosConfig CRD

rauno56:
This will remove usage of `odigosconfigurations.odigos.io` CRD and replace it with a plain ConfigMap to remove race conditions when installing named CRD and CR at the same time during the install process. New ConfigMap will look as following:

```
[...]
data:
  config.yaml: { ... serialized JSON of odigosconfiguration.spec ... }
[...]
```

With the exception of removal of `SupportedSDKs`(as per @blumamir request to remove it from CM) the contents of `config.yaml` value and the `Spec` field are identical and serialize from the same string and parse to the same object.

Installs now just install ConfigMap instead of the CR of OdigosConfiguration, Helm upgrades are unchanged, as they used to just overwrite the CR anyway. CLI parses the old config unless there's already a ConfigMap and creates it as a ConfigMap.

The old CRD is left there unused for now to be able to keep on using the API for it(for migration) - can be removed when we feel we are ready to remove the migration path.


helm/odigos/templates/instrumentor/clusterrole.yaml
```diff
@@ -25,6 +25,18 @@ rules:
...
+    resources:
+      - configmaps
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - patch
+      - update
+      - watch
```
blumamir, line 39: I think we only read from the configmap, right? can we keep here only `get`, `list` and `watch`?

helm/odigos/templates/ui/clusterrole.yaml
```diff
@@ -12,6 +12,18 @@ rules:
...
+    resources:
+      - configmaps
+    verbs:
+      - get
+      - list
+      - watch
+      - patch
+      - create
+      - delete
+      - update
```
blumamir, line 26: same here, I think this should be a role for the namespace only instead of clusterrole

common/consts/consts.go
```diff
@@ -8,6 +8,7 @@ const (
 	CurrentNamespaceEnvVar        = "CURRENT_NS"
 	DefaultOdigosNamespace        = "odigos-system"
 	OdigosConfigurationName       = "odigos-config"
+	OdigosConfigurationFileName   = "config.yaml"
```
blumamir, line 11: calling it `config.yaml` makes it should as if it is mapped to a real file somewhere. do you think it might be confusing for people reading it in the future?

.goreleaser.yaml
```diff
@@ -40,7 +40,7 @@ changelog:
       - "^docs:"
       - "^test:"
 brews:
-  - tap:
+  - repository:
```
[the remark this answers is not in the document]
rauno56: CI said `tap` is deprecated: https://goreleaser.com/deprecations/?h=tap+repository#brewstap

common/consts/consts.go
rauno56: It definitely **could** be mapped to a real file as well, but I don't think that's necessarily something that would confuse me if I saw this.

Happy to change it, but since the configmap is called "odigos-config", the only key, that would make more sense than "config.yaml" to me would be "content", which conceptually fits the "idea" of "odigos-config / data.content" better, but feels less practical, because 1. being overly generic 2. removing slightly more convenient volume mounting option.

What would you name it?

cli/cmd/resources/applyresources.go
```diff
@@ -41,6 +43,19 @@ func DeleteOldOdigosSystemObjects(ctx context.Context, client *kube.Client, ns s
...
+	var odigosConfig common.OdigosConfiguration
+	if err := yaml.Unmarshal([]byte(configMap.Data[consts.OdigosConfigurationFileName]), &odigosConfig); err != nil {
+		return nil, err
+	}
+	return &odigosConfig, nil
 }
+
+func GetCurrentOdigosConfig(ctx context.Context, client *kube.Client, ns string) (*v1alpha1.OdigosConfiguration, error) {
+  return client.OdigosClient.OdigosConfigurations(ns).Get(ctx, consts.OdigosConfigurationName, metav1.GetOptions{})
+}
```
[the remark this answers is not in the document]
rauno56: Changing it to `GetDeprecatedConfig`.

autoscaler/controllers/gateway/root.go
```diff
@@ -44,21 +47,26 @@ func Sync(ctx context.Context, k8sClient client.Client, scheme *runtime.Scheme,
...
+	var configMap v1.ConfigMap
+	if err := k8sClient.Get(ctx, types.NamespacedName{Namespace: odigosSystemNamespaceName, Name: consts.OdigosConfigurationName}, &configMap); err != nil {
 		logger.Error(err, "failed to get odigos config")
 		return err
 	}
+	var odigosConfig common.OdigosConfiguration
+	if err := yaml.Unmarshal([]byte(configMap.Data[consts.OdigosConfigurationFileName]), &odigosConfig); err != nil {
+		logger.Error(err, "failed to parse odigos config from configmap")
+		return err
+	}
```
edeNFed, line 62: This looks like a code block we would probably need in multiple places, maybe it's worth extracting it to `k8sutils`?

cli/cmd/resources/applyresources.go
```diff
@@ -41,6 +43,19 @@ func DeleteOldOdigosSystemObjects(ctx context.Context, client *kube.Client, ns s
...
-	return client.OdigosClient.OdigosConfigurations(ns).Get(ctx, consts.OdigosConfigurationName, metav1.GetOptions{})
+func GetCurrentConfig(ctx context.Context, client *kube.Client, ns string) (*common.OdigosConfiguration, error) {
+	configMap, err := client.CoreV1().ConfigMaps(ns).Get(ctx, consts.OdigosConfigurationName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	var odigosConfig common.OdigosConfiguration
+	if err := yaml.Unmarshal([]byte(configMap.Data[consts.OdigosConfigurationFileName]), &odigosConfig); err != nil {
+		return nil, err
+	}
```
edeNFed, line 54: `k8sutils`?

edeNFed:
looks great. In addition to Amir's great comments:
- Will we need a similar PR in `odigos-enterprise` repo?
- Update e2e tests assertion to validate ConfigMap is created during installation
- Today we have a retry loop in e2e tests for helm install to avoid the OdigosConfig CRD bug, can we delete it after this?

rauno56:
Although Amir's comments about the permissions are technically correct about being too broad. I'll leave the those the same for now - the current implementation mimics exactly what OdigosConfig had.

I will narrow down the permissions in a future PR.

> Will we need a similar PR in odigos-enterprise repo?

I will check that out.

> Update e2e tests assertion to validate ConfigMap is created during installation

I have a feeling it wasn't there previously because there's just no way that anything works if the services cannot find the config. Added it!

> Today we have a retry loop in e2e tests for helm install to avoid the OdigosConfig CRD bug, can we delete it after this?

I will remove it in this PR - the expectation is that the install will not fail and that workaround is not required.

cli/cmd/resources/applyresources.go
```diff
@@ -41,6 +43,19 @@ func DeleteOldOdigosSystemObjects(ctx context.Context, client *kube.Client, ns s
...
+	return &odigosConfig, nil
 }
```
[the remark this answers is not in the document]
rauno56: Refactored the use cases in controllers. Left the others intact.

rauno56:
@blumamir, @edeNFed, I really appreciate your comments. I think I've now addressed all of them.

blumamir (approved):
that's a lot of places to follow and clean up, Thank you @rauno56 for the thorough work

so happy to have this fixed 🎉 

Left just one comment about deleting the old resource to keep the cluster clean and consistent with helm. other than this, I think we can merge

k8sutils/pkg/utils/config_util.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/odigos-io/odigos/common"
+	"github.com/odigos-io/odigos/common/consts"
+	"github.com/odigos-io/odigos/k8sutils/pkg/env"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/yaml"
+)
+
+func GetCurrentConfig(ctx context.Context, k8sClient client.Client) (common.OdigosConfiguration, error) {
```
blumamir, line 15: nit: "Config" can mean few things, perhaps we can rename it `GetCurrentOdigosConfig` so that it communicate what is being returned in the function name

PR #1407, opened 2024-07-30
feat: odigos describe sub commands for auto completion 

blumamir:


cli/cmd/describe.go
```diff
@@ -43,55 +46,7 @@ func wrapTextSuccessOfFailure(text string, success bool) string {
...
-		}
-		return &daemonSet.Spec.Template, metav1.FormatLabelSelector(daemonSet.Spec.Selector), nil
-
-	default:
-		return nil, "", fmt.Errorf("unsupported kind: %s", gvk.Kind)
-	}
-}
-
-func getInstrumentationLabelTexts(workload *unstructured.Unstructured, ns *v1.Namespace) (workloadText, nsText, decisionText string, instrumented bool) {
+func getInstrumentationLabelTexts(workload *K8sSourceObject, ns *v1.Namespace) (workloadText, nsText, decisionText string, instrumented bool) {
```
yodigos, line 49: Looks like we don't return anything in this function - so we can remove the return values from the function signature

cli/cmd/describe.go
```diff
@@ -110,17 +65,17 @@ func getInstrumentationLabelTexts(workload *unstructured.Unstructured, ns *v1.Na
 	if workloadFound {
 		instrumented = workloadLabel == consts.InstrumentationEnabled
 		if instrumented {
-			decisionText = "Workload is instrumented because the " + workload.GetKind() + " contains the label '" + consts.OdigosInstrumentationLabel + "=" + workloadLabel + "'"
+			decisionText = "Workload is instrumented because the " + workload.Kind + " contains the label '" + consts.OdigosInstrumentationLabel + "=" + workloadLabel + "'"
```
yodigos, line 68: IMO, you can create a template of this text and just fill the needed values (including with/out -> NOT)

yodigos: Potentially if someone will edit this - he can forget to do it in one the places

cli/cmd/describe.go
```diff
@@ -448,44 +393,134 @@ func printPodsInfo(pods *corev1.PodList, instrumentationInstances *odigosv1.Inst
...
+			fmt.Printf("Error: %v\n", err)
 			return
 		}
+		workloadObj := &K8sSourceObject{
+			Kind:            "deployment",
+			ObjectMeta:      deployment.ObjectMeta,
+			PodTemplateSpec: &deployment.Spec.Template,
+			LabelSelector:   deployment.Spec.Selector,
+		}
+		printDescribeSource(cmd.Context(), client, workloadObj)
```
yodigos, line 449: change to ctx -> it is parsed at line 435

cli/cmd/describe.go
```diff
@@ -448,44 +393,134 @@ func printPodsInfo(pods *corev1.PodList, instrumentationInstances *odigosv1.Inst
...
 			kube.PrintClientErrorAndExit(err)
 		}
+
 		ctx := cmd.Context()
+		name := args[0]
 		ns := cmd.Flag("namespace").Value.String()
-
-		if len(args) != 2 {
-			fmt.Println("Usage: odigos describe <kind> <name>")
+		deployment, err := client.AppsV1().Deployments(ns).Get(ctx, name, metav1.GetOptions{})
```
yodigos, line 438: I see this function happens also at sources.go, login.go, uninstall.go.
Maybe we can create a util file and put the function over there.
Your call..

cli/cmd/describe.go
```diff
@@ -448,44 +393,134 @@ func printPodsInfo(pods *corev1.PodList, instrumentationInstances *odigosv1.Inst
...
+		if err != nil {
+			fmt.Printf("Error: %v\n", err)
 			return
 		}
+		workloadObj := &K8sSourceObject{
+			Kind:            "deployment",
```
yodigos, line 444: Not sure but I think deployment.Kind returns this string.
Also other places

yodigos (approved):

cli/cmd/describe.go
blumamir: return values are used by name

cli/cmd/describe.go
blumamir: mmm, each line is a bit different, and this function is the only place where we manage this text, so I think it's ok. perhaps if it gets more complicated in the future, we can refactor this

cli/cmd/describe.go
blumamir: good catch, fixed 👍

cli/cmd/describe.go
blumamir: Since it's just one line, I prefer to use the k8s client directly and not introduce more complexity to the code

cli/cmd/describe.go
blumamir: Unfortunately, The `Kind` property is not populated by the client when instantiating a `Deployment` object. This is why I needed to add it explicitly

PR #1411, opened 2024-08-01
--platform in dotnet phase

RonFed:


odiglet/Dockerfile
```diff
@@ -52,7 +52,7 @@ RUN yarn compile
 
 
 
-FROM busybox:1.36.1 AS dotnet-builder
+FROM --platform=$BUILDPLATFORM busybox:1.36.1 AS dotnet-builder
```
edeNFed, line 55: This will cause to always download amd64 version no?

RonFed: It will pull the image native to where docker is running.
I think the error we got is because the CI is running in AMD and was emulating wget for arm64 - which causes some weird errors.

edeNFed (approved):

yodigos (approved):

PR #1423, opened 2024-08-05
feat: FE e2e tests

yodigos:


frontend/webapp/cypress/e2e/test-overview.cy.ts
```diff
@@ -0,0 +1,23 @@
+describe('Overview Page Tests', () => {
+
+    beforeEach(() => {
+        cy.visit('localhost:3000')
```
alonkeyval, line 4: i think you can put the url inside environment variables, and then if we switch route it will be only in one place
`export default defineConfig({
  
  env: {
    overview: 'http://localhost:3000/overview',
    base_url: 'http://localhost:3000/'
  },
})`

alonkeyval (approved):

PR #1424, opened 2024-08-05
fix: node collector restarts when adding cluster processor

blumamir:


autoscaler/controllers/datacollection/daemonset.go
```diff
@@ -91,7 +92,8 @@ func syncDaemonSet(ctx context.Context, dests *odigosv1.DestinationList, datacol
 		return nil, err
 	}
 
-	desiredDs, err := getDesiredDaemonSet(datacollection, configMap.String(), scheme, imagePullSecrets, odigosVersion, odigletDaemonsetPodSpec)
```
blumamir, line 94: here is the original issue, we used `configMap.String()` which includes metadata that might change even when the content didn't.

edeNFed (approved):

PR #1426, opened 2024-08-06
feat: add receiver signals to collectors group CRD

blumamir:


api/odigos/v1alpha1/collectorsgroup_types.go
```diff
@@ -37,6 +38,11 @@ type CollectorsGroupSpec struct {
 // CollectorsGroupStatus defines the observed state of Collector
 type CollectorsGroupStatus struct {
 	Ready bool `json:"ready,omitempty"`
+
+	// Receiver Signals are the signals (trace, metrics, logs) that the collector has setup
+	// an otlp receiver for, thus it can accept data from an upstream component.
+	// this is used to determine if a workload should export each signal or not.
```
RonFed, line 44: They are determined by the destinations, right? worth adding to the comment

autoscaler/controllers/datacollection/daemonset.go
```diff
@@ -125,6 +130,12 @@ func syncDaemonSet(ctx context.Context, dests *odigosv1.DestinationList, datacol
 		return nil, err
 	}
 
+	err = common.UpdateCollectorGroupReceiverSignals(ctx, c, datacollection, signals)
+	if err != nil {
+		logger.Error(err, "Failed to update collector group received signals")
```
RonFed, line 135:
```suggestion
		logger.Error(err, "Failed to update node collector group received signals")
```

autoscaler/controllers/datacollection/configmap.go
```diff
@@ -283,3 +285,42 @@ func getConfigMap(ctx context.Context, c client.Client, namespace string) (*v1.C
 
 	return configMap, nil
 }
+
+func getSignalsFromOtelcolConfig(otelcolConfigContent string) ([]common.ObservabilitySignal, error) {
+	config := config.Config{}
+	err := yaml.Unmarshal([]byte(otelcolConfigContent), &config)
```
RonFed, line 291: This unmarshilng can have bad impact on memory consumption and CPU if the reconciler is out of control.

blumamir: This function should be called at most once every 10 seconds https://github.com/odigos-io/odigos/blob/0f80c84faefd58373912e50f402b0d5e3712076e/autoscaler/controllers/datacollection/root.go#L69

To be honest, I would love for the collector to automatically reload configuration in the future and then we can get rid of this code altogether

RonFed (approved):

PR #1430, opened 2024-08-06
feat: FE test action

yodigos:


frontend/webapp/cypress/e2e/action-testing.cy.ts
```diff
@@ -0,0 +1,20 @@
...
+        cy.url().should('eq', `${Cypress.config('baseUrl')}/choose-action`);
+        cy.get('[data-cy="choose-action-ProbabilisticSampler"]').should('exist').click()
+
+        cy.url().should('eq', `${Cypress.config('baseUrl')}/create-action?type=ProbabilisticSampler`);
+        cy.get('[data-cy="create-action-input-name"]').type('action-test');
+        cy.get('[data-cy="create-action-sampling-percentage"]').type('0.5');
+        cy.get('[data-cy="create-action-onclick"]').click()
+
+        cy.url().should('eq', `${Cypress.config('baseUrl')}/actions`);
+        cy.get('[data-cy="actions-action-name"]').should('have.text', 'action-test ')
```
alonkeyval, line 18: i think you have an extra space at the end

frontend/webapp/cypress/e2e/action-testing.cy.ts
```diff
@@ -0,0 +1,20 @@
+describe('Action Addition Testing', () => {
+    it('Add an Action', () => {
+        cy.visit('/')
+        cy.get('[data-cy="menu-Actions"]').should('exist').click()
+        cy.url().should('eq', `${Cypress.config('baseUrl')}/actions`);
```
alonkeyval, line 5:
```suggestion
const baseUrl = Cypress.config('baseUrl');
        cy.url().should('eq', `${baseUrl}//actions`);
```
calling only once to the cypress API and using it in all the places

frontend/webapp/cypress/e2e/action-testing.cy.ts
yodigos: It's for some reason what is written in the text in the UI, that is why

PR #1432, opened 2024-08-07
Report opamp health

tamirdavid1:


agents/python/configurator/__init__.py
```diff
@@ -80,3 +88,12 @@ def shutdown():
     atexit.register(shutdown)
 
     return client
+
+
+def is_supported_python_version():
+    current_version = sys.version_info
```
yodigos, line 94:
```suggestion
 return sys.version_info>= min_version
```

agents/python/opamp/http_client.py
```diff
@@ -50,14 +58,41 @@ def run(self):
...
+        self.send_agent_to_server_message(first_disconnect_message)
+        
     def send_first_message_with_retry(self) -> None:
         max_retries = 5
         delay = 2
         for attempt in range(1, max_retries + 1):
             try:
-                first_message_server_to_agent = self.send_full_state()
+                # Send first message to OpAMP server, Health is false as the component is not initialized
+                agent_health = self.get_agent_health(component_health=False, last_error="Component not initialized", status="Starting")
```
yodigos, line 84: Enum of all the status will avoid mistakes and help understand the different types

agents/python/opamp/http_client.py
```diff
@@ -50,14 +58,41 @@ def run(self):
...
+        
+        agent_description = self.get_agent_description()
+        
+        first_disconnect_message.agent_description.CopyFrom(agent_description)
+        
+        agent_disconnect = self.get_agent_disconnect()
+        first_disconnect_message.agent_disconnect.CopyFrom(agent_disconnect)
+    
+        python_version = f'{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}'
+        agent_health = self.get_agent_health(component_health=False, last_error=f"Opentelemetry SDK require Python in version 3.8 or higher, [{python_version}] not supported",
```
blumamir, line 72: should we rename the function to so it's clear that it is used only for `UnsupportedRuntimeVersion` status?

agents/python/opamp/http_client.py
```diff
@@ -35,7 +36,14 @@ def __init__(self, event, condition: threading.Condition):
         self.instance_uid = uuid7().__str__()
         self.remote_config_status = None
 
-    def start(self):
+
+    def start(self, python_version_supported: bool = None):
+        if not python_version_supported:
+            opamp_logger.warning("Python version not supported, sending disconnect message to OpAMP server...")
```
blumamir, line 42: might be useful to add the python version that we found and what is the minimum we support.

perhaps reuse the `last_error` below?

agents/python/opamp/http_client.py
blumamir: The text "Component not initialized" will end up being showed to the user, and I wonder what he can understand from it

Which component?
Is it permanent or transient? 

In node I used: "Node.js OpenTelemetry agent is starting".

agents/python/opamp/http_client.py
```diff
@@ -166,9 +220,9 @@ def mandatory_env_vars_set(self):
     
     def shutdown(self):
         self.running = False
-        
         opamp_logger.info("Sending agent disconnect message to OpAMP server...")
-        disconnect_message = opamp_pb2.AgentToServer(agent_disconnect=opamp_pb2.AgentDisconnect())
+        agent_health = self.get_agent_health(component_health=False, last_error="Component shutting down", status="ProcessTerminated")
```
blumamir: Here too - what is component? what can the user understand from this text?

We will not send any further update, thus this text will report "shutting down" even after the process has exited.

agents/python/opamp/http_client.py
```diff
@@ -76,7 +111,13 @@ def worker(self):
 
                     if server_to_agent.flags & opamp_pb2.ServerToAgentFlags_ReportFullState:
                         opamp_logger.info("Received request to report full state")
-                        server_to_agent = self.send_full_state()
+                        
+                        agent_description = self.get_agent_description()
+                        agent_health = self.get_agent_health(component_health=True, status="Healthy")
```
blumamir: Is this code guaranteed to be executed only when the agent is healthy?

blumamir (approved):

blumamir (approved):

PR #1433, opened 2024-08-07
Update Odigos collector to v0.106.0

RonFed:
Relevant breaking changes in the collector that are addressed in this PR (from the [contribe releases](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases) and [collector release](https://github.com/open-telemetry/opentelemetry-collector/releases).

> The otlpreceiver now uses localhost by default instead of 0.0.0.0. This may break the receiver in containerized environments like Kubernetes. If you depend on 0.0.0.0 disable the component.UseLocalHostAsDefaultHost feature gate or explicitly set the endpoint to 0.0.0.0

> confmap: When passing configuration for a string field using any provider, use the verbatim string representation as the value. (https://github.com/open-telemetry/opentelemetry-collector/issues/10605, https://github.com/open-telemetry/opentelemetry-collector/issues/10405)
This matches the behavior of ${ENV} syntax prior to the promotion of the confmap.unifyEnvVarExpansion feature gate
to beta. It changes the behavior of the ${env:ENV} syntax with escaped strings


RonFed:
- [ ] Check with destinations which requires an API key/tolen

edeNFed (approved):

PR #1439, opened 2024-08-11
Shutdown opamp when agent fail

tamirdavid1:


agents/python/opamp/http_client.py
```diff
@@ -52,16 +52,32 @@ def start(self, python_version_supported: bool = None):
...
-            return
-        
-        self.send_first_message_with_retry()
-        
-        self.event.set()
+            
+            self.worker()
+        except Exception as e:
+            opamp_logger.error(f"Error running OpAMP client: {e}")
+            self.send_agent_failure_disconnect_message(error_message=str(e))
```
blumamir, line 67: LGTM, but I wonder how this bolock gets terminated when the process goes down nicely.
I assume `self.worker()` returns and the function is terminated without calling the exception code, right?

agents/python/opamp/http_client.py
```diff
@@ -52,16 +52,32 @@ def start(self, python_version_supported: bool = None):
...
+            
             self.event.set()
-            return
-        
-        self.send_first_message_with_retry()
-        
-        self.event.set()
+            
+            self.worker()
+        except Exception as e:
```
blumamir, line 65: It's always better to have these protections in place, just in case, to make sure we do not cause any issue with the instrumented application.

I do wonder, if there are things in the `_initialize_components` of the `confgiurator` which can potentially fail? for example, in nodejs the instrumentation packages might through if imported in old node versions

agents/python/opamp/health_status.py
```diff
@@ -4,4 +4,5 @@ class AgentHealthStatus(str, Enum):
     HEALTHY = "Healthy"
     STARTING = "Starting"
     UNSUPPORTED_RUNTIME_VERSION = "UnsupportedRuntimeVersion"
-    TERMINATED = "ProcessTerminated"
\ No newline at end of file
+    TERMINATED = "ProcessTerminated"
+    AGENT_FAILURE = "AgentFailure"
```
blumamir, line 8: Please also add it [here](https://github.com/odigos-io/odigos/blob/880f2ce4febd5309048dccb7caad737908fd0991/common/agent_health_status.go#L25) 🙏

blumamir (approved):
LGTM 💯 
left few nits/questions

agents/python/opamp/http_client.py
tamirdavid1: correct

agents/python/configurator/__init__.py
```diff
@@ -20,40 +20,45 @@ def _configure(self, **kwargs):
...
+        resource_attributes_event.wait(timeout=30)  # Wait for the resource attributes to be received for 30 seconds
 
-    # Reorder the python sys.path to ensure that the user application's dependencies take precedence over the agent's dependencies.
-    # This is necessary because the user application's dependencies may be incompatible with those used by the agent.
-    reorder_python_path()
-    # Reload distro modules to ensure the new path is used.
-    reload_distro_modules()
+        received_value = client.resource_attributes
+        
+        if received_value:
```
blumamir, line 31: what happens if this condition is not met? perhaps we should write some warning or notify opamp server? what can make it happen in practice?

agents/python/configurator/__init__.py
```diff
@@ -20,40 +20,45 @@ def _configure(self, **kwargs):
...
+            initialize_logging_if_enabled(log_exporters, resource)
+
+        # Reorder the python sys.path to ensure that the user application's dependencies take precedence over the agent's dependencies.
+        # This is necessary because the user application's dependencies may be incompatible with those used by the agent.
+        reorder_python_path()
+        # Reload distro modules to ensure the new path is used.
+        reload_distro_modules()
+        
+    except Exception as e:
+        client.shutdown(custom_failure_message=str(e))
```
blumamir, line 60: Do we want to also log something to the logger here? will it be logged anywhere else?

blumamir, line 60: is it possible at this point that `client` is not yet initialized and is `nil`?

tamirdavid1: It wont logged anywhere, if you remember i've added in the http-client logger but we need to manually enable it (via code change)

tamirdavid1: Yes it is great point thanks!
In case agent failed within the client creation it can be nil
for example, failing within start_opamp_client function [line 26]
fixed.

agents/python/configurator/__init__.py
tamirdavid1: In practice it is basically means that server was not able to receive it's first_message and report back the resource_attributes.

tamirdavid1: In this case, probably the opamp-server is down so application wont be instrumented

blumamir: should we have something written to log in this case? so if it happens we can diagnose it better

agents/python/configurator/__init__.py
```diff
@@ -20,40 +20,48 @@ def _configure(self, **kwargs):
...
+            initialize_logging_if_enabled(log_exporters, resource)
+
+        # Reorder the python sys.path to ensure that the user application's dependencies take precedence over the agent's dependencies.
+        # This is necessary because the user application's dependencies may be incompatible with those used by the agent.
+        reorder_python_path()
+        # Reload distro modules to ensure the new path is used.
+        reload_distro_modules()
+        
+    except Exception as e:
+        if client is not None:
```
blumamir, line 62: should we also write the exception to the logger here?
or will it be logger somewhere else?

agents/python/configurator/__init__.py
tamirdavid1: There is no logger available for the configurator

agents/python/configurator/__init__.py
tamirdavid1: there is no logger available in the configurator

PR #1441, opened 2024-08-13
fix: remove only protected c files and their directories

tamirdavid1:
During an issue with eBPF instrumentation breaking upon odiglet restarts, I implemented a "smart" recreation of instrumentation files. The goal was to retain compiled C files when odiglet restarted. However, this approach led to other files not being properly recreated in certain cases. For example, if the instrumentation version was upgraded, the host filesystem could end up with both versions, which caused problems, especially in Python.

This PR addresses the issue by ensuring that everything is deleted and recreated except for the C files and their directories. For instance, in `/var/odigos/python-ebpf/`, everything will be recreated except for `pythonUSDT.abi3.so`.


odiglet/pkg/instrumentation/fs/copy.go
```diff
@@ -85,28 +85,25 @@ func worker(fileChan <-chan string, sourceDir, destDir string, wg *sync.WaitGrou
...
 		if !d.IsDir() {
-
 			if !CopyCFiles {
-				// filter out C files in ebpf directories
-				if strings.Contains(filepath.Dir(path), "ebpf") {
-					switch ext := filepath.Ext(path); ext {
-					case ".so", ".node", ".node.d", ".a":
-						return nil
-					}
+				if _, found := filesToKeepMap[strings.Replace(path, "/instrumentations/", "/var/odigos/", 1)]; found {
```
edeNFed, line 96: Can we avoid calling `strings.Replace` on every file by saving everything in `filesToKeepMap` as `/instrumentations/.....` instead of `/var/odigos/....`

tamirdavid1: Hey @edeNFed,
I'm using also in `removeFilesInDir` (which remove it from the hostDir which is /var/odigos)

edeNFed: OK Got it

odiglet/pkg/instrumentation/fs/remove.go
```diff
@@ -4,43 +4,59 @@ import (
...
-	"strings"
 
 	"github.com/odigos-io/odigos/odiglet/pkg/log"
 )
 
-func removeFilesInDir(hostDir string) error {
+func removeFilesInDir(hostDir string, filesToKeep map[string]struct{}) error {
 	shouldRecreateCFiles := ShouldRecreateAllCFiles()
-	log.Logger.V(0).Info(fmt.Sprintf("Removing files in directory: %s, shouldRecreateCFiles: %s", hostDir, fmt.Sprintf("%t", shouldRecreateCFiles)))
+	log.Logger.V(0).Info(fmt.Sprintf("Removing files in directory: %s, shouldRecreateCFiles: %t", hostDir, shouldRecreateCFiles))
```
edeNFed, line 13: nit: probably better to use structured logging (arguments as key-value pairs instead of one formatted string)

PR #1445, opened 2024-08-15
Head sampling crd

tamirdavid1:
I've added @RonFed CRD additions for Head-sampling + change in the relevant places in the code.


api/odigos/v1alpha1/instrumentatioconfig_types.go
```diff
@@ -41,16 +42,76 @@ type SdkConfig struct {
...
+	// Used as a fallback if all rules evaluate to false,
+	// it may be empty - in this case the default value will be 1 - all spans are sampled.
+	// it should be a float value in the range [0, 1] - the fraction of spans to sample.
+	// a value of 0 means no spans are sampled if none of the rules evaluate to true.
+	FallbackFraction *float64 `json:"fallbackFraction,omitempty"`
 }
 
 type InstrumentationLibraryConfig struct {
 	// The name of the instrumentation library
 	// - Node.js: The name of the npm package: `@opentelemetry/instrumentation-<name>`
```
RonFed, line 103: This comment should be updated to describe the ID and it's differences between the SDKs

api/odigos/v1alpha1/instrumentatioconfig_types.go
```diff
@@ -41,16 +42,76 @@ type SdkConfig struct {
...
+}
+
+// 'Operand' represents the attributes and values that an operator acts upon in an expression
+type AttributeCondition struct {
+	// attribute key (e.g. "url.path")
+	Key attribute.Key `json:"key"`
+	// currently only string values are supported.
+	Val string `json:"val"`
+	// The operator to use to compare the attribute value.
+	Operator Operator `json:"operator"`
```
RonFed, line 59: Maybe we should have `omitempty` here for the operator. This will result in the default operation being `equals` if we define a default value for the `Operator` as suggested in the other comment

api/odigos/v1alpha1/instrumentatioconfig_types.go
```diff
@@ -41,16 +42,76 @@ type SdkConfig struct {
...
+}
+
+type Operator string
+
+const (
+	Equals    Operator = "equals"
+	NotEquals Operator = "notEquals"
+	Contain   Operator = "contain"
```
blumamir, line 67: I wonder, since this operator cannot be implemented with eBPF, perhaps we can omit it now and consider it in the future when we meet a real usecase. then we need to make sure it can only be used with non eBPF SDKs.

api/odigos/v1alpha1/instrumentatioconfig_types.go
```diff
@@ -41,16 +42,76 @@ type SdkConfig struct {
...
+// AttributesAndSampler is a set of attribute compare samplers that are ANDed together.
+// If all attribute compare samplers evaluate to true, the AND sampler evaluates to true,
+// and the fraction is used to determine the sampling decision.
+// If any of the attribute compare samplers evaluate to false,
+// the fraction is not used and the rule is skipped.
+//
+// An "empty" AttributesAndSampler with no operands is considered to always evaluate to true.
+// and the fraction is used to determine the sampling decision.
+type AttributesAndSamplerRule struct {
+	AttributeConditions []AttributeCondition `json:"operands"`
```
blumamir, line 81:
```suggestion
	AttributeConditions []AttributeCondition `json:"attributeConditions"`
```

api/odigos/v1alpha1/instrumentatioconfig_types.go
```diff
@@ -41,16 +42,76 @@ type SdkConfig struct {
...
+)
+
+// AttributesAndSampler is a set of attribute compare samplers that are ANDed together.
+// If all attribute compare samplers evaluate to true, the AND sampler evaluates to true,
+// and the fraction is used to determine the sampling decision.
+// If any of the attribute compare samplers evaluate to false,
+// the fraction is not used and the rule is skipped.
+//
+// An "empty" AttributesAndSampler with no operands is considered to always evaluate to true.
+// and the fraction is used to determine the sampling decision.
```
blumamir, line 79:
```suggestion
// AttributesAndSamplerRule is a set of AttributeCondition that are ANDed together.
// If all attribute conditions evaluate to true, the AND sampler evaluates to true,
// and the fraction is used to determine the sampling decision.
// If any of the attribute compare samplers evaluate to false,
// the fraction is not used and the rule is skipped.
//
// An "empty" AttributesAndSamplerRule with no attribute conditions is considered to always evaluate to true.
// and the fraction is used to determine the sampling decision.
```

api/odigos/v1alpha1/instrumentatioconfig_types.go
```diff
@@ -41,16 +42,76 @@ type SdkConfig struct {
...
+	// If the fraction is 1, all spans are sampled.
+	Fraction float64 `json:"fraction"`
+}
+
+// HeadSamplingConfig is a set of attribute rules.
+// The first attribute rule that evaluates to true is used to determine the sampling decision based on its fraction.
+//
+// If none of the rules evaluate to true, the fallback fraction is used to determine the sampling decision.
+type HeadSamplingConfig struct {
+	AttributesAndSamplerRules []AttributesAndSamplerRule `json:"rules"`
```
blumamir, line 93:
```suggestion
	AttributesAndSamplerRules []AttributesAndSamplerRule `json:"attributesAndSamplerRules"`
```

api/odigos/v1alpha1/instrumentatioconfig_types.go
```diff
@@ -41,16 +42,76 @@ type SdkConfig struct {
...
+
+const (
+	Equals    Operator = "equals"
+	NotEquals Operator = "notEquals"
+	Contain   Operator = "contain"
+	EndWith   Operator = "endWith"
+	StartWith Operator = "startWith"
+)
+
+// AttributesAndSampler is a set of attribute compare samplers that are ANDed together.
```
blumamir, line 72: Probably we can add here that this entity is refered to as a rule in our internal terminology for sampling

blumamir (commented):
LGTM.
added few nit suggestions about updating someplace to the new names, and consider removing the `contains` operand for now, as it might get us into problems with eBPF implementation which we can postpone

api/odigos/v1alpha1/instrumentatioconfig_types.go
```diff
@@ -41,16 +42,76 @@ type SdkConfig struct {
...
+// the fraction is not used and the rule is skipped.
+//
+// An "empty" AttributesAndSampler with no operands is considered to always evaluate to true.
+// and the fraction is used to determine the sampling decision.
+type AttributesAndSamplerRule struct {
+	AttributeConditions []AttributeCondition `json:"operands"`
+	// The fraction of spans to sample, in the range [0, 1].
+	// If the fraction is 0, no spans are sampled.
+	// If the fraction is 1, all spans are sampled.
+	Fraction float64 `json:"fraction"`
```
edeNFed, line 85: If no fraction is specified it will be 0.0 and all spans will be dropped, maybe we should change it to `*float64` to differentiate between empty and 0?

api/odigos/v1alpha1/instrumentatioconfig_types.go
edeNFed: nit: I think it should be `Contains` (with an s), or drop the s in the equals/notEquals

edeNFed:
Great job @tamirdavid1 
A few comments on naming (nothing really important, feel free to skip all of them):
- `.spec.config` and `.spec.sdkConfigs` is confusing. if I understand correctly, one is for sampling and the other is for dynamic instrumentation, so the names should reflect that (`sampling`/`dynamicAttributes`/etc)
- All the nested objects are fields in `InstrumentationConfig` object, I think we can drop the word `instrumentation` in multiple places in the json fields as it clear that everything is related to instrumentation.
- `operands`does not make it clear that there is an AND between all of them and that all should return true in order to pass. Maybe we can give it a better name to explain this behavior 

Another thing to consider:
I see both `.spec.config` and `.spec.sdkConfigs` eventually point to a list of target libraries, 
Would it make more sense to unify them both and make the libraries the root object?
Maybe something like:
```yaml
spec:
    runtimeDetailsInvalidated: false
    configs:
        - name: "net/http"
          traceConfig:
              enbaled: true
          headSampling:
              fallbackFraction: 0.4
              rules:
                  - fraction: 1
                    operands:
                        - key: "http.method"
                          operation: "equal"
                          value: "POST"
          dynamicAttributes:
              - spanKind: "server"
                optionValueBoolean: true
                optionKey: "http.request.body"
    global:
        headSampling:
            fallbackFraction: 0.9
```

It is a big change and I am not sure how much value it adds, but maybe now is a good time to do it if you find it valuable. (cc @blumamir @RonFed )

RonFed (approved):

PR #1446, opened 2024-08-16
test: add e2e test for odigos cli upgrade from latest to HEAD

blumamir:
Since odigos upgrade is a potentially sensative process which can fail for various reasons, this PR adds an e2e test that starts with the latest odigos version from github, and then calls cli upgrade to the current version in HEAD.

Then it generates traffic and verify that traces are created.

- This is a basic test for the most straight forward case. while it can catch many issues, we can consider extending it in the future to run and assert more aspects after the upgrade
- It currently only runs cli upgrade and not helm upgrade. since we are soon going to migrate cli to helm as well, I think it can be taken care of after the refactor. @rauno56 
- I duplicated many of the existing yamls from existing tests, with slight modifications. looks like we can extract some of it to a common directory and reuse it?


cli/cmd/upgrade.go
```diff
@@ -42,52 +42,58 @@ and apply any required migrations and adaptations.`,
...
-			os.Exit(1)
-		}
+		var operation string
 
-		currOdigosVersion := cm.Data["ODIGOS_VERSION"]
-		if currOdigosVersion == "" {
-			fmt.Println("Odigos upgrade failed - unable to read the current Odigos version for migration")
-			os.Exit(1)
-		}
+		if !cmd.Flag("skip-version-check").Changed {
```
blumamir, line 47: the change here is adding this new flag since the current upgrade behavior requires a valid formatted version `v1.2.3` while the tests use the tag `e2e-test`. this flag allows the test to bypass the checks

tests/e2e/cli-upgrade/assert-odigos-upgraded.yaml
```diff
@@ -0,0 +1,158 @@
...
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    app.kubernetes.io/name: odigos-autoscaler
+  namespace: odigos-test-cli-upgrade
+spec:
+  containers:
+    - name: manager
+      image: keyval/odigos-autoscaler:e2e-test
```
blumamir, line 17: review note: i tested it with an illegal tag value to make sure this is asserted correctly

edeNFed (approved):
looks great!

PR #1447, opened 2024-08-16
feat: polish last bits after OdigosConfig CR removal

rauno56:
Check the commits for list of changes. I added `odigos.io/config` and `odigos.io/system-object` labels to all CRDs, not just the tested.


cli/pkg/autodetect/kind.go
```diff
@@ -22,7 +22,7 @@ var CurrentKubernetesVersion KubernetesVersion
 const (
 	KindUnknown   Kind = "Unknown"
 	KindMinikube  Kind = "Minikube"
-	KindKind      Kind = "KinD"
```
blumamir, line 25: I think this was written like so intentionally - "**K**ubernetes in **D**ocker".
@edeNFed

api/actions/v1alpha1/addclusterinfo_types.go
```diff
@@ -56,6 +56,8 @@ type AddClusterInfoStatus struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:resource:path=addclusterinfos,scope=Namespaced,shortName=aci
+//+kubebuilder:metadata:labels=metadata.labels.odigos.io/config=true
+//+kubebuilder:metadata:labels=metadata.labels.odigos.io/system-object=1
```
blumamir, line 60: I think they are opposite, should it be?
```suggestion
//+kubebuilder:metadata:labels=metadata.labels.odigos.io/config=1
//+kubebuilder:metadata:labels=metadata.labels.odigos.io/system-object=true
```

cli/pkg/autodetect/kind.go
rauno56: Nowhere on their site, they use that form.

edeNFed: I am ok with both

blumamir (approved):

PR #1449, opened 2024-08-17
refactor: cleanups for k8sutils workload package. create enums and move common functions

blumamir:
This is another step in the way to move everything related to common things into the common packages, and avoid duplications and untyped usages in the project.

This PR:
- adds more files to the `workload` package in k8sutils to differentiate between the different utils and enums
- enums for the workload kinds
- function to handle and convert workload kinds
- rename function to common format
- replace usages in controllers code to use enums where possible

There are countless places to go over and improve. This PR is touching on the most easy and straight forward cases I spotted.


k8sutils/pkg/workload/workloadkinds.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func IgnoreErrorKindNotSupported(err error) error {
+	if IsErrorKindNotSupported(err) {
+		return nil
+	}
+	return err
+}
+
+func WorkloadKindLowerCaseFromKind(pascalCase WorkloadKind) WorkloadKindLowerCase {
```
RonFed, line 47: It is equivalent to `strings.ToLower`. I'm not sure why we need types for both pascal and lowercase. We can have just the pascal one and use `ToLower` when needed in the utils functions

odiglet/pkg/kube/instrumentation_ebpf/pods.go
```diff
@@ -106,20 +106,16 @@ func (p *PodsReconciler) instrumentWithEbpf(ctx context.Context, pod *corev1.Pod
...
 	for _, owner := range pod.OwnerReferences {
-		name, kind, err := kubeutils.GetWorkloadNameFromOwnerReference(owner)
+		workloadName, workloadKind, err := workload.GetWorkloadFromOwnerReference(owner)
 		if err != nil {
-			if kubeutils.IsErrorKindNotSupported(err) {
-				return nil, nil
-			}
-
-			return nil, err
+			return nil, workload.IgnoreErrorKindNotSupported(err)
```
RonFed, line 113: We used to ignore (not return the error) the ErrorKindNotSupported. Is this change in behavior intentional?

RonFed (approved):

k8sutils/pkg/workload/workloadkinds.go
blumamir: yes, since it's used in several places, I choose to have a type for it as enum

PR #1454, opened 2024-08-19
Use address 0.0.0.0 to ui

brunocascio:
by using localhost the ui is only accesible through localhost and non-routable from the outside


blumamir (approved):
Thanks!

Out of curiosity, where do try to access the ui pod from?

brunocascio:
@blumamir I'm not accesing the pod directly but through the service and then the ingress.

The thing is the ui is listen to loopback address, so there's no way to access it through other net interface

PR #1455, opened 2024-08-20
Move opamp to different repo

tamirdavid1:


odiglet/Dockerfile
```diff
@@ -4,9 +4,9 @@
 FROM python:3.11 AS python-builder
 ARG ODIGOS_VERSION
 WORKDIR /python-instrumentation
-COPY ../agents/python ./agents/python
-RUN echo "VERSION = \"$ODIGOS_VERSION\";" > ./agents/python/configurator/version.py
-RUN mkdir workspace && pip install ./agents/python/  --target workspace
+COPY ../agents/python ./agents/configurator
```
blumamir, line 7: Since we removed the `agents` directory in this PR - how can this work?

agents/python/configurator/__init__.py
```diff
@@ -1,102 +1,15 @@
...
-import sys
-import os
 import opentelemetry.sdk._configuration as sdk_config
-
-from opentelemetry.sdk.resources import Resource
-from opentelemetry.sdk.resources import ProcessResourceDetector, OTELResourceDetector
-from .lib_handling import reorder_python_path, reload_distro_modules
-from .version import VERSION
-from opamp.http_client import OpAMPHTTPClient
+from initializer.components import _initialize_components
```
blumamir, line 2: I think that something starting with underscore in python is meant to be private and not imported from outside

blumamir (approved):

agents/python/configurator/__init__.py
tamirdavid1: Thanks for that comment, changed.

blumamir (approved):

PR #1462, opened 2024-08-25
feat: NGINX Version 

yodigos:


procdiscovery/pkg/inspectors/nginx/nginx.go
```diff
@@ -13,12 +15,40 @@ type NginxInspector struct{}
...
+	version, err := GetNginxVersion(podIp)
+	if err != nil {
+		return ""
+	}
+
+	return version
+}
+
+func GetNginxVersion(podIP string) (string, error) {
+	resp, err := http.Get("http://" + podIP)
```
edeNFed, line 36: I think this assumes port is always 80?

procdiscovery/pkg/inspectors/nginx/nginx.go
```diff
@@ -13,12 +15,40 @@ type NginxInspector struct{}
...
+	if err != nil {
+		return "", nil
+	}
+	defer resp.Body.Close()
+
+	serverHeader := resp.Header.Get("Server")
```
edeNFed, line 42: This may be a potential security issue (a malicious server returns a 1000Gb header for example) can we read only the first X chars?

procdiscovery/pkg/inspectors/nginx/nginx.go
```diff
@@ -13,12 +15,40 @@ type NginxInspector struct{}
...
+	if serverHeader == "" {
+		return "", nil
+	}
+
+	re := regexp.MustCompile(`nginx/(\d+\.\d+\.\d+)`)
```
edeNFed, line 47: Worth making a const

procdiscovery/pkg/inspectors/langdetect.go
```diff
@@ -23,7 +23,8 @@ func (e ErrLanguageDetectionConflict) Error() string {
 }
 
 type inspector interface {
-	Inspect(process *process.Details) (common.ProgramLanguageDetails, bool)
+	Inspect(process *process.Details) (common.ProgrammingLanguage, bool)
+	GetRuntimeVersion(process *process.Details, podIp string) string
```
edeNFed, line 27: Maybe we should make this a separate interface that inherits this interface? This way we can avoid the empty implementation on all the non supported languages

RonFed, line 27: Maybe worth considering returning a  [version object](https://pkg.go.dev/github.com/hashicorp/go-version#Version) instead of a string 
This way we can do checks like GraterThan out of the box

RonFed, line 27: Also, pod IP seems specific to be included in the generic interface

procdiscovery/pkg/inspectors/nginx/nginx.go
yodigos: I'm reading a specific key in this call. You think that the Value of this key in header can be big? Or all the Header?

procdiscovery/pkg/inspectors/langdetect.go
yodigos: Let's talk on this f2f

odiglet/pkg/kube/runtime_details/inspection.go
```diff
@@ -119,10 +120,15 @@ func runtimeInspection(pods []corev1.Pod, ignoredContainers []string) ([]odigosv
 				}
 			}
 
+			runtimeVersion := ""
```
RonFed, line 123:
```suggestion
			var runtimeVersion string
```

procdiscovery/pkg/inspectors/golang/golang.go
```diff
@@ -3,25 +3,36 @@ package golang
...
 
-	programLanguageDetails.Language = common.GoProgrammingLanguage
-	if buildInfo != nil {
-		programLanguageDetails.RuntimeVersion = buildInfo.GoVersion
+	return common.GoProgrammingLanguage, true
+}
+
+func (g *GolangInspector) GetRuntimeVersion(p *process.Details, containerURL string) *version.Version {
+	file := fmt.Sprintf("/proc/%d/exe", p.ProcessID)
+	buildInfo, err := buildinfo.ReadFile(file)
```
RonFed, line 28: nit: We are calling ReadFile twice for each Go process, if it is possible to reuse the buildInfo from the Inspect function it will be better

procdiscovery/pkg/inspectors/java/java.go
```diff
@@ -10,18 +12,22 @@ import (
...
+		return common.JavaProgrammingLanguage, true
+	}
+
+	return "", false
+}
 
-		return programLanguageDetails, true
+func (j *JavaInspector) GetRuntimeVersion(proc *process.Details, containerURL string) *version.Version {
+	if value, exists := proc.GetDetailedEnvsValue(process.JavaVersionConst); exists {
+		re := regexp.MustCompile(JavaVersionRegex)
```
RonFed, line 27: define as a global const, avoiding re-compiliong the regex each time

procdiscovery/pkg/inspectors/nginx/nginx.go
```diff
@@ -11,14 +14,45 @@ import (
...
+		return "", nil
+	}
+	defer resp.Body.Close()
+
+	serverHeader := resp.Header.Get("Server")
+	if serverHeader == "" {
+		return "", nil
+	}
+
+	re := regexp.MustCompile(NginxVersionRegex)
```
RonFed, line 51: Same as the comment for the other regex - use as a const

procdiscovery/pkg/inspectors/golang/golang.go
```diff
@@ -3,25 +3,36 @@ package golang
...
+	if err != nil || buildInfo == nil {
+		return nil
 	}
+	match := re.FindStringSubmatch(buildInfo.GoVersion)
```
RonFed, line 34: I'm not sure why we need this regex here. The buildinfo package is an official go package which should provide a valid go version if one exists.

RonFed (approved):

PR #1463, opened 2024-08-25
build Odigos collector and run unit tests in PRs

RonFed:


edeNFed (approved):

collector/exporters/azureblobstorageexporter/metadata.yaml
```diff
@@ -6,4 +6,7 @@ status:
     beta: [traces, logs]
   distributions: [contrib]
   codeowners:
-    active: [edeNFed]
\ No newline at end of file
+    active: [edeNFed]
+
+tests:
+  expect_consumer_error: true
```
tamirdavid1, line 12: What the idea of it?

tamirdavid1 (commented):
general question

collector/exporters/azureblobstorageexporter/metadata.yaml
RonFed: When using the unit tests which are auto-generated by the `mdatagen` tool, we can't really export to azure/googlecloud - so this is a way to configure the tool to avoid this check in the tests.
In general, the exporters' tests did problems and they are not being run in CI (currently only the processors tests are running in CI).

PR #1464, opened 2024-08-25
fix: cli-install is broken post work file deletion

tamirdavid1:


Makefile
```diff
@@ -192,7 +192,7 @@ check-clean-work-tree:
 .PHONY: cli-install
 cli-install:
 	@echo "Installing odigos from source. version: $(ODIGOS_CLI_VERSION)"
-	go run -tags=embed_manifests ./cli install --version $(ODIGOS_CLI_VERSION)
+	cd ./cli ; go run -tags=embed_manifests . install --version $(ODIGOS_CLI_VERSION)
```
RonFed, line 195: same for cli-upgrade below

tamirdavid1: Thanks, fixed

RonFed (approved):

PR #1472, opened 2024-08-27
Updated the docs page to include http.route attribute support warning

dovzhikova:


docs/instrumentations/golang/golang.mdx
```diff
@@ -19,7 +19,10 @@ The following go modules will be auto instrumented by Odigos:
 
 ### HTTP Frameworks
 
-- [`github.com/gin-gonic/gin`](https://pkg.go.dev/github.com/gin-gonic/gin) <sup>*</sup>. HTTP web framework
+<Warning>Please note that in order for the `http.route` parameter to be properly extracted by Go auto instrumentation, 
+you should use Go 1.22 or above.</Warning>
```
blumamir, line 23: This section contains 2 list items, but the comment is only relevant to `net/http`. we need to structure it correctly

blumamir, line 23: go 1.22 is not enough to make this attribute appear.

The code also needs to be written in a specific way

dovzhikova: Amir, I addressed your concerns and updated the doc

PR #1475, opened 2024-08-27
Nginx UI condition

yodigos:


instrumentor/controllers/instrumentationdevice/common.go
```diff
@@ -210,6 +210,12 @@ func reconcileSingleWorkload(ctx context.Context, kubeClient client.Client, runt
 		return err
 	}
 
+	if runtimeDetails.Spec.RuntimeDetails[0].Language == "nginx" {
```
yodigos, line 213: It looks like it may do bug for other sources.. Checking

yodigos, line 213: I think I may need to run all the runtimeDetails.Spec.RuntimeDetails, even though I'm not sure when it's more then one.

instrumentor/controllers/instrumentationdevice/common.go
```diff
@@ -26,6 +27,7 @@ const (
 	ApplyInstrumentationDeviceReasonNoRuntimeDetails       ApplyInstrumentationDeviceReason = "NoRuntimeDetails"
 	ApplyInstrumentationDeviceReasonErrApplying            ApplyInstrumentationDeviceReason = "ErrApplyingInstrumentationDevice"
 	ApplyInstrumentationDeviceReasonErrRemoving            ApplyInstrumentationDeviceReason = "ErrRemovingInstrumentationDevice"
+	ApplyInstrumentationDeviceReasonNotSupported           ApplyInstrumentationDeviceReason = "ErrInstrumentationNotSupported"
```
blumamir, line 30: It's not the instrumentation that is not supported. The version is not supported

procdiscovery/pkg/inspectors/nginx/nginx.go
```diff
@@ -11,14 +14,45 @@ import (
...
+func (j *NginxInspector) GetRuntimeVersion(p *process.Details, containerURL string) *version.Version {
+	nginxVersion, err := GetNginxVersion(containerURL)
+	if err != nil {
+		return nil
+	}
+
+	return common.GetVersion(nginxVersion)
+}
+
+func GetNginxVersion(containerURL string) (string, error) {
```
blumamir, line 39: The `error` in return types is always `nil`, do we want to return errors from this function when it fails?

api/go.mod
```diff
@@ -16,6 +16,7 @@ require (
 
 require (
 	github.com/fxamacker/cbor/v2 v2.7.0 // indirect
+	github.com/hashicorp/go-version v1.7.0 // indirect
```
blumamir, line 19: Do you understand where this dependency is coming from?
Does the `api` module has any dependency on procdiscovery?

odiglet/pkg/kube/utils/utils.go
```diff
@@ -50,3 +51,11 @@ func GetResourceAttributes(podWorkload *workload.PodWorkload, podName string) []
 
 	return attrs
 }
+
+func GetPodExternalURL(ip string, ports []corev1.ContainerPort) string {
+	if ports != nil && len(ports) > 0 {
```
blumamir, line 56: I think `ports != nil` is redundant since `len(ports) > 0` already covers this case

procdiscovery/pkg/inspectors/langdetect.go
```diff
@@ -22,11 +22,15 @@ func (e ErrLanguageDetectionConflict) Error() string {
...
 
-type inspector interface {
-	Inspect(process *process.Details) (common.ProgramLanguageDetails, bool)
+type LanguageInspector interface {
+	Inspect(process *process.Details) (common.ProgrammingLanguage, bool)
 }
 
-var inspectorsList = []inspector{
+type VersionInspector interface {
+	GetRuntimeVersion(process *process.Details, containerURL string) *version.Version
```
blumamir, line 30: nit: By returning `version.Version` here, we need to transform the version string in every inspector instead of once where it is invoked.

Did you consider simplifying the api and returning string here?

procdiscovery/pkg/inspectors/nginx/nginx.go
```diff
@@ -11,14 +14,45 @@ import (
 // but in order to avoid huge refactoring we are adding it as a language for now
 type NginxInspector struct{}
 
-const NginxProcessName = "nginx"
+const (
+	NginxProcessName  = "nginx"
+	NginxVersionRegex = `nginx/(\d+\.\d+\.\d+)`
+)
```
blumamir, line 20: Should these be private?

```suggestion
const (
	nginxProcessName  = "nginx"
	nginxVersionRegex = `nginx/(\d+\.\d+\.\d+)`
)
```

common/lang_detection.go
```diff
@@ -23,3 +25,11 @@ const (
...
 	IgnoredProgrammingLanguage ProgrammingLanguage = "ignored"
 )
+
+func GetVersion(versionString string) *version.Version {
+	runtimeVersion, err := version.NewVersion(versionString)
+	if err != nil {
+		return nil
+	}
+	return runtimeVersion
+}
```
blumamir, line 35: Perhaps?

```suggestion
func ParseVersion(versionString string) *version.Version {
	runtimeVersion, err := version.NewVersion(versionString)
	if err != nil {
		return nil
	}
	return runtimeVersion
}
```

common/lang_detection.go
```diff
@@ -23,3 +25,11 @@ const (
 	// Ignored is used when the odigos is configured to ignore the process/container
 	IgnoredProgrammingLanguage ProgrammingLanguage = "ignored"
 )
+
+func GetVersion(versionString string) *version.Version {
+	runtimeVersion, err := version.NewVersion(versionString)
+	if err != nil {
+		return nil
```
blumamir, line 32: Do we want to log it somewhere?

blumamir (approved):
Overall looks good.

Added few nit comments

PR #1476, opened 2024-08-28
chore: migrate docs from odigos-charts

rauno56:


blumamir (approved):
Awesome thanks @rauno56

PR #1477, opened 2024-08-29
refactor: instrumentation libraries config in CRD

blumamir:


api/odigos/v1alpha1/instrumentationconfig_types.go
```diff
@@ -24,11 +24,6 @@ type InstrumentationConfigSpec struct {
 	// true when the runtime details are invalidated and should be recalculated
 	RuntimeDetailsInvalidated bool `json:"runtimeDetailsInvalidated,omitempty"`
 
-	// config for this workload.
-	// the config is a list to allow for multiple config options and values to be applied.
-	// the list is processed in order, and the first matching config is applied.
-	Config []WorkloadInstrumentationConfig `json:"config,omitempty"`
```
RonFed, line 30: If we remove this, we should remove the struct definition for `WorkloadInstrumentationConfig`.

api/odigos/v1alpha1/instrumentationconfig_types.go
```diff
@@ -105,6 +100,10 @@ type InstrumentationLibraryConfig struct {
 	InstrumentationLibraryId InstrumentationLibraryId `json:"libraryId"`
 
 	TraceConfig *InstrumentationLibraryConfigTraces `json:"traceConfig,omitempty"`
+
+	// Configuration boolean options which are enabled for this instrumentation library.
+	// All options are disabled by default. To enable an option, add it's name to this list.
+	EnabledOptions []string `json:"enabledOptions,omitempty"`
```
RonFed, line 106: I think that a string alone for an option might be too specific, we can have specific options in the future which might have an integer value for them.
An option can have a string key and an *optional* string value - WDYT?

RonFed (changes requested):

api/odigos/v1alpha1/instrumentationconfig_types.go
blumamir: I think in the future, we will have options with various types: numbers, strings, string arrays, etc.
Since they have different types, I want to have different properties in the CRD for each type.

For boolean options, I was thinking we only need to have them if they are `true` so the `false` value is the default.

That is why I choose to write it this way and postpone the other types for the future. 
But perhaps it's better to have the final version here so that we don't need to make other changes down the road

api/odigos/v1alpha1/instrumentationconfig_types.go
blumamir: right!
good catch. fixed

api/odigos/v1alpha1/instrumentationconfig_types.go
blumamir: made it better with support for multiple types

tamirdavid1 (approved):

api/odigos/v1alpha1/instrumentationconfig_types.go
```diff
@@ -101,10 +96,63 @@ type HeadSamplingConfig struct {
 	FallbackFraction float64 `json:"fallbackFraction"`
 }
 
+// The InstrumentationLibraryCapabilityParameter represents a single configuration property that can be set for an instrumentation library option.
+// For example, the capability might be payload collection, and the parameter can be the maximum size of the payload to record,
+// if to skip recording incomplete payloads, conditions for recording payload (for example only for specific mime types - "only application/json"),
+// additional attributes to record original length, etc.
+type InstrumentationLibraryCapabilityParameter struct {
+	// The name of the property that should be configured for the instrumentation library option
+	ParameterName string `json:"parameterName"`
```
RonFed, line 105: What is the difference between this and `CapabilityName`?
If possible, adding an example in the comments will make it clearer

PR #1481, opened 2024-09-01
eBPF config provider

RonFed:
* Add an extension to the existing `OtelEbpfSdk` which allows to send configuration to a running SDK - the extended interface is called `ConfigurableOtelEbpfSdk`.
* Implement a basic config provider which is generic but currently only used by the Go instrumentation.
* Update the dependency of `go.opentelemetry.io/auto`
* Add `Languages` method to `InstrumentationConfig` which returns the relevant languages specified in the CRD - this allows filtering the call to the relevant eBPF directors.


odiglet/pkg/ebpf/configprovider.go
```diff
@@ -0,0 +1,69 @@
...
+	if c.stopped {
+		return nil
+	}
+
+	close(c.configChan)
+	c.stopped = true
+	return nil
+}
+
+func (c *ConfigProvider[C]) Watch() <-chan C {
```
edeNFed, line 45: Shouldn't we pass `context.Context` here as well? And maybe cancel it when we want to stop watching?

odiglet/pkg/ebpf/configprovider.go
```diff
@@ -0,0 +1,69 @@
...
+		log.Logger.Info("SendConfig called on stopped configProvider, the supplied config will be ignored")
+		return nil
+	}
+
+	// send a config or potentially return an error on timeout
+	// TODO: we should decide on a timeout value for sending config.
+	select {
+	case c.configChan <- newConfig:
+		return nil
+	case <-ctx.Done():
```
edeNFed, line 63: We assume the caller passed `ctx` which has a timeout, maybe a better design would be to get both context and `time.Time` or something as arguments and wrap the context in this function instead of assuming the caller did it

odiglet/pkg/ebpf/configprovider.go
RonFed: I'm not sure since it is ment to be a getter for the channel. 
The calling code in the instrumentation looks like:
```
	for {
		select {
		case <-ctx.Done():
			return
		case c, ok := <-m.cp.Watch():
			if !ok {
				m.logger.Info("Configuration provider closed, configuration updates will no longer be received")
				return
			}
			err := m.applyConfig(c)
			if err != nil {
				m.logger.Error(err, "Failed to apply config")
				continue
			}
			m.currentConfig = c
		}
	}
```
The way to stop watching should be to call `Shutdown` - which is called by the instrumentation once it is closing

edeNFed (approved):

tamirdavid1 (approved):

PR #1489, opened 2024-09-05
feat: send relevant instrumentationConfig to the opamp agents

tamirdavid1:


opampserver/pkg/sdkconfig/configsections/instrumentationconfig.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	return instrumentationConfig, nil
+}
+
+func FilterRelevantSdk(instrumentationConfig *v1alpha1.InstrumentationConfig, programmingLanguage string) (*protobufs.AgentConfigFile, error) {
+	relevantSdkConfig := v1alpha1.SdkConfig{}
+
+	for _, sdkConfig := range instrumentationConfig.Spec.SdkConfigs {
+		if string(sdkConfig.Language) == programmingLanguage {
```
blumamir, line 28: the sdkConfig is odigos internal enum as specified [here](https://github.com/odigos-io/odigos/blob/ab1af291b400a213d975a4cc0058a54bf9107258/common/lang_detection.go#L11) while the `programmingLanguage` conforms to the OpenTelemetry semantic conventions [here](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/resource#telemetry-sdk).

Looks like we call it `javascript` while OTEL calls it `nodejs`.

opampserver/pkg/sdkconfig/configsections/instrumentationconfig.go
```diff
@@ -0,0 +1,45 @@
...
+		return nil, err
+	}
+
+	return instrumentationConfig, nil
+}
+
+func FilterRelevantSdk(instrumentationConfig *v1alpha1.InstrumentationConfig, programmingLanguage string) (*protobufs.AgentConfigFile, error) {
+	relevantSdkConfig := v1alpha1.SdkConfig{}
+
+	for _, sdkConfig := range instrumentationConfig.Spec.SdkConfigs {
```
blumamir, line 27: What if we do not find and sdkConfig for this language? should we return an empty config or error?

opampserver/pkg/sdkconfig/sdkconfig.go
```diff
@@ -80,10 +80,21 @@ func (m *SdkConfigManager) GetFullConfig(ctx context.Context, remoteResourceAttr
 		return nil, err
 	}
 
+	// We are moving towards passing all Instrumentation capabilities unchanged within the instrumentationConfig to the opamp client.
+	// Gradually, we will migrate the InstrumentationLibraryConfigs and SDK remote config into the instrumentationConfig and the agents to use it.
+	instrumentationConfig, err := configsections.GetWorkloadInstrumentationConfig(ctx, m.mgr.GetClient(), instrumentedAppName, podWorkload.Namespace)
```
blumamir, line 85: `instrumentationConfig` can be `nil` in case we got `apierrors.IsNotFound(err)`. in which case, the `FilterRelevantSdk` function below will panic when trying to access it

opampserver/pkg/connection/conncache.go
```diff
@@ -142,3 +143,12 @@ func (c *ConnectionsCache) UpdateAllConnectionConfigs(connConfigEvaluator func(c
 		}
 	}
 }
+
+func (c *ConnectionsCache) GetConnectionInfoByWorkload(podWorkload workload.PodWorkload) (*ConnectionInfo, error) {
```
blumamir, line 147: we can have multiple connections for the same workload, right? should this be a slice?

opampserver/pkg/connection/conncache.go
```diff
@@ -142,3 +143,12 @@ func (c *ConnectionsCache) UpdateAllConnectionConfigs(connConfigEvaluator func(c
...
+	for _, conn := range c.liveConnections {
```
blumamir, line 148: I think we should protect this `liveConnection` access with:

```go
	c.mux.Lock()
	defer c.mux.Unlock()
```

as it can be accessed by multiple goroutiens at the same time

opampserver/pkg/connection/conncache.go
```diff
@@ -142,3 +143,12 @@ func (c *ConnectionsCache) UpdateAllConnectionConfigs(connConfigEvaluator func(c
...
+		if conn.Workload == podWorkload {
+			return conn, nil
+		}
+	}
+	return nil, fmt.Errorf("no connection found for workload %v", podWorkload)
```
blumamir, line 153: I think this is not an error

blumamir (approved):

common/lang_detection.go
```diff
@@ -26,6 +31,16 @@ const (
 	IgnoredProgrammingLanguage ProgrammingLanguage = "ignored"
 )
 
+func MapOdigosToSemConv(odigosPrograminglang ProgrammingLanguage) string {
+	fmt.Println(semconv.TelemetrySDKLanguageNodejs.Value.AsString())
```
blumamir, line 35: oops :)

common/lang_detection.go
```diff
@@ -26,6 +31,16 @@ const (
...
+	switch odigosPrograminglang {
+	case JavascriptProgrammingLanguage:
+		return semconv.TelemetrySDKLanguageNodejs.Value.AsString()
+	default:
```
blumamir, line 39: While we currently only care about python and nodejs, this function might be called in the future with other odigos programming languages which has no otel equivalents (`mysql`, `nginx`, etc).

Should we handle this here and make sure we only take what we can map correctly?

opampserver/pkg/connection/conncache.go
```diff
@@ -142,3 +154,18 @@ func (c *ConnectionsCache) UpdateAllConnectionConfigs(connConfigEvaluator func(c
 		}
 	}
 }
+
+func (c *ConnectionsCache) GetConnectionsInfoByWorkload(podWorkload workload.PodWorkload) []*ConnectionInfo {
```
blumamir, line 158: I think we can delete this one, right?

blumamir (approved):

common/lang_detection.go
tamirdavid1: added a comment for this as discussed

PR #1499, opened 2024-09-14
feat: otel sdk instrumentation rule

blumamir:
Add a instrumentation rule which allows to configure the SDKs to use.

This replaces the need to use `odigos-config` which must be applied as a whole which is inconvenient and error prone.

By using the new instrumentation rule, one can specify the usecase in a separate manifest, and gain more flexibility on how it is applied (workloads list, languages, multiple levels, etc)


edeNFed (approved):
LGTM

PR #1512, opened 2024-09-16
fix: reconcile in case pod killed abruptly

tamirdavid1:


RonFed (approved):

odiglet/pkg/kube/instrumentation_ebpf/manager.go
```diff
@@ -37,6 +37,25 @@ func (i *podPredicate) Update(e event.UpdateEvent) bool {
...
 	}
 
+	// Sum the restart counts for both oldPod and newPod containers, then compare them.
+	// If the newPod has a higher restart count than the oldPod, we need to re-instrument it.
+	// This happens because the pod was abruptly killed, which caused an increment in the restart count.
+
+	oldPodRestartCount := 0
+	newPodRestartCount := 0
+
+	for _, containerStatus := range oldPod.Status.ContainerStatuses {
```
yodigos, line 47: NIT: Can change this to a func "getSumRestarts(pod)" and then do one liner 
return getSumRestarts(newPod) > getSumRestarts(oldPod)

blumamir (approved):

odiglet/pkg/kube/instrumentation_ebpf/manager.go
```diff
@@ -37,7 +37,10 @@ func (i *podPredicate) Update(e event.UpdateEvent) bool {
 		return true
 	}
 
-	return false
+	// Sum the restart counts for both oldPod and newPod containers, then compare them.
+	// If the newPod has a higher restart count than the oldPod, we need to re-instrument it.
+	// This happens because the pod was abruptly killed, which caused an increment in the restart count.
```
RonFed, line 42:
```suggestion
	// Sum the restart counts for both oldPod and newPod containers, then compare them.
	// If the newPod has a higher restart count than the oldPod, we need to re-instrument it.
	// This happens because the pod was abruptly killed, which caused an increment in the restart count.
	// This check is required because the pod will remain running during the process kill and re-launch.
```

RonFed (approved):

odiglet/pkg/kube/instrumentation_ebpf/manager.go
tamirdavid1: Thanks!

PR #1513, opened 2024-09-17
fix: robust runtime details detection

blumamir:
This PR introduces a new way to apply runtime detection.

## Motivation

Our current language detection has the following issues:
- triggering the runtime detection is done via setting the `runtimeDetailsInvalidated` property on the `InstrumentationConfigSpec` CR. Since this resource is being written to by both the sdk configs controller in instromentor, and the odiglets once they finish runtime detection, the ping-pong is venerable to race conditions and is very hard to reason about.
- we trigger runtime inspection only once, possibly missing future changes in the following info: `runtimeVersion`, `envVars` which does not represent the current state of the workload. since odigos itself touches some env variables, this inconsistent can lead to really hard to find bugs. having the runtime detection up to date when new generations of the workload are applied, helps eliminate edge and advanced use-cases.
- we currently have 2 CRDs to represent instrumented workload in odigos: `instrumentationconfig` and `instrumentedapplication`. They can be merged into one CRD, making the maintenance a bit simpler. 
- the env override mechanism currently relies on the env values that are detected from runtime inspection to patch it with odigos content. For instrumentor to allow manifest changes in the env vars, it required to know if the runtime detection is up-to-date and avoid updating old values. Capturing the workload generation against which the runtime details were calculated, allows it to ignore this data when not up to date.

## Objectives

- collect runtime details for new generations of the workload (recalculate when a pod with newer generation starts)
- avoid calculating runtime details if the inspection is already made for this generation before.
- Do not pull in any new objects into the contoller-runtime cache, possibly blowing up the memory and cpu.
- apply the change in small steps, only changing one thing at a time for smaller PRs and safer transit

## Changes

- Write the new runtime inspection info into the status of the `instrumentationconfig` object. It will live alongside the existing runtime detection, and is not used anywhere in this step. We can then use it to compare the two values to gain confident in the process before start using this info.
- Add 2 new controllers to `odiglet`. one for detecting the language when a pod becomes running only if the geeneration in the CR requires to calculate again. the second one is for instrumentationconfig to do the initial language detection when a new source is added in odigos.
- For generation calculation, we need to get the replicaset for pods that were generated from deployments. For that we need to add the relevant RBAC permissions. the reeeplicaset is accessed only from a clientset, so not to pull all cluster replicasets into the controller-runtime cache.
- enhanced the e2e tests to assert the values we expect in existing tests


api/odigos/v1alpha1/instrumentationconfig_types.go
```diff
@@ -16,7 +16,16 @@ type InstrumentationConfig struct {
...
+	Spec   InstrumentationConfigSpec   `json:"spec,omitempty"`
+	Status InstrumentationConfigStatus `json:"status,omitempty"`
+}
+
+type InstrumentationConfigStatus struct {
+	// Capture Runtime Details for the workloads that this CR applies to.
+	RuntimeDetailsByContainer []RuntimeDetailsByContainer `json:"runtimeDetailsByContainer,omitempty"`
+
+	// The generation of the workload object for which these runtime details were captured.
+	ObserverWorkloadGeneration int64 `json:"observerWorkflowGeneration,omitempty"`
```
RonFed, line 28:
```suggestion
	ObservedWorkloadGeneration int64 `json:"observerWorkflowGeneration,omitempty"`
```

odiglet/pkg/kube/runtime_details/pods_controller.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	podGeneration, err := GetPodGeneration(ctx, p.Clientset, &pod)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// prevent runtime inspection on pods for which we already have the runtime details for this generation
+	if podGeneration == 0 || podGeneration <= instrumentationConfig.Status.ObserverWorkloadGeneration {
+		logger.Info("skipping redundant runtime details detection for pod with generation ", "name", request.Name, "namespace", request.Namespace, "currentPodGeneration", podGeneration, "observerWorkloadGeneration", instrumentationConfig.Status.ObserverWorkloadGeneration)
```
RonFed, line 100:
```suggestion
		logger.Info("skipping redundant runtime details detection for pod with old generation ", "name", request.Name, "namespace", request.Namespace, "currentPodGeneration", podGeneration, "observerWorkloadGeneration", instrumentationConfig.Status.ObserverWorkloadGeneration)
```

odiglet/pkg/kube/runtime_details/instrumentationconfigs_controller.go
```diff
@@ -0,0 +1,109 @@
...
+		// when a instrumentation config is created, many nodes may not have any running pods for it
+		return reconcile.Result{}, nil
+	}
+
+	odigosConfig, err := k8sutils.GetCurrentOdigosConfig(ctx, r.Client)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	runtimeResults, err := runtimeInspection(pods, odigosConfig.IgnoredContainers)
```
RonFed, line 86: Filter pods by generation and take only one

odiglet/pkg/kube/runtime_details/instrumentationconfigs_controller.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// persist the runtime results into the status of the instrumentation config
+	patchStatus := odigosv1.InstrumentationConfig{
+		Status: odigosv1.InstrumentationConfigStatus{
+			RuntimeDetailsByContainer: runtimeResults,
```
RonFed, line 94: Add generation

odiglet/pkg/kube/runtime_details/instrumentationconfigs_controller.go
```diff
@@ -0,0 +1,109 @@
...
+		},
+	}
+	patchData, err := json.Marshal(patchStatus)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	err = r.Client.Status().Patch(ctx, &instrumentationConfig, client.RawPatch(types.MergePatchType, patchData), client.FieldOwner("odiglet-runtimedetails"))
```
RonFed, line 101: Does this patch type will override existing runtime details?

blumamir: This works because we are using json merge patch, and this is a list, so it is replaced entirely

Makefile
```diff
@@ -200,10 +200,25 @@ cli-upgrade:
 	@echo "Installing odigos from source. version: $(ODIGOS_CLI_VERSION)"
 	cd ./cli ; go run -tags=embed_manifests . upgrade --version $(ODIGOS_CLI_VERSION) --yes
 
+.PHONY: cli-tests-artifacts
+cli-tests-artifacts:
```
RonFed, line 204: nit
```suggestion
.PHONY: cli-build
cli-build:
```

cli/cmd/resources/odiglet.go
```diff
@@ -184,6 +184,15 @@ func NewOdigletClusterRole(psp bool) *rbacv1.ClusterRole {
...
 				},
 			},
+			{
+				Verbs: []string{
+					"get",
+				},
+				APIGroups: []string{"apps"},
+				Resources: []string{
+					"replicasets",
+				},
```
RonFed, line 194: Does this also imply that Odiglet will now hold all the ReplicaSets in the cache? That used to be the case and when it was removed it saved a lot of memory.

odiglet/pkg/kube/runtime_details/instconfig_controller.go
```diff
@@ -19,12 +19,15 @@ import (
...
+// replaced by odiglet/pkg/kube/runtime_details/instrumentationconfigs_controller.go
+// which does not rely on the RuntimeDetailsInvalidated flag
+// left here until the migration is complete
+type DeprecatedInstrumentationConfigReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
 }
 
-func (i *InstrumentationConfigReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
+func (i *DeprecatedInstrumentationConfigReconciler) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
```
RonFed, line 30: nit:
You can mark this function with a comment of `Deprected:` -  some of the IDEs (like vscode) identify this and mark any call to those functions.

odiglet/pkg/kube/runtime_details/instrumentationconfigs_controller.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func (p *instrumentationConfigPredicate) Generic(e event.GenericEvent) bool {
+	return false
+}
+
+type InstrumentationConfigReconciler struct {
+	client.Client
+	Scheme    *runtime.Scheme
+	Clientset *kubernetes.Clientset
```
RonFed, line 51: What are the main reasons for using this client? saving cache space? correctness of the observed generations? If it is the latter, does it really affect correctness or does the cache have guarantees that are enough for us?
In any case. worth adding a comment about why this client is used.

odiglet/pkg/kube/runtime_details/instrumentationconfigs_controller.go
```diff
@@ -0,0 +1,136 @@
...
+			logger.Error(err, "Failed to get pod generation")
+			return reconcile.Result{}, err
+		}
+		if podGeneration == 0 {
+			// 0 means the pod is not relevant for runtime detection
+			continue
+		}
+
+		if podGeneration > instrumentationConfig.Status.ObservedWorkloadGeneration {
+			if podGeneration > selectedPodGeneration {
```
RonFed, line 94: `selectedPodGeneration` is going to be the latest one, right? If so, worth renaming it.

odiglet/pkg/kube/runtime_details/instrumentationconfigs_controller.go
```diff
@@ -0,0 +1,136 @@
...
+	pods, err := kubeutils.GetRunningPods(ctx, labels, workload.GetNamespace(), r.Client)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// we need to apply runtime detection for just one pod with newer generation
+	var podForRuntimeDetails *corev1.Pod
+	selectedPodGeneration := int64(0)
+	for _, pod := range pods {
+		podGeneration, err := GetPodGeneration(ctx, r.Clientset, &pod)
```
RonFed, line 83: Here we are using the "slow" client. We can have these loop iterations done in parallel and save time. I'm not sure if this will become very slow in large clusters or not. Assuming 10ms per API server query and 100 pods will result in 1 second for this loop.

odiglet/pkg/kube/runtime_details/pods_controller.go
```diff
@@ -0,0 +1,153 @@
...
+		return reconcile.Result{}, err
+	}
+
+	// prevent runtime inspection on pods for which we already have the runtime details for this generation
+	if podGeneration == 0 || podGeneration <= instrumentationConfig.Status.ObservedWorkloadGeneration {
+		logger.Info("skipping redundant runtime details detection since generation is not newer", "name", request.Name, "namespace", request.Namespace, "currentPodGeneration", podGeneration, "observedWorkloadGeneration", instrumentationConfig.Status.ObservedWorkloadGeneration)
+		return reconcile.Result{}, nil
+	}
+
+	// The reconcile should be called every time the pod is updated and changes from not running to running
```
RonFed, line 104: This comment seems not related to the code here
```suggestion
	// Perform runtime inspection once we know the pod is newer that the latest runtime inspection performed and saved.
```

odiglet/pkg/kube/runtime_details/pods_controller.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	patchData, err := json.Marshal(patchStatus)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	err = p.Client.Status().Patch(ctx, &instrumentationConfig, client.RawPatch(types.MergePatchType, patchData), client.FieldOwner("odiglet-runtimedetails"))
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
```
RonFed, line 131: This part is common to this and the instrumentationConfigReconciler, right? maybe worth creating a function for it (`PersistRuntimeInspection`)?

RonFed:
Also, worth adding in the PR description a small explanation about the change and the motivation.

odiglet/pkg/kube/runtime_details/instrumentationconfigs_controller.go
blumamir: It's only for pods of this workload, that run on the current node.
I think this will most times be 0, or 1-2, and in rare cases it might be more, but not something we should be worried about.

odiglet/pkg/kube/runtime_details/instrumentationconfigs_controller.go
blumamir: This is the generation of the pod we selected for `podForRuntimeDetails`.
It will be the higher generation we found, not the last one

blumamir: renamed the pod variable to "selectedPodForInspection"

odiglet/pkg/kube/runtime_details/pods_controller.go
blumamir: good idea, updated

RonFed (approved):

cli/cmd/resources/odiglet.go
blumamir: Notice that the verbs for this rule is only "get", and controller-runtime cache requires a "watch" and "list", thus even if someone uses it with `client.Client`, it will not work.

odiglet/pkg/kube/runtime_details/instrumentationconfigs_controller.go
```diff
@@ -0,0 +1,128 @@
...
+			return reconcile.Result{}, err
+		}
+		if podGeneration == 0 {
+			// 0 means the pod is not relevant for runtime detection
+			continue
+		}
+
+		if podGeneration > instrumentationConfig.Status.ObservedWorkloadGeneration && podGeneration > selectedPodGeneration {
+			selectedPodGeneration = podGeneration
+			selectedPodForInspection = &pod
```
edeNFed, line 97: I would verify that it is not always takes the last pod: https://blog.devtrovert.com/p/go-tricky-a-bug-with-for-range-loops

I think go changed that behavior in Go 1.22 but still worth checking manually once

odiglet/pkg/kube/runtime_details/pods_controller.go
```diff
@@ -0,0 +1,145 @@
...
+	podGeneration, err := GetPodGeneration(ctx, p.Clientset, &pod)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// prevent runtime inspection on pods for which we already have the runtime details for this generation
+	if podGeneration == 0 || podGeneration <= instrumentationConfig.Status.ObservedWorkloadGeneration {
+		logger.Info("skipping redundant runtime details detection since generation is not newer", "name", request.Name, "namespace", request.Namespace, "currentPodGeneration", podGeneration, "observedWorkloadGeneration", instrumentationConfig.Status.ObservedWorkloadGeneration)
```
edeNFed, line 102: This has potential to fill Odiglet log I suggest logging this as V(3) or similar

PR #1514, opened 2024-09-17
fix: avoid panic in case instrumentationConfig deleted

tamirdavid1:


opampserver/pkg/sdkconfig/instrumentationconfigs_controller.go
```diff
@@ -24,7 +24,7 @@ func (i *InstrumentationConfigReconciler) Reconcile(ctx context.Context, req ctr
 
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			instrumentationConfig = nil
+			return ctrl.Result{}, nil
 		} else {
 			return ctrl.Result{}, err
 		}
```
blumamir, line 30: Thanks!

If you like, this can be refactored to:

```golang
```

blumamir (approved):

opampserver/pkg/sdkconfig/instrumentationconfigs_controller.go
tamirdavid1: nice, thanks

PR #1521, opened 2024-09-18
feat: Reading Logs in daignose cli command

yodigos:


cli/cmd/diagnose.go
```diff
@@ -0,0 +1,151 @@
...
+	return nil
+}
+
+func fetchOdigosComponentsLogs(ctx context.Context, client *kube.Client) error {
+	odigosNamespace, err := resources.GetOdigosNamespace(client, ctx)
+	if err != nil {
+		return err
+	}
+
+	pods, err := client.CoreV1().Pods(odigosNamespace).List(context.TODO(), metav1.ListOptions{})
```
yodigos, line 78: Add here Previous option for if Odiglet falls

cli/cmd/diagnose.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	return nil
+}
+
+func createAllDirs() error {
+	if err := os.RemoveAll(mainDir); err != nil {
+		return err
+	}
+	if err := os.MkdirAll(mainDir, 0755); err != nil {
```
edeNFed, line 57: I think this will create the directory in the same place where `odigos diagnose` is run. Maybe better to always create it at some temp directory. I think Go has a built in function like `ioutil.TempDir` or something similar

cli/cmd/diagnose.go
```diff
@@ -0,0 +1,148 @@
...
+
+func fetchPodLogs(ctx context.Context, client *kube.Client, odigosNamespace string, pod v1.Pod) error {
+	for _, container := range pod.Spec.Containers {
+		fmt.Printf("Fetching logs for Pod: %s, Container: %s\n", pod.Name, container.Name)
+
+		// Define the log file path for saving compressed logs
+		logFilePath := filepath.Join(mainDir, logDir, pod.Name+"_"+container.Name+".log.gz")
+		logFile, err := os.OpenFile(logFilePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0666)
+		if err != nil {
+			return err
```
edeNFed, line 97: Should we stop on error or continue collecting logs of the rest of the pods anyway?

cli/cmd/diagnose.go
yodigos: Why don't we want it to be saved where the user run  odigos diagnose?

cli/cmd/diagnose.go
```diff
@@ -0,0 +1,148 @@
...
+
+	if err := fetchOdigosComponentsLogs(ctx, client); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func createAllDirs() error {
+	if err := os.RemoveAll(mainDir); err != nil {
```
RonFed, line 54: Also related to Eden's comment, you can have a look at [the TempDir function](https://pkg.go.dev/os#TempDir).
I think having the files created in a temp location and only saving the final zip in the CWD is the accepted behavior

yodigos: TempDir is deprecated from go 1.16 - the new way is with os.MkdirTemp

yodigos: ohhh my bad, ioutil.TempDir() is deprecated

cli/cmd/diagnose.go
```diff
@@ -0,0 +1,216 @@
...
+}
+
+func fetchPodLogs(ctx context.Context, client *kube.Client, odigosNamespace string, pod v1.Pod, logDir string) {
+	for _, container := range pod.Spec.Containers {
+		logPrefix := fmt.Sprintf("Fetching logs for Pod: %s, Container: %s, Node: %s", pod.Name, container.Name, pod.Spec.NodeName)
+		fmt.Printf(logPrefix + "\n")
+
+		// Define the log file path for saving compressed logs
+		logFilePath := filepath.Join(logDir, pod.Name+"_"+container.Name+"_"+pod.Spec.NodeName+".log.gz")
+		logFile, err := os.OpenFile(logFilePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0666)
```
RonFed, line 107: should add here
`defer logFile.Close()`
Instead of the current code which only closes the file at the happy-path where no errors occur.
This will guarantee the file is close when this function returns.

cli/cmd/diagnose.go
```diff
@@ -0,0 +1,216 @@
...
+		if err != nil {
+			fmt.Printf(logPrefix+" - Failed - Error creating log file: %v\n", err)
+			return
+		}
+
+		req := client.CoreV1().Pods(odigosNamespace).GetLogs(pod.Name, &v1.PodLogOptions{})
+		logStream, err := req.Stream(ctx)
```
RonFed, line 114: Same comment as for the file close

cli/cmd/diagnose.go
yodigos: **Why Not Use defer Inside the Loop?**
In a loop, using defer to close a file might seem like a convenient option, but deferred calls are not executed until the surrounding function returns. If you use defer inside the loop, it will accumulate all the defer calls, and the files will remain open until the function (or main program) exits. This can lead to excessive memory or resource usage, especially with a large number of iterations.

yodigos: func() {
```suggestion
		logFile, err := os.OpenFile(logFilePath, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0666)
```
			defer func() {
				if err := file.Close(); err != nil {
					fmt.Printf("Error closing file %s: %v\n", fileName, err)
				}
			}()

yodigos: Ok I read a bit and this is a real cool way to solve this.. The anonymous function will be called after opening the file and because it uses defer and created inside the scope of the for-loop -> it will trigger in the end of the loop (succesfully / with error)
```
func() {
			defer func() {
				if err := file.Close(); err != nil {
					fmt.Printf("Error closing file %s: %v\n", fileName, err)
				}
			}()
```

RonFed: You can also extract the code in the loop body to a function and use a defer within that function, that way there is no risc for a long queue of defer functions.

RonFed (approved):

PR #1523, opened 2024-09-19
Fix eBPF director cleanups

RonFed:
Fixed cleanup in eBPF director:
* Delete `instrumentationInstance` correctly (passing the correct instrumentation instance name which includes the pod name and the PID).
* Add a periodic cleanup routine to detect processes that do not exist and clear all the instrumentation resources associated with them.

Refine the director data structures by introducing `InstrumentedProcess`:
* make sure we only try to instrument a given PID in a given pod once with `sync.Once`.
* Once an instrumented process is closed use an atomic flag to mark it as closed.
* Refactor the cleanup of an instrumented process to a common function.
* Move the go instrumentation to an `sdks` folder which allows separating the generic director to its package and testing it without having dependencies on the go instrumentation repo.
* Add unit testing for the director.


odiglet/pkg/ebpf/director.go
```diff
@@ -201,95 +245,92 @@ func (d *EbpfDirector[T]) Instrument(ctx context.Context, pid int, pod types.Nam
...
-		} else {
-			d.mux.Unlock()
-			// we attempted to instrument this process, but it was already cleaned up
-			// so we need to clean up the instrumentation we just created
-			err = inst.Close(ctx)
-			if err != nil {
-				log.Logger.Error(err, "error cleaning up instrumentation for process", "pid", pid)
+			if ip.closed.Load() {
+				log.Logger.Info("Instrumentation already closed before running, stopping instrumentation", "pid", pid)
+				return
```
blumamir, line 314: above we called `d.instrumentationFactory.CreateEbpfInstrumentation` and here we return without cleaning up anything. Just making sure that the ebpf instrumentation does not hold any resources after creation

blumamir (approved):

odiglet/pkg/ebpf/director.go
RonFed: If it is closed, it means the cleanup function was called for that instrumented process

PR #1527, opened 2024-09-20
ci: avoid installing webapp unused deps

blumamir:


blumamir:
apparently it is used for cypress tests

PR #1539, opened 2024-09-24
feat: Java workload e2e

yodigos:


tests/e2e/workload-lifecycle/services/java-http-server/java-supported-version.Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM maven:3.8.5-openjdk-17 AS build
+COPY src /home/app/src
+COPY pom.xml /home/app
+RUN mvn -f /home/app/pom.xml clean package
+
+FROM eclipse-temurin:17-jre-jammy
+COPY --from=build /home/app/target/*.jar /app/java-supported-version.jar
+USER 15000
+CMD ["java","-jar", "/app/java-supported-version.jar"]
```
blumamir, line 9: Are there any tests for unsupported version?

tests/e2e/workload-lifecycle/services/java-http-server/java-supported-docker-env.Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM maven:3.8.5-openjdk-17 AS build
+COPY src /home/app/src
+COPY pom.xml /home/app
+RUN mvn -f /home/app/pom.xml clean package
+
+FROM eclipse-temurin:17-jre-jammy
+ENV JAVA_OPTS="-Dthis.does.not.exist=true"
```
blumamir, line 7: Where do we assert that this value indeed exists in the application?
If odigos remove it when env overwriting, will the test fail?

blumamir (commented):
Please write a description about the scenarios [here](https://github.com/odigos-io/odigos/tree/main/tests/e2e/workload-lifecycle).

I see we don't have any tests for unsupported or old versions. Since this is where many problems are manifested, I think it even more important to test these then the mainstream version.

Also, please add a test for java latests version, as if something breaks with new versions, we will not have any other indication. Should we also test the minimum version?

blumamir (approved):

PR #1541, opened 2024-09-25
feat: odigos describe remote flag to run the cluster

blumamir:
This can be used to run `odigos describe` in machines where `kubectl` may lacks permissions for some of the resources.

It runs the describe in the ui deployment which can access the relevant resources, and return it with http to the cli


edeNFed (approved):
🚀

PR #1547, opened 2024-09-25
ci: arrange make build targets to start with the longest ones

blumamir:
When we run `make -j 3`, make will start with the ones in the beginning of the list, so this PR arrange them so that the heavy builds are first (ui, odiglet, collector) and only then comes the lighter ones (autoscaler, instrumentor, schedualr)

Hopeful, it will decrease a bit from CI time, better using the resources available

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ build-ui:
 
 .PHONY: build-images
 build-images:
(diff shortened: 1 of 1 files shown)


blumamir:
Looks like it saves around 20-40 seconds from the build time.
not dramatic, but worth doing

RonFed (approved):

PR #1548, opened 2024-09-25
ci: build the cli in concurrent to the docker images

blumamir:
Currently we are building the cli and docker images one after the other.
Since the cli takes ~30 seconds to build, we can benefit from doing it in parallel to the docker images, reducing 30 seconds from CI time


blumamir:
According to my check, now the "build-odigos-images" job takes 6m 32s instead of 7m 27s previously, reducing the CI time in almost 1 minute

RonFed (approved):
🥇

PR #1564, opened 2024-10-01
Addind e2e tests for python

tamirdavid1:


tests/e2e/workload-lifecycle/services/python-http-server/myapp/__pycache__/__init__.cpython-310.pyc
blumamir, line 1: should these be checked in?
I think we need to have them in `.gitignore`

blumamir (approved):
🥇

PR #1595, opened 2024-10-17
feat: odigos describe command

blumamir:
Changes:

- Add `odigos describe` command to describe the general status of odigos itself. Currently implemented Cluster Collector Information
- Add conditions to collectors group status to record if the collectors deployment failed to succeed. For example - this can happen in the present of validation webhook that rejects the deployment if the image is not signed according to some requirements
- move some code from `cli` to `k8sutils` to make it available for all modules in the code.
- some cleanups on logging, so if the deployment fails to succeed, it is only printed once to autoscaler logs


cli/cmd/describe.go
```diff
@@ -20,6 +20,32 @@ var describeCmd = &cobra.Command{
 	Use:   "describe",
 	Short: "Show details of a specific odigos entity",
 	Long:  `Print detailed description of a specific odigos entity, which can be used to troubleshoot issues`,
```
RonFed, line 22: Do we need to update these descriptions? or have different ones for `odigos describe` and `odigos describe source`?

RonFed (approved):

PR #1597, opened 2024-10-17
feat: node collector info in odigos describe

blumamir:
This PR:

- record the status of autoscaler node collector info into the collector group condition to make it available for display.
- enhance the `odigos describe` command to include info about the node collector deployment:

image

### DIFF
--- a/k8sutils/pkg/describe/common.go
+++ b/k8sutils/pkg/describe/common.go
@@ -26,3 +26,10 @@ func describeText(sb *strings.Builder, indent int, printftext string, args ...in
 	lineText := fmt.Sprintf(printftext, args...)
 	sb.WriteString(fmt.Sprintf("%s%s\n", indentText, lineText))
 }
+
+func boolToText(b bool) string {
+	if b {
+		return "True"
+	}
+	return "False"
+}
--- a/k8sutils/pkg/describe/common.go
+++ b/k8sutils/pkg/describe/common.go
@@ -26,10 +26,3 @@ func describeText(sb *strings.Builder, indent int, printftext string, args ...in
 	lineText := fmt.Sprintf(printftext, args...)
 	sb.WriteString(fmt.Sprintf("%s%s\n", indentText, lineText))
 }
-
-func boolToText(b bool) string {
-	if b {
-		return "True"
-	}
-	return "False"
-}
--- a/k8sutils/pkg/describe/odigos.go
+++ b/k8sutils/pkg/describe/odigos.go
@@ -22,6 +22,11 @@ type clusterCollectorResources struct {
 	LatestRevisionPods *corev1.PodList
 }
 
+type nodeCollectorResources struct {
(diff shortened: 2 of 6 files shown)


k8sutils/pkg/describe/common.go
```diff
@@ -26,3 +26,10 @@ func describeText(sb *strings.Builder, indent int, printftext string, args ...in
 	lineText := fmt.Sprintf(printftext, args...)
 	sb.WriteString(fmt.Sprintf("%s%s\n", indentText, lineText))
 }
+
+func boolToText(b bool) string {
+	if b {
+		return "True"
+	}
+	return "False"
+}
```
RonFed, line 35: you can use %t to print boolean values: https://go.dev/play/p/k5QSFJBYiJs

k8sutils/pkg/describe/odigos.go
```diff
@@ -22,6 +22,11 @@ type clusterCollectorResources struct {
 	LatestRevisionPods *corev1.PodList
 }
 
+type nodeCollectorResources struct {
+	CollectorsGroup *odigosv1.CollectorsGroup
+	DaemonSet       *appsv1.DaemonSet
```
RonFed, line 27: I think we should also include the status for the node collector pods similar to what was done for the cluster collector. In this case we will actually have pod in each node so reporting their status is useful (If a some nodes are not in good shape).
Or at least to write `x/y node collectors are healthy`

k8sutils/pkg/describe/odigos.go
```diff
@@ -77,13 +82,35 @@ func getClusterCollectorResources(ctx context.Context, kubeClient kubernetes.Int
...
+
+	nodeCollector.DaemonSet, err = kubeClient.AppsV1().DaemonSets(odigosNs).Get(ctx, consts.OdigosNodeCollectorDaemonSetName, metav1.GetOptions{})
+	if err != nil && !apierrors.IsNotFound(err) {
+		return
+	}
+
+	return
+}
+
+func getRelevantOdigosResources(ctx context.Context, kubeClient kubernetes.Interface, odigosClient odigosclientset.OdigosV1alpha1Interface, odigosNs string) (clusterCollector clusterCollectorResources, nodeCollector nodeCollectorResources, destinations *odigosv1.DestinationList, instrumentationConfigs *odigosv1.InstrumentationConfigList, err error) {
```
RonFed, line 102: nit:  maybe we should return here a single struct of `OdigosResources` which contains everything  and an error  instead of having 5 returned values.

k8sutils/pkg/describe/odigos.go
blumamir: I added this info based on the ds status fields.
It doesn't read as nicely, but it's suppose to give indication if something is not running.
See the screenshot in the PR description

RonFed (approved):

PR #1599, opened 2024-10-18
docs: add clickhouse destination documentation

blumamir:


docs/backends/clickhouse.mdx
```diff
@@ -0,0 +1,172 @@
...
+- Click `Add New Destination`
+
+- Select `ClickHouse` and follow the on-screen instructions.
+
+2. **Using kubernetes manifests**
+
+Save the YAML below to a file (e.g., `destination.yaml`) and apply it using `kubectl`:
+
+```bash
+kubectl apply -f destination.yaml
```
RonFed, line 131: when created from the UI, we put it in the Odigos ns, does it matter here if it will be applied in the default one?

RonFed (approved):

PR #1602, opened 2024-10-19
chore(scheduler): use instrumentation config instead of application

blumamir:
This PR moves the scheduler component to use `instrumentationconfigs` instead of `instrumentedapplications`. There is an effort to merge them any only use one CRD - the `instrumentationconfigs`.

The content of the object is not examined, it only checks if we have any `instrumentationconfigs` or not. Thus it should be safe.

### DIFF
--- a/scheduler/controllers/instrumentationconfig_controller.go
+++ b/scheduler/controllers/instrumentationconfig_controller.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"context"
+
 	odigosv1 "github.com/odigos-io/odigos/api/odigos/v1alpha1"
 	"github.com/odigos-io/odigos/k8sutils/pkg/consts"
 	"github.com/odigos-io/odigos/k8sutils/pkg/env"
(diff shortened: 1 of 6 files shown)


scheduler/controllers/instrumentationconfig_controller.go
```diff
@@ -14,27 +15,26 @@ import (
...
-	err := r.List(ctx, &instrumentedApps)
+	var instrumentedConfigs odigosv1.InstrumentationConfigList
+	err := r.List(ctx, &instrumentedConfigs)
 	if err != nil {
-		logger.Error(err, "failed to list InstrumentedApplications")
+		logger.Error(err, "failed to list InstrumentationConfigs")
 		return ctrl.Result{}, err
 	}
-	numberOfInstrumentedApps := len(instrumentedApps.Items)
+	numberOfInstrumentedApps := len(instrumentedConfigs.Items)
```
RonFed, line 37: numberOfInstrumentedApps -> numberOfInstrumentationConfigs

RonFed (approved):

scheduler/controllers/instrumentationconfig_controller.go
blumamir: It's intentional. Each instrumentation config object represents an app

PR #1618, opened 2024-10-25
feat: move default node collector port to avoid conflicts and add setting in config

blumamir:
Since node collector runs as a DaemonSet with host network (so that otlp ports 4317 and 4318 are accessible from the pods exporters), it also shares the ports namespace with the node.

The ports it binds to might conflict with ports of other tools installed in the cluster that does similar things, and in this case data collection will fail to start with error: "listen tcp 0.0.0.0:8888: bind: address already in use".

The collectors currently expose and scrape their own metrics on port `8888` which is good candidate for collistions.

This PR:

- makes the default port for own telemetry endpoint for node collector `55682` to have low chance of collision with something that already runs on the node (instead of using `8888` as default).
- Adds an option to odigos config to set the port, in case one needs it hard-coded to a specific value to manage the public ports on the node).

The port to use is written on the collectors group CRD by the scheduler to abstract away the default and odigos config resolving, and allow the resolved value to be more easer to consume.

For the cluster collector, the collector runs as deployment with k8s service, thus we can safely use `8888` without colliding with anything else


scheduler/controllers/collectorgroups/datacollection.go
```diff
@@ -2,23 +2,37 @@ package collectorgroups
...
 import (
 	odigosv1 "github.com/odigos-io/odigos/api/odigos/v1alpha1"
+	"github.com/odigos-io/odigos/common"
 	"github.com/odigos-io/odigos/k8sutils/pkg/consts"
 	"github.com/odigos-io/odigos/k8sutils/pkg/env"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func NewNodeCollectorGroup() *odigosv1.CollectorsGroup {
+const NodeCollectorDefaultOwnMetricsPort = int32(55682)
```
RonFed, line 11: see another comment about moving this const to common.
Also, this default value should be defined in helm, right?

k8sutils/pkg/utils/collectorgroup_util.go
```diff
@@ -2,15 +2,24 @@ package utils
...
-	return c.Create(ctx, collectorGroup)
+func ApplyCollectorGroup(ctx context.Context, c client.Client, collectorGroup *odigosv1.CollectorsGroup) error {
+	logger := log.FromContext(ctx).WithValues("collectorGroupName", collectorGroup.Name)
+	logger.Info("Applying collector group", "collectorGroupName", collectorGroup.Name)
+
+	err := c.Patch(ctx, collectorGroup, client.Apply, client.ForceOwnership, client.FieldOwner("scheduler"))
+	if err != nil {
+		logger.Error(err, "Failed to apply collector group")
+		return err
+	}
```
RonFed, line 20: Why do we need the scheduler ownership here?

RonFed (approved):
Wrote a few small bob blocking comments

autoscaler/controllers/collectorsgroup_controller.go
```diff
@@ -102,6 +102,8 @@ func (r *CollectorsGroupReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 func (r *CollectorsGroupReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&odigosv1.CollectorsGroup{}).
-		WithEventFilter(&onlyCreatePredicate{}).
```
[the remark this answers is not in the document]
blumamir: this file is removed after refactor

RonFed (approved):

PR #1626, opened 2024-10-26
Remove default SDKs and supported SDKs from Odigos config

RonFed:
Removing the default SDKs from odigos config.
The updated way of changing instrumentation SDK is with rules and profiles.

This change includes:
* Remove the setting of the SDKs by the CLI.
* Add to instrumentor: checking the tier and setting the default SDKs map once.
* Add role to instrumentor to `get` secrets in odigos ns.
* Remove odigosconfig role for instrumentor.
* Adjust `envtest` for env overwriter - instead of reading writing to mock of `odigosconfig` - mock the default SDK for the new code and simulate a changing SDK by mocking instrumentation rule.


api/odigos/v1alpha1/odigosconfig_types.go
```diff
@@ -45,8 +45,6 @@ type OdigosConfigurationSpec struct {
 	OdigletImage      string                                          `json:"odigletImage,omitempty"`
 	InstrumentorImage string                                          `json:"instrumentorImage,omitempty"`
 	AutoscalerImage   string                                          `json:"autoscalerImage,omitempty"`
-	SupportedSDKs     map[common.ProgrammingLanguage][]common.OtelSdk `json:"supportedSDKs,omitempty"`
-	DefaultSDKs       map[common.ProgrammingLanguage]common.OtelSdk   `json:"defaultSDKs,omitempty"`
```
blumamir, line 49: 🎉

cli/cmd/resources/instrumentor.go
```diff
@@ -68,6 +68,54 @@ func NewInstrumentorRoleBinding(ns string) *rbacv1.RoleBinding {
...
+			Namespace: ns,
+		},
+		Rules: []rbacv1.PolicyRule{
+			{
+				Verbs: []string{
+					"get",
+				},
+				APIGroups: []string{""},
+				Resources: []string{
+					"secrets",
```
blumamir, line 112: I think it's good for now.
In the future, we might want to consider storing the `tier` in the `odigos-deployment` configmap, so we can read the tier easily without needing permissions to secrets. We can also use `envFrom` to automatically have these as env variables in the code

blumamir (approved):

PR #1636, opened 2024-10-27
fix: new runtime inspection to run only when all containers are ready…

tamirdavid1:
… and started

### DIFF
--- a/odiglet/pkg/kube/runtime_details/pods_controller.go
+++ b/odiglet/pkg/kube/runtime_details/pods_controller.go
@@ -5,6 +5,7 @@ import (
 
 	odigosv1 "github.com/odigos-io/odigos/api/odigos/v1alpha1"
 	"github.com/odigos-io/odigos/common"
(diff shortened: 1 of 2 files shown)


odiglet/pkg/kube/runtime_details/pods_controller.go
```diff
@@ -18,28 +19,74 @@ import (
...
+	// If pod has no containers, return false as we can't determine readiness
+	if len(pod.Status.ContainerStatuses) == 0 {
+		return false
+	}
+
+	// Iterate over all containers in the pod
+	// Return false if any container is:
+	// 1. Not Ready
+	// 2. Started is nil or false
+	for _, containerStatus := range pod.Status.ContainerStatuses {
```
blumamir, line 41: we do not check the previous event here, thus after this state is achieved, every new update will trigger language detection.
We should compare old and new value and only trigger when not all containers are ready but now they are

blumamir (approved):
🥇

RonFed (approved):

PR #1645, opened 2024-10-28
Use new runtime detector for logging process events

RonFed:
This is the first step of using the new runtime detector module for triggering eBPF instrumentations.
Currently, it is set up by the Odiglet and only logs the events.


edeNFed (approved):
Lgtm

odiglet/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/odigos-io/odigos/odiglet
 
-go 1.22.0
+go 1.22.5
```
tamirdavid1, line 3: @RonFed why these changed ?

RonFed: I think its because I ran `go mod tidy` locally, I'll try to revert this part and see if the CI is happy this it

blumamir (approved):
💯

tamirdavid1 (approved):

PR #1649, opened 2024-10-28
refactor: describe source use 2 phases

blumamir:
this PR makes it so that `odigos describe source` uses 2 phases: analyze and print.

This allows the UI and tools to fetch the describe result and use it, without converting it to human text


edeNFed (approved):
🔥

PR #1652, opened 2024-10-29
[GEN-1521]: update actions

BenElferink:


frontend/webapp/containers/main/actions/choose-action-body/index.tsx
```diff
@@ -3,7 +3,7 @@ import styled from 'styled-components';
 import { type ActionInput } from '@/types';
 import ActionCustomFields from './custom-fields';
 import { type ActionOption } from '../choose-action-modal/action-options';
-import { DocsButton, Input, Text, TextArea } from '@/reuseable-components';
+import { Checkbox, DocsButton, Input, Text, TextArea, Toggle } from '@/reuseable-components';
```
alonkeyval, line 6: merge imports

frontend/webapp/graphql/mutations/action.ts
```diff
@@ -7,3 +7,16 @@ export const CREATE_ACTION = gql`
     }
   }
 `;
+
+export const UPDATE_ACTION = gql`
+  mutation UpdateAction($id: ID!, $action: ActionInput!) {
+    updateAction(id: $id, action: $action) {
+      id
```
alonkeyval, line 14: Do we need to get back all those fields?

frontend/webapp/hooks/actions/useActionFormData.ts
```diff
@@ -43,10 +44,42 @@ export function useActionFormData() {
     return ok;
   };
 
+  const loadFormWithDrawerItem = (drawerItem: DrawerBaseItem) => {
```
alonkeyval, line 47: build update action locally and set the state once

```suggestion
const loadFormWithDrawerItem = (drawerItem: DrawerBaseItem) => {
  const { type, spec } = drawerItem.item as ActionDataParsed;
  
  const updatedData: ActionInput = {
    ...INITIAL,
    type,
  };

  Object.entries(spec).forEach(([k, v]) => {
    switch (k) {
      case 'actionName':
        updatedData.name = v as string;
        break;

      case 'disabled':
        updatedData.disable = v as boolean;
        break;

      case 'notes':
      case 'signals':
        updatedData[k] = v;
        break;

      default:
        updatedData.details = JSON.stringify({ [k]: v });
        break;
    }
  });

  // Set the accumulated changes in a single state update
  setFormData(updatedData);
};
```

alonkeyval (approved):

PR #1664, opened 2024-10-30
fix: autocompleted notification type

BenElferink:


BenElferink:
This fix was required due to the change of the `NotificationNote` component.
Type "info" was changed from blue to grey, type "default" got blue from the previous.

alonkeyval (approved):

PR #1666, opened 2024-10-31
fix: check for all containers running instead of pod phase

tamirdavid1:


odiglet/pkg/kube/runtime_details/instconfig_controller.go
```diff
@@ -56,7 +56,11 @@ func (i *DeprecatedInstrumentationConfigReconciler) Reconcile(ctx context.Contex
 	}
 	err = inspectRuntimesOfRunningPods(ctx, &logger, labels, i.Client, i.Scheme, workload)
 	if err != nil {
-		return reconcile.Result{}, ignoreNoPodsFoundError(err)
+		// If no pods are found, requeue the request. This can happen because not AllContainersReady for the workload
+		if ignoreNoPodsFoundError(err) == nil {
+			return reconcile.Result{RequeueAfter: consts.DefaultRequeueAfter}, ignoreNoPodsFoundError(err)
```
RonFed, line 61: As we talked about, I think we should have some way to make sure the requeue won't happen forever

RonFed (approved):

PR #1673, opened 2024-11-03
fix: handle missing container language for instrumentation device

blumamir:


instrumentor/instrumentation/instrumentation.go
```diff
@@ -40,6 +40,11 @@ func ApplyInstrumentationDevicesToPodTemplate(original *corev1.PodTemplateSpec,
 		if containerLanguage == nil || *containerLanguage == common.UnknownProgrammingLanguage || *containerLanguage == common.IgnoredProgrammingLanguage || *containerLanguage == common.NginxProgrammingLanguage {
 			// always patch the env vars, even if the language is unknown or ignored.
 			// this is necessary to sync the existing envs with the missing language if changed for any reason.
+			// containerLanguage should not be nil, but if it somehow is, treat it as unknown.
+			if containerLanguage == nil {
```
RonFed, line 44: I prefer to change `getLanguageOfContainer` to return `UnknownProgrammingLanguage` if noting is found instead of returning nil.

blumamir: Thanks! refactored

tamirdavid1 (approved):

RonFed (approved):

PR #1685, opened 2024-11-04
feat: Adding Profiling for the Odiglets

yodigos:


cli/cmd/diagnose_util/crd_utils.go
```diff
@@ -0,0 +1,150 @@
...
+		CRDName:  "instrumentationconfigs",
+		CRDGroup: odigosGroupName,
+	},
+	{
+		CRDName:  "instrumentationrules",
+		CRDGroup: odigosGroupName,
+	},
+	{
+		CRDName:  "instrumentationinstances",
+		CRDGroup: odigosGroupName,
```
tamirdavid1, line 68: Can you please add destinations as we discussed?

cli/cmd/diagnose_util/crd_utils.go
```diff
@@ -0,0 +1,150 @@
...
+	crdYAML, err := yaml.Marshal(crd)
+	if err != nil {
+		return err
+	}
+
+	_, err = gzipWriter.Write(crdYAML)
+	if err != nil {
+		return err
+	}
+	if err = gzipWriter.Flush(); err != nil {
```
tamirdavid1, line 145: why we tar each dir of logs/crds and not only the bundle? it make it a bit harder to investigate using this 
e.g: 
image

cli/cmd/diagnose_util/crd_utils.go
yodigos: sure sure, I separated this task for different PRs

tamirdavid1 (approved):

PR #1689, opened 2024-11-05
feat: enable gateway hpa based on memory

tamirdavid1:


autoscaler/main.go
```diff
@@ -262,3 +274,23 @@ func main() {
...
+	// Use APIReader (uncached client) for direct access to the API server
+	// uses because mgr not cache the metrics-server deployment
+	err := mgr.GetAPIReader().Get(context.TODO(), types.NamespacedName{Name: "metrics-server", Namespace: "kube-system"}, &metricsServerDeployment)
+	if err != nil {
+		if client.IgnoreNotFound(err) != nil {
+			logger.Error(err, "Failed to get metrics-server deployment")
+			return false
+		}
+
+		logger.Error(err, "Metrics-server deployment not found")
```
blumamir, line 290: Is it an error? 
I guess it will no be found in some clusters, and the error can be missleading

autoscaler/main.go
```diff
@@ -262,3 +274,23 @@ func main() {
...
 }
+
+func isMetricsServerInstalled(mgr ctrl.Manager, logger logr.Logger) bool {
+	var metricsServerDeployment appsv1.Deployment
+
+	// Use APIReader (uncached client) for direct access to the API server
+	// uses because mgr not cache the metrics-server deployment
+	err := mgr.GetAPIReader().Get(context.TODO(), types.NamespacedName{Name: "metrics-server", Namespace: "kube-system"}, &metricsServerDeployment)
+	if err != nil {
+		if client.IgnoreNotFound(err) != nil {
```
blumamir, line 285: nit: the `IgnoreNotFound` is for when you want to return this error.

You can use `apierrors.IsNotFound(err)` which is more appropriate in this case

blumamir (approved):

PR #1690, opened 2024-11-05
Expand timeouts to runtime detection

tamirdavid1:


odiglet/pkg/kube/runtime_details/inspection.go
```diff
@@ -45,6 +45,7 @@ func inspectRuntimesOfRunningPods(ctx context.Context, logger *logr.Logger, labe
 	}
 
 	if len(pods) == 0 {
+		logger.Info("No Running pods found for the workload", "name", object.GetName())
```
blumamir, line 48: Will we get this in the logs of each odiglet that does not run the workload?
I guess this can be quite a lot

blumamir (approved):

PR #1710, opened 2024-11-08
fix(instrumentor): reconcile device on restarts

blumamir:
At the moment, the sequence of events is:

1. user deleted a source from ui
2. the label is removed from the workload object (deployment etc).
3. the instrumentor detects the label is removed and delete instrumented application and instrumentation config.
4. when instrumented application is deleted, the device is cleared from workload pod spec with a relevant controller.

Device should be removed if the instrumented application is deleted, and specifically in these cases:
- if instrumentor restarts when the instrumented application has been deleted but the device cleanup where not applied yet.
- when 2 controllers perform the device logic at the same time, and one is deleting the device from the workload, while the other one, which might be using old state, will attempt to reintroduce it.

I reproduced these cases locally with old version, observed the device is not cleaned up properly in these edge cases, and then run the changes in this PR to make sure these are fixed.


edeNFed (approved):
🚀

instrumentor/controllers/instrumentationdevice/manager.go
```diff
@@ -27,7 +27,12 @@ type workloadPodTemplatePredicate struct {
 }
 
 func (w workloadPodTemplatePredicate) Create(e event.CreateEvent) bool {
-	return false
+	// when instrumentor restarts, this case will be triggered as workloads objects are being added to the cache.
+	// in this case, we need to reconcile the workload, and guarantee that the device is injected or removed
+	// based on the current state of the cluster.
+	// if the instrumented application is deleted but the device is not cleaned,
+	// the instrumented application controller will not be invoked after restart, which is why we need to handle this case here.
+	return true
```
RonFed, line 35: Maybe we should add a check here that the odigos label is present?
Since after this change, when the instrumentor is starting this will cause it to have an event for each workload in the cluster - even those which odigos does not care about.
For those the new check added in this PR below `if apierrors.IsNotFound(err)` will be true.
That means the even for workloads which were not instrumented at all we will call `removeInstrumentationDeviceFromWorkload`.

RonFed (changes requested):
I think we should refine the Create condtion

instrumentor/controllers/instrumentationdevice/manager.go
blumamir: We need to handle workloads which has no label.

Consider the following sequence:
- label is removed from a workload which has the device
- instrumented application deleted
- instrumentor went down

When the instrumentor starts, we need to remove the device from this workload. if we add the filter you suggested, nothing will remove the device from this workload

RonFed: How about checking if there is a device present in the pod spec? Since we have the workload object here I think we can check that.

instrumentor/controllers/instrumentationdevice/common.go
```diff
@@ -157,46 +148,46 @@ func removeInstrumentationDeviceFromWorkload(ctx context.Context, kubeClient cli
...
-	err := kubeClient.Get(ctx, client.ObjectKey{
-		Namespace: namespace,
-		Name:      workloadName,
-	}, workloadObj)
-	if err != nil {
-		return client.IgnoreNotFound(err)
-	}
-
-	result, err := controllerutil.CreateOrPatch(ctx, kubeClient, workloadObj, func() error {
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
RonFed, line 151: According to https://pkg.go.dev/k8s.io/client-go/util/retry#pkg-variables

> DefaultBackoff is the recommended backoff for a conflict where a client may be attempting to make an unrelated modification to a resource under active management by one or more controllers.

> DefaultRetry is the recommended retry for a conflict where multiple clients are making changes to the same resource.

So maybe `DefaultBackoff` is more fitting? I'm not sure

RonFed, line 151: I think we should verify that this behaves as we expect by doing a large batch of un-instrumentation and adding some logs here for debug

instrumentor/controllers/instrumentationdevice/common.go
```diff
@@ -157,46 +148,46 @@ func removeInstrumentationDeviceFromWorkload(ctx context.Context, kubeClient cli
...
+		err = kubeClient.Update(ctx, workloadObj)
+		if err != nil {
+			// if the update fails due to a conflict, we will retry
+			return err
+		}
 
-	modified := result != controllerutil.OperationResultNone
-	if modified {
 		logger := log.FromContext(ctx)
 		logger.V(0).Info("removed instrumentation device from workload", "namespace", workloadObj.GetNamespace(), "kind", workloadObj.GetObjectKind(), "name", workloadObj.GetName(), "reason", uninstrumentReason)
```
RonFed, line 185: Will this log happen even for workloads where we didn't change anything?

instrumentor/controllers/instrumentationdevice/common.go
```diff
@@ -157,46 +148,46 @@ func removeInstrumentationDeviceFromWorkload(ctx context.Context, kubeClient cli
...
 		if err != nil {
 			return err
 		}
 
-		return nil
-	})
+		// if we didn't change anything, we don't need to update the object
+		// skip the api-server call, return no-op and skip the log message
+		if !webhookLabelRemoved && !deviceRemoved && !envChanged {
+			return nil
```
RonFed, line 175: Maybe we should return here a custom error indicating that nothing has been done?
This is for logging below - to avoid logging for un-relevant workloads.

Another option is to set a bool `updatedWorkload` in the `removeInstrumentationDeviceFromWorkload` scope - initialize it to true - here we will set it to false. This will allow us to know if nil is returned because of a successful update or because of this case.

instrumentor/controllers/instrumentationdevice/workload_controllers.go
```diff
@@ -44,9 +45,19 @@ func reconcileSingleInstrumentedApplicationByName(ctx context.Context, k8sClient
...
+			// if there is no instrumented application, make sure the device is removed from the workload pod template manifest
+			workloadName, workloadKind, err := workload.ExtractWorkloadInfoFromRuntimeObjectName(instrumentedAppName)
+			if err != nil {
+				return err
+			}
+			err = removeInstrumentationDeviceFromWorkload(ctx, k8sClient, namespace, workloadKind, workloadName, ApplyInstrumentationDeviceReasonNoRuntimeDetails)
+			return err
+		} else {
+			return err
+		}
```
RonFed, line 58: nit:

```suggestion
		if apierrors.IsNotFound(err) {
			// if there is no instrumented application, make sure the device is removed from the workload pod template manifest
			workloadName, workloadKind, err := workload.ExtractWorkloadInfoFromRuntimeObjectName(instrumentedAppName)
			if err != nil {
				return err
			}
			err = removeInstrumentationDeviceFromWorkload(ctx, k8sClient, namespace, workloadKind, workloadName, ApplyInstrumentationDeviceReasonNoRuntimeDetails)
		}
		return err
```

instrumentor/controllers/instrumentationdevice/workload_controllers.go
```diff
@@ -37,16 +39,26 @@ type StatefulSetReconciler struct {
...
-		// changes in workload when there is no instrumented application is not interesting
-		return client.IgnoreNotFound(err)
+		if apierrors.IsNotFound(err) {
+			// if there is no instrumented application, make sure the device is removed from the workload pod template manifest
+			workloadName, workloadKind, err := workload.ExtractWorkloadInfoFromRuntimeObjectName(instrumentedAppName)
+			if err != nil {
+				return err
+			}
+			err = removeInstrumentationDeviceFromWorkload(ctx, k8sClient, namespace, workloadKind, workloadName, ApplyInstrumentationDeviceReasonNoRuntimeDetails)
+			return err
```
RonFed, line 56: this means that a conflict error will be printed by the controller runtime - right?

RonFed (approved):

PR #1711, opened 2024-11-08
fix(odiglet): never return err from device allocate

blumamir:
When a new pod is scheduled on a node, and the pod contains resource request for odigos device, then kubelet will trigger the Allocate function in odiglet.

Odiglet will try to `Get` the collectors group to calculate enabled signals, which are forwarded to the agent creation to enable / disable specific signals and avoid sending data from applications to a non-existing endpoint.

We must never return an error from the `Allocate` function, even if for some reason the collectors group had error, as it will make the pod non schedulable  with error: `UnexpectedAdmissionError`.

This PR makes it so if we have some error, we log it and start the agent with no enable signals, so odigos will never interfere with the pod start sequence

### DIFF
--- a/odiglet/pkg/instrumentation/plugin.go
+++ b/odiglet/pkg/instrumentation/plugin.go
@@ -10,6 +10,7 @@ import (
 	"github.com/odigos-io/odigos/k8sutils/pkg/env"
 	"github.com/odigos-io/odigos/odiglet/pkg/instrumentation/devices"
 	"github.com/odigos-io/odigos/odiglet/pkg/log"
+	"github.com/pingcap/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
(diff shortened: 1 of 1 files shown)


edeNFed (approved):

odiglet/pkg/instrumentation/plugin.go
```diff
@@ -82,16 +83,25 @@ func (p *plugin) GetPreferredAllocation(ctx context.Context, request *v1beta1.Pr
...
-		return nil, err
-	}
-
-	enabledSignals := make(map[common.ObservabilitySignal]struct{})
-	for _, signal := range nodeCollectorGroup.Status.ReceiverSignals {
-		enabledSignals[signal] = struct{}{}
+		// we should have collectors group created for odigos device to trigger.
+		// however if we don't, just log and do not populate the enabled signals.
+		if apierrors.IsNotFound(err) {
+			log.Logger.V(3).Info("pod with odigos device started, but collectors group not created. disabling all signals for this pod", "collectorGroupName", k8sconsts.OdigosNodeCollectorCollectorGroupName)
```
RonFed, line 97: I think this should have a V(0) level, no? to make sure we see it

blumamir: changed this to be an error

RonFed (approved):

PR #1714, opened 2024-11-09
refactor: extract common predicates to k8s utils

blumamir:
Using the right event filters for controllers is very important to reduce resource utilization, avoid un-needed calls to api-server, and reduce chance to race conditions when multiple things changes at the same time.

We keep refining our controller event filters to achieve better stability for odigos.

Right now, we define our custom predicates near the controller where they are used, but this make it less easy to reuse existing common predicates and share code. This PR move some of the common and reuseable predicates to `k8sutils` module and refactor how they are being set up

### DIFF
--- a/k8sutils/pkg/predicate/objectname.go
+++ b/k8sutils/pkg/predicate/objectname.go
@@ -65,7 +65,7 @@ var OdigosConfigMapPredicate = ObjectNamePredicate{
 //		For(&odigosv1.CollectorsGroup{}).
 //		WithEventFilter(&odigospredicates.OdigosCollectorsGroupCluster).
 //		Complete(r)
-var OdigosCollectorsGroupNode = ObjectNamePredicate{
+var OdigosCollectorsGroupNodePredicate = ObjectNamePredicate{
 	AllowedObjectName: odigosk8sconsts.OdigosNodeCollectorCollectorGroupName,
 }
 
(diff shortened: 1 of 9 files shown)


edeNFed (approved):

k8sutils/pkg/predicate/objectname.go
```diff
@@ -0,0 +1,72 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	cr_predicate "sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+type ObjectNamePredicate struct {
+	AllowedObjectName string
+}
+
+func (o ObjectNamePredicate) Create(e event.CreateEvent) bool {
+	return e.Object.GetName() == o.AllowedObjectName
```
blumamir, line 15: check for nil

scheduler/controllers/nodecollector_controller.go
```diff
@@ -178,7 +95,8 @@ func (r *NodeCollectorsGroupReconciler) SetupWithManager(mgr ctrl.Manager) error
 	err = ctrl.NewControllerManagedBy(mgr).
 		For(&odigosv1.CollectorsGroup{}).
 		Named("nodecollectorgroup-collectorsgroup").
-		WithEventFilter(&clusterCollectorBecomesReadyPredicate{}).
+		WithEventFilter(&odigospredicates.OdigosCollectorsGroupCluster).
+		WithEventFilter(&odigospredicates.CgBecomesReadyPredicate{}).
```
blumamir, line 99: use `And` predicate

RonFed (approved):

PR #1716, opened 2024-11-09
fix: ns reconcile race condition on delete crds

blumamir:
The delete instrumented application controllers in instrumentor are responsible to delete the instrumentated application and instrumentation config once the workload and namespace labels are changed.

Specifically, it has a ns reconcile, and once detects labels change for a ns and the ns is uninstrumented, it loops on all deployments, ds and ss to delete any instrumentations for workloads that inheritaed instrumentation from ns.

There were few bugs fixed in this PR:
- the ns reconciler would list the workloads, and then iterate them one by one. This can take time. When UI instruments workloads, it starts by setting the label on the ns and then iterates the workloads. This mean that the ns reconciler might have read stale data and use it to delete instrumentation, even if the workload has already being labeld. It is fixed by always `Get`ing a fresh workload object when iterating the list.
- the reconciler was attempting to delete any uninstrumented workload. In the ns case, we should only care for workloads without label which inherited the instrumentation from the ns (and ignore those that are already marked with "disabled" which are handled in another controller. This fix is to narrow the check for only those cases.
- the event filter allowed any label changes, it was narrowed to only allow ns label changed from `enabled` to something else, making this reconciler being called less times. For that the cache is modified to store all ns so we can know when the labels has changed. This should not be significant compared to other objects we store in cache.

### DIFF
--- a/instrumentor/controllers/deleteinstrumentedapplication/manager.go
+++ b/instrumentor/controllers/deleteinstrumentedapplication/manager.go
@@ -53,7 +53,7 @@ func SetupWithManager(mgr ctrl.Manager) error {
 		ControllerManagedBy(mgr).
 		Named("deleteinstrumentedapplication-namespace").
 		For(&corev1.Namespace{}).
-		WithEventFilter(predicate.LabelChangedPredicate{}).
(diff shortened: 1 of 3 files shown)


RonFed (approved):
💯

PR #1726, opened 2024-11-10
Node affinity to odiglet nodes

tamirdavid1:


odiglet/cmd/main.go
```diff
@@ -118,6 +126,10 @@ func main() {
 		log.Logger.Error(err, "Failed to stop runtime detector")
 		os.Exit(-1)
 	}
+	// Remove the label before exiting
```
tamirdavid1, line 129: Need to remove if we want pods to still be scheduled on nodes that odiglet was one time on it

PR #1734, opened 2024-11-11
enhance reconcileWorkload to only get the workload kind

RonFed:
Instead of passing an empty object and the object kind, pass only the object kind. Use `workload.ClientObjectFromWorkloadKind(objKind)` within the function.


blumamir (approved):
🥇

PR #1737, opened 2024-11-12
handle conflict error in instrumentation config updates

RonFed:
A typical error we see when trying to update the instrumentation config CR is:

> 2024-11-12T08:43:28Z	ERROR	error updating instrumentation config	{"controller": "instrumentor-instrumentationconfig-instrumentedapplication", "controllerGroup": "odigos.io", "controllerKind": "InstrumentedApplication", "InstrumentedApplication": {"name":"deployment-pricing","namespace":"simple-demo19"}, "namespace": "simple-demo19", "name": "deployment-pricing", "reconcileID": "8c980fc2-ead9-47c7-b2ac-69be6e880002", "workload": "deployment-pricing", "error": "Operation cannot be fulfilled on instrumentationconfigs.odigos.io \"deployment-pricing\": the object has been modified; please apply your changes to the latest version and try again"}

This is due to the Get and Update pattern we use. Following #1710 , use the update error util for this case as well.
Adding to the error handling util - a check for `IsNotFound` error and ignoring it.


instrumentor/controllers/instrumentationconfig/instrumentedapplication_controller.go
```diff
@@ -55,19 +56,18 @@ func (r *InstrumentedApplicationReconciler) Reconcile(ctx context.Context, req c
 
 	instrumentationRules := &odigosv1.InstrumentationRuleList{}
 	err = r.Client.List(ctx, instrumentationRules)
+	if client.IgnoreNotFound(err) != nil {
```
blumamir, line 59: can a list operation return not found? or is it only for safety?

If it's not found, don't we want to return and not cause any updates?

k8sutils/pkg/utils/retryonconflict.go
```diff
@@ -5,12 +5,21 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-func RetryOnConflict(err error) (reconcile.Result, error) {
+// K8SUpdateErrorHandler is a helper function to handle k8s update errors.
+// It returns a reconcile.Result and an error
+// If the error is a conflict error, it returns a requeue result without an error
+// If the error is a not found error, it returns an empty result without an error
+// For other errors, it returns the error as is - which will cause a requeue
```
blumamir, line 12: like 👍

blumamir (approved):

instrumentor/controllers/instrumentationconfig/instrumentedapplication_controller.go
RonFed: Thich check is for errors which are not `NotFound`

PR #1749, opened 2024-11-13
[GEN-1705]: fast-select for sources

BenElferink:


frontend/webapp/containers/main/sources/choose-sources/choose-sources-body/choose-sources-body-simple/index.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
alonkeyval, line 1: this file is duplicated

alonkeyval (approved):

PR #1751, opened 2024-11-14
Support Kubernetes version 1.20.15

david336362:


autoscaler/controllers/datacollection/daemonset.go
```diff
@@ -369,3 +379,29 @@ func patchDaemonSet(existing *appsv1.DaemonSet, desired *appsv1.DaemonSet, ctx c
...
+func isVersionSupported(cfg *rest.Config, minMajor int, minMinor int) (bool, error) {
+	discoveryClient, err := discovery.NewDiscoveryClientForConfig(cfg)
+	if err != nil {
+		return false, err
+	}
+
+	serverVersion, err := discoveryClient.ServerVersion()
+	if err != nil {
+		return false, err
+	}
```
RonFed, line 392: Since this will be called on every reconcile, we need to avoid this re-calculation and querying the API server and do it once in main.

autoscaler/controllers/datacollection/daemonset.go
```diff
@@ -369,3 +379,29 @@ func patchDaemonSet(existing *appsv1.DaemonSet, desired *appsv1.DaemonSet, ctx c
...
+		return false, fmt.Errorf("failed to parse major version: %w", err)
+	}
+
+	minor, err := strconv.Atoi(strings.TrimSuffix(serverVersion.Minor, "+"))
+	if err != nil {
+		return false, fmt.Errorf("failed to parse minor version: %w", err)
+	}
+
+	// Check if the server version meets or exceeds minMajor.minMinor
+	return major > minMajor || (major == minMajor && minor >= minMinor), nil
```
RonFed, line 406: This should be replaced by the standard functions of `apimachinery` for comparing versions

cli/pkg/kube/client.go
```diff
@@ -197,7 +200,53 @@ func (c *Client) DeleteOldOdigosSystemObjects(ctx context.Context, resourceAndNa
 	labelSelector := k8slabels.NewSelector().Add(*systemObject).Add(*notLatestVersion).String()
 	resource := resourceAndNamespace.Resource
 	ns := resourceAndNamespace.Namespace
-	return c.Dynamic.Resource(resource).Namespace(ns).DeleteCollection(ctx, metav1.DeleteOptions{}, metav1.ListOptions{
-		LabelSelector: labelSelector,
-	})
+	cfg := ctrl.GetConfigOrDie()
+	versionSupported, err := isVersionSupported(cfg, 1, 23)
```
RonFed, line 204: `v1.23.0` should be put in a variable of type version with a name that suggests what feature is supported from that version

cli/pkg/kube/client.go
```diff
@@ -197,7 +200,53 @@ func (c *Client) DeleteOldOdigosSystemObjects(ctx context.Context, resourceAndNa
...
+		return false, fmt.Errorf("failed to parse major version: %w", err)
+	}
+
+	minor, err := strconv.Atoi(strings.TrimSuffix(serverVersion.Minor, "+"))
+	if err != nil {
+		return false, fmt.Errorf("failed to parse minor version: %w", err)
+	}
+
+	// Check if the server version meets or exceeds minMajor.minMinor
+	return major > minMajor || (major == minMajor && minor >= minMinor), nil
```
RonFed, line 251: same as other comment about using the `apimachinary` function for comparing versions

autoscaler/controllers/datacollection/daemonset.go
```diff
@@ -179,7 +184,15 @@ func getDesiredDaemonSet(datacollection *odigosv1.CollectorsGroup, configData st
 	maxUnavailable := intstr.FromString("50%")
 	// maxSurge is the number of pods that can be created above the desired number of pods.
 	// we do not want more then 1 datacollection pod on the same node as they need to bind to oltp ports.
-	maxSurge := intstr.FromInt(0)
+	rollingUpdate := &appsv1.RollingUpdateDaemonSet{
+		MaxUnavailable: &maxUnavailable,
+	}
+	cfg := ctrl.GetConfigOrDie()
+	versionSupported, err := isVersionSupported(cfg, 1, 22)
```
RonFed, line 191: `v1.22.0` should be put in a variable of type version with a name suggesting what feature is supported from that version

RonFed (commented):
Added a few comments.
In addition, I think we can move some of the functions to `k8sutils` and some of the versions which are used for feature checking to the `consts.go` file in `k8sutils`.

autoscaler/migration.go
```diff
@@ -50,3 +50,42 @@ func MigrateCollectorsWorkloadToNewLabels(ctx context.Context, c client.Client,
 	return nil
 
 }
+func OldVerKubernetesMigrateCollectorsWorkloadToNewLabels(ctx context.Context, c client.Client, ns string) error {
```
blumamir, line 53: The purpose of this function is to migrage pre 91 versions due to some breaking changes.

Since no one can currently sun version `v1.0.91` with old k8s versions, I think this functionallity can simply be skipped for old versions (and later removed for all versions once we deprecate `v1.0.91`

autoscaler/main.go
```diff
@@ -161,11 +169,30 @@ func main() {
...
+	if k8sVersion != nil && k8sVersion.GreaterThan(version.MustParse("v1.23")) {
+		// Use the cached client for versions >= 1.23
+		err = MigrateCollectorsWorkloadToNewLabels(context.Background(), mgr.GetClient(), odigosNs)
+		if err != nil {
+			setupLog.Error(err, "unable to migrate collectors workload to new labels")
+			os.Exit(1)
+		}
+	} else {
+		// Create a non-caching client for versions < 1.23
+		migrationClient, err = client.New(cfg, client.Options{Scheme: scheme})
```
blumamir, line 185: What makes this client non caching?

cli/cmd/install.go
```diff
@@ -80,23 +80,21 @@ This command will install k8s components that will auto-instrument your applicat
...
+			fmt.Println("Unknown Kubernetes cluster detected, proceeding with installation")
+		}
+
+		if !errors.Is(err, autodetect.ErrCannotDetectK8sVersion) {
+			autodetect.CurrentKubernetesVersion.Version = details.K8SVersion
 			if details.K8SVersion.LessThan(minK8SVersionForInstallation) {
 				fmt.Printf("\033[31mERROR\033[0m Odigos requires Kubernetes version %s or higher but found %s, aborting\n", minK8SVersionForInstallation.String(), details.K8SVersion.String())
 				os.Exit(1)
 			}
 			fmt.Printf("Detected cluster: %s Kubernetes version: %s\n", details.Kind, details.K8SVersion.String())
```
blumamir, line 95: If we failed to detect the cluster kind, can this print some confusing message?

cli/go.mod
```diff
@@ -15,22 +15,32 @@ require (
 	k8s.io/apiextensions-apiserver v0.31.0
 	k8s.io/apimachinery v0.31.0
 	k8s.io/client-go v0.31.0
+	sigs.k8s.io/controller-runtime v0.19.0
```
blumamir, line 18: should be removed I think

cli/cmd/install.go
david336362: fixed

blumamir (approved):

PR #1757, opened 2024-11-15
refactor(autoscaler): use common predicates for event filters

blumamir:
This PR:

- shifts to use the common predicates in `k8sutils` over writing them specifically in the controllers.
- splits logic into smaller parts and combine them with `predicate.And` for better readability.
- adds few straight forward predicates to places where they are currently missing

### DIFF
--- a/autoscaler/controllers/secret_controller.go
+++ b/autoscaler/controllers/secret_controller.go
@@ -5,10 +5,10 @@ import (
 
 	controllerconfig "github.com/odigos-io/odigos/autoscaler/controllers/controller_config"
 	"github.com/odigos-io/odigos/autoscaler/controllers/gateway"
+	odigospredicate "github.com/odigos-io/odigos/k8sutils/pkg/predicate"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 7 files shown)


autoscaler/controllers/secret_controller.go
```diff
@@ -62,10 +35,12 @@ func (r *SecretReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 
-// SetupWithManager sets up the controller with the Manager.
 func (r *SecretReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&corev1.Secret{}).
-		WithEventFilter(&secretPredicate{}).
+		// we need to handle secrets only when they are updated.
+		// this is to trigger redeployment of the cluster collector in case of destination secret change.
+		// when the secret was just created (via auto-scaler restart or initial deployment), the cluster collector will be reconciled by other controllers.
+		WithEventFilter(predicate.And(&odigospredicate.OnlyUpdatesPredicate{}, &predicate.ResourceVersionChangedPredicate{})).
```
RonFed, line 44: Should we use a generation change predicate instead of a resource version predicate here?

RonFed (approved):

autoscaler/controllers/secret_controller.go
blumamir: I think secrets has no spec, thus no generation

PR #1758, opened 2024-11-15
Include `-0` for the min version to allow EKS versions

RonFed:
For versions such as `v1.23.17-eks-ce1d5eb` we would get the following error:

> Error: chart requires kubeVersion: >= 1.20.15 which is incompatible with Kubernetes v1.23.17-eks-ce1d5eb


edeNFed (approved):

esara:
@RonFed thanks for fixing this, just figured out the same problem

ps: for reference, awkward conversation with AWS https://github.com/helm/helm/issues/10375

PR #1759, opened 2024-11-15
Add PersistentPreRun to the root CLI cmd to set the kube client and cluster details once for all commands

RonFed:
Add `PersistentPreRun` which will be inherited by all sub-commands of the root command.
Perform a client initialzation in that function and set the cluster kind and k8s version as well - so they can be used by all commands. Both the kube client and the cluster details are set on the command context and can be used by all the commands.
The helper function from k8sutils `GetKubernetesVersion` can't be used outside of the cluster by the CLI since it as using a rest client which states:

> // InClusterConfig returns a config object which uses the service account
// kubernetes gives to pods. It's intended for clients that expect to be
// running inside a pod running on kubernetes. It will return ErrNotInCluster
// if called from a process not running in a kubernetes environment.
func InClusterConfig() (*Config, error) {

Hence, removing the use of this function from the CLI, and using the k8s version which was calculated in the `PersistentPreRun` function.
Before this change `odigos upgrade` resulted in errors such as:

> Syncing customresourcedefinitionsDeleteOldOdigosSystemObjects failed to get k8s version, proceeding.. :unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined                 ✔
Syncing deploymentsDeleteOldOdigosSystemObjects failed to get k8s version, proceeding.. :unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined                               ✔
Syncing daemonsetsDeleteOldOdigosSystemObjects failed to get k8s version, proceeding.. :unable to load in-cluster

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -23,7 +23,6 @@ import (
 	"github.com/spf13/cobra"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/version"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
@@ -46,12 +45,6 @@ var (
 	imagePrefix       string
 )
 
-var (
-	// minK8SVersionForInstallation is the minimum Kubernetes version required for Odigos installation
-	// this value must be in sync with the one defined in the kubeVersion field in Chart.yaml
-	minK8SVersionForInstallation = version.MustParse("v1.20.15")
-)
-
 type ResourceCreationFunc func(ctx context.Context, cmd *cobra.Command, client *kube.Client, ns string) error
 
 // installCmd represents the install command
@@ -61,9 +54,8 @@ var installCmd = &cobra.Command{
 	Long: `Install Odigos in your kubernetes cluster.
 This command will install k8s components that will auto-instrument your applications with OpenTelemetry and send traces, metrics and logs to any telemetry backend`,
 	Run: func(cmd *cobra.Command, args []string) {
-
-		client := kube.GetCLIClient()
 		ctx := cmd.Context()
+		client := kube.KubeClientFromContextOrExit(ctx)
 		ns := cmd.Flag("namespace").Value.String()
 
 		// Check if Odigos already installed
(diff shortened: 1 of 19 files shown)


cli/pkg/autodetect/kind.go
```diff
@@ -48,18 +42,23 @@ type ClusterDetails struct {
...
+
+func SetK8SClusterDetails(ctx context.Context, kc string, client *kube.Client)  {
 	details := k8sutils.GetCurrentClusterDetails(kc)
 	serverVersion, err := client.Discovery().ServerVersion()
 	if err != nil {
-		return ClusterDetails{}, errors.Join(ErrCannotDetectK8sVersion, err)
+		currentClusterDetails.K8SVersion = nil
 	}
 	ver := version.MustParse(serverVersion.String())
+	currentClusterDetails.K8SVersion = ver
```
blumamir, line 60: Above, we now set `currentClusterDetails.K8SVersion = nil` and override this value here. I guess it works because `serverVersion` is empty and `MustParse` it returns nil?

Anyway, I think it's better to have this in an `else`.

Also - why `MustParse` and not `Parse` that can also report error?

cli/pkg/autodetect/kind.go
```diff
@@ -48,18 +42,23 @@ type ClusterDetails struct {
...
-func DetectK8SClusterDetails(ctx context.Context, kc string, client *kube.Client) (ClusterDetails, error) {
+func GetClusterKind() Kind {
+	return currentClusterDetails.Kind
+}
+
+func SetK8SClusterDetails(ctx context.Context, kc string, client *kube.Client)  {
 	details := k8sutils.GetCurrentClusterDetails(kc)
 	serverVersion, err := client.Discovery().ServerVersion()
 	if err != nil {
-		return ClusterDetails{}, errors.Join(ErrCannotDetectK8sVersion, err)
```
blumamir, line 60: Should we have any visibility for this error anywhere?

cli/pkg/autodetect/kind.go
```diff
@@ -48,18 +42,23 @@ type ClusterDetails struct {
...
+func GetK8SVersion() *version.Version {
+	return currentClusterDetails.K8SVersion
+}
 
-func DetectK8SClusterDetails(ctx context.Context, kc string, client *kube.Client) (ClusterDetails, error) {
+func GetClusterKind() Kind {
+	return currentClusterDetails.Kind
+}
+
+func SetK8SClusterDetails(ctx context.Context, kc string, client *kube.Client)  {
```
blumamir, line 53: Perhaps a small comment, mentioning that this function will set global variables, and is a requirement for the `GetK8SVersion` and `GetClusterKind` above to work.

cli/pkg/kube/client.go
```diff
@@ -193,15 +210,14 @@ func (c *Client) ApplyResource(ctx context.Context, configVersion int, obj Objec
...
 	systemObject, _ := k8slabels.NewRequirement(odigoslabels.OdigosSystemLabelKey, selection.Equals, []string{odigoslabels.OdigosSystemLabelValue})
 	notLatestVersion, _ := k8slabels.NewRequirement(odigoslabels.OdigosSystemConfigLabelKey, selection.NotEquals, []string{strconv.Itoa(configVersion)})
 	labelSelector := k8slabels.NewSelector().Add(*systemObject).Add(*notLatestVersion).String()
 	resource := resourceAndNamespace.Resource
 	ns := resourceAndNamespace.Namespace
-	k8sVersion, err := odigosver.GetKubernetesVersion()
-	if err != nil {
-		fmt.Printf("DeleteOldOdigosSystemObjects failed to get k8s version, proceeding.. :%v", err)
+	if k8sVersion == nil {
+		fmt.Printf("DeleteOldOdigosSystemObjects failed to get k8s version, proceeding.. ")
```
blumamir, line 220: The `DeleteOldOdigosSystemObjects` function is called many times, and will print this message to the console in case the version is missing.
Can we alternatively print the message just once and ignore it here?

cli/pkg/kube/client.go
```diff
@@ -193,15 +210,14 @@ func (c *Client) ApplyResource(ctx context.Context, configVersion int, obj Objec
...
 }
 
-func (c *Client) DeleteOldOdigosSystemObjects(ctx context.Context, resourceAndNamespace ResourceAndNs, configVersion int) error {
+func (c *Client) DeleteOldOdigosSystemObjects(ctx context.Context, resourceAndNamespace ResourceAndNs, configVersion int, k8sVersion *version.Version) error {
 	systemObject, _ := k8slabels.NewRequirement(odigoslabels.OdigosSystemLabelKey, selection.Equals, []string{odigoslabels.OdigosSystemLabelValue})
 	notLatestVersion, _ := k8slabels.NewRequirement(odigoslabels.OdigosSystemConfigLabelKey, selection.NotEquals, []string{strconv.Itoa(configVersion)})
 	labelSelector := k8slabels.NewSelector().Add(*systemObject).Add(*notLatestVersion).String()
 	resource := resourceAndNamespace.Resource
 	ns := resourceAndNamespace.Namespace
-	k8sVersion, err := odigosver.GetKubernetesVersion()
```
blumamir, line 202: I suggest adding a comment to this function, which was the source of the error.
We can mention that it uses `"k8s.io/client-go/rest"` and can only be used for pods that runs inside the cluster.

cli/pkg/kube/client.go
```diff
@@ -45,7 +44,25 @@ type Object interface {
 	runtime.Object
 }
 
-func CreateClient(cmd *cobra.Command) (*Client, error) {
+var cliClient *Client
```
blumamir, line 47: I know having globals is convenient, but can be considered problematic for several reasons.

cli/cmd/install.go
```diff
@@ -78,26 +74,45 @@ This command will install k8s components that will auto-instrument your applicat
...
 				os.Exit(1)
 			}
-			fmt.Printf("Detected cluster: Kubernetes version: %s\n", details.K8SVersion.String())
+			fmt.Printf("Detected cluster: Kubernetes version: %s\n", k8sVersion.String())
 		} else {
 			fmt.Println("Unknown Kubernetes version detected, proceeding with installation")
 		}
 
+
+		// kc := cmd.Flag("kubeconfig").Value.String()
```
blumamir, line 96: Do we want to keep this?

blumamir (approved):
🥇

PR #1825, opened 2024-11-22
feat: include memory limit in memory settings

blumamir:
For cluster collector, we currently populate the resource request for memory with default value of `500MiB`.

However, we do not set a limit on the resource request.

This PR adds a limit for memory request for the cluster collector, set by default to `1.25` times the memory request, `625MiB` by default. It is not expected to be reached, but set as protection to guarantee we supply a small slack for extreme bursts, but capping the memory at a reasonable size to protect the cluster


api/odigos/v1alpha1/collectorsgroup_types.go
```diff
@@ -40,6 +40,13 @@ type CollectorsGroupMemorySettings struct {
 	// it will be embedded in the as a resource request of the form "memory: <value>Mi"
 	MemoryRequestMiB int `json:"memoryRequestMiB"`
 
+	// This option sets the limit on the memory usage of the collector.
+	// since the memory limiter mechanism is heuristic, and operates on fixed intervals,
+	// while it cannot fully prevent OOMs, it can help in reducing the chances of OOMs in edge cases.
+	// the settings should prevent the collector from exceeding the memory request,
+	// so one can set this to the same value as the memory request or higher to allow for some buffer for bursts.
+	MemoryLimitMiB int `json:"memoryLimitMiB"`
```
RonFed, line 48: Do you think it is better to configure this as relative to the MemoryRequest field? (i/e with precentages or fractions relative to the request amount.)

api/odigos/v1alpha1/collectorsgroup_types.go
```diff
@@ -40,6 +40,13 @@ type CollectorsGroupMemorySettings struct {
 	// it will be embedded in the as a resource request of the form "memory: <value>Mi"
 	MemoryRequestMiB int `json:"memoryRequestMiB"`
 
+	// This option sets the limit on the memory usage of the collector.
+	// since the memory limiter mechanism is heuristic, and operates on fixed intervals,
+	// while it cannot fully prevent OOMs, it can help in reducing the chances of OOMs in edge cases.
+	// the settings should prevent the collector from exceeding the memory request,
+	// so one can set this to the same value as the memory request or higher to allow for some buffer for bursts.
```
RonFed, line 47: Should we write the k8s value that will map into? and state that it will impact when the collector is OOMed?

RonFed (approved):
Added a nit about the comments.

Also discussed offline about the need to make all the precntages sizes configurable

PR #1847, opened 2024-11-25
feat: add running field to ii + set it to false once opamp shutdown

tamirdavid1:


RonFed (changes requested):
I'm not sure about this change, would prefer to discuss it first

PR #1879, opened 2024-11-27
Refactor Odiglet main function

RonFed:
This PR refactors the top-level initialization of Odiglet:

* Split to `newOdilget` and `run` functions. The first initializes all the different components, the second launches them, and handles closing them all gracefully on shutdown.
* Some of the components were slightly refactored to support this:  `OpAmp`, `runtime-detector`, `pprof` server and kube manager.


odiglet/cmd/main.go
```diff
@@ -106,25 +81,102 @@ func main() {
...
+		clientset:     clientset,
+		mgr:           mgr,
+		ctx:           ctx,
+		ebpfDirectors: ebpfDirectors,
 	}
+}
 
-	<-ctx.Done()
-	for _, director := range ebpfDirectors {
+func (o *odiglet) run() {
```
blumamir, line 92: I think it's a bit non-conventional to have the `ctx` as property.

Consider sending it to the `run` function

odiglet/cmd/main.go
```diff
@@ -39,24 +41,14 @@ func odigletInitPhase() {
...
-		os.Exit(1)
-	}
+type odiglet struct {
+	clientset     *kubernetes.Clientset
+	mgr           ctrl.Manager
+	ctx           context.Context
+	ebpfDirectors ebpf.DirectorsMap
+}
 
+func newOdiglet() *odiglet {
```
blumamir, line 51: Consider returning an error here and let the `main` call the logger and `os.Exit`

odiglet/cmd/main.go
```diff
@@ -106,25 +81,102 @@ func main() {
...
-	<-ctx.Done()
-	for _, director := range ebpfDirectors {
+func (o *odiglet) run() {
+	var wg sync.WaitGroup
+
+	// Start pprof server
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		common.StartPprofServer(o.ctx, log.Logger)
```
blumamir, line 99: should we log/use the `error` returned from this function?

blumamir (approved):

odiglet/cmd/main.go
RonFed: Right, currently I keep it that way since we have the `initEbpf` part which gets a context, but that will be removed soon.

PR #1886, opened 2024-11-29
feat: run python agent alogside NR

tamirdavid1:


edeNFed (approved):

blumamir (approved):
Looks good 👍 

How will it look in the UI now? I remember there is a message regarding other agent. just checking if we are expecting to see it or not, and if we do - what does it say

PR #1903, opened 2024-12-02
Reload Collectors without restart when ConfigMap changes

damemi:
Kubernetes automatically updates mounted configmaps in a pod (see https://kubernetes.io/docs/tutorials/configuration/updating-configuration-via-a-configmap/). But, the workload needs to watch for the update itself. So in our case, if the collector knows to watch for file changes, it can automatically signal a dynamic config reload without restarting the collector.

This forks the default fileprovider in our collectors to create a new odigosfileprovider that uses fsnotify to watch for updates to the collector ConfigMap. When the ConfigMap is updated, the new fileprovider will signal the collector to hot-reload its config. Technically, we watch for `fsnotify.Remove` events, because the projected configmap data is a symlink, not an actual copy of the configmap (meaning that watching `fsnotify.Write` doesn't trigger any update).

This means we don't need to restart the collector deployments or daemonsets for basic config updates, so those controllers have been updated to no longer update deployments when just the configmap has changed. They can of course still be manually redeployed with `kubectl`.

In my manual testing, it took about 1 minute for the change to be reflected, which is due to the default kubelet sync period (https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically). Not sure how we could test for this automatically but working on that


collector/providers/odigosfileprovider/provider.go
```diff
@@ -0,0 +1,96 @@
...
+				}
+				// k8s configmaps are mounted as symlinks; need to watch for remove, not write
+				if event.Has(fsnotify.Remove) {
+					watcher.Remove(file)
+					watcher.Add(file)
+					wf(&confmap.ChangeEvent{})
+				}
+			case err, ok := <-watcher.Errors:
+				if !ok {
+					wf(&confmap.ChangeEvent{Error: fmt.Errorf("error watching event %+v", err)})
```
blumamir, line 81: We will it this line if `!ok` meaning that the channel is closed. In this case, I think `err` will be nil, no?

collector/providers/odigosfileprovider/provider.go
```diff
@@ -0,0 +1,96 @@
...
+					wf(&confmap.ChangeEvent{Error: fmt.Errorf("error watching event")})
+				}
+				// k8s configmaps are mounted as symlinks; need to watch for remove, not write
+				if event.Has(fsnotify.Remove) {
+					watcher.Remove(file)
+					watcher.Add(file)
+					wf(&confmap.ChangeEvent{})
+				}
+			case err, ok := <-watcher.Errors:
+				if !ok {
```
blumamir, line 80: What if `ok` is not false? then it's an error from the watcher. Should we log it somewhere? should we return it on the `wf` callback?

blumamir (commented):
Thanks mike, added few comments.

I wonder if we should now revisit this mechanism: https://github.com/odigos-io/odigos/blob/89aa4d0a6147ee2d096bf66acd3f6289f52fefcc/autoscaler/controllers/datacollection/daemonset.go#L49

It was added to prevent a burst of restarts to the node collectors when sources a reconciled in a batch one-by-one. we can do it also as a followup PR. If we can remove it, we will have odigos pipeline starting 5 seconds earlier.

collector/odigosotelcol/go.mod
```diff
@@ -85,13 +85,12 @@ require (
 	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/zipkinreceiver v0.106.0
 	github.com/stretchr/testify v1.9.0
 	go.opentelemetry.io/collector/component v0.106.0
-	go.opentelemetry.io/collector/confmap v0.106.0
+	go.opentelemetry.io/collector/confmap v1.20.0
```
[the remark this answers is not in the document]
edeNFed: Or configure dependabot to automatically do it

collector/builder-config.yaml
```diff
@@ -106,11 +106,19 @@ connectors:
   - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/connector/servicegraphconnector v0.106.0
   - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/connector/spanmetricsconnector v0.106.0
 
+providers:
+  - gomod: go.opentelemetry.io/collector/odigos/providers/odigosfileprovider v0.106.0 # fork default file provider for config reloading
+  - gomod: go.opentelemetry.io/collector/confmap/provider/envprovider v1.20.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/httpprovider v1.20.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/httpsprovider v1.20.0
```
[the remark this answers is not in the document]
damemi: Not that I know of, so I included them to be safe. The reason is because right now, these are all silently included by default since we don't explicitly set any `providers` in our builder config.

Now that I'm adding the `odigosfileprovider` here, not providing these means they'll be removed from our build. If that's okay then for sure we can clean that up!

collector/providers/odigosfileprovider/provider.go
damemi: Yeah I misunderstood what was being returned from the channels here

collector/providers/odigosfileprovider/provider.go
damemi: yup, updated

collector/providers/odigosfileprovider/provider.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	// Clean the path before using it.
+	file := filepath.Clean(uri[len(schemeName)+1:])
+	content, err := os.ReadFile(file)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read the file %v: %w", uri, err)
+	}
+
+	watcher, err := fsnotify.NewWatcher()
```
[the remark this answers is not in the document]
damemi: I was supposed to do that, added handling around shutdown and making sure the routine closes

collector/providers/odigosfileprovider/provider.go
```diff
@@ -0,0 +1,96 @@
...
+//	local-path		= [ drive-letter ] file-path
+//	drive-letter	= ALPHA ":"
+//
+// The "file-path" can be relative or absolute, and it can be any OS supported format.
+//
+// Examples:
+// `file:path/to/file` - relative path (unix, windows)
+// `file:/path/to/file` - absolute path (unix, windows)
+// `file:c:/path/to/file` - absolute path including drive-letter (windows)
+// `file:c:\path\to\file` - absolute path including drive-letter (windows)
```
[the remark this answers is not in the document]
damemi: done

collector/providers/odigosfileprovider/provider.go
```diff
@@ -0,0 +1,138 @@
...
+		return nil, fmt.Errorf("unable to read the file %v: %w", uri, err)
+	}
+
+	err = fmp.watcher.Add(file)
+	if err != nil {
+		return nil, err
+	}
+
+	// start a new watcher routine only if one isn't already running, since Retrieve could be called multiple times
+	if !fmp.running {
```
RonFed, line 91: can we use `sync.Once` for this pattern? (instead of boolean and mutex)

collector/providers/odigosfileprovider/provider.go
```diff
@@ -0,0 +1,138 @@
...
+		fmp.wg.Add(1)
+		go func() {
+			fmp.running = true
+			defer fmp.wg.Done()
+		LOOP:
+			for {
+				select {
```
RonFed, line 98: should the context argument be passed for this select, and having `<-ctx.Done` as one of the cases?

collector/providers/odigosfileprovider/provider.go
```diff
@@ -0,0 +1,138 @@
...
+
+	"github.com/fsnotify/fsnotify"
+	"go.opentelemetry.io/collector/confmap"
+	"go.uber.org/zap"
+)
+
+const schemeName = "file"
+
+type provider struct {
+	wg sync.WaitGroup
```
RonFed, line 23: nit: since we are aiming for one goroutine, this can be defined as:
```suggestion
	done chan struct{}
```

tests/e2e/workload-lifecycle/chainsaw-test.yaml
RonFed, line 1: I'm not sure if the `workload-lifcycle` is the best place to add these
@blumamir WDYT?

collector/builder-config.yaml
```diff
@@ -106,11 +106,19 @@ connectors:
   - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/connector/servicegraphconnector v0.106.0
   - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/connector/spanmetricsconnector v0.106.0
 
+providers:
+  - gomod: go.opentelemetry.io/collector/odigos/providers/odigosfileprovider v0.106.0 # fork default file provider for config reloading
+  - gomod: go.opentelemetry.io/collector/confmap/provider/envprovider v0.106.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/httpprovider v0.106.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/httpsprovider v0.106.0
+  - gomod: go.opentelemetry.io/collector/confmap/provider/yamlprovider v0.106.0
```
RonFed, line 114: Why are those needed?

tests/e2e/workload-lifecycle/chainsaw-test.yaml
damemi: @RonFed I was actually thinking we should add a full suite of tests specifically for actions, processors, etc. I think that is its own bigger task though (reevaluating our tests)

collector/builder-config.yaml
damemi: See https://github.com/odigos-io/odigos/pull/1903#discussion_r1870208029

collector/providers/odigosfileprovider/provider.go
damemi: I'm not sure `sync.Once` is what we want. I think that will only let the routine be called once, ever, for this instance of the provider. What I'm doing here considers the case where the watcher closes (for whatever reason) and ends the goroutine, then `Retrieve` is called again on the same Provider. In that case, we should start up the watcher again. `sync.Once` won't start it a 2nd time

collector/providers/odigosfileprovider/provider.go
damemi: good call

collector/providers/odigosfileprovider/provider.go
damemi: good idea, done

collector/builder-config.yaml
RonFed: I think we can remove them now that we have a provider and we don't use them.

autoscaler/controllers/gateway/deployment.go
```diff
@@ -42,8 +42,8 @@ func syncDeployment(dests *odigosv1.DestinationList, gateway *odigosv1.Collector
 		return nil, errors.Join(err, errors.New("failed to get secrets hash"))
 	}
 
-	// Calculate the hash of the config data and the secrets version hash, this is used to make sure the gateway will restart when the config changes
-	configDataHash := common.Sha256Hash(fmt.Sprintf("%s-%s", configData, secretsVersionHash))
+	// Use the hash of the secrets  to make sure the gateway will restart when the secrets (mounted as environment variables) changes
+	configDataHash := common.Sha256Hash(secretsVersionHash)
```
RonFed, line 46: Can we remove the `configData` argument from this function?

collector/providers/odigosfileprovider/provider.go
```diff
@@ -0,0 +1,146 @@
...
+	err = fmp.watcher.Add(file)
+	if err != nil {
+		return nil, err
+	}
+
+	// start a new watcher routine only if one isn't already running, since Retrieve could be called multiple times
+	if !fmp.running {
+		go func() {
+			fmp.running = true
+			defer func() { fmp.done <- struct{}{} }()
```
RonFed, line 95: nit
```suggestion
			defer close(fmp.done)
```

collector/providers/odigosfileprovider/provider.go
```diff
@@ -0,0 +1,146 @@
...
+
+	err = fmp.watcher.Add(file)
+	if err != nil {
+		return nil, err
+	}
+
+	// start a new watcher routine only if one isn't already running, since Retrieve could be called multiple times
+	if !fmp.running {
+		go func() {
+			fmp.running = true
```
RonFed, line 94: I think there is possible non-safe concurrent access to this flag since this is set inside a goroutine, and being read from the outer `Retrieve`

RonFed (approved):
Looks great, left a few non-blocking comments (except maybe the concurrent access to the `running` flag in the provider)

collector/builder-config.yaml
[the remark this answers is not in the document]
blumamir: I am pretty sure they are not used, so it's good opportunity to clean this unused component.
We can do it later on a different PR if you prefer

collector/odigosotelcol/go.mod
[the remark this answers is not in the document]
blumamir: The PR still shows `v1.20.0`, do we need to revisit this?

collector/providers/odigosfileprovider/provider.go
blumamir, line 95: If this code is invoked from a flow which isn't Shutdown, this call will block indefinitely, right?

collector/builder-config.yaml
damemi: We probably will still want the env provider at least, others I think can be removed

autoscaler/controllers/gateway/deployment.go
damemi: that's what I did, since we don't want a change in the configdata to trigger a new rollout, we don't need to hash the config data anymore. is that what you were asking?

RonFed: I was referring to the function signature of `syncDeployment`:

```Go
func syncDeployment(dests *odigosv1.DestinationList, gateway *odigosv1.CollectorsGroup, configData string,
	ctx context.Context, c client.Client, scheme *runtime.Scheme, imagePullSecrets []string, odigosVersion string) (*appsv1.Deployment, error)
```

collector/providers/odigosfileprovider/go.mod
```diff
@@ -0,0 +1,23 @@
+module odigosfileprovider
+
+go 1.22.0
+
+toolchain go1.22.6
+
+require (
+	github.com/fsnotify/fsnotify v1.8.0
+	go.opentelemetry.io/collector/confmap v1.20.0
```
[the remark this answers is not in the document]
damemi: missed this, sorry

collector/providers/odigosfileprovider/provider.go
```diff
@@ -0,0 +1,146 @@
...
+
+const schemeName = "file"
+
+type provider struct {
+	done chan struct{}
+	mu   sync.Mutex
+
+	watcher fsnotify.Watcher
+	running bool
+	logger  zap.Logger
```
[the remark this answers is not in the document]
damemi: yup done

collector/providers/odigosfileprovider/provider.go
damemi: good point, I switched this back to using a waitgroup which I think will fix that

collector/providers/odigosfileprovider/provider.go
damemi: updated this to lock around the variable. this feels a little like "communicating by sharing memory" but I'm not sure of a better way to do what we want here

autoscaler/controllers/gateway/deployment.go
damemi: Ah, updated in https://github.com/odigos-io/odigos/pull/1903/commits/6d53ab0d054b12fd942e23174945c2b6c9abcefc

PR #1928, opened 2024-12-05
[GEN-1927]: add a Cypress test for onboarding-flow

BenElferink:


BenElferink:
WIP - struggling to get namespaces in GitHub environment

PR #1955, opened 2024-12-08
Add missing destination providers

breadluvr:
I found that the destinations in `backends-overview.mdx` didn't match those in this file so I added them. Also alphabetically sorted the cards.


BenElferink (approved):
Looks great, thanks! 😊

PR #1957, opened 2024-12-09
feat: odigos to run in parallel to otheragents using config

tamirdavid1:


cli/cmd/resources/odigosconfig.go
```diff
@@ -51,6 +51,13 @@ func (a *odigosConfigResourceManager) InstallFromScratch(ctx context.Context) er
 	collectorGatewayConfig := GetGatewayConfigBasedOnSize(sizingProfile)
 	a.config.CollectorGateway = collectorGatewayConfig
 
+	if AgentsCanRunConcurrently(a.config.Profiles) {
+		value := true
+		a.config.AllowConcurrentAgents = &value
+	} else {
+		a.config.AllowConcurrentAgents = nil
+	}
```
blumamir, line 59: If the user sets the value for this in config, wouldn't this code overwrite it's configuration even when a relevant profile is not set at all?

blumamir (approved):

blumamir (approved):

PR #1980, opened 2024-12-12
Generalize instrumentation manager

RonFed:
A follow up to #1776:
* Make the instrumentation manager generic, and move it to a new module.
* Add a k8s-oriented implementation of the manager's different interfaces under `odiglet/pkg/ebpf`


instrumentation/manager.go
```diff
@@ -0,0 +1,327 @@
...
+	if err != nil {
+		return errors.Join(err, errFailedToGetConfigGroup)
+	}
+
+	factory, found := m.factories[otelDisto]
+	if !found {
+		return errNoInstrumentationFactory
+	}
+
+	// Fetch initial config based on the InstrumentationConfig CR
```
RonFed, line 229: update comment

instrumentation/manager.go
```diff
@@ -0,0 +1,348 @@
...
+				if err != nil && !errors.Is(err, errNoInstrumentationFactory) {
+					m.logger.Error(err, "failed to handle process exec event")
+				}
+			case detector.ProcessExitEvent:
+				m.cleanInstrumentation(ctx, e.PID)
+			}
+		case configUpdate := <-m.configUpdates:
+			if len(configUpdate) == 0 {
+				break
+			}
```
blumamir, line 171: nit: this is redundant as the following loop will not run in this case

instrumentation/manager.go
```diff
@@ -0,0 +1,348 @@
...
+}
+
+func (m *manager[Details, ConfigGroup]) cleanInstrumentation(ctx context.Context, pid int) {
+	details, found := m.detailsByPid[pid]
+	if !found {
+		m.logger.V(3).Info("no instrumentation found for exiting pid, nothing to clean", "pid", pid)
+		return
+	}
+
+	m.logger.Info("cleaning instrumentation resources", "pid", pid)
```
blumamir, line 205: nit: we can add the podWorkload here which can be more useful for debugging

instrumentation/manager.go
```diff
@@ -0,0 +1,348 @@
...
+	errFailedToGetDetails       = errors.New("failed to get details for process event")
+	errFailedToGetDistribution  = errors.New("failed to get otel distribution for details")
+	errFailedToGetConfigGroup   = errors.New("failed to get config group")
+)
+
+// ConfigUpdate is used to send a configuration update request to the manager.
+// The manager will apply the configuration to all instrumentations that match the config group.
+type ConfigUpdate[configGroup ConfigGroup] map[configGroup]Config
+
+type instrumentationDetails[details Details, configGroup ConfigGroup] struct {
```
blumamir, line 26: Since we already have `Details`, and now `instrumentationDetails` with a property `details` inside it, can we have better names so it's less confusing?

This creates code like `details.details`

blumamir (approved):

instrumentation/manager.go
RonFed: Renamed `Details` to `ProcessGroup` and added comments explaining the meaning behind it

PR #1984, opened 2024-12-13
feat: add nop destination and cleanup deprecated components

blumamir:
This PR:

- removes the deprecated and unused logging exporter, which is replaced with the debug exporter which we do use.
- removed the deprecated `ballast` extension which  we don't use and is deprecated in favor of `GOMEMLIMIT` which we do use.
- adds the `nop` exporter which is useful for development and testing (for example - load scenarios)
- adds documentation in `CONTRIBUTING.md` regarding how to use the debug and nop destination for development.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -251,3 +251,49 @@ spec:
 ```
 
 Once this is done, you can use the .vscode/launch.json configuration and run instrumentor local for debugging.
+
+## Odigos Collector Distribution
+
+### Debugging and Trouble Shooting
+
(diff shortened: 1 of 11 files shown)


edeNFed (approved):
Cool idea! 
Not: I think adding it as a target in Makefile will be more convenient than documenting in markdown file

RonFed (approved):

blumamir:
> Cool idea! nit: I think adding it as a target in Makefile will be more convenient than documenting in markdown file

good idea. Added and updated the markdown

PR #1986, opened 2024-12-13
chore: have crds in helm chart instead of copying

blumamir:
The odigos CRD yamls are currently generated in the api module. Whenever the helm chart should be used for development, we needed to run `cp -r ./api/config/crd/bases/* ./helm/odigos/templates/crds/` to have the helm repo working.

I think it's nicer to have all the files right in the helm. template, and remove the need to `cp` in release and deplpoyment. Added a check in CI to make sure the files are aligned


edeNFed (approved):

RonFed (changes requested):
Can we tell `controller-gen` to generate the files directly in the helm directory instead of having two copies in git?

RonFed (approved):

PR #1989, opened 2024-12-15
Use GOMAXPROCS in gateway based on limits.cpu

RonFed:
Relevant references on why this is important:
https://www.ardanlabs.com/blog/2024/02/kubernetes-cpu-limits-go.html
https://engineering.grab.com/performance-bottlenecks-go-apps


tests/e2e/cli-upgrade/chainsaw-test.yaml
```diff
@@ -103,10 +103,6 @@ spec:
       try:
         - apply:
             file: ../../common/apply/add-tempo-traces-destination.yaml
-    - name: Odigos pipeline ready
-      try:
-        - assert:
-            file:  ../../common/assert/pipeline-ready.yaml
```
blumamir, line 109: This guaranteed that we do not progress until the pipeline is ready.
Wonder if removing the assert can cause issues

tests/common/assert/pipeline-ready.yaml
```diff
@@ -55,12 +55,19 @@ spec:
...
+                  containerName: gateway
+                  divisor: "0"
+                  resource: limits.cpu
           name: gateway
           resources:
             requests:
               (memory != null): true
             limits:
               (memory != null): true
+              (cpu != null): true
```
blumamir, line 70: since we have it for both limits and requests, can be nice to add it in both places :)

blumamir (approved):

tests/e2e/cli-upgrade/chainsaw-test.yaml
RonFed: I added this step back, but only to make sure all the pods are ready, not asserting all the pods spec.
The full pipeline ready assert will be after the upgrade as before

blumamir (approved):

PR #1990, opened 2024-12-15
feat: add configurable cpu request and limit to node collector settings

blumamir:


helm/odigos/values.yaml
```diff
@@ -77,6 +77,17 @@ collectorNode:
...
+  # the CPU request for the node collector daemonset.
+  # it will be embedded in the daemonset as a resource request
+  # of the form "cpu: <value>m".
+  # default value is 250m
+  requestCPUm: 250
+  # the CPU limit for the node collector daemonset.
+  # it will be embedded in the daemonset as a resource limit
+  # of the form "cpu: <value>m".
+  # default value is 500m
+  limitCPUm: 555
```
RonFed, line 89:
```suggestion
  limitCPUm: 550
```

tamirdavid1 (approved):

scheduler/controllers/nodecollectorsgroup/common.go
```diff
@@ -78,12 +83,23 @@ func getMemorySettings(odigosConfig common.OdigosConfiguration) odigosv1.Collect
...
 	}
 
+	cpuRequestm := defaultRequestCPUm
+	if nodeCollectorConfig != nil && nodeCollectorConfig.RequestCPUm > 0 {
+		cpuRequestm = nodeCollectorConfig.RequestCPUm
+	}
+	cpuLimitm := defaultLimitCPUm
+	if nodeCollectorConfig != nil && nodeCollectorConfig.LimitCPUm > 0 {
+		cpuLimitm = nodeCollectorConfig.LimitCPUm
+	}
```
RonFed, line 93: If a user provides request > limit do we want to check for this here or just let k8s return the error once we deploy?

blumamir: We are not handling it at the moment, can be addressed as a separate PR later on

PR #2012, opened 2024-12-16
feat: configurable mock destination for pipeline testing

blumamir:
This new exporter mocks a destination, and allows one to simulate various back-pressure scenarios, like long time to get a response for the export, and simulating some export requests being rejected.

It is useful for pipeline development and testing


RonFed (approved):
Very cool

collector/exporters/mockdestinationexporter/exporter.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+type MockDestinationExporter struct {
+	config *Config
+	logger *zap.Logger
+}
+
+func NewMockDestinationExporter(config *Config,
+	params exporter.Settings) (*MockDestinationExporter, error) {
+
```
tamirdavid1, line 25: nit: maybe worth to add small config validation ?
ResponseDuration > 0 
1 > RejectFraction > 0

tamirdavid1 (approved):

collector/exporters/mockdestinationexporter/exporter.go
blumamir: added in config.go under `Validate` function

collector/exporters/mockdestinationexporter/exporter.go
```diff
@@ -0,0 +1,62 @@
...
+	return e.mockExport(ctx)
+}
+
+func (e *MockDestinationExporter) ConsumeLogs(ctx context.Context, logs plog.Logs) error {
+	return e.mockExport(ctx)
+}
+
+func (e *MockDestinationExporter) mockExport(context.Context) error {
+	// not taking care of ctx cancel and shutdown as this is a dummy exporter and not used in production
+	<-time.After(e.config.ResponseDuration)
```
edeNFed, line 57: nit: I think time.Sleep is better as time.After create goroutine every time it is called so this has the potential to blow up CPU according to the number of parallel invocation to this function

PR #2015, opened 2024-12-16
cOdiglet to use criwrapper to create ic

tamirdavid1:


api/config/crd/bases/odigos.io_instrumentationconfigs.yaml
```diff
@@ -423,21 +423,15 @@ spec:
...
-                format: int64
-                type: integer
               runtimeDetailsByContainer:
                 description: Capture Runtime Details for the workloads that this CR
                   applies to.
                 items:
                   properties:
                     containerName:
                       type: string
+                    criErrorMessage:
```
edeNFed, line 433: isnt status a better place for error messages instead of spec?

k8sutils/pkg/cri/criwrapper.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+// Connect attempts to establish a connection to a CRI runtime.
+func (rc *CriClient) Connect() error {
+	var err error
+
+	for _, endpoint := range defaultRuntimeEndpoints {
+		rc.Logger.Info("Attempting to connect to CRI runtime", "endpoint", endpoint)
+
+		rc.conn, err = grpc.NewClient(
```
edeNFed, line 36: Are you sure this is actually trying to connect and not just creating a client?
Few thoughts:
- As only one of the default Endpoints will respond - maybe worth putting a timeout when opening a connection as some of them will just time out so we can bail quickly
- I think the CRI implementation is part of the container id in Kubernetes maybe we can use that (for example getting the container id of the current Odiglet Pod) and connector only to the right one instead of trying all the options

k8sutils/pkg/cri/criwrapper.go
```diff
@@ -0,0 +1,142 @@
...
+
+// GetContainerInfo retrieves the "info" field of the specified container.
+func (rc *CriClient) GetContainerInfo(ctx context.Context, containerID string) (map[string]string, error) {
+	containerID = extractContainerID(containerID)
+	if rc.client == nil {
+		return nil, fmt.Errorf("runtime client is not connected")
+	}
+
+	// Call the ContainerStatus API
+	response, err := rc.client.ContainerStatus(ctx, &criapi.ContainerStatusRequest{
```
edeNFed, line 62: Maybe we want to put a timeout on that? In case the remote CRI does not respond in time / unavailable?

odiglet/pkg/kube/runtime_details/inspection.go
```diff
@@ -176,7 +183,91 @@ func runtimeInspection(pods []corev1.Pod, ignoredContainers []string) ([]odigosv
...
+// and updates the RuntimeDetailsByContainer accordingly.
+func updateRuntimeDetailsWithDockerFileEnvs(ctx context.Context, criClient criwrapper.CriClient, pod corev1.Pod, container corev1.Container,
+	programLanguageDetails common.ProgramLanguageDetails, resultsMap *map[string]odigosv1.RuntimeDetailsByContainer) {
+	// Retrieve environment variable names for the specified language
+	envVarNames, exists := envOverwrite.EnvVarsForLanguage[programLanguageDetails.Language]
+	if !exists {
+		return
+	}
+
+	// Verify if environment variables already exist in the container manifest.
```
edeNFed, line 196: Why not always collect the relevant env vars from Dokcerfile?
I don't have a good scenario in my head, but my feeling is that the more stateless we can make everything the simpler it will be

odiglet/pkg/kube/runtime_details/inspection.go
```diff
@@ -176,7 +183,91 @@ func runtimeInspection(pods []corev1.Pod, ignoredContainers []string) ([]odigosv
...
 func persistRuntimeDetailsToInstrumentationConfig(ctx context.Context, kubeclient client.Client, instrumentationConfig *odigosv1.InstrumentationConfig, newStatus odigosv1.InstrumentationConfigStatus) error {
+	// This come to make sure we're updating instrumentationConfig only once (at the first time)
+	currentConfig := &odigosv1.InstrumentationConfig{}
+	err := kubeclient.Get(ctx, client.ObjectKeyFromObject(instrumentationConfig), currentConfig)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve current InstrumentationConfig: %w", err)
+	}
+
+	// Check if status is already set
+	if len(currentConfig.Status.RuntimeDetailsByContainer) != 0 {
```
edeNFed, line 267: Why not update always? This means only the first Odiglet that runs this wins, so we will not update the value if the user changed the docker image?

api/odigos/v1alpha1/instrumentedapplication_types.go
```diff
@@ -45,12 +45,14 @@ type OtherAgent struct {
...
-	OtherAgent     *OtherAgent                `json:"otherAgent,omitempty"`
-	LibCType       *common.LibCType           `json:"libCType,omitempty"`
+	ContainerName         string                     `json:"containerName"`
+	Language              common.ProgrammingLanguage `json:"language"`
+	RuntimeVersion        string                     `json:"runtimeVersion,omitempty"`
+	EnvVars               []EnvVar                   `json:"envVars,omitempty"`
+	OtherAgent            *OtherAgent                `json:"otherAgent,omitempty"`
+	LibCType              *common.LibCType           `json:"libCType,omitempty"`
+	CriErrorMessage       *string                    `json:"criErrorMessage,omitempty"`
+	EnvVarsFromDockerFile []EnvVar                   `json:"envVarsFromDockerFile,omitempty"`
```
RonFed, line 55: Maybe we should rename this to something like
`EnvFromContainerRuntime` ?
Since containers can be created using other ways from Dockerfile

k8sutils/pkg/cri/criwrapper.go
tamirdavid1: 1. I Warp it up as a Connect() -  The function’s primary purpose is to connect to the CRI runtime, which it achieves by establishing and validating a gRPC connection.

I change the logic so it will try to find the unix domain socket in the FS so we wont need to iterate over them and try to connect to each one of them

api/odigos/v1alpha1/instrumentedapplication_types.go
tamirdavid1: Thanks for sharing Ron, will update

api/odigos/v1alpha1/instrumentedapplication_types.go
```diff
@@ -43,14 +43,25 @@ type OtherAgent struct {
...
-	LibCType       *common.LibCType           `json:"libCType,omitempty"`
+	ContainerName           string                     `json:"containerName"`
+	Language                common.ProgrammingLanguage `json:"language"`
+	RuntimeVersion          string                     `json:"runtimeVersion,omitempty"`
+	EnvVars                 []EnvVar                   `json:"envVars,omitempty"`
+	OtherAgent              *OtherAgent                `json:"otherAgent,omitempty"`
+	LibCType                *common.LibCType           `json:"libCType,omitempty"`
+	CriErrorMessage         *string                    `json:"criErrorMessage,omitempty"`
+	EnvFromContainerRuntime []EnvVar                   `json:"envFromContainerRuntime,omitempty"`
+	RuntimeUpdateState      *ProcessingState           `json:"runtimeUpdateState,omitempty"` // Tracks whether the new runtime detection process has been executed. If empty, the process has not been executed.
```
blumamir, line 64: Add a comment that this property is temporary and only used for the migration and showed be removed after some time

odiglet/pkg/kube/runtime_details/inspection.go
```diff
@@ -72,7 +74,7 @@ func inspectRuntimesOfRunningPods(ctx context.Context, logger *logr.Logger, labe
 	return nil
 }
 
-func runtimeInspection(pods []corev1.Pod, ignoredContainers []string) ([]odigosv1.RuntimeDetailsByContainer, error) {
+func runtimeInspection(ctx context.Context, pods []corev1.Pod, ignoredContainers []string, criClient *criwrapper.CriClient) ([]odigosv1.RuntimeDetailsByContainer, error) {
```
blumamir, line 77: `ignoredContainers` - make sure it still works after the refactor

k8sutils/pkg/cri/criwrapper.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create gRPC connection: %v", err)
+	}
+
+	// Create a new RuntimeService client
+	rc.client = criapi.NewRuntimeServiceClient(rc.conn)
+
+	// Validate the connection by invoking a lightweight method
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*2)
+	defer cancel()
```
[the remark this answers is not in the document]
tamirdavid1: According to what i've read:
You need `defer cancel()` to properly release resources associated with the context. It's a best practice whenever you create a context with WithTimeout or WithCancel, even if the timeout expires or the function exits early. This habit helps prevent subtle resource leaks and keeps your program efficient.

In addition to it, we're doing that in everyplace we're creating context with timeout (config_provider, kubelet communication and more)

k8sutils/pkg/cri/criwrapper.go
```diff
@@ -0,0 +1,167 @@
...
+func (rc *CriClient) GetContainerInfo(ctx context.Context, containerID string) (map[string]string, error) {
+	containerID = extractContainerID(containerID)
+	if rc.client == nil {
+		return nil, fmt.Errorf("runtime client is not connected")
+	}
+
+	// Call the ContainerStatus API
+	response, err := rc.client.ContainerStatus(ctx, &criapi.ContainerStatusRequest{
+		ContainerId: containerID,
+		Verbose:     true,
```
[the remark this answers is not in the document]
tamirdavid1: No, without this the cri output wont include response.GetInfo()

k8sutils/pkg/cri/criwrapper.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	return result, nil
+}
+
+// Close closes the gRPC connection.
+func (rc *CriClient) Close() {
+	if rc.conn != nil {
+		rc.Logger.V(0).Info("gRPC connection is closed")
+		rc.conn.Close()
```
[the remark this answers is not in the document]
tamirdavid1: Thanks!

odiglet/pkg/kube/runtime_details/inspection.go
tamirdavid1: It works :)

blumamir (approved):

odiglet/pkg/kube/runtime_details/inspection.go
tamirdavid1: correct, it wont work with cases where user changed the value exists in Dockerfile.
we design this solution with this gap in mind

k8sutils/pkg/cri/criwrapper.go
tamirdavid1: done

k8sutils/pkg/cri/criwrapper.go
```diff
@@ -0,0 +1,180 @@
...
+	return ""
+}
+
+// Connect attempts to establish a connection to a CRI runtime.
+func (rc *CriClient) Connect(ctx context.Context) error {
+	var err error
+
+	endpoint := detectRuntimeSocket()
+	if endpoint == "" {
+		return fmt.Errorf("unable to detect CRI runtime endpoint")
```
RonFed, line 48: nit: (as the error doesn't include formatting)
```suggestion
		return errors.New("unable to detect CRI runtime endpoint")
```

We could also export this by declaring above a package level error
```Go
var ErrDetectingCRI = errors.New("unable to detect CRI runtime endpoint")
```
This will allow the caller of this function hande this specific error

k8sutils/pkg/cri/criwrapper.go
```diff
@@ -0,0 +1,180 @@
...
+	// Create a new RuntimeService client
+	rc.client = criapi.NewRuntimeServiceClient(rc.conn)
+
+	// Validate the connection by invoking a lightweight method
+	ctx, cancel := context.WithTimeout(ctx, time.Second*2)
+	defer cancel()
+
+	_, err = rc.client.Version(ctx, &criapi.VersionRequest{})
+	if err != nil {
+		return fmt.Errorf("Failed to validate CRI runtime connection")
```
RonFed, line 71: Same comment as above, I think we can declare this as `ErrFailedToConnectToCRI` and the caller can handle this accordingly

RonFed (approved):

PR #2016, opened 2024-12-16
Track un-active instrumentation for better reporting

RonFed:
This PR changes when the instrumentation manager tracks an instrumentation.
Once an instrumentation is initialized, we should start tracking it even if it fails to load.
The reason is, that we need to call the reporter once the process exits (so it can report/clean it) - however, if we only track the PID once the instrumentation is loaded we won't be able to invoke the reporter properly once a failed-to-load instrumentation needs to be cleaned.

Each PID we track is marked whether the instrumentation is active (loaded successfully) or not - by the inst field being nil or not.
For un-active PIDs we won't apply config updates.

In addition, in case we get an exec event on a non-active PID, we try to instrument it again. This is helpful in cases of chain-loading where the first executable is written in a language we can't instrument, while the second is valid for instrumentation.


instrumentation/manager.go
```diff
@@ -266,10 +273,15 @@ func (m *manager[ProcessGroup, ConfigGroup]) handleProcessExecEvent(ctx context.
...
-	err = inst.Load(ctx)
-	// call the reporter regardless of the load result - as we want to report the load status
-	reporterErr := m.handler.Reporter.OnLoad(ctx, e.PID, err, pg)
-	if err != nil {
+	loadErr := inst.Load(ctx)
+
+	// we need to track the instrumentation even if the load failed.
+	// consider a reporter which writes a persistent record for a failed/successful load
+	// we need to notify the reporter once that PID exists to clean up the resources
+	m.startTrackInstrumentation(e.PID, inst, pg, configGroup, loadErr == nil)
```
blumamir, line 281: if we had an error with the `Load`, it means that at this point the `inst` is not longer relevant. We will not need to call `Close` or apply config to it since it's not loaded. Thus, I recommend to consider using `nil` instead of `inst` in this case so that it doesn't get called accidentally and cause hard to find bug

instrumentation/manager.go
```diff
@@ -27,6 +27,11 @@ type instrumentationDetails[processGroup ProcessGroup, configGroup ConfigGroup]
 	inst Instrumentation
 	pg   processGroup
 	cg   configGroup
+
+	// active is used to track if the instrumentation is loaded successfully or not.
+	// we want to track the instrumentation even if it failed to load, to be able to report the error
+	// and clean up the instrumentation resources and the reporter resources once the process exits.
+	active bool
```
blumamir, line 34: consider:

```suggestion
	loaded bool
```

to maintain a consistent terminology with existing terms.

blumamir (approved):

PR #2020, opened 2024-12-18
Common Odiglet creation and Run functions

RonFed:
This PR makes the Odiglet creation and Run function shared so we can re-use them.
In addition, the Run function is updated to use [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup) to manage the different components.

### DIFF
--- a/odiglet/cmd/main.go
+++ b/odiglet/cmd/main.go
@@ -7,8 +7,8 @@ import (
 	"github.com/odigos-io/odigos/odiglet/pkg/ebpf/sdks"
 	"github.com/odigos-io/odigos/odiglet/pkg/instrumentation/fs"
 
-	commonInstrumentation "github.com/odigos-io/odigos/instrumentation"
 	"github.com/odigos-io/odigos/common"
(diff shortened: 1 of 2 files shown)


odiglet/cmd/main.go
```diff
@@ -182,24 +48,20 @@ func main() {
...
 	log.Logger.V(0).Info("odiglet exiting")
 }
 
-func runDeviceManager(clientset *kubernetes.Clientset) {
-	log.Logger.V(0).Info("Starting device manager")
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-
-	otelSdkLsf := map[common.ProgrammingLanguage]map[common.OtelSdk]instrumentation.LangSpecificFunc{
+func otelSdkLsf() instrumentation.OtelSdksLsf{
```
tamirdavid1, line 63: nit: think need better name then Lsf

tamirdavid1 (approved):

blumamir (approved):
💯

PR #2024, opened 2024-12-18
chore: narrow down the rbac permissions for schedualer

blumamir:
In this PR:

- move destinations and collectorsgroup rbac permission to be role (on odigos ns) instead of clusterrole. These live anyway only in our ns, but it looks better in rbac reviews.
- removed unused create/delete/update/patch for the destinations, as we only read them.
- currently the leader election role brings in configmaps permissions, but I also added the permissions we need for the reconcilers so not to mix unrelated things. we will need to also understand and address the leader election role later on.
- removed the finalizers permissions, as we don't use finalizers and it's unused.
- created some consts in cli files for better structure to the references
- synced helm files with the changes
- tested both helm and cli locally to make sure it works

InstrumentationConfig is left as a clusterrole, since they belong to the various namespaces where the sources reside.

### DIFF
--- a/cli/cmd/resources/scheduler.go
+++ b/cli/cmd/resources/scheduler.go
@@ -7,6 +7,7 @@ import (
 	"github.com/odigos-io/odigos/cli/pkg/containers"
 	"github.com/odigos-io/odigos/cli/pkg/kube"
 	"github.com/odigos-io/odigos/common"
+	"github.com/odigos-io/odigos/common/consts"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


cli/cmd/resources/scheduler.go
```diff
@@ -31,13 +37,13 @@ func NewSchedulerServiceAccount(ns string) *corev1.ServiceAccount {
...
-			Name:      "odigos-scheduler",
+			Name:      SchedulerServiceAccountName,
 			Namespace: ns,
 		},
 	}
 }
 
-func NewSchedulerRoleBinding(ns string) *rbacv1.RoleBinding {
+func NewSchedulerLeaderElectionRoleBinding(ns string) *rbacv1.RoleBinding {
 	return &rbacv1.RoleBinding{
```
damemi, line 47: could we just embed the role and rolebinding yamls and marshal them to the API types? Then we can keep them in one spot that's a little more readable

blumamir: Yes, they are currently duplicated between helm and cli. we sync them manually but need to find a better way to have them just once

RonFed (approved):

cli/cmd/resources/scheduler.go
damemi: Opened https://github.com/odigos-io/odigos/issues/2047 to track (gh issue since it's not high priority)

PR #2056, opened 2024-12-23
Fix .NET instrumentation for .NET 6.0

edeNFed:


tests/e2e/workload-lifecycle/03-wait-for-trace.yaml
RonFed, line 1: needs also to update `01-wait-for-traces` with the .NET apps

edeNFed: fixed

RonFed (approved):

PR #2065, opened 2024-12-24
Enable webhook for greatwall inject otel service by webhook2

tamirdavid1:


edeNFed (commented):
looks good, only have one comment:
- Webhook should be really quick as it is called for every pod in the cluster.
Maybe we should calculate the app service name (which involves reading from API server) to only after we now this is a relevant pod (.NET or Java community)
what do you think?

blumamir (approved):

tamirdavid1:
> looks good, only have one comment:
> 
> * Webhook should be really quick as it is called for every pod in the cluster.
>   Maybe we should calculate the app service name (which involves reading from API server) to only after we now this is a relevant pod (.NET or Java community)
>   what do you think?

Thanks for the comment @edeNFed,
I change the logic please double verify the change, i will also do it on OSS

edeNFed (approved):

PR #2066, opened 2024-12-24
feat: add cli command to update onprem token

AvihuHenya:
Currently, users cannot update their on-prem token and are limited to reinstalling Odigos in their cluster. This PR introduces the ability to update the odigos-pro secret's data using the odigos cli


cli/cmd/update.go
```diff
@@ -0,0 +1,75 @@
...
+
+	secret.Data["odigos-onprem-token"] = []byte(tokenValue)
+
+	// Apply the updated secret
+	_, err = client.CoreV1().Secrets(namespace).Update(ctx, secret, metav1.UpdateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to update secret %s in namespace %s: %w", consts.OdigosProSecretName, namespace, err)
+	}
+
+	fmt.Printf("Updated secret %s in namespace %s with new %s token\n", consts.OdigosProSecretName, namespace, tokenType)
```
edeNFed, line 66: I think we should add to the message something like:
The new token will take affect only after Odiglets restart. To trigger odiglets restart run:
`kubectl rollout restart daemonset odiglet -n <odigos-namespace>`

edeNFed (approved):

cli/cmd/update.go
```diff
@@ -0,0 +1,75 @@
...
+var updateTokenCmd = &cobra.Command{
+	Use:   "update-token",
+	Short: "Update a client's token for Odigos",
+	Long: `Update the token used for authenticating with Odigos.
+This command is useful for updating your on-prem or cloud token for an existing installation.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		ctx := cmd.Context()
+		client := cmdcontext.KubeClientFromContextOrExit(ctx)
+		ns, err := resources.GetOdigosNamespace(client, ctx)
+		if err != nil {
```
blumamir, line 26: In other cli command we differentiate between:
- odigos is not installed in the cluster
- some general error (permissions, network etc)

I think it makes sense to add it here to give users a usable message

cli/cmd/update.go
```diff
@@ -0,0 +1,75 @@
...
+			fmt.Errorf("No Odigos installation found in cluster to upgrade")
```
blumamir, line 27:
```suggestion
			fmt.Errorf("No Odigos installation found in cluster")
```

cli/cmd/update.go
```diff
@@ -0,0 +1,76 @@
...
+
+func updateOdigosToken(ctx context.Context, client *kube.Client, namespace string, tokenType, tokenValue string) error {
+
+	// Retrieve the existing secret
+	secret, err := client.CoreV1().Secrets(namespace).Get(ctx, consts.OdigosProSecretName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get secret %s in namespace %s: %w", consts.OdigosProSecretName, namespace, err)
+	}
+
+	secret.Data["odigos-onprem-token"] = []byte(tokenValue)
```
blumamir, line 58: There is a const for that as well [here](https://github.com/odigos-io/odigos/blob/94e1e05a829344e69e70bd7b395d3355bf84a803/cli/cmd/resources/odigospro/consts.go#L8). consider moving it to `k8sutils` and reusing it from there

cli/cmd/update.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Apply the updated secret
+	_, err = client.CoreV1().Secrets(namespace).Update(ctx, secret, metav1.UpdateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to update secret %s in namespace %s: %w", consts.OdigosProSecretName, namespace, err)
+	}
+
+	fmt.Printf("Updated secret %s in namespace %s with new %s token\n", consts.OdigosProSecretName, namespace, tokenType)
```
blumamir, line 66: Will this message be shown in the terminal?

I think those details are our internal implementation, and we can write something more user friendly like: "onprem token updated successfully"

cli/cmd/update.go
```diff
@@ -0,0 +1,76 @@
...
+		ns, err := resources.GetOdigosNamespace(client, ctx)
+		if err != nil {
+			fmt.Errorf("No Odigos installation found in cluster to upgrade")
+			os.Exit(1)
+		}
+
+		// Retrieve token
+		onPremToken := cmd.Flag("onprem-token").Value.String()
+
+		if onPremToken == "" {
```
blumamir, line 34: since we have `MarkFlagRequired`, do we need to check here?

cli/cmd/update.go
```diff
@@ -0,0 +1,76 @@
...
+			fmt.Errorf("\033[31mERROR\033[0m --onprem-token is required")
+			os.Exit(1)
+		}
+
+		var tokenType, tokenValue string = "onprem", onPremToken
```
blumamir, line 39: We prefer to initialize variables like this:

```go
tokenType := "foo"
tokenValue := "bar"
```

cli/cmd/update.go
RonFed: Do we want users to rollout Odiglet or we can do it here after the token is updated?

cli/cmd/update.go
```diff
@@ -0,0 +1,75 @@
...
+	cmdcontext "github.com/odigos-io/odigos/cli/pkg/cmd_context"
+	"github.com/odigos-io/odigos/cli/pkg/kube"
+	"github.com/odigos-io/odigos/k8sutils/pkg/consts"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// updateTokenCmd represents the update-token command
+var updateTokenCmd = &cobra.Command{
+	Use:   "update-token",
```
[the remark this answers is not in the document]
AvihuHenya: I agree that update-token is not the best naming choice. Another advantage of this change is the opportunity to centralize all potential operations that users can perform within the on-prem environment.

I will proceed with the change.

cli/cmd/update.go
AvihuHenya: I agree. fixed.

cli/cmd/update.go
AvihuHenya: fixed

cli/cmd/update.go
AvihuHenya: fixed. i will ask someone about the rest of the code changes.

cli/cmd/pro.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/odigos-io/odigos/cli/pkg/kube"
+	"github.com/odigos-io/odigos/k8sutils/pkg/consts"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+
+var proCmd = &cobra.Command{
+	Use:   "pro",
+	Short: "operations for pro clients of Odigos",
```
blumamir, line 19: WDYT about?
```suggestion
	Short: "manage odigos pro",
```

cli/cmd/pro.go
```diff
@@ -0,0 +1,79 @@
...
+		fmt.Println("kubectl rollout restart daemonset odiglet -n", ns)
+	},
+}
+
+func updateOdigosToken(ctx context.Context, client *kube.Client, namespace string, tokenType, tokenValue string) error {
+
+	// Retrieve the existing secret
+	secret, err := client.CoreV1().Secrets(namespace).Get(ctx, consts.OdigosProSecretName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("Tokens are not available in the open-source version of Odigos. Please use the on-premises version.")
```
blumamir, line 60: This will get printed for any `err`. both if the secret is not found and for network / permissions or any other general error.

You can differentiate between them with 

```go
apierrors.IsNotFound(err)
```

cli/cmd/pro.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	secret.Data[consts.OdigosOnpremTokenSecretKey] = []byte(tokenValue)
+
+	// Apply the updated secret
+	_, err = client.CoreV1().Secrets(namespace).Update(ctx, secret, metav1.UpdateOptions{})
+	if err != nil {
+		return fmt.Errorf("Tokens are not available in the open-source version of Odigos. Please use the on-premises version.")
```
blumamir, line 67: This message will be printed after we already know there is a secret, but failed to update it.

I think we should log the err itself here

blumamir (approved):
Great job!

Left few final nits and then we can merge

cli/cmd/pro.go
```diff
@@ -0,0 +1,82 @@
...
+	},
+}
+
+func updateOdigosToken(ctx context.Context, client *kube.Client, namespace string, tokenType, tokenValue string) error {
+	secret, err := client.CoreV1().Secrets(namespace).Get(ctx, consts.OdigosProSecretName, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsForbidden(err) {
+			return fmt.Errorf("Insufficient permissions to retrieve the token")
+		}
+		return fmt.Errorf("Tokens are not available in the open-source version of Odigos. Please use the on-premises version.")
```
blumamir, line 61: This condition assumes that any error which is not `Forbidden` is `NotExists`. please use `apierrors.IsNotFound(err)` which indicates this case explicitly, and print all other errors as is

cli/cmd/pro.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	secret.Data[consts.OdigosOnpremTokenSecretKey] = []byte(tokenValue)
+
+	_, err = client.CoreV1().Secrets(namespace).Update(ctx, secret, metav1.UpdateOptions{})
+	if err != nil {
+		if apierrors.IsForbidden(err) {
+			return fmt.Errorf("Insufficient permissions to update the token")
```
blumamir, line 68: If this happens, should we also print what the error is?

cli/cmd/pro.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+		return fmt.Errorf("%v", err)
```
blumamir, line 70:
```suggestion
		return err
```

.vscode/launch.json
```diff
@@ -64,6 +65,27 @@
...
       "buildFlags": "-tags=embed_manifests"
+    }, 
+    {
+      "name": "cli install",
+      "type": "go",
+      "request": "launch",
+      "mode": "debug",
+      "program": "${workspaceFolder}/cli",
+      "cwd": "${workspaceFolder}/cli",
+      "args": ["install", "--version", "<version_number>"], // Specify the desired version as the third argument in the command
```
blumamir, line 76: this is not valid `json`. Just making sure it was tested to work

blumamir: It looks like you can use env for that:
https://code.visualstudio.com/docs/editor/variables-reference#_environment-variables

.vscode/launch.json
```diff
@@ -64,7 +65,28 @@
...
+      "request": "launch",
+      "mode": "debug",
+      "program": "${workspaceFolder}/cli",
+      "cwd": "${workspaceFolder}/cli",
+      "args": ["pro", "--onprem-token", "${env:ONPREM_TOKEN}"],
+      "buildFlags": "-tags=embed_manifests"
     }
-  ]
+  ],
+  "envFile": "${workspaceFolder}/.env"
```
blumamir, line 90: The `.env` file is in the `cli` directory, does it work?

cli/cmd/pro.go
```diff
@@ -0,0 +1,86 @@
...
+	Use:   "pro",
+	Short: "manage odigos pro",
+	Long:  `The pro command provides various operations and functionalities specifically designed for enterprise users. Use this command to access advanced features and manage your pro account.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		ctx := cmd.Context()
+		var onPremToken string
+		if err := godotenv.Load(); err != nil {
+			fmt.Println("\033[31mERROR\033[0m Failed to load .env file")
+		}
+		if os.Getenv("DEBUG") == "true" {
```
blumamir, line 28: I thought the vsc debugger handles the `.env` file itself.

blumamir (changes requested):

blumamir (approved):

RonFed (approved):

cli/cmd/pro.go
```diff
@@ -0,0 +1,93 @@
...
+	},
+}
+
+func updateOdigosToken(ctx context.Context, client *kube.Client, namespace string, onPremToken string) error {
+	secret, err := client.CoreV1().Secrets(namespace).Get(ctx, consts.OdigosProSecretName, metav1.GetOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return err
+		}
+		return fmt.Errorf("Tokens are not available in the open-source version of Odigos. Please use the on-premises version.")
```
blumamir, line 54: The conditions are reversed. If not found it means that it's not odigos pro, and all other errors should be returned as is

cli/cmd/pro.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	secret.Data[consts.OdigosOnpremTokenSecretKey] = []byte(onPremToken)
+
+	_, err = client.CoreV1().Secrets(namespace).Update(ctx, secret, metav1.UpdateOptions{})
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return err
+		}
+		return err
```
blumamir, line 63: if we always return the `err` here, we don't need the if

cli/cmd/pro.go
```diff
@@ -0,0 +1,93 @@
...
+		return fmt.Errorf("failed to get DaemonSet odiglet in namespace %s: %v", namespace, err)
+	}
+	
+	// Modify the DaemonSet spec.template to trigger a rollout
+	if daemonSet.Spec.Template.Annotations == nil {
+		daemonSet.Spec.Template.Annotations = make(map[string]string)
+	}
+	daemonSet.Spec.Template.Annotations["rollout-trigger"] = time.Now().Format(time.RFC3339)
+
+	_, err = client.AppsV1().DaemonSets(namespace).Update(context.TODO(), daemonSet, metav1.UpdateOptions{})
```
blumamir, line 77:
```suggestion
	_, err = client.AppsV1().DaemonSets(namespace).Update(ctx, daemonSet, metav1.UpdateOptions{})
```

cli/cmd/pro.go
```diff
@@ -0,0 +1,93 @@
...
+	daemonSet, err := client.AppsV1().DaemonSets(namespace).Get(ctx, "odiglet", metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get DaemonSet odiglet in namespace %s: %v", namespace, err)
+	}
+	
+	// Modify the DaemonSet spec.template to trigger a rollout
+	if daemonSet.Spec.Template.Annotations == nil {
+		daemonSet.Spec.Template.Annotations = make(map[string]string)
+	}
+	daemonSet.Spec.Template.Annotations["rollout-trigger"] = time.Now().Format(time.RFC3339)
```
blumamir, line 75: consider adding the label name as const [here](https://github.com/odigos-io/odigos/blob/main/common/consts/consts.go)

cli/cmd/pro.go
```diff
@@ -0,0 +1,88 @@
...
+		onPremToken := cmd.Flag("onprem-token").Value.String()
+		err = updateOdigosToken(ctx, client, ns, onPremToken)
+		if err != nil {
+			fmt.Println("\033[31mERROR\033[0m Failed to update token:")
+			fmt.Println(err)
+			os.Exit(1)
+		}
+
+		fmt.Println()
+		fmt.Printf("\u001B[32mSUCCESS:\u001B[0m Token updated successfully in namespace %s\n", ns)
```
blumamir, line 43:
```suggestion
		fmt.Println("\u001B[32mSUCCESS:\u001B[0m Token updated successfully")
```

blumamir (approved):

PR #2081, opened 2024-12-26
Test other agent

tamirdavid1:


tests/e2e/workload-lifecycle/services/python-http-server/Dockerfile.python-other-agent
```diff
@@ -0,0 +1,22 @@
...
+RUN apt-get update && apt-get install sqlite3 -y
+
+RUN pip install --no-cache-dir -r requirements.txt
+
+# Supress health check endpoint from tracing
+ENV OTEL_PYTHON_DJANGO_EXCLUDED_URLS=health/
+
+COPY entrypoint.sh /entrypoint.sh
+
+ENV NEW_RELIC_CONFIG_FILE=/app/newrelic.ini
```
RonFed, line 18: is the `/newrelic.ini` file present in the repo? couldn't find it

RonFed (commented):
Do we need to update the count in `wait-for-traces` files?

tamirdavid1:
> Do we need to update the count in `wait-for-traces` files?

We detect is using env var so don't really need it in the test

RonFed (approved):

PR #2082, opened 2024-12-26
Migration to new runtimedetails approach

tamirdavid1:


instrumentor/runtimemigration/runtimemigration.go
```diff
@@ -0,0 +1,314 @@
...
+
+		IcName := item.GetName()
+		parts := strings.Split(IcName, "-")
+		if len(parts) < 2 {
+			continue
+		}
+
+		IcNamespace := item.GetNamespace()
+		workloadType := parts[0] // deployment/statefulset/daemonset
+		workloadName := strings.Join(parts[1:], "-")
```
RonFed, line 56: We have `ExtractWorkloadInfoFromRuntimeObjectName` which can work here to get the workload name and type

instrumentor/runtimemigration/runtimemigration.go
```diff
@@ -0,0 +1,314 @@
...
+		if _, exists := workloadNamespaces[workloadType]; !exists {
+			workloadNamespaces[workloadType] = make(map[string]map[string]*v1alpha1.InstrumentationConfig)
+		}
```
RonFed, line 59: nit: this check can be removed if we initialize the map with entries for `deployment` `statefulset` and `daemonset`

instrumentor/runtimemigration/runtimemigration.go
```diff
@@ -0,0 +1,314 @@
...
+			}
+		default:
+			fmt.Printf("Unknown workload type: %s\n", workloadType)
+		}
+	}
+
+	return nil
+}
+
+func fetchAndProcessDeployments(ctx context.Context, clientset client.Client, namespaces map[string]map[string]*v1alpha1.InstrumentationConfig) error {
```
RonFed, line 92: nit: this is the controller-runtime client
```suggestion
func fetchAndProcessDeployments(ctx context.Context, client client.Client, namespaces map[string]map[string]*v1alpha1.InstrumentationConfig) error {
```

instrumentor/runtimemigration/runtimemigration.go
```diff
@@ -0,0 +1,314 @@
...
+		err := clientset.List(ctx, &statefulSets, &client.ListOptions{Namespace: namespace})
+		if err != nil {
+			return fmt.Errorf("failed to list statefulsets in namespace %s: %v", namespace, err)
+		}
+
+		for _, sts := range statefulSets.Items {
+			// Checking if the statefulset is in the list of statefulsets that need to be processed
+			if contains(workloadNames, sts.Name) {
+
+				originalWorkloadEnvVar, _ := envoverwrite.NewOrigWorkloadEnvValues(sts.Annotations)
```
RonFed, line 145: should we check for an error and continue if one happens? otherwise we risk panic in `*originalWorkloadEnvVar`

instrumentor/runtimemigration/runtimemigration.go
```diff
@@ -0,0 +1,314 @@
...
+				}
+			}
+		}
+	}
+	return nil
+}
+func handleContainerRuntimeDetailsUpdate(
+	containerObject v1.Container,
+	originalWorkloadEnvVar envoverwrite.OrigWorkloadEnvValues,
+	runtimeDetailsByContainer *[]v1alpha1.RuntimeDetailsByContainer,
```
RonFed, line 222: I don't thinks we should pass a pointer to the slice here since we are not modifying the slice itself but its elements
```suggestion
	runtimeDetailsByContainer []v1alpha1.RuntimeDetailsByContainer,
```

instrumentor/runtimemigration/runtimemigration.go
```diff
@@ -0,0 +1,314 @@
...
+
+				// Update runtimeDetailsByContainer in workloadInstrumentationConfigReference
+				workloadInstrumentationConfigReference.Status.RuntimeDetailsByContainer = runtimeDetailsByContainer
+
+				// Update the InstrumentationConfig status
+				err := clientset.Status().Update(
+					ctx,
+					workloadInstrumentationConfigReference,
+				)
+				if err != nil {
```
RonFed, line 168: What would happen if another component is editing this instrumentation config while the migration is running?
I assume we can get a conflict error from this Update.
This is more likely in this case in large clusters since we are listing all the instrumentation configs and this migration can take some time, which incrases the probability of handling a stale instrumentation config being updated.

instrumentor/runtimemigration/runtimemigration.go
```diff
@@ -0,0 +1,314 @@
...
+		for envKey, envValue := range annotationEnvVarsForContainer {
+			isEnvVarAlreadyExists := isEnvVarPresent(containerRuntimeDetails.EnvFromContainerRuntime, envKey)
+			// The runtimeDetails might already contain the environment variable if Odiglet started before the migration was executed and modified the env vars.
+			// In this case, we want to overwrite the value set by Odiglet with a new one.
+			if isEnvVarAlreadyExists {
+				containerRuntimeDetails.EnvFromContainerRuntime = removeEnvVar(containerRuntimeDetails.EnvFromContainerRuntime, envKey)
+			}
+
+			// Handle runtime-originated environment variables
+			if envValue == nil {
```
RonFed, line 243: I might be missing something.
If an env var originated from a Dockerfile, it won't be part of the annotation map, right?
If that is the case, the key will not be part of the map.
This code assumes that env vars that are not from the k8s manifest will appear in the annotation with nil value - is this the case?

tamirdavid1:
```
odigos.io/manifest-env-original-val: '{"simple-app":{"PYTHONPATH":null}}'
odigos.io/manifest-env-original-val: '{"coupon":{"NODE_OPTIONS":null}}' 
```

instrumentor/runtimemigration/runtimemigration.go
tamirdavid1: i've added another get call right before preforming the logic of the migration.
For testing, i run 35 namespace cluster with 180~ deployments and check for conflict errors without success

RonFed (approved):

instrumentor/runtimemigration/runtimemigration.go
```diff
@@ -0,0 +1,393 @@
...
+			// Checking if the deployment is in the list of deployments that need to be processed
+			if contains(workloadNames, dep.Name) {
+
+				originalWorkloadEnvVar, err := envoverwrite.NewOrigWorkloadEnvValues(dep.Annotations)
+				if err != nil {
+					m.Logger.Error(err, "Failed to get original workload environment variables")
+					continue
+				}
+
+				workloadInstrumentationConfigReference := workloadNames[dep.Name]
```
blumamir, line 113: check if we get `nil` here

instrumentor/runtimemigration/runtimemigration.go
```diff
@@ -0,0 +1,393 @@
...
+	for i := range runtimeDetailsByContainer {
+		containerRuntimeDetails := &(runtimeDetailsByContainer)[i]
+
+		// Find the relevant container in runtimeDetailsByContainer
+		if containerRuntimeDetails.ContainerName != containerObject.Name {
+			continue
+		}
+		// Skip if the container has already been processed
+		if containerRuntimeDetails.RuntimeUpdateState != nil {
+			continue
```
blumamir, line 299:
```suggestion
			return nil
```
?

instrumentor/runtimemigration/runtimemigration.go
```diff
@@ -0,0 +1,393 @@
...
+func isEnvVarPresent(envVars []v1alpha1.EnvVar, envVarName string) bool {
+	for _, envVar := range envVars {
+		if envVar.Name == envVarName {
+			return true
+		}
+	}
+	return false
+}
+
+func removeEnvVar(envVars []v1alpha1.EnvVar, key string) []v1alpha1.EnvVar {
```
blumamir, line 385: is it used?

instrumentor/runtimemigration/runtimemigration.go
```diff
@@ -0,0 +1,393 @@
...
+		// clean up the value from all possible odigos additions
+		for _, value := range envOverwrite.GetPossibleValuesPerEnv(manifestEnvVarKey) {
+			manifestEnvVarValue = strings.ReplaceAll(manifestEnvVarValue, value, "")
+		}
+		withoutTrailingColon := cleanTrailingChar(manifestEnvVarValue, ":")
+		withoutTrailingAndLeadingSpace := strings.TrimSpace(withoutTrailingColon)
+		return withoutTrailingAndLeadingSpace
+	} else {
+		// manifestEnvVarKey does not exist in the EnvValuesMap
+		return ""
```
blumamir, line 364: This is just for safety and should never happen, right?

blumamir (approved):

instrumentor/runtimemigration/runtimemigration.go
tamirdavid1: correct, it is not really needed

PR #2097, opened 2024-12-29
Reduce usage of inst app CRD, move definition to inst config types

RonFed:
* Reduce usage of `instrumentedApplication` CRD where only the `runtimeDetailsByContainers` are used.
* Move definition of `RuntimeDetailsByContainer` `OtherAgent` and `EnvVar` to instrumentationConfig types.


instrumentor/controllers/instrumentationdevice/common.go
```diff
@@ -133,7 +133,7 @@ func addInstrumentationDeviceToWorkload(ctx context.Context, kubeClient client.C
 			agentsCanRunConcurrently = *odigosConfiguration.AllowConcurrentAgents
 		}
 
-		err, deviceApplied, deviceSkippedDueToOtherAgent := instrumentation.ApplyInstrumentationDevicesToPodTemplate(podSpec, runtimeDetails, otelSdkToUse, obj, logger, agentsCanRunConcurrently)
+		err, deviceApplied, deviceSkippedDueToOtherAgent := instrumentation.ApplyInstrumentationDevicesToPodTemplate(podSpec, instApp.Spec.RuntimeDetails, otelSdkToUse, obj, logger, agentsCanRunConcurrently)
```
BenElferink, line 136: Should we use the instru. config instead?
`instruConfig.Status.RuntimeDetailsByContainer`

BenElferink (approved):

instrumentor/controllers/instrumentationdevice/common.go
RonFed: I'll rename it in the next PR after the passed argument will be of type instConfig

blumamir (approved):

PR #2107, opened 2024-12-31
feat: webhook to set agent injection env var with odigos additions

tamirdavid1:
This is the initial phase of enabling the webhook to set agent injection-related environment variables (e.g., PYTHONPATH). The plan is to merge this after the migration and the new runtime inspection have been running for a while.

Once enough time has passed, I will merge this change to handle simpler cases where the envOverwriter is not involved. After this phase runs for a sufficient period, we can fully remove the envOverwriter and rely solely on this logic.

Finally, we can clean up the remaining code related to the envOverwriter.


instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -0,0 +1,168 @@
...
+	return nil
+}
+
+func prepareEnvVars(runtimeDetails *v1alpha1.RuntimeDetailsByContainer, envVarName string, otelsdk common.OtelSdk) []corev1.EnvVar {
+	var envVars []corev1.EnvVar
+	if runtimeDetails.EnvFromContainerRuntime == nil {
+		odigosValueForOtelSdk := envOverwrite.GetPossibleValuesPerEnv(envVarName)
+		if odigosValueForOtelSdk != nil {
+			valueToInject := odigosValueForOtelSdk[otelsdk]
+			patchedEnvVarValue := updatedGetPatchedEnvValue(envVarName, "", valueToInject) // empty observedValue
```
blumamir, line 103: isn't it simply `valueToInject`?

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -0,0 +1,168 @@
...
+			envVars = append(envVars, corev1.EnvVar{Name: envVarName, Value: *patchedEnvVarValue})
+		}
+
+	} else {
```
blumamir, line 107: what if `runtimeDetails.EnvFromContainerRuntime` is not nil, but the env var name is not in the list? looks like we will not add the env var in this case, right?

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -0,0 +1,168 @@
...
+		for _, envVar := range runtimeDetails.EnvFromContainerRuntime {
+			// Get the relevant envVar that we're iterating over
+			if envVar.Name != envVarName {
+				continue
+			}
+			odigosValueForOtelSdk := envOverwrite.GetPossibleValuesPerEnv(envVarName)
+			if odigosValueForOtelSdk != nil {
+				valueToInject := odigosValueForOtelSdk[otelsdk]
```
blumamir, line 115: can `valueToInject` be `nil`?

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -0,0 +1,168 @@
...
+	if !shouldInject(runtimeDetails, logger, container.Name) {
+		return nil
+	}
+
+	envVarsToInject := prepareEnvVars(runtimeDetails, envVarName, otelsdk)
+	container.Env = append(container.Env, envVarsToInject...)
+	return nil
+}
+
+func prepareEnvVars(runtimeDetails *v1alpha1.RuntimeDetailsByContainer, envVarName string, otelsdk common.OtelSdk) []corev1.EnvVar {
```
blumamir, line 97: there are so many env vars involved (manifest, envOverwriter, runtime detection). what does it mean "prepare"?

From it's name, it is impossible to understand it's role.

instrumentor/controllers/instrumentationdevice/pods_webhook.go
```diff
@@ -34,6 +37,14 @@ type PodsWebhook struct {
 var _ webhook.CustomDefaulter = &PodsWebhook{}
 
 func (p *PodsWebhook) Default(ctx context.Context, obj runtime.Object) error {
+	logger := log.FromContext(ctx)
+
+	// In certain scenarios, the raw request can be utilized to retrieve missing details, like the namespace.
+	req, err := admission.RequestFromContext(ctx)
```
edeNFed, line 43: Maybe worth mentioning this is relevant to Kubernetes <1.24 so we will know to delete this code once we no longer need to support version that old

instrumentor/controllers/instrumentationdevice/pods_webhook.go
```diff
@@ -43,96 +47,68 @@ func (p *PodsWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
-			container.Env = append(container.Env, corev1.EnvVar{
-				Name:  otelResourceAttributesEnvVarName,
-				Value: resourceAttributesEnvValue,
-			})
 		}
+
+		resourceAttributes := getResourceAttributes(podWorkload, container.Name)
+		resourceAttributesEnvValue := getResourceAttributesEnvVarValue(resourceAttributes)
+
+		container.Env = append(container.Env, corev1.EnvVar{
```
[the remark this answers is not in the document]
edeNFed: Please wait for me for that discussion 😀 I am also touching this area and wonder why you are against it

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -0,0 +1,168 @@
...
+
+func updatedGetPatchedEnvValue(envName string, observedValue string, desiredOdigosAddition string) *string {
+	_, ok := envOverwrite.EnvValuesMap[envName]
+	if !ok {
+		// Odigos does not manipulate this environment variable, so ignore it
+		return nil
+	}
+
+	// In case observedValue is exists but empty, we just need to set the desiredOdigosAddition without delim before
+	if observedValue == "" {
```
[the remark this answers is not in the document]
tamirdavid1: sounds reasonable

api/odigos/v1alpha1/instrumentationconfig_types.go
```diff
@@ -52,6 +53,15 @@ type InstrumentationConfigStatus struct {
 	RuntimeDetailsByContainer []RuntimeDetailsByContainer `json:"runtimeDetailsByContainer,omitempty"`
 }
 
+func (in *InstrumentationConfigStatus) GetRuntimeDetailsForContainer(container v1.Container) *RuntimeDetailsByContainer {
```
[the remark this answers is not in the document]
tamirdavid1: As we discussed let's discuss on this with @RonFed and take decision here

instrumentor/controllers/instrumentationdevice/pods_webhook.go
```diff
@@ -43,96 +47,68 @@ func (p *PodsWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
-func (p *PodsWebhook) getServiceNameForEnv(ctx context.Context, pod *corev1.Pod) (*string, *workload.PodWorkload) {
-
-	logger := log.FromContext(ctx)
+func (p *PodsWebhook) injectOdigosEnvVars(ctx context.Context, logger logr.Logger, pod *corev1.Pod) {
+	// Environment variables that remain consistent across all containers
+	commonEnvVars := getCommonEnvVars()
 
 	podWorkload, err := workload.PodWorkloadObject(ctx, pod)
 	if err != nil {
 		logger.Error(err, "failed to extract pod workload details from pod. skipping OTEL_SERVICE_NAME injection")
```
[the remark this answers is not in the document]
tamirdavid1: In case we wont be able to get the PodWorkloadObject, we wont be able to set Odigos additions to env var such as PYTHONPATH fromRuntime (manifest works) so instrumentation wont work basically.
i'll adjust the log message

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+}
+
+func getEnvVarNamesForLanguage(pl common.ProgrammingLanguage) []string {
+	return envOverwrite.EnvVarsForLanguage[pl]
+}
+
+// Return true if further processing should be skipped, either because it was already handled or due to a potential error (e.g., missing possible values)
+// Return false if the env was not processed using the manifest value and requires further handling by other methods.
+func handleManifestEnvVar(container *corev1.Container, envVarName string, otelsdk common.OtelSdk, logger logr.Logger) bool {
```
RonFed, line 43:
```suggestion
func ShouldIgnoreInjection(container *corev1.Container, envVarName string, otelsdk common.OtelSdk, logger logr.Logger) bool {
```

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// In case observedValue is exists but empty, we just need to set the desiredOdigosAddition without delim before
+	if strings.TrimSpace(observedValue) == "" {
+		return &desiredOdigosAddition
+	} else {
+		// In case observedValue is not empty, we need to append the desiredOdigosAddition with the delim
+		mergedEnvValue := observedValue + envValues.Delim + desiredOdigosAddition
+		return &mergedEnvValue
+	}
+}
```
RonFed, line 143: maybe this function should be under the `overwriter.go`?
That way we won't need to export the `Delim` and we could use that module in the future from the vm-agent

instrumentor/controllers/instrumentationdevice/pods_webhook.go
```diff
@@ -43,63 +56,36 @@ func (p *PodsWebhook) Default(ctx context.Context, obj runtime.Object) error {
 		pod.Annotations = map[string]string{}
 	}
 
-	serviceName, podWorkload := p.getServiceNameForEnv(ctx, pod)
-
 	// Inject ODIGOS environment variables into all containers
-	injectOdigosEnvVars(pod, podWorkload, serviceName)
+	p.injectOdigosEnvVars(ctx, logger, pod, req)
```
damemi, line 60: since you're passing the context here, do you need to pass the request too? You could just call `requestFromContext` in `injectOdigosEnvVars`

damemi:
Would you please update the test cases re:reported-name like I did in https://github.com/odigos-io/odigos/pull/2122/commits/8b60b1d187ded761db40be0833c6c12c35d551cf ? You can probably just cherry pick that commit

instrumentor/controllers/instrumentationdevice/pods_webhook.go
tamirdavid1: Yea technically you right, but i thought to put this on the main func of the webhook with the comment so further users of this webhook will know they can retrieve missing details from the raw request.
WDYT?

tamirdavid1:
> Would you please update the test cases re:reported-name like I did in [8b60b1d](https://github.com/odigos-io/odigos/commit/8b60b1d187ded761db40be0833c6c12c35d551cf) ? You can probably just cherry pick that commit
Yea I can do that, can you elaborate about why 26 spans reduced to 13?

instrumentor/controllers/instrumentationdevice/pods_webhook.go
damemi: I see what you mean, but in that case I would say just parse the namespace out of the request in the main func and not have to pass anything. It's unnecessary to pass both the context and the request

damemi:
> Yea I can do that, can you elaborate about why 26 spans reduced to 13?

The original test used the same resource names for 2 runs of traffic generation. So the first time it would generate 13 spans and then 13 more. But now the test updates the reported-name after the first set is generated, so the new set of spans is just 13 entirely new resources.

instrumentor/controllers/instrumentationdevice/pods_webhook.go
tamirdavid1: Agree, changed to what you've originally suggested requestFromContext in injectOdigosEnvVars

tamirdavid1:
> > Yea I can do that, can you elaborate about why 26 spans reduced to 13?
> 
> The original test used the same resource names for 2 runs of traffic generation. So the first time it would generate 13 spans and then 13 more. But now the test updates the reported-name after the first set is generated, so the new set of spans is just 13 entirely new resources.

Done, the requested commit has been cherry-picked.

damemi (approved):

PR #2111, opened 2025-01-01
Resource Attributes - refactor and add missing attributes

edeNFed:
This PR do the following changes:
- Unify all resource attributes logic to one package
- Remove anything related to device id and communication with kubelet from OpAMP / data-collector
- Modify eBPF director, Pods Webhook and OpAMP server to use the new unified logic.
- Use `k8sattributeprocessor` in `odigos-data-collection` to fill the rest of the needed kubernetes resource attributes


common/resourceattributes/root.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// AfterPodStart returns the resource attributes that should be set after the container starts.
+// This function is called by the eBPF director or by the OpAMP server after the container starts.
+// Currently, the attributes returned are the minimal set of attributes that should be set to identify the container.
+// Other attributes (k8s related, cloud provider, etc) are set later by a dedicated processor in odigos-data-collector.
+// Notice that this function is not always called, only OpAMP clients or eBPF-based clients will call this function.
+// Vanilla OpenTelemetry clients will only call BeforePodStart.
+// OpAMP/eBPF clients WILL OVERWRITE the attributes set by BeforePodStart with the attributes returned by this function.
+func AfterPodStart(identifier *ContainerIdentifier) Attributes {
```
RonFed, line 44: The `AfterPodStart` and `BeforePodStart` are identical.
Are we assuming they will differ in the future?

instrumentation/factory.go
RonFed, line 1: This will require an update in the vm-agent which also depends on this

instrumentor/controllers/instrumentationdevice/manager.go
```diff
@@ -91,7 +93,7 @@ func SetupWithManager(mgr ctrl.Manager) error {
 		ControllerManagedBy(mgr).
 		Named("instrumentationdevice-collectorsgroup").
 		For(&odigosv1.CollectorsGroup{}).
-		WithEventFilter(predicate.And(&odigospredicate.OdigosCollectorsGroupNodePredicate, &odigospredicate.CgBecomesReadyPredicate{})).
+		WithEventFilter(predicate.And[client.Object](&odigospredicate.OdigosCollectorsGroupNodePredicate, &odigospredicate.CgBecomesReadyPredicate{})).
```
RonFed, line 96: ?

common/resourceattributes/root.go
edeNFed: Yes, the main goal of this file is to explain how resource attributes are handled in Odigos.

This split into two phases is because:
A. This is what happens, it is just an identical value
B. To better explain it
C. Prepare for the future when we would want to add "slow" resource attribute

odiglet/pkg/ebpf/settings_getter.go
```diff
@@ -29,9 +30,13 @@ func (ksg *k8sSettingsGetter) Settings(ctx context.Context, kd K8sProcessDetails
 	}
 
 	return instrumentation.Settings{
-		ServiceName:        OtelServiceName,
-		ResourceAttributes: utils.GetResourceAttributes(kd.pw, kd.pod.Name),
-		InitialConfig:      sdkConfig,
+		ServiceName: OtelServiceName,
+		ResourceAttributes: resourceattributes.AfterPodStart(&resourceattributes.ContainerIdentifier{
```
RonFed, line 34: I think this change will drop the `deployment.name` attribute we used to have - now the processor will add them. Why not add more attributes here if we can get them, instead of relying on the processor?

autoscaler/controllers/datacollection/configmap.go
```diff
@@ -144,14 +144,64 @@ func calculateConfigMapData(nodeCG *odigosv1.CollectorsGroup, sources *odigosv1.
...
+				"k8s.daemonset.name",
+				"k8s.cronjob.name",
+				"k8s.job.name",
+				"k8s.cluster.uid",
+				"k8s.node.name",
+				"container.id",
+				"container.image.name",
+				"container.image.tag",
+			},
+		},
```
RonFed, line 184: Does this make the `add cluster info` action obsolete?

common/resourceattributes/attributes.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"go.opentelemetry.io/otel/attribute"
+	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
+)
+
+type Attributes []attribute.KeyValue
```
RonFed, line 11: This type definition introduces a breaking change to the `factory` API for eBPF instrumentation.
It adds some better syntax, but I prefer to just use the `Otel` type. We could still have the util functions here.

instrumentor/controllers/webhook/env_vars.go
```diff
@@ -0,0 +1,67 @@
...
+		return newVal
+	}
+
+	AppendWithSpace modificationFunc = func(origVal string, newVal string) string {
+		return origVal + " " + newVal
+	}
+
+	AppendWithComma modificationFunc = func(origVal string, newVal string) string {
+		return origVal + "," + newVal
+	}
```
RonFed, line 20: If those are just used for testing, better to move them to the `_test` file

PR #2112, opened 2025-01-01
Feature/gen 2030 group sources to destination

alonkeyval:


Makefile
```diff
@@ -206,6 +206,11 @@ cli-uninstall:
 	@echo "Uninstalling odigos from source. version: $(ODIGOS_CLI_VERSION)"
 	cd ./cli ; go run -tags=embed_manifests . uninstall
 
+.PHONY: cli-uninstall
```
BenElferink, line 209: This carried over as a duplicate, with typo in the "echo" message, please remove :)

cli/cmd/resources/autoscaler.go
```diff
@@ -172,6 +172,15 @@ func NewAutoscalerClusterRole() *rbacv1.ClusterRole {
...
 				Verbs:     []string{"get", "list", "watch"},
 			},
+			{
+				Verbs: []string{
+					"get",
+					"list",
+					"watch",
+				},
+				APIGroups: []string{"odigos.io"},
+				Resources: []string{"sources"},
```
BenElferink, line 182: Add a comment for why this is needed, and add it to the `cli/cmd/resources/README.md` file.
Also add resources for helm: `helm/odigos/templates/autoscaler`

cli/cmd/resources/instrumentor.go
```diff
@@ -201,6 +201,16 @@ func NewInstrumentorClusterRole() *rbacv1.ClusterRole {
 				Resources: []string{"sources/finalizers"},
 				Verbs:     []string{"update"},
 			},
+			{
+				APIGroups: []string{"odigos.io"},
+				Resources: []string{"sources"},
+				Verbs:     []string{"create", "delete", "get", "list", "patch", "update", "watch"},
```
BenElferink, line 207: Add a comment for why this is needed, and add it to the `cli/cmd/resources/README.md` file.
Also add resources for helm: `helm/odigos/templates/instrumentor`

autoscaler/controllers/common/processors.go
```diff
@@ -70,3 +78,90 @@ func GetGenericBatchProcessor() odigosv1.Processor {
...
+		matchConditions := []string{}
+		if len(dest.Spec.SourceSelector.Namespaces) > 0 {
+			for _, namespace := range dest.Spec.SourceSelector.Namespaces {
+				matchConditions = append(matchConditions, fmt.Sprintf("%s/*/*", namespace))
+			}
+		}
+		if len(dest.Spec.SourceSelector.Groups) > 0 {
+			matchedSources := fetchSourcesByGroups(ctx, kubeClient, dest.Spec.SourceSelector.Groups, logger)
+			for _, source := range matchedSources {
+				key := fmt.Sprintf("%s/%s/%s", source.Spec.Workload.Namespace, source.Spec.Workload.Name, source.Spec.Workload.Kind)
```
damemi, line 111: how will this work with Namespace Sources? You can create one Source to instrument an entire namespace, so won't you need to do some sort of linking between the workload and the namespace source?

damemi (commented):
Can you add an e2e test for this? You should be able to add it to [tests/e2e/source](https://github.com/odigos-io/odigos/tree/main/tests/e2e/source) to show:

* Workload Source with `odigos.io/group-` label works with grouping
* Namespace Source with `odigos.io/group-` label includes all the workloads in the namespace
* Workload Source with `disableInstrumentation=true` excludes that Source from the grouping (? not sure what the behavior should be here)

tamirdavid1 (commented):
In general I think that we should do it with connector as discussed

autoscaler/controllers/common/processors.go
alonkeyval: Currently we rely on the source crd to represent a group. What do you think about creating the source crd if it doesn't exist (because the source is represented by source crd from namespace kind) once the UI creates a group?

in this case we will have the source crd that represent the workload and one more for the namespace

collector/processors/odigosroutingfilterprocessor/processor.go
```diff
@@ -0,0 +1,152 @@
...
+
+		for j := 0; j < ilSpans.Len(); j++ {
+			scopeSpan := ilSpans.At(j)
+			spans := scopeSpan.Spans()
+
+			spans.RemoveIf(func(span ptrace.Span) bool {
+				namespace, name, kind := extractResourceDetails(resourceAttributes)
+				return !fp.matches(name, namespace, kind)
+			})
+		}
```
blumamir, line 45: since you make the decision based on the resource attributes only, you can remove the spans (metrics/logs) iterations and simply remove or keep the entire scopeSpans

collector/processors/odigosroutingfilterprocessor/processor.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func extractResourceDetails(attributes pcommon.Map) (namespace, name, kind string) {
+	namespace = getAttribute(attributes, k8sNamespaceNameAttr)
+	if namespace == "" {
+		return "", "", ""
+	}
+
+	name, kind = getDynamicNameAndKind(attributes)
+	if name == "" || kind == "" {
```
blumamir, line 121: same here, return "found" boolean to indicate this

autoscaler/controllers/common/processors.go
blumamir, line 111: we usually write kind before name. e.g. `deployment-foo` and not `foo-deployment`
```suggestion
				key := fmt.Sprintf("%s/%s/%s", source.Spec.Workload.Namespace, source.Spec.Workload.Kind, source.Spec.Workload.Name)
```

collector/processors/odigosroutingfilterprocessor/config.go
```diff
@@ -0,0 +1,23 @@
+package odigosroutingfilterprocessor
+
+import (
+	"go.opentelemetry.io/collector/component"
+)
+
+type Config struct {
+	MatchConditions []string `mapstructure:"match_conditions"`
+	MatchMap        map[string]struct{}
```
blumamir, line 9: this is not part of the config. I think it should go directly as a member of the `filterProcessor`

blumamir (commented):
Looks good 👍 
Added few comments most are nit

autoscaler/controllers/common/processors.go
```diff
@@ -70,3 +78,90 @@ func GetGenericBatchProcessor() odigosv1.Processor {
...
+		}
+		sources = append(sources, sourceList.Items...)
+	}
+	return sources
+}
+
+func fetchSourcesByGroups(ctx context.Context, kubeClient client.Client, groups []string, logger logr.Logger) []odigosv1.Source {
+	sourceMap := make(map[string]odigosv1.Source)
+	for _, group := range groups {
+		labelSelector := labels.Set{fmt.Sprintf("odigos.io/group-%s", group): "true"}.AsSelector()
```
damemi, line 144: What about making the label name constant, and the group is the value of the label? So instead of 

```
odigos.io/group-foo: true
```

it's 

```
odigos.io/source-group: foo
```

Labels are meant to hold variable string values

autoscaler/controllers/common/processors.go
damemi: We should minimize how often we create Sources on behalf of the user, since it's supposed to be a user-created object. Also, even the UI doesn't create Namespace Sources so if there is one it means the user likely intentionally created it instead of using Workload Sources.

I think it would be better here to handle a Namespace Source as a special case, where you loop through all the workloads in the namespace and check if they should inherit the group from the namespace (like we do in the Instrumentor).

The next question is what if a workload Source already exists with a different group than the namespace Source? Should the workload be part of both groups? Or should the workload Source take priority? I think to begin with, it should take priority over the namespace.

So, when you see a Namespace Source:

1. List all other Sources in that Namespace to check for different groups or excluded workloads
2. Remember any workloads with a different group or excluded
3. List through all the deployments/statefulsets/daemonsets in the namespace
4. If the workload isn't in the list from earlier, add it to the group in the Namespace Source

The pseudo-go would look like this:

```go
matchedSources := fetchSourcesByGroups(ctx, kubeClient, dest.Spec.SourceSelector.Groups, logger)
for _, source := range matchedSources {
  if source.Spec.Workload.Kind == workload.WorkloadKindNamespace {
    excludedWorkloads := make(map[string]struct{}{})
    // exclude disabled Sources or Sources with a different group
    nsSources := &odigosv1.SourceList{}
    err := kubeClient.List(ctx, sourceList, client.InNamespace(source.Spec.Workload.Namespace))
    if err != nil {
      continue
    }
    for _, nsSource := range nsSources.Items {
      val, exists := nsSource.Labels["odigos.io/source-group"]
      // if the nsSource has a group label that's different from the original matchedSource, or the Source is disabled, exclude it
      if (exists && val != source.Labels["odigos.io/source-group"]) || !nsSource.Spec.DisableInstrumentation {
        excludedWorkloads[fmt.Sprintf("%s/%s/%s", source.Spec.Workload.Namespace, source.Spec.Workload.Name, source.Spec.Workload.Kind] = struct{}{}
      }

    // get workload keys for non-excluded workloads in the namespace
    deployments := &v1.DeploymentList{}
    err := kubeClient.List(ctx, deployments, client.InNamespace(source.Spec.Workload.Namespace))
      if err != nil {
        continue
      } 
    }
    
    for _, dep := range deployments.Items {
      key := fmt.Sprintf("%s/%s/%s", deployment.GetNamespace(), deployment.GetName(), workload.WorkloadKindDeployment)
      if _, exists := excludedWorkloads[key]; !exists {
        matchConditions = append(matchConditions, key)
      }
    }
    // repeat for statefulsets and daemonsets

  } else { //not a namespace source, follow the flow you have now
    key := fmt.Sprintf("%s/%s/%s", source.Spec.Workload.Namespace, source.Spec.Workload.Name, source.Spec.Workload.Kind)
    matchConditions = append(matchConditions, key)
  }
}
```

This should be a test case too.

Technically, you probably don't even need to handle any workload Sources with `disableInstrumentation: true`, because they won't be sending instrumentation anyway. But I guess it doesn't hurt here.

autoscaler/controllers/common/processors.go
alonkeyval: The idea was to do it this way so we could implement the behavior of disabling/enabling a group without deleting it.

autoscaler/controllers/common/processors.go
BenElferink: Correction:
The UI does create namespace sources, when toggling `future apps` to true.
See [PersistK8sNamespace](https://github.com/odigos-io/odigos/blob/b2673ab20b78b08e7a04b4bc9d51b2e6c212442e/frontend/graph/schema.resolvers.go#L357) for example :)

damemi: Ah thanks @BenElferink, I didn't realize that!

My point is still though, even in that case, the user is doing something that they expect to create a Source (instrumenting the Namespace). I'm not sure that in this case a user will expect to see a bunch of new Sources show up in their namespace just by adding the `source-group` label. That also would effectively make `source-group` a replacement for `odigos-instrumentation: true`.

I think there is also still the question of a conflicting group on the namespace and an existing workload Source. You'll have to check for that anyway, right? I'm open to other ideas for this or any reasoning that I'm misunderstanding btw

alonkeyval: I think the question here is how do I create a group for a source that is defined as a namespace instrument? It doesn't have a source crd.

damemi: The point of a Namespace Source is that you don't need additional Sources for each workload in the namespace. They are implicitly instrumented by the Namespace Source (unless they have their own explicit Workload Source, such as for disabling instrumentation).

So to create a group for a Namespace, list all the workloads in the namespace and build a key for each one the same way you are with normal workloads. 

The catch is that you need to check if any of those workloads *do* have their own explicit Workload Sources, and if so, skip them. Since your function here is listing all the available Source objects, you'll get to those skipped Workloads eventually anyway, where you can handle them like you do now

damemi (approved):
Talked to @alonkeyval offline and I misunderstood a couple things about this

* The UI should definitely create Sources with a group for the user when necessary. For example, if a Namespace is instrumented and the user chooses X workloads to create a group. The Frontend will create workload Sources for those workloads.
* I didn't realize the connector/processor has both a workload and a namespace mode. So, having to do the work to figure out inherited workloads shouldn't be necessary since the processor can just see the workload
* For the case where a workload has a different group than its namespace, both groups apply.
* The `odigos.io/source-group` label won't work if a workload should be in multiple groups. Labels aren't supposed to have list/array values, so it's kind of unfortunate. But going back to `odigos.io/group-*` seems like the best option

autoscaler/controllers/common/processors.go
```diff
@@ -70,3 +78,76 @@ func GetGenericBatchProcessor() odigosv1.Processor {
...
+		}
+		if len(dest.Spec.SourceSelector.Groups) > 0 {
+			matchedSources, err := fetchSourcesByGroups(ctx, kubeClient, dest.Spec.SourceSelector.Groups, logger)
+
+			if err != nil {
+				return nil, err
+			}
+
+			for _, source := range matchedSources {
+				sourceSelectorKey := fmt.Sprintf("%s/%s/%s", source.Spec.Workload.Namespace, source.Spec.Workload.Kind, source.Spec.Workload.Name)
```
blumamir, line 111: we already calculated this key in the `fetchSourcesByGroups`.

Can we reuse this key instead of calculating it again?

autoscaler/controllers/common/processors.go
```diff
@@ -70,3 +78,76 @@ func GetGenericBatchProcessor() odigosv1.Processor {
...
+				matchConditions = append(matchConditions, sourceSelectorKey)
+			}
+		}
+
+		sanitizedProcessorName := strings.ReplaceAll(dest.GetID(), ".", "-")
```
blumamir, line 116: How come this isn't an issue with other keys in the configmap?

collector/processors/odigossourcesfilterprocessor/processor.go
```diff
@@ -0,0 +1,147 @@
...
+func (fp *filterProcessor) processTraces(ctx context.Context, td ptrace.Traces) (ptrace.Traces, error) {
+	rspans := td.ResourceSpans()
+
+	rspans.RemoveIf(func(resourceSpan ptrace.ResourceSpans) bool {
+		resourceAttributes := resourceSpan.Resource().Attributes()
+		namespace, name, kind, found := extractResourceDetails(resourceAttributes)
+		if found {
+			return !fp.matches(name, namespace, kind)
+		}
+		return false
```
blumamir, line 55: If the attributes are not found, should we remove it or not?

collector/processors/odigossourcesfilterprocessor/processor.go
```diff
@@ -0,0 +1,147 @@
...
+	})
+
+	return ld, nil
+}
+
+func (fp *filterProcessor) matches(name, namespace, kind string) bool {
+	if namespace == "" {
```
blumamir, line 92: Can it be empty? aren't we verifying it exists in `getAttribute`?
same for name and kind

tests/e2e/source-webhooks/chainsaw-test.yaml
blumamir, line 1: are there any real changes in this file?

autoscaler/controllers/common/processors.go
alonkeyval: Processors have stricter naming rules compared to other OpenTelemetry components like Exporters. While Exporters can include dots (.) in their names since they are used as hierarchical identifiers, Processors require flat, valid identifiers. This is why dots (.) must be replaced with dashes (-) in processor names, while they work fine in Exporters or Go maps.

blumamir (approved):

PR #2122, opened 2025-01-04
Use request.Namespace in Pod webhook

damemi:
In older versions of k8s (prior to https://github.com/kubernetes/kubernetes/pull/94637 in ~1.24), the Pod object received by an admission webhook isn't guaranteed to have its namespace set. Instead, we should use the namespace from the admission request.

I found this while adding tests that use the `odigos.io/reported-name` annotation, and saw that the service name wasn't taking effect on some pods. The instrumentor would fail to parse some of them with the following error:

```
2025-01-03T20:38:36Z	ERROR	admission	failed to get workload object from cache. cannot check for workload annotation. using workload name as OTEL_SERVICE_NAME	{"webhookGroup": "", "webhookKind": "Pod", "Pod": {"name":"","namespace":"default"}, "namespace": "default", "name": "", "resource": {"group":"","version":"v1","resource":"pods"}, "user": "system:serviceaccount:kube-system:replicaset-controller", "requestID": "8e9331f3-b5fa-4334-a238-a755a32d7b62", "error": "Deployment.apps \"inventory\" not found"}
github.com/odigos-io/odigos/instrumentor/controllers/instrumentationdevice.(*PodsWebhook).getServiceNameForEnv
	/workspace/instrumentor/controllers/instrumentationdevice/pods_webhook.go:67
github.com/odigos-io/odigos/instrumentor/controllers/instrumentationdevice.(*PodsWebhook).Default
	/workspace/instrumentor/controllers/instrumentationdevice/pods_webhook.go:46
```

(https://github.com/odigos-io/odigos/pull/2008#issuecomment-2569829498)

This uses the request namespace instead, which is always set(https://github.com/kubernetes/kubernetes/issues/113952#issuecomment-1317042694).

### DIFF
--- a/instrumentor/controllers/instrumentationdevice/pods_webhook.go
+++ b/instrumentor/controllers/instrumentationdevice/pods_webhook.go
@@ -14,6 +14,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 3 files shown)


damemi:
Didn't realize this is covered by https://github.com/odigos-io/odigos/pull/2107/files#diff-09f9653d9c6c1a8bb7d52e7f3bf0060e0176369df0ee04d04bfb6b4d620794ddR40-R45, closing

PR #2124, opened 2025-01-04
fix: remove finalizer only after instrumentation config is deleted

blumamir:
this is an attempt to address this comment:

> 			// Remove the finalizer first, because if the InstrumentationConfig is not found we
>			// will deadlock on the finalizer never getting removed.
>			// On the other hand, this could end up deleting a Source with an orphaned InstrumentationConfig.

It will basically first try to remove the instrumentation config, report any real errors to let the controller retry the event, and ignore `NotFound`.
Then it will remove the finalizer (but only after the config is verified to be removed).

### DIFF
--- a/instrumentor/controllers/startlangdetection/source_controller.go
+++ b/instrumentor/controllers/startlangdetection/source_controller.go
@@ -3,6 +3,7 @@ package startlangdetection
 import (
 	"context"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


damemi:
@blumamir thanks for this! But, this will be covered by https://github.com/odigos-io/odigos/pull/2105

For context, when I added this I didn't know that `deleteinstrumentedapplication` controller deleted InstrumentationConfigs too. In #2105, I moved all the deletion to that controller (which makes more sense anyway since this is the startlangdetection controller). That removed the need for this weird finalizer deadlock check

So, I think we can close this. Please correct me if I misunderstood anything

blumamir:
> @blumamir thanks for this! But, this will be covered by #2105
> 
> For context, when I added this I didn't know that `deleteinstrumentedapplication` controller deleted InstrumentationConfigs too. In #2105, I moved all the deletion to that controller (which makes more sense anyway since this is the startlangdetection controller). That removed the need for this weird finalizer deadlock check
> 
> So, I think we can close this. Please correct me if I misunderstood anything

makes sense 👍

PR #2125, opened 2025-01-04
[DO NOT MERGE] Remove device ID processor

edeNFed:
As we move to the new webhook mechanism, we no longer need the device id processor (which has many drawbacks such as connectivity to kubelet). 
This PR removes it and anything dependent on it.

### Before merging this PR the following tasks needs to be completed:
- [ ] Wait a few weeks to allow users to upgrade their Odigos version and restart their pods
- [ ] Communicate this change over Slack
- [ ] Change Python Instrumentation to remove device id 
- [ ] Change NodeJS instrumentation to remove device id


blumamir:
replaced by #2413

PR #2146, opened 2025-01-07
Remove cert-manager certificate generation, use self sign instead

edeNFed:


BenElferink (approved):

damemi:
@edeNFed I think we should require cert-manager for OpenShift certification. The [requirements for certification](https://docs.redhat.com/en/documentation/red_hat_software_certification/8.64/html/red_hat_openshift_software_certification_policy_guide/assembly-products-managed-by-an-operator_openshift-sw-cert-policy-container-images#con-operator-requirements_openshift-sw-cert-policy-products-managed) include generating an [Operator Bundle](https://sdk.operatorframework.io/docs/cli/operator-sdk_generate_bundle/), which needs a set of ready-to-deploy manifests.

We can generate these manifests with `helm template`, but they will have the same self-signed cert generated. We obviously shouldn't check that into our open source code, but even if we only do it in the release pipeline for the Operator Bundle image it means every install of Odigos will use the same cert.

Do you think we could make cert-manager an optional dependency? I imagine some users might want to use cert-manager anyway.

PR #2158, opened 2025-01-08
Run device manager outside of the error group

RonFed:
The device manager library doesn't support passing a context, hence it can't be canceled by the error group.
To avoid blocking Odiglet in case of a fatal error, the device manager is changed to run in a "regular" goroutine.


odiglet/odiglet.go
```diff
@@ -108,11 +109,15 @@ func (o *Odiglet) Run(ctx context.Context) {
...
 	// however, internally it uses a context to cancel the device manager once SIGTERM or SIGINT is received.
-	g.Go(func() error {
+	// We run it outside of the error group to avoid blocking on Wait() in case of a fatal error.
+	go func()  {
 		err := runDeviceManager(o.clientset, o.deviceInjectionCallbacks)
 		log.Logger.V(0).Info("Device manager exited")
-		return err
-	})
+		if err != nil {
+			log.Logger.Error(err, "Device manager exited with error")
```
blumamir, line 117: nit: when existed with error, we will have 2 log lines for this:

```
Device manager exited
Device manager exited with error
```

blumamir (approved):

PR #2164, opened 2025-01-08
Verify the language of a process before instrumenting

RonFed:
Since our language detection is done in the container level but instrumentation is done in the process level it is possible that we'll try to instrument a process as being in certain language even though it is written in a different one - this can happen when multiple processes exist in the same container.
To handle this, the distribution matcher is updated to call the language inspector of the language written in the device - and verify the language match.

In addition, the manager is updated to call the reporter `OnInit` method regardless of an init error. Also tracking the PID in case of an init error - this allows the reporter to clean the instrumentation instance once the PID is exiting.


edeNFed (approved):

blumamir (approved):
Looks good 💯

PR #2170, opened 2025-01-08
Merge `feature/source-crd` into `main`

damemi:
This merges the work done on https://github.com/odigos-io/odigos/tree/feature/source-crd into `main`, including:

* Workload instrumentation with Source
* Namespace instrumentation with Source
* Workload exclusion from Namespace instrumentation with Source
* Webhooks for validating and defaulting Source
* UI changes to rely solely on creating and deleted Source to instrument and uninstrument applications
* e2es for Sources

This does not remove backend support for instrumentation labels, and the behavior of using Sources alongside labels is undefined. The next PR will remove backend support for instrumentation labels and change the behavior of the label to simply create a Source.

This PR uses a branch based off `feature/source-crd` so we could re-run generators locally to resolve conflicts.


BenElferink (approved):

instrumentor/sources_webhooks.go
```diff
@@ -0,0 +1,263 @@
...
+	if source.Spec.Workload.Kind == "Namespace" &&
+		(source.Spec.Workload.Name != source.Spec.Workload.Namespace) {
+		allErrs = append(allErrs, field.Invalid(
+			field.NewPath("spec").Child("workload").Child("namespace"),
+			source.Spec.Workload.Namespace,
+			"namespace Source must have matching workload.name and workload.namespace",
+		))
+	}
+
+	workloadKind := workload.WorkloadKindFromString(string(source.Spec.Workload.Kind))
```
blumamir, line 225: This will allow the kind to be with any casing (e.g. `Deployment`, `deployment`, `DEPLOYMENT`). do you think we should enforce a specific casing so that we don't end up with mismatches downstream when we try to use it?

instrumentor/sources_webhooks.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+	if new.Spec.Workload != old.Spec.Workload {
+		allErrs = append(allErrs, field.Invalid(
+			field.NewPath("spec").Child("workload"),
+			new.Spec.Workload,
+			"Source workload is immutable",
+		))
+	}
+
+	errs := s.validateSourceFields(ctx, new)
```
blumamir, line 162: If it's an update, and we've already verified that the labels and spec workload is not changed, what value do we get from calling it here? or is it only for safety?

instrumentor/sources_webhooks.go
```diff
@@ -0,0 +1,263 @@
...
+func (s *SourcesValidator) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) (admission.Warnings, error) {
+	var allErrs field.ErrorList
+	old, ok := oldObj.(*v1alpha1.Source)
+	if !ok {
+		return nil, fmt.Errorf("expected a Source but got a %T", old)
+	}
+	new, ok := newObj.(*v1alpha1.Source)
+	if !ok {
+		return nil, fmt.Errorf("expected a Source but got a %T", new)
+	}
```
blumamir, line 115: nit: if this error ends up hitting, the 2 lines has the same text, which can make it difficult for us later on to understand which of the 2 caused it.

will we benefit from adding `old` and `new` to the error text? or will we get stacktrace here?

instrumentor/controllers/deleteinstrumentationconfig/namespace_controller.go
```diff
@@ -165,11 +169,24 @@ func syncGenericWorkloadListToNs(ctx context.Context, c client.Client, kind work
...
+		return false, err
+	}
+
+	if sourceList.Namespace != nil && sourceList.Namespace.DeletionTimestamp.IsZero() {
+		return true, nil
+	}
+
+	if sourceList.Workload != nil && sourceList.Workload.DeletionTimestamp.IsZero() {
+		return false, nil
+	}
```
blumamir, line 184: from previous review - since we first check the namespace and returns `true` if it exists, we will return wrong value here if the `sourceList.Workload` also exists (and then the workload is not inheriting)

consider updating the implementation or rename the function to reflect it's return value)

instrumentor/controllers/deleteinstrumentationconfig/namespace_controller.go
```diff
@@ -165,11 +169,24 @@ func syncGenericWorkloadListToNs(ctx context.Context, c client.Client, kind work
...
 // instrumented due to the label on the namespace. These are workloads with the label missing.
 // (they inherit the instrumentation from the namespace this way)
-func isInheritingInstrumentationFromNs(obj client.Object) bool {
+func isInheritingInstrumentationFromNs(ctx context.Context, c client.Client, obj client.Object) (bool, error) {
+	sourceList, err := v1alpha1.GetWorkloadSources(ctx, c, obj)
+	if err != nil {
+		return false, err
+	}
+
+	if sourceList.Namespace != nil && sourceList.Namespace.DeletionTimestamp.IsZero() {
```
blumamir, line 178: WDYT about adding a function to utils that gets a source and returns if it's being deleted or not. could be more fluent to read, e.g.

```suggestion
	if sourceList.Namespace != nil && !k8sutils.IsTerminating(sourceList.Namespace) {
```

or:
```suggestion
	if sourceList.Namespace != nil && k8sutils.IsActive(sourceList.Namespace) {
```

We use it in many places and the double negative can be confusing to read :)

The utils function can also check the `nil` and make usage even shorter and more easy to understand.

instrumentor/controllers/deleteinstrumentationconfig/namespace_controller.go
```diff
@@ -165,11 +169,24 @@ func syncGenericWorkloadListToNs(ctx context.Context, c client.Client, kind work
 // when reconciling the namespace, the usecase is to delete instrumentation for workloads that were only
 // instrumented due to the label on the namespace. These are workloads with the label missing.
 // (they inherit the instrumentation from the namespace this way)
-func isInheritingInstrumentationFromNs(obj client.Object) bool {
+func isInheritingInstrumentationFromNs(ctx context.Context, c client.Client, obj client.Object) (bool, error) {
```
blumamir, line 172: nit: update the comment for this function to reflect what it currently does

instrumentor/controllers/deleteinstrumentationconfig/source_controller.go
```diff
@@ -0,0 +1,150 @@
...
+	source := &v1alpha1.Source{}
+	err := r.Get(ctx, req.NamespacedName, source)
+	if err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// If this is a regular Source that's being deleted, or a workload Exclusion Source
+	// that's being created, try to uninstrument relevant workloads.
+	if source.DeletionTimestamp.IsZero() == v1alpha1.IsWorkloadExcludedSource(source) {
+		logger.Info("Reconciling workload for Source object", "name", req.Name, "namespace", req.Namespace)
```
blumamir, line 55: consider also adding here:
- it's kind
- if it's being deleted or not (active/terminating)
- if excluded or not

Since it affects how this function branches, I think it has value to record

instrumentor/controllers/deleteinstrumentationconfig/source_controller.go
```diff
@@ -0,0 +1,150 @@
...
+
+		if source.Spec.Workload.Kind == "Namespace" {
+			logger.V(2).Info("Uninstrumenting workloads for Namespace Source", "name", req.Name, "namespace", req.Namespace)
```
blumamir, line 58: we already have an info log before the if, can we drop this one?

instrumentor/controllers/deleteinstrumentationconfig/source_controller.go
blumamir, line 55: There is an info record in the beginning of the function saying "Reconciling Source object". can we merge them into one so the logs are less chatty?

instrumentor/controllers/deleteinstrumentationconfig/source_controller.go
```diff
@@ -0,0 +1,150 @@
...
+			!source.DeletionTimestamp.IsZero() &&
+			controllerutil.ContainsFinalizer(source, consts.DeleteInstrumentationConfigFinalizer) {
+			controllerutil.RemoveFinalizer(source, consts.DeleteInstrumentationConfigFinalizer)
+			if err := r.Update(ctx, source); err != nil {
+				return k8sutils.K8SUpdateErrorHandler(err)
+			}
+		}
+	}
+
+	return ctrl.Result{}, client.IgnoreNotFound(err)
```
blumamir, line 82: nit (already checking before):

```suggestion
	return ctrl.Result{}, nil
```

blumamir (commented):
Thanks mike, this is awesome!

I am still in the process of reviewing, but adding some comments and suggestions in the meanwhile so we can keep the wheels in motion.

Will continue the review tomorrow

cli/cmd/resources/ui.go
```diff
@@ -241,6 +241,11 @@ func NewUIClusterRole() *rbacv1.ClusterRole {
 				Resources: []string{"instrumentationconfigs", "instrumentationinstances"},
 				Verbs:     []string{"get", "list", "watch"},
 			},
+			{ // Needed to instrument / uninstrument sources
```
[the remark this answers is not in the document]
damemi: @blumamir Sources are created in the namespace for the workload they instrument. So, cluster admins can see Sources in all namespaces but individual teams can be restricted to only Sources in their own namespaces (just like today, they can be limited to labelling workloads in their own namespaces).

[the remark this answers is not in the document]
blumamir: I was under the impression that we only store source objects in `odigos-system` namespace.

in this case, why do we need to store the workload namespace label in the object? isn't it redundant and adds extra complexity?

What happens if a source is created in namespace `foo` and contains other namespace `bar` in the spec?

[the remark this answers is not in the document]
blumamir: also, is there anything that delete the source objects when uninstalling odigos?

If they are not in odigos ns, they need to be specifically deleted I think

instrumentor/controllers/startlangdetection/source_controller.go
```diff
@@ -0,0 +1,140 @@
...
+		if source.Spec.Workload.Kind == "Namespace" {
+			// pre-process existing Sources for specific workloads so we don't have to make a bunch of API calls
+			// This is used to check if a workload already has an explicit Source, so we don't overwrite its InstrumentationConfig
+			sourceList := v1alpha1.SourceList{}
+			err := r.Client.List(ctx, &sourceList, client.InNamespace(source.Spec.Workload.Name))
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			namespaceKindSources := make(map[workload.WorkloadKind]map[string]struct{})
+			for _, source := range sourceList.Items {
```
RonFed, line 47: Thie name shadows the variable defined in `source := &v1alpha1.Source{}` - maybe better to rename it to avoid confusion

instrumentor/controllers/startlangdetection/workload_controllers.go
```diff
@@ -58,9 +58,21 @@ func reconcileWorkload(ctx context.Context, k8sClient client.Client, objKind wor
...
+		}
+		if sourceList.Workload == nil && sourceList.Namespace == nil {
+			return ctrl.Result{}, nil
+		}
+		// if this is explicitly excluded (and the excluded Source isn't being deleted), skip
+		if sourceList.Workload != nil {
+			if odigosv1.IsWorkloadExcludedSource(sourceList.Workload) && sourceList.Workload.DeletionTimestamp.IsZero() {
+				return ctrl.Result{}, nil
+			}
+		}
```
RonFed, line 74: Maybe worth to consider adding a function similar to `IsWorkloadInstrumentationEffectiveEnable` which will do the same logic. I assume that it could be useful also for the UI

instrumentor/sources_webhooks.go
```diff
@@ -0,0 +1,263 @@
...
+	if !v1alpha1.IsWorkloadExcludedSource(source) &&
+		source.DeletionTimestamp.IsZero() &&
+		!controllerutil.ContainsFinalizer(source, consts.DeleteInstrumentationConfigFinalizer) {
+		controllerutil.AddFinalizer(source, consts.DeleteInstrumentationConfigFinalizer)
+	}
+	if v1alpha1.IsWorkloadExcludedSource(source) &&
+		source.DeletionTimestamp.IsZero() &&
+		!controllerutil.ContainsFinalizer(source, consts.StartLangDetectionFinalizer) {
+		controllerutil.AddFinalizer(source, consts.StartLangDetectionFinalizer)
+	}
```
RonFed, line 74: Do we want to enforce that the excluded label can only have true value? Do we support having excluded as false?

instrumentor/sources_webhooks.go
```diff
@@ -0,0 +1,263 @@
...
+	if len(allErrs) == 0 {
+		return nil, nil
+	}
+
+	return nil, apierrors.NewInvalid(
+		schema.GroupKind{Group: "odigos.io", Kind: "Source"},
+		new.Name, allErrs)
+}
+
+func (s *SourcesValidator) ValidateDelete(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
```
RonFed, line 176: Is this used?

frontend/services/sources.go
```diff
@@ -278,3 +265,84 @@ func updateAnnotations(annotations map[string]string, reportedName string) map[s
...
+
+	if err != nil {
+		return nil, err
+	}
+	if len(list.Items) == 0 {
+		return nil, fmt.Errorf(`source "%s" not found`, workloadName)
+	}
+	if len(list.Items) > 1 {
+		return nil, fmt.Errorf(`expected to get 1 source "%s", got %d`, workloadName, len(list.Items))
+	}
```
RonFed, line 286: This is a duplication of the function in `source_types.go` - I think it is better to re-use

frontend/services/sources.go
```diff
@@ -278,3 +265,84 @@ func updateAnnotations(annotations map[string]string, reportedName string) map[s
...
+
+	crdName := list.Items[0].Name
+	crd, err := kube.DefaultClient.OdigosClient.Sources(nsName).Get(ctx, crdName, metav1.GetOptions{})
```
RonFed, line 289: Why do we need to do a `Get` here if we did a `List` ins the function start? we have the object in `Items[0]`

cli/cmd/resources/instrumentor.go
```diff
@@ -189,6 +191,16 @@ func NewInstrumentorClusterRole() *rbacv1.ClusterRole {
 				Resources: []string{"instrumentationconfigs"},
 				Verbs:     []string{"create", "delete", "get", "list", "patch", "update", "watch"},
 			},
+			{
+				APIGroups: []string{"odigos.io"},
+				Resources: []string{"sources"},
+				Verbs:     []string{"create", "delete", "get", "list", "patch", "update", "watch"},
```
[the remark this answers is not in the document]
damemi: We'll need `create` to continue with the shorthand support for the current label

cli/cmd/uninstall.go
```diff
@@ -387,6 +387,20 @@ func uninstallCRDs(ctx context.Context, cmd *cobra.Command, client *kube.Client,
 		return err
 	}
 
+	// Clear finalizers from Source objects so they can be uninstalled
```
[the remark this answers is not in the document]
damemi: this is required to completely uninstall the CRDs, I think it makes sense to keep it in this function to signal that

api/odigos/v1alpha1/source_types.go
```diff
@@ -62,6 +71,68 @@ type SourceList struct {
...
+
+type WorkloadSources struct {
+	Workload  *Source
+	Namespace *Source
+}
+
+// GetWorkloadSources returns a WorkloadSources listing the Workload and Namespace Source
+// that currently apply to the given object. In theory, this should only ever return at most
+// 1 Namespace and/or 1 Workload Source for an object. If more are found, an error is returned.
+func GetWorkloadSources(ctx context.Context, kubeClient client.Client, obj client.Object) (*WorkloadSources, error) {
```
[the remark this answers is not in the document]
damemi: sgtm

cli/cmd/resources/instrumentor.go
```diff
@@ -242,7 +254,132 @@ func NewInstrumentorService(ns string) *corev1.Service {
 	}
 }
 
-func NewMutatingWebhookConfiguration(ns string, caBundle []byte) *admissionregistrationv1.MutatingWebhookConfiguration {
+func NewSourceValidatingWebhookConfiguration(ns string, caBundle []byte) *admissionregistrationv1.ValidatingWebhookConfiguration {
```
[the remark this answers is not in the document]
damemi: 👍

instrumentor/controllers/deleteinstrumentationconfig/common.go
```diff
@@ -38,24 +49,25 @@ func reconcileWorkloadObject(ctx context.Context, kubeClient client.Client, work
...
 		},
 	})
 
 	if instConfigErr != nil {
 		return client.IgnoreNotFound(instConfigErr)
 	}
 
-	logger := log.FromContext(ctx)
-	logger.V(1).Info("instrumented application deleted", "namespace", ns, "name", name, "kind", kind)
+	logger.V(1).Info("instrumentationconfig deleted", "namespace", ns, "name", name, "kind", kind)
```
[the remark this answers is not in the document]
damemi: 👍

instrumentor/sources_webhooks.go
damemi: Good point. Updated the validating webhook to enforce consistent casing on admission

instrumentor/sources_webhooks.go
damemi: Yeah I figured it doesn't hurt to double check that they are also valid

instrumentor/sources_webhooks.go
damemi: updated

common/consts/consts.go
```diff
@@ -15,6 +15,10 @@ const (
 	OdigosInstrumentationLabel   = "odigos-instrumentation"
 	InstrumentationEnabled       = "enabled"
 	InstrumentationDisabled      = "disabled"
+	OdigosNamespaceAnnotation    = "odigos.io/workload-namespace"
+	OdigosWorkloadKindAnnotation = "odigos.io/workload-kind"
+	OdigosWorkloadNameAnnotation = "odigos.io/workload-name"
```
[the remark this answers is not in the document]
blumamir: The thing with common is that it's also being used by the vm-agent, so ideally we should place anything specific to k8s (like consts that says "namespace" and "kind" in `k8sutils`)

cli/cmd/resources/ui.go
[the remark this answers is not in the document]
damemi: > I was under the impression that we only store source objects in odigos-system namespace.

There are a few reasons not to do this:

* Source is a user-facing object; any application team should be able to create a Source for their apps. `odigos-system` is a core namespace that only admins should have access to.
* Putting all Sources in one namespace (or cluster-scoping) means all tenants that can create a Source will have access to all other tenants' Source configurations.

Both of these hinge on Odigos being usable by both end-user teams and cluster admins. 

> in this case, why do we need to store the workload namespace label in the object? isn't it redundant and adds extra complexity?

We don't need it, the requirement is inherited by the fact that [workload.PodWorkload is required in the spec](https://github.com/damemi/odigos/blob/46d917e356eecf2eb95aa96d645e65bd56615f17/api/odigos/v1alpha1/source_types.go#L52-L53), which makes all of its fields required too. I decided to leave it for now, but we can relax the requirement.

That said, it is easier (from an API perspective) to eventually make something optional than to turn around and make it required.

> What happens if a source is created in namespace foo and contains other namespace bar in the spec?

This is a good point and comes back to the multitenancy issue. In the code right now, I think it would just work. That's a bug -- users shouldn't be able to create a Source in their own namespace that affects workloads in other namespaces.

Taking all this into consideration, I think the lightest solution right now is to ensure that `metadata.namespace == speec.workload.namespace` in Source validation.  I'd like to re-evaluate whether `spec.workload.namespace` needs to be required later on, when not so much stuff is moving around. What do you think?

cli/cmd/resources/instrumentor.go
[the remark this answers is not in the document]
blumamir: right!

cli/cmd/uninstall.go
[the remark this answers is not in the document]
damemi: yup, that sounds good to me

common/consts/consts.go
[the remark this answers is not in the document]
damemi: Ah, I didn't realize that's what "common" was referring to.

Either way, besides the excluded label these are all used by frontend, so that will be on @BenElferink to update

instrumentor/controllers/deleteinstrumentationconfig/namespace_controller.go
damemi: good idea, done

tests/e2e/source-webhooks/chainsaw-test.yaml
```diff
@@ -0,0 +1,168 @@
...
+              - check:
+                  ($error != null): true
+    - name: "[Defaulting] - Normal Source has deleteinstrumentationconfig finalizer added"
+      try:
+        - apply:
+            file: 04-deleteinstrumentationconfig-finalizer.yaml
+        - assert:
+            timeout: 2m
+            file: 04-assert-deleteinstrumentationconfig-finalizer.yaml
+    - name: "[Defaulting] - Finalizer does not block normal Source from being deleted"
```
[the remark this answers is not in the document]
damemi: Yes this is a check for deadlocks, even though it doesn't return "immediately" it should be very quick, within the timeout for this check

api/odigos/v1alpha1/source_types.go
```diff
@@ -62,6 +71,68 @@ type SourceList struct {
...
+	}
+
+	namespaceSourceList := SourceList{}
+	namespaceSelector := labels.SelectorFromSet(labels.Set{
+		consts.WorkloadNameLabel:      obj.GetNamespace(),
+		consts.WorkloadNamespaceLabel: obj.GetNamespace(),
+		consts.WorkloadKindLabel:      "Namespace",
+	})
+	err = kubeClient.List(ctx, &namespaceSourceList, &client.ListOptions{LabelSelector: namespaceSelector})
+	if err != nil {
```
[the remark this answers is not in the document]
damemi: it will return an empty list if none are found. but it could still return an error in other cases (like network or api server issues)

instrumentor/controllers/startlangdetection/manager.go
```diff
@@ -74,5 +76,18 @@ func SetupWithManager(mgr ctrl.Manager) error {
 		return err
 	}
 
+	err = builder.
+		ControllerManagedBy(mgr).
+		Named("startlangdetection-source").
+		For(&v1alpha1.Source{}).
+		WithEventFilter(predicate.Or(&odigospredicate.CreationPredicate{}, &odigospredicate.OnlyUpdatesPredicate{})).
```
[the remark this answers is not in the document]
damemi: When a Source is deleted, it is first updated to add a DeletionTimestamp. We have the update predicate here to catch that Update event, and do any necessary uninstrumentation. We then remove the finalizer, and the k8s garbage collector fully deletes the Source.

The finalizer pauses deletion so that we can read the Source before it's deleted, then once we are done with our logic we say okay, we're done with it, go ahead and delete it now by removing the finalizer.

So in theory, a specific predicate could check for a DeletionTimestamp. But I think that's unnecessary, is it's one of the first things we check in the reconcile function anyway. So every event still gets processed the same amount.

instrumentor/controllers/startlangdetection/workload_controllers.go
damemi: `IsWorkloadInstrumentationEffectiveEnable` will be removed or refactored to essentially be this check when the backend support for labels is removed (next PR)

frontend/services/sources.go
damemi: @BenElferink

instrumentor/controllers/deleteinstrumentationconfig/source_controller.go
damemi: You are referring to the top of the file, where we try to get the Source right?

This is to check here if any of the workload reconcile calls returned NotFound

instrumentor/controllers/startlangdetection/manager.go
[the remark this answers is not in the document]
RonFed: The advantage in the predicate is that we can see the changes which we can see in the reconcile itself. 
My concern here is that: suppose a spec change happens (in the future the spec will expand I assume) - from the reconciler perspective, will it decide to trigger language detection (and possibly overwriting the instrumentation config) because some change happened and there is not deletion timestamp.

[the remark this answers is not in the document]
damemi: Operators should not reconcile based on "what" changed. That is an edge-based design. The state of the cluster should be enough for it to decide if there is any action to take. Events are just a way to tell the operator "hey something changed", and it is the operator's job to compare the desired state with the current state.

That's one of the main goals for the Source CRD: moving away from how we currently have to check if a label went from true->false or vice versa.

> suppose a spec change happens - from the reconciler perspective, will it decide to trigger language detection (and possibly overwriting the instrumentation config) because some change happened and there is not deletion timestamp.

This should not happen, if it does then our reconcile logic needs to be more robust to decide, for example, if language detection should be triggered at all or if the output of that detection should overwrite what already exists. In other words, that's a bug, but it's an API/resource issue, not an event watching issue.

common/consts/consts.go
```diff
@@ -15,6 +15,10 @@ const (
...
+	OdigosWorkloadExcludedLabel  = "odigos.io/workload-excluded"
```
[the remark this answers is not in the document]
damemi: I like it, updated

instrumentor/controllers/startlangdetection/manager.go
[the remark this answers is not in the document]
RonFed: I see.
the missing part for me was `return client.IgnoreAlreadyExists(err)` when we create the instrumentation config - which means we won't overwrite the config.
Thanks for the explanation!

instrumentor/sources_webhooks.go
damemi: based on @blumamir's suggestion, moved the excluded label to a mutable field on the Source object. true and false(empty) are both supported, and a Source can toggle between them.

common/consts/consts.go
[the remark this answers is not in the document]
BenElferink: On it 👍

frontend/services/sources.go
BenElferink: On it 👍

frontend/services/sources.go
BenElferink: I cannot use the `GetSources` function from `source_types.go` because of client-differences, check this error for example:
```
cannot use kube.DefaultClient.OdigosClient (variable of type "github.com/odigos-io/odigos/api/generated/odigos/clientset/versioned/typed/odigos/v1alpha1".OdigosV1alpha1Interface) as "sigs.k8s.io/controller-runtime/pkg/client".Client value in argument to v1alpha1.GetSources: "github.com/odigos-io/odigos/api/generated/odigos/clientset/versioned/typed/odigos/v1alpha1".OdigosV1alpha1Interface does not implement "sigs.k8s.io/controller-runtime/pkg/client".Client (missing method Create)
```

RonFed: I see.
@blumamir  Do you know if we have an example for utils function that is used both by the frontend and by the controllers? I was pretty sure that we have examples for the describe but not sure, and which client interface can fit both use cases.

tests/e2e/workload-lifecycle/02-wait-for-trace.yaml
RonFed, line 1: What's the reason for this change? same question for the other file under `workload-lifecycle`

RonFed: Thanks!

RonFed (approved):
🎉🚀🥇

frontend/services/sources.go
blumamir: We don't have this kind of capability at the moment. where we need to do something with 2 utils one for each client

blumamir (approved):
Great work 🥇

api/odigos/v1alpha1/source_types.go
```diff
@@ -62,6 +73,64 @@ type SourceList struct {
...
+	if len(namespaceSourceList.Items) == 1 {
+		workloadSources.Namespace = &namespaceSourceList.Items[0]
+	}
+
+	return workloadSources, nil
+}
+
+// IsWorkloadExcludedSource returns true if the Source is disabling instrumentation.
+func IsWorkloadExcludedSource(source *Source) bool {
+	return source.Spec.InstrumentationDisabled
```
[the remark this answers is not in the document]
damemi: updated the name

instrumentor/controllers/deleteinstrumentationconfig/source_controller.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func (r *SourceReconciler) syncWorkload(ctx context.Context, source *v1alpha1.Source) error {
+	// This is a Source for a specific workload, not an entire namespace
+	obj := workload.ClientObjectFromWorkloadKind(source.Spec.Workload.Kind)
+	err := r.Client.Get(ctx, types.NamespacedName{Name: source.Spec.Workload.Name, Namespace: source.Spec.Workload.Namespace}, obj)
+	if err != nil {
+		// TODO: Deleted objects should be filtered in the event filter
```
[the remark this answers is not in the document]
damemi: bad copypasta

k8sutils/pkg/utils/object.go
```diff
@@ -0,0 +1,9 @@
+package utils
+
+import "sigs.k8s.io/controller-runtime/pkg/client"
+
+// IsTerminating returns true if a client.Object has a non-zero DeletionTimestamp.
+// Otherwise, it returns false.
+func IsTerminating(obj client.Object) bool {
+	return !obj.GetDeletionTimestamp().IsZero()
```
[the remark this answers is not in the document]
damemi: good catch, updated this to check for nil

PR #2173, opened 2025-01-09
feat: cli remote update token

AvihuHenya:
declaring remote flag in cli pro command for enabling update token via Odigos ui


k8sutils/pkg/consts/consts.go
```diff
@@ -69,4 +69,9 @@ var (
 const (
 	OdigosCloudApiKeySecretKey = "odigos-cloud-api-key"
 	OdigosOnpremTokenSecretKey = "odigos-onprem-token"
+)
+
+const (
+	OdigosUiServiceName = "ui"
+	OdigosUiServicePort = 3000
```
blumamir, line 76: great!

Can you please also update [here](https://github.com/odigos-io/odigos/blob/main/cli/cmd/resources/ui.go)

k8sutils/pkg/pro/common.go
```diff
@@ -0,0 +1,62 @@
...
+	if daemonSet.Spec.Template.Annotations == nil {
+		daemonSet.Spec.Template.Annotations = make(map[string]string)
+	}
+	daemonSet.Spec.Template.Annotations[odigosconsts.RolloutTriggerAnnotation] = time.Now().Format(time.RFC3339)
+
+	_, err = client.AppsV1().DaemonSets(namespace).Update(ctx, daemonSet, metav1.UpdateOptions{})
+	if err != nil {
+		fmt.Printf("Failed to restart Odiglets. Reason: %s\n", err)
+		fmt.Printf("To trigger a restart manually, run the following command:\n")
+		fmt.Printf("kubectl rollout restart daemonset odiglet -n %s\n", daemonSet.Namespace)
```
blumamir, line 59: Don't `fmt.Printf` in  `k8sutils` as it might get called from cli (and seens) but also from ui pod.

Return an error and print in cli based on the error (either remote or local)

blumamir (approved):

PR #2178, opened 2025-01-10
chore(odiglet): remove unused rbac permission in replicasets

blumamir:
It was used in old runtime detection. Now that this is removed, we no longer need replicasets permissions for odiglets and can clean this up


edeNFed (approved):
nice

PR #2192, opened 2025-01-11
Add golang linters

edeNFed:
Enable linters for `common` and `k8sutils` modules. Rest of the modules will be added gradually


blumamir (approved):
can we have a make target to lint the code?

PR #2209, opened 2025-01-13
feat: delete instrumentation instance if exit with healthy state

tamirdavid1:


blumamir (approved):

opampserver/pkg/server/handlers.go
```diff
@@ -164,14 +164,19 @@ func (c *ConnectionHandlers) UpdateInstrumentationInstanceStatus(ctx context.Con
...
+	// 1. In case disconnect with healthy status, we will delete the instrumentationInstnace CR
+	// 2. Otherwise, we will keep the CR in unhealthy state so it can be used for troubleshooting
 	if isAgentDisconnect {
 		if !hasHealth {
 			return fmt.Errorf("missing health in agent disconnect message")
 		}
+		// [1] - agent disconnects with healthy status, delete the instrumentation instance CR
 		if message.Health.Healthy {
-			return fmt.Errorf("agent disconnect message with healthy status")
+			c.logger.Info("Agent disconnected with healthy status, deleting instrumentation instance", "workloadNamespace", connectionInfo.Workload.Namespace, "workloadName", connectionInfo.Workload.Name, "workloadKind", connectionInfo.Workload.Kind)
```
RonFed, line 176: This log can happen a lot for certain applications, right? (i.e python server that runs processes periodically)?

RonFed (approved):

opampserver/pkg/server/handlers.go
tamirdavid1: @RonFed correct, but there is already a log for opamp connect meaning we will have logs for each process connect as well

PR #2210, opened 2025-01-13
chore: reconcile profiles in schedualer

blumamir:
Instead of only calculating them in cli.

This PR makes it so that profiles can be added by helm, ui, gitOps. anything that will end up updating the profiles list in `odigos-config` will cause profiles to update in the cluster and get applied (unlike today where only cli is supported)


scheduler/controllers/odigosconfig/profileutils.go
```diff
@@ -0,0 +1,66 @@
...
+
+		// if this profile has dependencies, add them to the list
+		if p.Dependencies != nil {
+			effectiveProfiles = append(effectiveProfiles, calculateEffectiveProfiles(p.Dependencies, availableProfiles)...)
+		}
+	}
+	return effectiveProfiles
+}
+
+func findProfileNameInAvailableList(profileName common.ProfileName, availableProfiles []profile.Profile) (profile.Profile, bool) {
```
tamirdavid1, line 58: I guess we can make it more generic
getMatchingProfile
and then use it also in 
cli/cmd/profile:110 
cli/cmd/profile:131

scheduler/main.go
```diff
@@ -109,6 +119,13 @@ func main() {
 		os.Exit(1)
 	}
 
+	// create dynamic k8s client to apply profile manifests
```
tamirdavid1, line 122: why we need this dynamic client?

scheduler/controllers/odigosconfig/odigosconfig_controller.go
```diff
@@ -117,45 +133,105 @@ func (r *odigosConfigController) persistEffectiveConfig(ctx context.Context, eff
...
+		annotations = make(map[string]string)
+	}
+	annotations[k8sconsts.OdigosProfileAnnotation] = string(profileName)
+	obj.SetAnnotations(annotations)
+
+	gvk := obj.GroupVersionKind()
+	gvr := schema.GroupVersionResource{
+		Group:    gvk.Group,
+		Version:  gvk.Version,
+		Resource: strings.ToLower(gvk.Kind) + "s", // TODO: this is a hack, might not always work
```
tamirdavid1, line 216: as we discussed i think better to do it with map,
so if someone will add a new resource that it wont work for it it will fail

tamirdavid1 (approved):

PR #2229, opened 2025-01-14
Enhance leader election config

tamirdavid1:


instrumentor/main.go
```diff
@@ -113,6 +114,31 @@ func main() {
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
 		LeaderElectionID:       "201bdfa0.odigos.io",
+		/*
+			Leader Election Parameters:
+
+			LeaseDuration (5s):
+			- Maximum time a pod can remain the leader after its last successful renewal.
```
RonFed, line 121: I'm not sure what that means,
if all goes well an instrumetor pod should remain the leader for all its lifetime. no? (not max of 5 seconds)

RonFed (approved):

blumamir (approved):

PR #2230, opened 2025-01-14
feat: add code attributes configuration to instrumentation rules

blumamir:


api/odigos/v1alpha1/instrumentationrules/code-attributes.go
```diff
@@ -0,0 +1,24 @@
+package instrumentationrules
+
+// +kubebuilder:object:generate=true
+// +kubebuilder:deepcopy-gen=true
+type CodeAttributes struct {
+
+	// Should record the `code.column` attribute.
+	Column *bool `json:"column,omitempty"`
```
RonFed, line 8: Maybe worth to add a general comment here about why we use pointer and the meaning of nil

api/odigos/v1alpha1/instrumentationrule_type.go
```diff
@@ -65,6 +65,9 @@ type InstrumentationRuleSpec struct {
 	// instrumentation libraries will be ignored if set.
 	// the rule will be used only for languages which are specified, and ignored otherwise.
 	OtelSdks *instrumentationrules.OtelSdks `json:"otelSdks,omitempty"`
+
+	// Configure which code attributes should be recorded as span attributes.
```
RonFed, line 69: add to the comment what it means if nil

RonFed (approved):

PR #2231, opened 2025-01-14
Migrate `odigos-instrumentation` label to Source CRD

damemi:
This adds a migration for the current `odigos-instrumentation` label to Source CRDs.

If a workload or namespace has `odigos-instrumentation: enabled`, the startlangdetection controller will automatically create a Source for it (if one doesn't already exist) and log a deprecation warning.

If a workload or namespace has `odigos-instrumentation: disabled`, the deleteinstrumentationconfig controller will automatically create a Source with `disableInstrumentation: true`.

In both cases, if a Source already exists, the controller will update the existing Source. This means that the label can be toggled back and forth.

However, deleting the label will not inherently un-instrument a workload that doesn't have inheriting Namespace instrumentation. We can't tell if the label was deleted without checking the change during the event, which is bad to do. This is the only drop in feature parity from the current label.


k8sutils/pkg/source/source.go
```diff
@@ -0,0 +1,151 @@
...
+			source = sources.Workload
+		}
+	}
+
+	create := false
+	if source == nil {
+		create = true
+		source = &v1alpha1.Source{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: fmt.Sprintf("source-%s-%s-", strings.ToLower(string(kind)), strings.ToLower(obj.GetName())),
```
RonFed, line 93: `CalculateWorkloadRuntimeObjectName` looks identical to this

k8sutils/pkg/source/source.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+	}
+	source.Spec.DisableInstrumentation = disableInstrumentation
+
+	if create {
+		log.FromContext(ctx).Info("creating source", "source", source.Spec)
+		return client.IgnoreAlreadyExists(k8sClient.Create(ctx, source))
+	}
+	log.FromContext(ctx).Info("updating source", "source", source.Spec)
+	return k8sClient.Update(ctx, source)
```
RonFed, line 112: Do you think we should check with `K8SUpdateErrorHandler` after each call to this? to handle conflict errors

k8sutils/pkg/source/source.go
```diff
@@ -0,0 +1,151 @@
...
+		return true, nil
+	}
+
+	return false, nil
+}
+
+// IsActiveSource returns true if the Source enables instrumentation.
+// Specifically, the Source must be either:
+// 1) Inclusive AND NOT terminating, or
+// 2) Exclusive AND terminating
```
RonFed, line 50: Can you explain the motivation for considering `Exclusive AND terminating` as active Source?
Assuming no namespace source:
*  If I delete a Source with `excluded=true` I would not expect my workload to get instrumented right away (only after a source without the exclusion is created)

On the other hand
* If I had an inclusion source and I delete it, It's true that it is expected to un-instrument.

instrumentor/controllers/deleteinstrumentationconfig/source_controller.go
```diff
@@ -25,20 +25,51 @@ import (
...
 	client.Client
 	Scheme *runtime.Scheme
 }
 
+// isDeleteInstrumentationConfigSource returns true if the Source object is relevant to starting language detection.
+// This means that the Source must be either:
+// 1) A normal (non-excluding) Source AND terminating, or
+// 2) An excluding Source AND NOT terminating
+// In either of these cases, we want to check if workloads should start to be instrumented.
+var SourcePredicates = predicate.Funcs{
```
RonFed, line 51: The comment does not match the name.
Don't we check in the reconciler for `IsActiveSource`? since we are not looking at the diff here, maybe worth just removing this predicate.

api/odigos/v1alpha1/source_types.go
```diff
@@ -87,14 +87,19 @@ func GetSources(ctx context.Context, kubeClient client.Client, obj client.Object
 	var err error
 	workloadSources := &WorkloadSources{}
 
-	if obj.GetObjectKind().GroupVersionKind().Kind != "Namespace" {
+	namespace := obj.GetNamespace()
+	if len(namespace) == 0 && obj.GetObjectKind().GroupVersionKind().Kind == string(workload.WorkloadKindNamespace) {
```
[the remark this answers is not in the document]
damemi: yes Namespace objects don't have a namespace

instrumentor/controllers/deleteinstrumentationconfig/source_controller.go
damemi: In this case, a predicate makes sense to filter out checking irrelevant Sources in the reconciler. We don't need to look at the diff to tell whether this Source is relevant. If anything I think we can remove the same check in the Reconcile function

Otherwise, both startlangdetection and deleteinstrumentationconfig simultaneously try to reconcile() every Source. It's a no-op on the irrelevant controller because of the check, but moving the check to a predicate shifts it further up the chain

k8sutils/pkg/source/source.go
RonFed: Yeah, that case is expected.
Does that mean that every time we call this function we need to make sure there is a namespace source?

k8sutils/pkg/source/source.go
damemi: Yeah, double check the change in https://github.com/odigos-io/odigos/pull/2231/commits/47dcd81f7c265708e0413f122e05702cc7638f7b

Rather than update this function to return a ctrl.Result (for the requeue), I decided to just update the return call in the controllers wherever this is used. let me know what you think

RonFed: Yeah, that looks good to me.

k8sutils/pkg/source/source.go
damemi: Not always, this function is just a helper to determine if you're looking at a source that should be used for instrumentation or un-instrumentation. It's easier to read than `terminating == excluding` / `terminating != excluding` everywhere.

For example, the deleteinstrumentationconfig controller uses this check on single workloads regardless of if there is namespace instrumentation. When a `!ActiveSource` is created (disable=true), that overrides namespace instrumentation for the workload, so we don't even check for it there.

Similarly, a namespace can technically be excluded too, with a workload that is explicitly instrumented within it (weird but we [have tests for it](https://github.com/odigos-io/odigos/blob/3cbb11747682f88fa24a7fa08c81ae3f56d59467/instrumentor/controllers/deleteinstrumentationconfig/namespace_controller_test.go#L93)). In that case, the startlangdetection controller checks if `IsActiveSource` and instruments.

RonFed: I see. 
My point is that calling this function might return an unexpected result (the case I describe above when there is no ns Source at all and an exclusion Source is being deleted) - As I understand it, we will decide to instrument that workload (in startlangdetection) based on this logic.
Maybe I'm missing something

damemi: This function doesn't make any decisions about instrumenting -- it's just a shorthand helper.

In the case you describe, an Exclusion Source being deleted will pass the startlangdetection predicate. The controller will then use `IsObjectInstrumentedBySource` (added in this PR) to decide if it should actually go through with instrumentation, based on checking the hierarchy of workload->exclusion->ns source.

This function is more lightweight for a gating check (the predicate), because it doesn't need to call the k8s api and doesn't rely on any information about the workload. So, some no-ops will still get through, but that is expected in an operator.

damemi: Ah that's a good point, I don't think `IsObjectInstrumentedBySource` should be relying on `IsActiveSource` in this case. I will update it

damemi: I confirmed this with an e2e test as an edge case that wasn't covered. Updated the check in https://github.com/odigos-io/odigos/pull/2231/commits/8a96dd9cef2d7f0b1473d63df28477ac553c4275

I think `IsActiveSource` is still useful as a lightweight predicate for these controllers like I mentioned, so I removed it from anywhere else that was using it. It could probably have a better name so I'm open to suggestions

tests/e2e/source/chainsaw-test.yaml
```diff
@@ -497,3 +497,39 @@ spec:
...
+        - script:
+            timeout: 60s
+            content: kubectl delete sources/membership-excluded
+        - assert:
+            timeout: 2m
+            file: 11-workloads.yaml
+        - script:
+            content: kubectl get instrumentationconfigs/deployment-membership
+            check:
+              ($error != null): true
```
RonFed, line 535: Do you think we should check this in a loop, to make sure the `instrumentationconfig` is not created?
Since it might take some time for the instrumentation config to get created, maybe worth to make sure it is not created

instrumentor/controllers/deleteinstrumentationconfig/source_controller.go
```diff
@@ -48,37 +79,26 @@ func (r *SourceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
-			controllerutil.RemoveFinalizer(source, consts.DeleteInstrumentationConfigFinalizer)
-			if err := r.Update(ctx, source); err != nil {
-				return k8sutils.K8SUpdateErrorHandler(err)
-			}
+	if !v1alpha1.IsExcludedSource(source) &&
+		k8sutils.IsTerminating(source) &&
+		controllerutil.ContainsFinalizer(source, consts.DeleteInstrumentationConfigFinalizer) {
+		controllerutil.RemoveFinalizer(source, consts.DeleteInstrumentationConfigFinalizer)
+		if err := r.Update(ctx, source); err != nil {
+			return k8sutils.K8SUpdateErrorHandler(err)
```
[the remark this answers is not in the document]
damemi: Yeah, good point. In that case, if either of those return an error, the Source will be stuck in terminating. I think that's what we want, since the instrumentation may still be active since the sync functions failed

instrumentor/controllers/startlangdetection/workload_predicate.go
```diff
@@ -48,48 +47,25 @@ func (i *WorkloadEnabledPredicate) Update(e event.UpdateEvent) bool {
...
-	oldEnabled := workload.IsObjectLabeledForInstrumentation(e.ObjectOld)
-	newEnabled := workload.IsObjectLabeledForInstrumentation(e.ObjectNew)
-	becameEnabled := !oldEnabled && newEnabled
-
 	newReplicas := wNew.AvailableReplicas()
 	oldReplicas := wOld.AvailableReplicas()
 
-	// 1. workload became enabled and has available (running) replicas
-	if becameEnabled && newReplicas > 0 {
+	// 1. workload has available (running) replicas
```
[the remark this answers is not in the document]
damemi: I was originally leaning away from calling `IsObjectInstrumentedBySource` because it means an extra API call before we even get to the reconciler. The concern is API server traffic. And since that's one of the first API calls we make inside the reconciler anyway, that tradeoff seemed like a wash.

But, for non-instrumented workloads, that would actually mean 3 unnecessary API calls: one to `Reconcile()`, one to `getWorkloadObject()`, and then `IsObjectInstrumentedBySource`. In contrast, we already have the object in the Predicate, so we could filter out non-instrumented workloads with just 1 API call, like you're suggesting.

However, that means an extra unnecessary API call for every instrumented workload. And instrumented workloads are really what we care about. In fact, we hope that instrumented workloads should largely outnumber non-instrumented workloads for Odigos users :)

So I think the efficiency gain for non-instrumented workloads isn't worth the efficiency loss for workloads we actually care about.

k8sutils/pkg/source/source.go
damemi: Updated the name and the comment about how this should be used

tests/e2e/source/chainsaw-test.yaml
damemi: I don't think there's a way for us to efficiently watch for "this never happens" in a test without just setting an arbitrary timer or running an async watchdog. Unfortunately that means this is probably the best we can do, but the generation checks in later tests will mismatch if something goes wrong within the timespan of the rest of the test.

instrumentor/controllers/startlangdetection/workload_predicate.go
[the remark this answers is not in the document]
RonFed: I agree with this analysis.
However, I have seen cases of instrumenting single or a few workloads in a cluster for a few reasons. It might be a new client just starting and doesn't want to commit to instrumenting all the applications, or maybe most of the apps are not relevant/ written in unsupported languages.
So, in those cases (which I think are not rare) - we will have some regression in terms of performance. I guess it also depends on how often deployments are updated in the cluster. If some automatic tools are involved I think that can also increase the amount of times the reconciler will trigger.
Not saying this should block or anything, just worth to consider.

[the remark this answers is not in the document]
damemi: It's definitely a good consideration. I also think that the "extra" API call should usually be using the controller's cache, while the event getting passed to `Reconcile()` is actual traffic every time. So I think my point above isn't always valid anyway.

I added a basic version of the previous check in https://github.com/odigos-io/odigos/pull/2231/commits/b6ad6ac91e783c8e67c770ac077537b61e33cb48, ptal. Some notes:

* We don't have a context to use in the requests, so the `IsObjectInstrumentedBySource` call needs to just use `context.Background`.
* We can't retry on an error, so we need to either ignore the error and let the object through (where the controller can retry it -- say a client hiccup) or just block the request entirely. I think the former is safer so we don't miss relevant objects.
* While we're migrating, we still need to check the label here. That's fine, but I simplified the check in `Update` to just look at if the new object is instrumented instead of comparing if the old object was not instrumented. There's no equivalent way with Sources to know if the old object didn't have a Source, so I think that's fine.

Let me know what you think

instrumentor/controllers/startlangdetection/workload_predicate.go
```diff
@@ -1,23 +1,35 @@
...
 	if err != nil {
 		return false
 	}
+
+	instrumented, _ := sourceutils.IsObjectInstrumentedBySource(context.Background(), i.Client, e.Object)
+	if !instrumented || !workload.IsObjectLabeledForInstrumentation(e.Object) {
+		return false
+	}
+
 	return w.AvailableReplicas() > 0
```
RonFed, line 35: nit: we can move this check before the API calls, to potentially save calls.

instrumentor/controllers/startlangdetection/workload_predicate.go
```diff
@@ -43,6 +55,11 @@ func (i *WorkloadAvailablePredicate) Update(e event.UpdateEvent) bool {
 		return false
 	}
 
+	instrumented, _ := sourceutils.IsObjectInstrumentedBySource(context.Background(), i.Client, e.ObjectNew)
+	if !instrumented || !workload.IsObjectLabeledForInstrumentation(e.ObjectNew) {
+		return false
+	}
+
```
RonFed: same nit as above: maybe better to do the replicas check before this.

instrumentor/controllers/startlangdetection/namespace_controller.go
```diff
@@ -3,17 +3,15 @@ package startlangdetection
...
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	sourceutils "github.com/odigos-io/odigos/k8sutils/pkg/source"
+	k8sutils "github.com/odigos-io/odigos/k8sutils/pkg/utils"
+	"github.com/odigos-io/odigos/k8sutils/pkg/workload"
 )
 
 type NamespacesReconciler struct {
```
RonFed, line 17: nit: maybe worth marking this as deprecated reconciler
just making sure, we are going to remove it once we fully remove support for labels, right?

RonFed (approved):
Left a few nits, this looks great 🥇

instrumentor/controllers/startlangdetection/namespace_controller.go
damemi: No, we should still have workload/namespace reconcilers. Since Sources are detached from the workloads, a workload can be redeployed or come up after the Source (this is by design). If that happens, we should trigger on the event for that workload too. Maybe we could deprecate the Namespace reconciler, since Sources need to exist in a namespace. But if we ever support cross-namespace Sources we'll have similar uses for it. It's also just good to have in case we miss a Source event

instrumentor/controllers/startlangdetection/workload_predicate.go
```diff
@@ -1,32 +1,41 @@
...
-func (i *WorkloadEnabledPredicate) Create(e event.CreateEvent) bool {
-	enabled := workload.IsObjectLabeledForInstrumentation(e.Object)
+func (i *WorkloadAvailablePredicate) Create(e event.CreateEvent) bool {
 	w, err := workload.ObjectToWorkload(e.Object)
 	if err != nil {
 		return false
 	}
-	return enabled && w.AvailableReplicas() > 0
+
+	instrumented, err := sourceutils.IsObjectInstrumentedBySource(context.Background(), i.Client, e.Object)
```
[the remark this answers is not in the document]
damemi: I agree with this, if this ends up putting too much load then we can add back in the instrumentedBySource check (cc @RonFed), but I think limiting it to just Create events will be a good compromise.

k8sutils/pkg/source/source.go
```diff
@@ -0,0 +1,151 @@
...
+	k8sutils "github.com/odigos-io/odigos/k8sutils/pkg/utils"
+	"github.com/odigos-io/odigos/k8sutils/pkg/workload"
+)
+
+// IsObjectInstrumentedBySource returns true if the given object has an active, non-excluding Source.
+// 1) Is the object actively included by a workload Source: true
+// 2) Is the object active excluded by a workload Source (overrides namespace instrumentation): false
+// 3) Is the object actively included by a namespace Source: true
+// 4) False
+func IsObjectInstrumentedBySource(ctx context.Context, k8sClient client.Client, obj client.Object) (bool, error) {
```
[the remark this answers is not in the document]
damemi: This is also used to check if a Namespace is instrumented, so I don't think we want to call this "Workload"

k8sutils/pkg/source/source.go
```diff
@@ -0,0 +1,151 @@
...
+// Specifically, the Source must be either:
+// 1) Inclusive AND NOT terminating, or
+// 2) Exclusive AND terminating
+// This function alone should not be used to determine any instrumentation changes.
+func IsSourceRelevant(source *v1alpha1.Source) bool {
+	return v1alpha1.IsExcludedSource(source) == k8sutils.IsTerminating(source)
+}
+
+// CreateOrUpdateSourceForObject creates a Source for an object if one does not exist and
+// applies a label to the object referencing the new Source.
```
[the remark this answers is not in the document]
damemi: It's used in the Migrate functions below

instrumentor/controllers/startlangdetection/namespace_controller.go
```diff
@@ -29,67 +27,18 @@ func (n *NamespacesReconciler) Reconcile(ctx context.Context, request ctrl.Reque
...
-				logger.Error(err, "error requesting runtime details from odiglets", "name", ds.Name, "namespace", ds.Namespace)
-			}
-		}
+	if !enabled {
+		return ctrl.Result{}, nil
 	}
 
-	return ctrl.Result{}, nil
+	logger.V(0).Info("Namespace enabled for instrumentation, recalculating runtime details of relevant workloads")
+	return ctrl.Result{}, syncNamespaceWorkloads(ctx, n.Client, n.Scheme, ns.GetName())
```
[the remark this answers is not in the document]
damemi: Right now it's not possible, since Sources are namespaced and we require them to be in the same namespace they refer to. But if we add cross-namespace Sources, it's a possibility

k8sutils/pkg/source/source.go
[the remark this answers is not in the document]
blumamir: I actually meant the comment "applies a label to the object referencing the new Source."

Do we apply it somewhere?

k8sutils/pkg/source/source.go
```diff
@@ -0,0 +1,151 @@
...
+		log.FromContext(ctx).Info("creating source", "source", source.Spec)
+		return client.IgnoreAlreadyExists(k8sClient.Create(ctx, source))
+	}
+	log.FromContext(ctx).Info("updating source", "source", source.Spec)
+	return k8sClient.Update(ctx, source)
+}
+
+// MigrateInstrumentationLabelToSource checks if an object is enabled by the legacy label.
+// If so, it will create (or update) a Source for the object.
+func MigrateInstrumentationLabelToSource(ctx context.Context, k8sClient client.Client, obj client.Object, kind workload.WorkloadKind) error {
```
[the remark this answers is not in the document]
damemi: moved to labelmigration controller

instrumentor/controllers/startlangdetection/source_controller.go
```diff
@@ -23,6 +26,33 @@ type SourceReconciler struct {
...
 }
 
+// StartLangDetectionSourcePredicate returns true if the Source object is relevant to starting language detection.
+// This means that the Source must be either:
+// 1) A normal (non-excluding) Source AND NOT terminating, or
+// 2) An excluding Source AND terminating
+// In either of these cases, we want to check if workloads should start to be instrumented.
+var StartLangDetectionSourcePredicate = predicate.Funcs{
+	UpdateFunc: func(e event.UpdateEvent) bool {
+		source := e.ObjectNew.(*v1alpha1.Source)
```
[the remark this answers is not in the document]
damemi: k8s controllers shouldn't rely on parsing transitions, because if you miss that event you might not ever recover.

instrumentor/controllers/startlangdetection/namespace_controller.go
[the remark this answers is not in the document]
damemi: I don't think we need to be worried about namespace events

k8sutils/pkg/workload/workloadkinds.go
```diff
@@ -19,6 +19,7 @@ const (
 	WorkloadKindDeployment  WorkloadKind = "Deployment"
 	WorkloadKindDaemonSet   WorkloadKind = "DaemonSet"
 	WorkloadKindStatefulSet WorkloadKind = "StatefulSet"
+	WorkloadKindNamespace   WorkloadKind = "Namespace"
```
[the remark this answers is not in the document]
damemi: I didn't like the name either. But we are checking if kind==Namespace in so many spots, almost always along the same paths that we check workload kinds, that I feel like we should rethink some of the naming here. Maybe not WorkloadKind, maybe SourceKind or something.

PR #2238, opened 2025-01-15
Add Source CRD docs

damemi:
Some things to follow up:

* We still mention InstrumentedApplication in some sections
* Odigos describe cli still lists `odigos-instrumentation`, this should be updated when the command is updated


BenElferink (approved):
🏆

RonFed (approved):

blumamir (approved):

PR #2248, opened 2025-01-16
Generate CRD API docs

damemi:
This adds a `make docgen` that uses [genref](https://github.com/kubernetes-sigs/reference-docs/tree/master/genref) to generate API docs based on our CRD types.

This also updates some API type comments that broke the generator by appearing to be html

Sample screenshot:
Screenshot 2025-01-16 at 1 26 10 PM


BenElferink (approved):
Love this! 🙏

api/markdown/type.tpl
```diff
@@ -0,0 +1,37 @@
...
+{{ end }}
+
+{{- with .References }}
+**Appears in:**
+{{ range . }}
+{{ if or .Referenced .IsExported -}}
+- [{{ .DisplayName }}]({{ .Link }})
+{{ end -}}
+{{- end -}}
+{{- end }}
```
BenElferink, line 16: This should render conditionally `**Appears in:**`:

```suggestion
{{- $hasReferences := false -}}
{{- with .References }}
{{- range . }}
  {{- if or .Referenced .IsExported -}}
    {{- $hasReferences = true -}}
  {{- end -}}
{{- end -}}
{{- if $hasReferences }}
**Appears in:**
{{- range . }}
  {{- if or .Referenced .IsExported -}}
  - [{{ .DisplayName }}]({{ .Link }})
  {{- end -}}
{{- end -}}
{{- end -}}
```

PR #2281, opened 2025-01-20
chore: move k8s consts from k8sutils module to api module

blumamir:
There is currently a dependency of `api` module on `k8sutils` and also `k8sutils` depends on `api`.

This sometimes creates cyclic import and is a bad practice.
Current PR moves the consts from `k8sutils` to `api` module, as the first step towards breaking this cycle.

The end result will be having the `k8sutils` depend on `api`, but not the other way around.

This PR simply moves things between modules, no renames, no changes (even though there are some things we can address to improve). wanted to not introduce any changes to this PR that touch so many files all over


BenElferink (approved):

damemi (approved):

api/k8sconsts/consts.go
```diff
@@ -0,0 +1,26 @@
...
+
+const (
+	OdigosUiServiceName = "ui"
+	OdigosUiServicePort = 3000
+)
+
+var (
+	// MinK8SVersionForInstallation is the minimum Kubernetes version required for Odigos installation
+	// this value must be in sync with the one defined in the kubeVersion field in Chart.yaml
+	MinK8SVersionForInstallation = version.MustParse("v1.20.15-0")
```
RonFed, line 20: This should be in the CLI

api/k8sconsts/consts.go
```diff
@@ -0,0 +1,26 @@
...
+
+const (
+	OdigletPprofEndpointPort    int32 = 6060
+	CollectorsPprofEndpointPort int32 = 1777
+)
+
+const (
+	OdigosUiServiceName = "ui"
+	OdigosUiServicePort = 3000
+)
```
RonFed, line 15: only relevant to one file in the CLI

RonFed (approved):

api/k8sconsts/consts.go
blumamir: Not in scope for this PR, it was in k8sutils to being with

api/k8sconsts/consts.go
blumamir: Not in scope for this PR, it was in k8sutils to being with

PR #2289, opened 2025-01-21
fix: prevent deleting user instrumentation rules on profile reconcile

blumamir:


BenElferink (approved):

api/k8sconsts/resources.go
```diff
@@ -11,6 +11,11 @@ const (
 	// and use the label to identify the resources that needs to be deleted.
 	OdigosProfilesHashLabel = "odigos.io/profiles-hash"
 
+	// this label is used to mark resources that are managed by a profile.
+	// when reconciling profiles, we can use this label to know which profiles needs to be deleted.
+	OdigosProfilesManagedByLabel = "odigos.io/managed-by"
+	OdigosProfilesManagedByValue = "profile"
```
damemi, line 17: would there be any benefit to the value being the name of the profile?

blumamir: I am setting the profile name as an annotation.

This label is used to identify which profiles needs to be deleted when updating the profiles list (removed or changed in code with new version)

PR #2292, opened 2025-01-21
cli: support custom scheduler, frontend, and collector images

damemi:
Working on openshift certification, I wanted to test custom images (based on the red hat base) for these components, but didn't see any way in the cli to set them. It follows the pattern we already have for custom odiglet, autoscaler, and instrumentor images.


RonFed:
@damemi Is this still relevant?

damemi:
@RonFed no, after we discussed I think we will just remove these flags and move that option to a makefile target (since they are only used for development anyway)

PR #2298, opened 2025-01-23
Add reported name field to Source, deprecate the odigos.io/reported-name annotation

RonFed:


api/odigos/v1alpha1/source_types.go
```diff
@@ -53,6 +53,10 @@ type SourceSpec struct {
 	// DisableInstrumentation excludes this workload from auto-instrumentation.
 	// +kubebuilder:validation:Optional
 	DisableInstrumentation bool `json:"disableInstrumentation,omitempty"`
+	// ReportedName determines "service.name" resource attribute which will be reported by the instrumentations of this source.
+	// If not set, the workload name will be used.
+	// It is not valid for namespace sources.
+	ReportedName string `json:"reportedName,omitempty"`
```
damemi, line 59: needs `// +kubebuilder:validation:Optional` (generated docs show it as required)

instrumentor/controllers/instrumentationconfig/source_controller.go
```diff
@@ -0,0 +1,64 @@
...
+	logger.Info("Reconciling Source object", "name", req.Name, "namespace", req.Namespace)
+	source := &odigosv1alpha1.Source{}
+	err := r.Get(ctx, req.NamespacedName, source)
+	if err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	if source.Spec.Workload.Kind == workload.WorkloadKindNamespace {
+		// Namespace Source does not contain configuration for a specific workload
+		return ctrl.Result{}, nil
```
damemi, line 35: It should be documented in the api field that this is only valid for workload Sources. A validation check in the webhook would be good too

instrumentor/controllers/instrumentationconfig/manager.go
```diff
@@ -35,39 +34,12 @@ func SetupWithManager(mgr ctrl.Manager) error {
...
-	// Watch for DaemonSet changes using DaemonSetReconciler
-	if err := builder.
-		ControllerManagedBy(mgr).
-		Named("instrumentor-instrumentationconfig-daemonset").
-		For(&appsv1.DaemonSet{}).
-		WithEventFilter(workloadReportedNameAnnotationChanged{}).
-		Complete(&DaemonSetReconciler{
+		Named("instrumentor-instrumentationconfig-source").
+		For(&odigosv1alpha1.Source{}).
+		Complete(&SourceReconciler{
```
damemi, line 42: @RonFed wdyt about doing this in the startlangdetection Source controller instead of adding a new controller here? It would be nice to converge toward a single controller for Sources where possible

I think startlangdetection will already pick up these changes, it just won't act on them because `requestOdigletToCalculateRuntimeDetails` only calls [k8sClient.Create](https://github.com/odigos-io/odigos/blob/5cac2ff25aab58a6b0ec0e186d9f7e88f2aba8c4/instrumentor/controllers/startlangdetection/workload_controllers.go#L70). But it could call `Update` to patch in the new service name.

Either way, the Source startlangdetection controller should also handle setting service name when a Source is created anyway

instrumentor/controllers/instrumentationconfig/source_controller.go
RonFed: I think these 2 are included in this PR. Let me know what you think.

api/odigos/v1alpha1/source_types.go
RonFed: It looks like we have `Required` for all of our fields and `// +kubebuilder:validation:Optional` is not doing anything.
Adding `// +optional` did the trick, which seems to be what the `genref`  [looks for](https://github.com/kubernetes-sigs/reference-docs/blob/015aaac611407c4fe591bc8700d2c67b7521efca/genref/types.go#L81)

I can do a follow up PR adding it to all our fields. (and maybe remove ``// +kubebuilder:validation:Optional`` which seems to not do anything?)

damemi: ah, weird. I think that's a bug in the genref, they should definitely be parsing `+kubebuilder` tags too

A follow up to add `// +optional` would be good. I think it will only affect the docs, because the controller-gen scripts should be respecting either tag. Meaning, I think our CRDs are accurate

instrumentor/controllers/instrumentationconfig/manager.go
damemi: Ah, I completely agree. Doing it in a separate refactor PR is smart, no need to overcomplicate this change

instrumentor/controllers/workloadmigrations/controllers.go
```diff
@@ -150,6 +153,16 @@ func CreateOrUpdateSourceForObject(ctx context.Context,
...
 	source.Spec.DisableInstrumentation = disableInstrumentation
+	// migrate the reported name from the workload to the Source only if the Source is not a namespace Source
+	// and the workload Source does not already have a reported name.
+	//
+	// If a user have set the annotation on the workload we will use it in the source only if no reported name is set.
+	// If that happens once, and the reported name was taken from the annotation,
+	// any more changes to the annotation will not be reflected in the source.
+	// This is valid, since the annotation is deprecated and we want to encourage users to use Source CR.
+	if kind != workload.WorkloadKindNamespace && source.Spec.ReportedName == "" {
+		source.Spec.ReportedName = reportedName
```
damemi, line 164: could this update the Source even if one is already set? So someone could update the label and the migration updates the Source for them.

instrumentor/sources_webhooks.go
```diff
@@ -260,6 +260,14 @@ func (s *SourcesValidator) validateSourceFields(ctx context.Context, source *v1a
 		))
 	}
 
+	if source.Spec.Workload.Kind == workload.WorkloadKindNamespace && source.Spec.ReportedName != "" {
+		allErrs = append(allErrs, field.Invalid(
+			field.NewPath("spec").Child("reportedName"),
+			source.Spec.ReportedName,
+			"Namespace Source must not have a reportedName, to configure a reported name for a workload edit its Source",
```
damemi, line 267:
```suggestion
			"Reported name is not valid for Namespace sources, only valid for Workload Sources",
```

just a little clearer imo

tests/e2e/cli-upgrade/chainsaw-test.yaml
```diff
@@ -92,8 +92,12 @@ spec:
             file: ../../common/assert/simple-demo-installed.yaml
     - name: Instrument Namespace
       try:
-        - apply:
-            file: ../../common/apply/instrument-default-ns.yaml
+      # Using the deprecated way of instrumenting a namespace,
+      # after v1.0.144, use ../../common/apply/instrument-default-ns.yaml
+        - script:
+            content: kubectl label namespace default odigos-instrumentation=enabled
```
damemi, line 98: I actually like this for now as another way to test the label migration in v1.0.144

tests/e2e/source/03-workloads.yaml
```diff
@@ -1,44 +1,34 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  annotations:
-    deployment.kubernetes.io/revision: "4"
-  generation: 5 # Note new generation after annotating service.name, but not new revision
```
damemi, line 6: huge +1 to not having to keep track of the annotation/generation skew anymore!

instrumentor/controllers/workloadmigrations/controllers.go
RonFed: I was thinking about it and I'm not sure.
My thought is that if the reported name is set using the new way - it should take precedence over the annotation.
Didn't want to overcomplicate it by trying to understand whether the reported name on a Source is the result of a migration or as a result of setting the field.
That means that the annotation will be migrated to the Source once and only if the Source doesn't already include a reported name.
WDYT?

api/odigos/v1alpha1/source_types.go
```diff
@@ -53,6 +53,12 @@ type SourceSpec struct {
 	// DisableInstrumentation excludes this workload from auto-instrumentation.
 	// +kubebuilder:validation:Optional
 	DisableInstrumentation bool `json:"disableInstrumentation,omitempty"`
+	// ReportedName determines the "service.name" resource attribute which will be reported by the instrumentations of this source.
+	// If not set, the workload name will be used.
+	// It is not valid for namespace sources.
+	// +kubebuilder:validation:Optional
+	// +optional
+	ReportedName string `json:"reportedName,omitempty"`
```
blumamir, line 61: suggestion:

```suggestion
	OtelServiceName string `json:"otelServiceName,omitempty"`
```

instrumentor/controllers/workloadmigrations/controllers.go
blumamir, line 164: add a log here for that

instrumentor/controllers/instrumentationconfig/source_controller.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	if odigosv1alpha1.IsDisabledSource(source) {
```
blumamir, line 38: consider removing.

We are already checking if instrumentationconfig is not found below.

If it's found, let's update the instrumentation config

instrumentor/controllers/instrumentationconfig/source_controller.go
```diff
@@ -0,0 +1,64 @@
...
+		// Exclude this workload from instrumentation, no need to update InstrumentationConfig
+		return ctrl.Result{}, nil
+	}
+
+	instConfigName := workload.CalculateWorkloadRuntimeObjectName(source.Spec.Workload.Name, source.Spec.Workload.Kind)
+	return r.updateInstrumentationConfigServiceName(ctx, instConfigName, req.Namespace, source.Spec.ReportedName)
+}
+
+func (r *SourceReconciler) updateInstrumentationConfigServiceName(ctx context.Context, instConfigName, namespace string, serviceName string) (reconcile.Result, error) {
```
blumamir, line 47: do we need it as a separate function here?

blumamir (approved):
🥇

blumamir:
Also update the docs: https://docs.odigos.io/pipeline/sources/update#service-name

docs/pipeline/sources/update.mdx
```diff
@@ -12,12 +12,23 @@ It is used to identify the service that emitted a span, metric or log record.
 
 Odigos will automatically populate this attribute based on the name of the workload. If there are multiple containers in the workload, the `service.name` will be set to the container name.
 
-If odigos fails to set a `service.name`, or if you want to override the default value, you can set it manually for a specific source.
-
+If odigos fails to set a `service.name`, or if you want to override the default value, you can also set the `service.name` attribute by adding the `otelServiceName` attribute to an exiting Source, or by creating a `Source` with that attribute:
```
blumamir, line 15: nit:

The "fails to set a `service.name`" is here because of the device id uuid which sometimes leaked into the telemetry. I think it is no longer relevant (mechanism replaced everywhere), so we can remove these negative words and just suggest the user to choose his own value.

docs/pipeline/sources/update.mdx
```diff
@@ -33,33 +44,11 @@ To undo the override, simply erase the value from the `service.name` field.
...
-```shell
-kubectl annotate deployment my-deployment odigos.io/reported-name=foo --overwrite
-```
-
-To view the changes, run:
+For example, to set the `service.name` attribute in telemetry for an existing source:
 
 ```shell
-kubectl get deployment my-deployment -o yaml
+kubectl patch source/<source-name> -n <namespace> --type=merge -p '{"spec":{"otelServiceName":"<serviceName>"}}'
```
blumamir, line 52: Users probably know if it's deployment / daemonset / statefulset, it's name, and it's namespace.
Do you have the `kubectl` command that translate user properties to relevant source? could be so awesome to have it here as reference and better usability.

instrumentor/controllers/workloadmigrations/controllers.go
```diff
@@ -150,11 +155,26 @@ func CreateOrUpdateSourceForObject(ctx context.Context,
...
 	}
 	source.Spec.DisableInstrumentation = disableInstrumentation
+	// migrate the reported name from the workload to the Source only if the Source is not a namespace Source
+	// and the workload Source does not already have a service name.
+	//
+	// If a user have set the annotation on the workload we will use it in the source only if no reported name is set.
+	// If that happens once, and the reported name was taken from the annotation,
+	// any more changes to the annotation will not be reflected in the source.
+	// This is valid, since the annotation is deprecated and we want to encourage users to use Source CR.
+	if kind != workload.WorkloadKindNamespace && source.Spec.OtelServiceName == "" {
```
blumamir, line 165: do we need to check here that `serviceName` is not empty string?
how will it act if the annotation is missing to begin with?

instrumentor/controllers/instrumentationconfig/source_controller.go
```diff
@@ -0,0 +1,58 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// These controllers handle update of the InstrumentationConfig's ServiceName
+// whenever there are changes in the associated Source object.
+type SourceReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
```
blumamir, line 21: not using Scheme

instrumentor/controllers/instrumentationdevice/pods_webhook.go
```diff
@@ -70,12 +70,22 @@ func (p *PodsWebhook) getServiceNameForEnv(ctx context.Context, pod *corev1.Pod)
...
-		logger.Error(err, "failed to get workload object from cache. cannot check for workload annotation. using workload name as OTEL_SERVICE_NAME")
+		logger.Error(err, "failed to get workload object from cache. cannot check for workload source. using workload name as OTEL_SERVICE_NAME")
 		return &podWorkload.Name, podWorkload
 	}
-	resolvedServiceName := workload.ExtractServiceNameFromAnnotations(workloadObj.GetAnnotations(), podWorkload.Name)
+
+	resolvedServiceName, err := sourceutils.OtelServiceNameBySource(ctx, p.Client, workloadObj)
+	if err != nil {
+		logger.Error(err, "failed to get OTel service name from source. using workload name as OTEL_SERVICE_NAME")
+		return &podWorkload.Name, podWorkload
```
blumamir, line 83: nit: instead of return, set `resolvedServiceName` so the function has better control structure

blumamir (approved):
Looks awesome, love all the cleanups this brings.
Did one last pass of review, added few nits

instrumentor/controllers/instrumentationconfig/source_controller.go
RonFed: removed

PR #2303, opened 2025-01-24
Add `--update` flag to `odigos install`

damemi:
This would allow `odigos install` to be re-run against an existing installation by passing `--update` (with a confirmation from the user). This could be used to update image prefixes or any other setting set at install time.

Requires odigos to already be installed in the specified namespace. This is to prevent accidental duplicate installs

### DIFF
--- a/cli/cmd/install.go
+++ b/cli/cmd/install.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/odigos-io/odigos/cli/pkg/autodetect"
+	"github.com/odigos-io/odigos/cli/pkg/confirm"
 	"github.com/odigos-io/odigos/common/consts"
 	"github.com/odigos-io/odigos/profiles"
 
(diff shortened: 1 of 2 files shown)


BenElferink:
Please update the docs too :)
https://github.com/odigos-io/odigos/blob/main/docs/cli/odigos_install.mdx

And is this supposed to be different than `odigos upgrade`?
https://docs.odigos.io/cli/odigos_upgrade

damemi:
@BenElferink thanks! I opened https://github.com/odigos-io/odigos/pull/2319 to make syncing our CLI docs easier and I'll use that to update the docs once it merges

> is this supposed to be different than odigos upgrade?

yes, the goal of this isn't to upgrade the version but to re-run the same installation again, making any config changes. I guess you could combine `--update` and `--version` to do a hacky upgrade, but that would not be advised as it won't do any migration work

damemi:
Going to hold on this to talk about alternatives first

damemi:
closing for https://github.com/odigos-io/odigos/pull/2372

PR #2310, opened 2025-01-26
fix: sanitize odigos overwritten envs in uninstall

blumamir:
This is to remove any odigos injected values from env values when attempting to revert them back to the workload manifest


edeNFed (approved):
May god help us 🙏🏻

PR #2312, opened 2025-01-26
Limit instrumentation instance

RonFed:
To limit the number of instrumentation instances, this PR limits them to 15 per pod.


k8sutils/pkg/instrumentation_instance/status.go
```diff
@@ -63,43 +72,70 @@ func InstrumentationInstanceName(ownerName string, pid int) string {
...
-	if err = kubeClient.Create(ctx, updatedInstance); err != nil {
-		if !apierrors.IsAlreadyExists(err) {
+		if !apierrors.IsNotFound(err) {
 			return err
 		}
-		err := kubeClient.Get(ctx, client.ObjectKeyFromObject(updatedInstance), updatedInstance)
+
+		// check if we can create a new instance
+		if instrumentationInstancesCountReachedLimit(ctx, owner, kubeClient) {
+			return fmt.Errorf("instrumentation instances count per pod is over the limit of %d", maxInstrumentationInstancesPerPod)
```
blumamir, line 88: I guess it's not really an error, but normal operation.
Will it show up in the logs as error with stack trace?
Perhaps we can just log as info here and return nil?

k8sutils/pkg/instrumentation_instance/status.go
```diff
@@ -63,43 +72,70 @@ func InstrumentationInstanceName(ownerName string, pid int) string {
...
+
+		err := controllerutil.SetControllerReference(owner, &instance, scheme)
 		if err != nil {
 			return err
 		}
-	}
 
-	updatedInstance.Status = updateInstrumentationInstanceStatus(updatedInstance.Status, options...)
+		err = kubeClient.Create(ctx, &instance)
+		if err != nil && !apierrors.IsAlreadyExists(err) {
```
blumamir, line 116: if we get `AlreadyExists` here, this function will not `return`, but continue to execute the code after the `if` with an uninitialized `instance` variable

RonFed: Do you think we should return here if we get `AlreadyExits`? I think it may lead to status updates to get lost.

k8sutils/pkg/instrumentation_instance/status.go
RonFed: I prefer to return an error and let the caller decide what to do with it.
I can change that to an exported type `ErrInstanceReachedLimit` - wdyt?

k8sutils/pkg/instrumentation_instance/status.go
```diff
@@ -63,43 +72,70 @@ func InstrumentationInstanceName(ownerName string, pid int) string {
...
 	if err != nil {
 		return err
 	}
+
 	return nil
 }
+
+func instrumentationInstancesCountReachedLimit(ctx context.Context, owner client.Object, kubeClient client.Client) bool {
+	instances := &odigosv1.InstrumentationInstanceList{}
+	err := kubeClient.List(ctx, instances,
```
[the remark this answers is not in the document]
RonFed: yep, it's the cache client

blumamir (approved):

PR #2325, opened 2025-01-28
List pods to find those relevant for runtime detection instead of caching the workload objects

RonFed:
This PR refactors the runtime detection controllers to avoid `Get` or `List` operation on workload objects (Deployments, StatefulSets and DaemonSets). As a result:
* The Odiglet permissions are narrowed.
* The cache won't contain workload objects which results in a significant memory consumption improvement (especially in large clusters).

This is done by modifying the `insttrumentationconfig` reconciler to list the Pods associated to the reconciled instrumentationConfig.
The Odiglet cache is already configured to only include Pods whithin the same node, hence this list operation will only return pods in the same node and in the same ns of the instrumentationConfig.


odiglet/pkg/kube/runtime_details/instrumentationconfigs_controller.go
```diff
@@ -73,27 +70,42 @@ func (r *InstrumentationConfigReconciler) Reconcile(ctx context.Context, request
...
 	if err != nil {
-		logger.Error(err, "Failed to get workload and labels from owner")
 		return reconcile.Result{}, err
 	}
 
-	pods, err := kubeutils.GetRunningPods(ctx, labels, workload.GetNamespace(), r.Client)
-	if err != nil {
-		return reconcile.Result{}, err
+	var selectedPods []corev1.Pod
+	for _, pod := range podList.Items {
```
edeNFed, line 83: consider using index and do direct access to avoid copying large structs

blumamir (approved):

PR #2334, opened 2025-01-29
Add docs page on Source details

damemi:
This adds a new page on some of the under-the-hood details of how sources work and how they are meant to be used


docs/pipeline/sources/working-with-sources.mdx
```diff
@@ -0,0 +1,104 @@
+---
+title: "Working with Sources"
+description: "Learn more about how Sources work and how to use them."
+sidebarTitle: "Working with Sources"
+icon: 'circle-check'
+---
+
+This page explains how Source objects work in more detail, including best practices and recommended workflows.
```
BenElferink, line 8: I think this page should be part of the "introduction" page.
Intro should explain the "what is this and how to approach it". Right now for Sources, the intro only has supported workloads and the labels .

damemi: My thinking was to add this as a deep dive page and keep the other pages like introduction/creating/deleting brief with only the key info. This goes a lot more into details that I think would be unnecessary or overwhelming in the intro. I did link to this page from the introduction though, what do you think?

docs/pipeline/sources/working-with-sources.mdx
```diff
@@ -0,0 +1,104 @@
...
+* `odigos.io/workload-kind`
+
+## Automatically created Sources
+
+In some scenarios, Odigos may create Sources for you automatically.
+
+* When instrumenting workloads in the UI, Odigos will create a Workload Source for each workload.
+* When choosing **Include future workloads** in the UI, Odigos will create a Namespace Source for the namespace.
+* When creating a Source grouping in the UI, Odigos will create a Workload Source for each Workload in the group.
+* In v1.0.144+, when setting the `odigos-instrumentation` label on a workload or namespace, Odigos will automatically create a Source to migrate from the legacy label to Sources. Note that deleting the label will not remove the Source once it is automatically created.
```
RonFed, line 101: Do you think this will need revisiting once we implement the Source ownership? wonder if should keep this out until we have an implementation

docs/pipeline/sources/working-with-sources.mdx
```diff
@@ -0,0 +1,104 @@
...
+
+## Source labels
+
+Odigos uses mutating webhooks to automatically copy the values of `spec.workload` to matching labels. It is
+encouraged to use these labels to manage Sources in a cluster:
+
+* `odigos.io/workload-name`
+* `odigos.io/workload-namespace`
+* `odigos.io/workload-kind`
+
```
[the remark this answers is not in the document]
BenElferink: That already exists in `introduction`:
https://github.com/odigos-io/odigos/blob/e263f5ce33efa84e7b4c2dee9e6704ebe5d982b6/docs/pipeline/sources/introduction.mdx?plain=1#L32

docs/pipeline/sources/working-with-sources.mdx
BenElferink: It only seems like a lot because some of the info is duplicated:

For example the "Source Labels" which exist here and in `introduction`:
https://github.com/odigos-io/odigos/blob/e263f5ce33efa84e7b4c2dee9e6704ebe5d982b6/docs/pipeline/sources/introduction.mdx?plain=1#L23

Or "Using multiple Sources" which exists here and in `create`:
https://github.com/odigos-io/odigos/blob/e263f5ce33efa84e7b4c2dee9e6704ebe5d982b6/docs/pipeline/sources/create.mdx?plain=1#L70

I think if the duplicates are stripped down, or merged into their existing sections, then the rest of this file can be inserted in `introduction`.

It'll also be harder to maintain the docs in the future if we have multiple pages mentioning the same things.

damemi: Thanks @BenElferink, I think you're right. I updated this to put the information in the Sources introduction page, ptal!

docs/pipeline/sources/working-with-sources.mdx
damemi: I think we should say exactly what happens right now, and then yes once we change how Sources work for ownership this should be updated

BenElferink (approved):
🙌🏆

PR #2339, opened 2025-01-29
feat: init commit

tamirdavid1:


instrumentor/controllers/instrumentationdevice/pods_webhook.go
```diff
@@ -50,7 +51,15 @@ func (p *PodsWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
 
 	// Inject ODIGOS environment variables into all containers
-	return p.injectOdigosEnvVars(ctx, logger, pod)
+	err := p.injectOdigosEnvVars(ctx, logger, pod)
+	if err != nil {
+		return err
+	}
+
+	// Add odiglet installed node-affinity to the pod
+	podutils.AddOdigletInstalledAffinity(pod, k8sconsts.OdigletInstalledLabel, k8sconsts.OdigletInstalledLabelValue)
```
tamirdavid1, line 60: move the consts into the function

api/k8sconsts/odiglet.go
```diff
@@ -11,4 +11,7 @@ const (
 	OdigletContainerName          = "odiglet"
 	OdigletImageName              = "keyval/odigos-odiglet"
 	OdigletEnterpriseImageName    = "keyval/odigos-enterprise-odiglet"
+	// Used to indicate that the odiglet is installed on a node.
+	OdigletInstalledLabel      = "odiglet-installed"
```
RonFed, line 15:
```suggestion
	OdigletInstalledLabel      = "odigos.io/odiglet-installed"
```

I think that's a better practice for naming kabels/annotations

k8sutils/pkg/pod/pod.go
```diff
@@ -0,0 +1,54 @@
+package pod
+
+import (
+	"github.com/odigos-io/odigos/api/k8sconsts"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func AddOdigletInstalledAffinity(pod *corev1.Pod) {
```
damemi, line 8: it looks like this is only called by the instrumentor right now, is there a plan for it to be used in other places? If not it could probably just be there instead of k8sutils

k8sutils/pkg/node/node.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"context"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+func AddLabelToNode(nodeName string, labelKey string, labelValue string) error {
```
damemi, line 12: same question as AddOdigletInstalledAffinity, is this only used in the one spot right now?

damemi (commented):
the overall approach lgtm, can you please update the e2es to check that the affinity is added and removed when expected?

Also, is the label ever cleaned up when Odigos is uninstalled?

cli/cmd/resources/odiglet.go
```diff
@@ -395,6 +395,16 @@ func NewOdigletDaemonSet(ns string, version string, imagePrefix string, imageNam
 							Args: []string{
 								"init",
 							},
+							Env: []corev1.EnvVar{
+								{
+									Name: "NODE_NAME",
```
RonFed, line 400: nit: consider using `k8sconsts.NodeNameEnvVar` here and also below in the main container

RonFed:
Looks great, +1 to Mike's comments, especially about cleaning the label

k8sutils/pkg/node/node.go
tamirdavid1: This one used only once but it is general function that could be in use later in other places

k8sutils/pkg/pod/pod.go
tamirdavid1: Thanks !

cli/cmd/resources/odiglet.go
tamirdavid1: Thanks!

tamirdavid1:
> the overall approach lgtm, can you please update the e2es to check that the affinity is added and removed when expected? (doesn't need to be in every test, just one add/remove should be good)
> 
> Also, is the label ever cleaned up when Odigos is uninstalled?

First of all thanks for the great comments :)

1. I've added e2e tests to check that the affinity is added (*)
2. Added cleanup label from node once Odigos uninstalled

* > it will be removed if app not instrumented (and webhook wont run to add it)

tests/common/assert/simple-demo-instrumented.yaml
damemi, line 1: This is great, could you please add an update to one of the uninstrumentation tests to make sure the affinity is removed? Since it's a required affinity, if we don't remove if then user workloads could end up unschedulable (like if odigos is uninstalled)

I think you could test that by basically copying this file (`simple-demo-instrumented`), asserting that affinity is empty with chainsaw, and then adding that assertion to some place like [here in the Source e2e](https://github.com/odigos-io/odigos/blob/ca13f98ab0e97eb7e0b6fa53269cd884b5aa5fb7/tests/e2e/source/chainsaw-test.yaml#L162-L179), after we've uninstrumented the workloads

after that, lgtm

damemi (approved):

tests/common/assert/simple-demo-instrumented.yaml
tamirdavid1: Currently as far as i know we dont have un-instrumentation tests.
and we're adding the affinity on the webhook which run only for pods of instrumented workloads

damemi: Sorry, I forgot that this is on the pods, not the workload. In that case we shouldn't need to worry about the scheduling issue I mentioned. Since when a workload is uninstrumented, it will roll out new pods.

odiglet/odiglet.go
```diff
@@ -175,3 +179,27 @@ func runDeviceManager(clientset *kubernetes.Clientset, otelSdkLsf instrumentatio
...
+		os.Exit(-1)
+	}
+
+	nn, ok := os.LookupEnv(k8sconsts.NodeNameEnvVar)
+	if !ok {
+		log.Logger.Error(fmt.Errorf("env var %s is not set", k8sconsts.NodeNameEnvVar), "Failed to load env")
+		os.Exit(-1)
+	}
+
+	if err := k8snode.AddLabelToNode(nn, k8sconsts.OdigletInstalledLabel, k8sconsts.OdigletInstalledLabelValue); err != nil {
```
blumamir, line 199: I wonder if we should specify here if it's community or enterprise.
Since these 2 options will end up with different content in the node fs, and if a community pod get's scheduled on onprem node or vice-versa, it will crash the pod.

blumamir (approved):

blumamir (approved):

PR #2341, opened 2025-01-29
removal of inject instrumentation label when uninstall using cli

AvihuHenya:


BenElferink (approved):

RonFed (approved):
Looks great to me, worth that @blumamir will also have a look

cli/cmd/uninstall.go
```diff
@@ -209,6 +216,12 @@ func getWorkloadRolloutJsonPatch(obj kube.Object, pts *v1.PodTemplateSpec) ([]by
 			})
 		}
 	}
+	if _, found := pts.ObjectMeta.Labels[k8sconsts.OdigosInjectInstrumentationLabel]; found {
```
blumamir, line 219: this should go into the `if obj.GetLabels() != nil` block above
if Labels is `nil` this might panic I think

blumamir (commented):
looks great @AvihuHenya 

one small issue: we should move the access to the label into the if that first checks this is not nil.

cli/cmd/uninstall.go
AvihuHenya: after checking, all resources from apps/v1 API group that we request for must have selector defined in spec.selector and matching labels in spec.template.metadata.labels, so this scenario dosent possible

blumamir (approved):

PR #1355, opened 2024-07-15
Avoid reconciling cluster-wide resources in odiglet

edeNFed:

PR #2348, opened 2025-01-29
Updates for OpenShift compatibility

damemi:
OpenShift seems to enable [OwnerReferencesPermissionEnforcement](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement) by default, which gives errors like this when setting OwnerReferences:

```
2025-01-29T19:42:38Z	ERROR	Reconciler error	{"controller": "startlangdetection-source", "controllerGroup": "odigos.io", "controllerKind": "Source", "Source": {"name":"source-xsbv2","namespace":"test-project"}, "namespace": "test-project", "name": "source-xsbv2", "reconcileID": "c36405f6-40bb-4eb0-9d5d-fe4bc7802994", "error": "instrumentationconfigs.odigos.io \"deployment-inventory\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"}
```

This basically means wherever we update an ownerreference with `BlockOwnerDeletion: true` (which is set by the controller-runtime helpers), we also need permission to update the finalizers for whatever we're making the owner.

I found this in these places:

* When the autoscaler creates a collector configmap, the owner of the ConfigMap is the CollectorsGroups. So we need to be able to update `collectorsgroups/finalizers`
* When the Instrumentor creates an InstrumentationConfig, the owner of the InstrumentationConfig is the workload. So we need finalizers for deployments,statefulsets,and daemonsets
* When the odiglet creates an InstrumentationInstance, the owner of the InstrumentationInstance is also the workload. Same as above

This also adds the SELinux volume mount from https://github.com/odigos-io/odigos/pull/1131 to the Odiglet init container added in https://github.com/odigos-io/odigos/pull/1355, along with making the init container privileged and adding logging for when the SELinux commands aren't found (this was failing silently)

This also updates the `--openshift` flag to automatically use the `*-ubi9` images. If no `--image-prefix` is set, the images will be pulled from `registry.connect.redhat.com/odigos`.


RonFed:
The owners of instrumentation instances are pods, not workloads.

Does setting the finalizers on the workload happen today for setups that are not OpenShift?
I wonder if the owner of instrumentation config should be the Source

damemi:
@RonFed I don't think this admission enforcement is on by default in most clusters. We don't even necessarily have to set the finalizers to trigger this, it's that we are trying to set ownerreferences with BlockDeletion.

Source shouldn't be the ownerreference for the instrumentationconfig, because Sources can persist beyond a workload existing. InstrumentationConfig should be deleted when the workload is deleted

Updated to use Pods instead of workloads for odiglet

autoscaler/main.go
```diff
@@ -92,7 +93,7 @@ func main() {
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.StringVar(&imagePullSecretsString, "image-pull-secrets", "",
 		"The image pull secrets to use for the collectors created by autoscaler")
-	flag.StringVar(&nameutils.ImagePrefix, "image-prefix", "", "The image prefix to use for the collectors created by autoscaler")
+	flag.StringVar(&nameutils.ImagePrefix, "image-prefix", "", "The image prefix to use for the collectors created by autoscaler (DEPRECATED: Use ODIGOS_COLLECTOR_IMAGE env variable)")
```
blumamir, line 96: I think it is safe to delete this option, no one is using it as a binary anywhere, only from odigos deployment

blumamir, line 96: let's also cleanup where it is being populated:

https://github.com/odigos-io/odigos/blob/dfff83dc69f80f88d9d623bece56c8d05fc06f60/cli/cmd/resources/autoscaler.go#L355

api/k8sconsts/consts.go
```diff
@@ -4,6 +4,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/version"
 )
 
+const RedHatImagePrefix = "registry.connect.redhat.com/odigos"
```
RonFed, line 7: nit: this looks specific to the CLI, maybe we should move this constant there,
edit: I see that a lot of the consts here are used only in the CLI, so I guess we can keep it here for consistency

blumamir (approved):
Looks like a difficult task, great work.

autoscaler/main.go
blumamir: and also helm here:
https://github.com/odigos-io/odigos/blob/dfff83dc69f80f88d9d623bece56c8d05fc06f60/helm/odigos/templates/autoscaler/deployment.yaml#L26

autoscaler/main.go
```diff
@@ -262,6 +269,7 @@ func main() {
 		OdigosVersion:        odigosVersion,
 		K8sVersion:           k8sVersion,
 		DisableNameProcessor: disableNameProcessor,
+		Config:               config,
```
RonFed, line 272: Since this is a constant after this point, I think it would be cleaner not to pass it around to all the functions, but to create a getter for it that can be called by whoever needs it

RonFed (approved):
lgtm,
left a few small comments.
Also, worth making sure that this change won't break users that use the `image prefix` option - it looks to me like it won't but just making sure.
In the CLI we are using the same `GetImageName` and in the helm there is the new util function which looks to preserve the current behavior.

autoscaler/main.go
damemi: updated

PR #2355, opened 2025-01-31
fix: collector metrics update after collector version update

RonFed:
This PR fixes the collector metrics package in the UI pod, after the collector version bump.
This deals with breaking changes added to the collector that include renaming and removing metrics.
Added a test script to make sure the metrics are collected by the UI, this script is called both in the multi-apps and in the helm-chart tests.
The `odigos installed` assert didn't assert the UI pod is ready, so this is added.


damemi (approved):
Thanks @RonFed, sorry this broke in the collector updates. Upstream changed a lot of things with self observability for components, so I'm glad you added some new tests

edeNFed (approved):

PR #2360, opened 2025-02-02
feat: add otel distribution concept with initial metadata per distro

blumamir:


edeNFed (approved):

RonFed (dismissed):
I don't think this should be another module,
prefer it under common

PR #2372, opened 2025-02-03
[GEN-2309] fix: config set command

alonkeyval:
This pull request introduces several changes to the `cli/cmd/config.go` file to enhance configuration management in Odigos. The changes include importing new packages, updating command arguments, and modifying the configuration update process.

### Imports and Dependencies:

* Added imports for `resources`, `odigospro`, `common`, and `consts` packages to support new functionalities.

### Command Argument Handling:

* Changed the `Args` for the `setConfigCmd` command from `cobra.ExactArgs(2)` to `cobra.MinimumNArgs(2)` to allow for multiple values.

### Configuration Update Process:

* Replaced the `updateConfigProperty` function with `setConfigProperty` to handle a broader range of configuration properties and to increment the `ConfigVersion`.
* Introduced a new process to fetch the current configuration and apply changes using `resources.GetCurrentConfig`, `resources.CreateResourceManagers`, and `resources.ApplyResourceManagers`.
* Added error handling for deprecated configurations and cleanup of old Odigos resources.

These changes improve the flexibility and robustness of the configuration management in Odigos.


cli/cmd/config.go
```diff
@@ -22,66 +24,124 @@ var configCmd = &cobra.Command{
...
+		}
+
+		l.Success()
+
 	},
 }
 
-func updateConfigProperty(ctx context.Context, client *kube.Client, ns, property, value string) error {
-	configMapName := "odigos-config"
+func setConfigProperty(config *common.OdigosConfiguration, property string, value []string) error {
```
blumamir, line 82: should we err in case the property expects 1 value but the `value` array has more than one?
instead of ignoring it

cli/cmd/config.go
```diff
@@ -22,66 +24,124 @@ var configCmd = &cobra.Command{
...
-		value := args[1]
+		value := args[1:]
 
 		ctx := cmd.Context()
 		client := cmdcontext.KubeClientFromContextOrExit(ctx)
 		ns, _ := cmd.Flags().GetString("namespace")
 		if ns == "" {
-			ns = "odigos-system"
+			ns = consts.DefaultOdigosNamespace
 		}
```
blumamir, line 37: please use

```
ns, err := resources.GetOdigosNamespace(client, ctx)
```

To automatically get the namespace instead of having the user supplying it

blumamir (approved):

damemi:
@alonkeyval @blumamir do you think this could replace https://github.com/odigos-io/odigos/pull/2303? My intent with that was to be able to override settings like image prefix. I had also considered a new `odigos` subcommand, but I didn't know about the `config set` command already. I think what you have here is probably a better way to do it, but let me know what you think

alonkeyval:
@damemi It sounds like we can use this command for the purposes you described as well.

PR #2375, opened 2025-02-03
Add cli subcommands for manipulating Sources

damemi:
This adds a new `odigos sources` subcommand for manipulating Sources:

* `odigos sources create ...`
* `odigos sources update ...`
* `odigos sources delete ...`

Each command takes the following flags:

* `--namespace`: The namespace of the Source to act on
* `--workload-kind`: Applies the command to all Sources with the matching label
* `--workload-name`: Applies the command to all Sources with the matching label
* `--workload-namespace`: Applies the command to all Sources with the matching label
* `--group`: Applies the command to all Sources in the group

## Create

This command creates a Source, for example:

```
odigos sources create my-source --workload-kind=Deployment --workload-name=FooBar --workload-namespace=default
```

It requires the `kind`, `name`, and `namespace` flags (enforced by the Source validation webhook, along with all other checks provided by that webhook).

It also provides the following flags:

* `--disable-instrumentation`: set `spec.disableInstrumentation: <value>` (default: false)

## Update

This command updates any Sources matching the label flags, for example:

```
odigos sources update --workload-kind=DaemonSet --disable-instrumentation
```

Because this can update many Sources at once, there is a confirmation prompt.

It also provides the following flags:

* `--all-namespaces`: apply the change to Sources in all namespaces
* `--yes`: Skip the confirmation prompt
* `--set-group`: Add the `odigos.io/group-<value>` label to the Source
* `--remove-group`: Remove the `odigos.io/group-<value>` label from the Source

## Delete

This command deletes all Sources matching the provided labels, for example:

```
odigos sources delete --group mygroup --all-namespaces
```

It also provides the following flags:

* `--all-namespaces`: apply the change to Sources in all namespaces
* `--yes`: Skip the confirmation prompt


cli/cmd/sources.go
```diff
@@ -0,0 +1,248 @@
...
+	return namespaceText, providedWorkloadFlags, namespaceList, labelSet
+}
+
+func init() {
+	sourceFlags = pflag.NewFlagSet("sourceFlags", pflag.ContinueOnError)
+	sourceFlags.StringVarP(&namespaceFlag, namespaceFlagName, "n", "default", "Kubernetes Namespace for Source")
+	sourceFlags.StringVar(&workloadKindFlag, workloadKindFlagName, "", "Kubernetes Kind for entity (one of: Deployment, DaemonSet, StatefulSet, Namespace)")
+	sourceFlags.StringVar(&workloadNameFlag, workloadNameFlagName, "", "Name of entity for Source")
+	sourceFlags.StringVar(&workloadNamespaceFlag, workloadNamespaceFlagName, "", "Namespace of entity for Source")
+	sourceFlags.StringVar(&sourceGroupFlag, sourceGroupFlagName, "", "Name of Source group to use")
```
RonFed, line 228: Wonder if we should add a short flag as well for these flags

RonFed (approved):
🥇

PR #2377, opened 2025-02-04
fix: remove pod label from workload manifest

blumamir:


instrumentor/controllers/agentenabled/sync.go
```diff
@@ -0,0 +1,369 @@
...
+}
+
+func applyRulesForDistros(defaultDistros map[common.ProgrammingLanguage]string,
+	instrumentationRules *[]odigosv1.InstrumentationRule) map[common.ProgrammingLanguage]string {
+
+	for _, rule := range *instrumentationRules {
+		if rule.Spec.OtelSdks == nil {
+			continue
+		}
+		// TODO: change this from otel sdks to distros and use distro name
```
RonFed, line 318: why not taking the value from the rule if avaialble?

instrumentor/controllers/agentenabled/sync.go
```diff
@@ -0,0 +1,369 @@
...
+//
+// 2. No runtime details for this workload
+//   - Can be transient state, until odiglet calculates the runtime details in a short while
+//   - Can be permanent state (odiglet is not running, workload has not running pods, etc.)
+//
+// The function returns
+// - waitingForPrerequisites: true if we are waiting for some prerequisites to be completed before injecting the agent
+// - reason: AgentInjectionReason enum value that represents the reason why we are waiting
+// - message: human-readable message that describes the reason why we are waiting
+func isPrerequisiteCompleted(cg *odigosv1.CollectorsGroup, ic *odigosv1.InstrumentationConfig) (bool, odigosv1.AgentInjectionReason, string) {
```
RonFed, line 342: what do you think of `isReadyForInjection` or `isReadyForInstrumentation` ?

instrumentor/controllers/agentenabled/sync.go
blumamir: This is missing, I still need to address this

instrumentor/controllers/workloadrollout/controller.go
damemi, line 1: do you think this needs to be its own controller? Or did you consider just adding this functionality to the existing instrumentationconfig controller?

blumamir: Me and @RonFed discussed it in the office, and we are not sure.

I think it's more maintainable when each controller is responsible for doing one thing, but it does end up with more controllers and things we need to follow. I am open to merging it with existing controllers if you think it's more appropriate.

damemi: Yeah I'm thinking about the race issues we were talking about the other day. Adding a new controller that's managing the same resource as another controller can cause problems like that. I think this fits well with the other controller, and that controller is still doing one thing - managing instrumentationconfigs

blumamir: It manages instrumentation configs, and additionally managing setting the annotation to the workload.

I personally don't think it's highly related, and we can soon want to add more logic around rollouts (like not to do them too often, or not to do it on the entire cluster at once), or initiate them as a finalizer on the instrumentation config.
Since it's already implemented this way, I suggest we keep it as is, and consider refactor after the important release of this urgent feature. But I am also good with merging them tomorrow morning if you guys prefer it this way.

BenElferink (approved):

RonFed (approved):

instrumentor/controllers/workloadrollout/controller.go
damemi: cool, thanks for the update @RonFed !

damemi (approved):

PR #2378, opened 2025-02-04
fix: update describe to rely on Sources

RonFed:
Update describe to inspect the Sources instead of relying on the deprecated labels.


alonkeyval:
frontend side looks good

BenElferink (approved):
🏆

blumamir (approved):

PR #2382, opened 2025-02-04
Reduce fp java inspection

AvihuHenya:


procdiscovery/pkg/inspectors/java/java.go
```diff
@@ -12,19 +15,64 @@ import (
...
+		if libjvmRegex.MatchString(scanner.Text()) {
+			return true
+		}
+	}
+	return false
+}
+
+// isJavaExecutable checks if the process binary name suggests it's a Java process.
+// This is useful for cases where "libjvm.so" isn't found in "/proc/<pid>/maps".
+func isJavaExecutable(procExe string) bool {
```
RonFed, line 67: I think we should inline this instead of creating a function for it

procdiscovery/pkg/inspectors/java/java.go
```diff
@@ -12,19 +15,64 @@ import (
...
 var re = regexp.MustCompile(JavaVersionRegex)
 
 func (j *JavaInspector) Inspect(proc *process.Details) (common.ProgrammingLanguage, bool) {
-	if strings.Contains(proc.ExePath, processName) || strings.Contains(proc.CmdLine, processName) {
+	if checkForLoadedJVM(proc.ProcessID) {
+		return common.JavaProgrammingLanguage, true
+	}
+
+	// TODO: do we need to cover the case that the process is a Java process but it dont loaded JVM?
+	if isJavaExecutable(proc.ExePath) {
```
RonFed, line 33: Wonder what is the right order for these checks:
The first check is more reliable than the second one but it's much more expensive - and it has the potential to be triggered for a lot of non-Java processes.
The second check is cheap, but less reliable

@edeNFed what do you think is the right order?

procdiscovery/pkg/inspectors/java/java.go
```diff
@@ -12,19 +15,64 @@ import (
...
+// This is useful for cases where "libjvm.so" isn't found in "/proc/<pid>/maps".
+func isJavaExecutable(procExe string) bool {
+	return filepath.Base(procExe) == processName
+}
+
+// func isGraalVMProcess(cmdline string) bool {
+// 	// GraalVM native images do not load libjvm.so but have Graal-specific arguments
+// 	return strings.Contains(cmdline, "-XX:+UseGraalVM") || strings.Contains(cmdline, "-H:+")
+// }
+
```
RonFed, line 75: Instead of having commented code, I think its better to just write a comment in the `Inspect` function stating that we could support in the future if we need, with a small explanation.

tests/e2e/workload-lifecycle/services/java-http-server/java-omit-in-entrypoint.Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY --from=build /home/app/target/*.jar /app/app.jar
+
+# Rename Java binary to remove "java" from the execution command
+RUN mv /opt/java/openjdk/bin/java /opt/java/openjdk/bin/app_runner
+
+# Set user for security
+USER 15000
+
+# Run the application without "java" keyword
+ENTRYPOINT ["/opt/java/openjdk/bin/app_runner", "-jar", "/app/app.jar"]
```
RonFed, line 20: We do have the `java` word here,
Maybe we should move the `app_runner` another location without any java name and run it from there?

tests/e2e/workload-lifecycle/01-wait-for-trace.yaml
```diff
@@ -9,15 +9,16 @@ query: |
...
+  { resource.service.name = "java-omit-in-entrypoint" } ||
   { resource.service.name = "java-supported-docker-env" } ||
   { resource.service.name = "java-supported-manifest-env" } ||
   { resource.service.name = "java-azul" } ||
-  { resource.service.name = "python-latest-version" && span.http.route = "insert-random/" } ||
-  { resource.service.name = "python-alpine" && span.http.route = "insert-random/" } ||
-  { resource.service.name = "python-min-version" && span.http.route = "insert-random/" }  ||
+  ({ resource.service.name = "python-latest-version" } && { span.http.route = "insert-random/" }) ||
+  ({ resource.service.name = "python-alpine" } && { span.http.route = "insert-random/" }) ||
+  ({ resource.service.name = "python-min-version" } && { span.http.route = "insert-random/" }) ||
```
RonFed, line 18: Why was this changed? it seems unrelated to this PR

distros/yamls/java-enterprise.yaml
```diff
@@ -20,6 +20,8 @@ spec:
     - envName: JAVA_TOOL_OPTIONS
       envValue: "-javaagent:{{ODIGOS_AGENTS_DIR}}/java-ext-ebpf/javaagent.jar -Dotel.javaagent.extensions={{ODIGOS_AGENTS_DIR}}/java-ext-ebpf/otel_agent_extension.jar"
       delimiter: ' '
+    - envName: OTEL_LOGS_EXPORTER
+      envValue: "none"
```
RonFed, line 24: Why is this added?

edeNFed:
I don't know if you are planning to do the performance improvements as part of this PR or a separate one but what I had in mind is:
- Change the `LanguageInspector` interface to be something like that:
```
ProcessName() string
MapEntry() string
```
- Do the heavy operations (iterating over `/proc/id/maps`) once in the `DetectLanguage` function and while iterating check with every Inspector if is has a match. 
- This way we can achieve multiple goals: A. Start with the most accurate operations first. B. improve performance by doing heavy operations once and not per language. C. fallback to the least accurate guess (process name) if nothing else worked.
- Another idea, I think it may be possible to change Python inspector to also look at `/proc/id/maps` instead of `/proc/id/exe` which will make it more suitable to the new interface (and the overall process more performant because we can skip one unique check). but this needs to be verified on a real Python app.

@AvihuHenya @RonFed what do you think?

PR #2383, opened 2025-02-04
reduce fp nodejs runtime inspection

AvihuHenya:


procdiscovery/pkg/inspectors/nodejs/nodejs.go
```diff
@@ -11,10 +12,15 @@ import (
 
 type NodejsInspector struct{}
 
-const nodeProcessName = "node"
+var v8Regex = regexp.MustCompile(`^(?:.*/)?node(\d+)?$`)
+
+var nodeExecutables = map[string]bool{
+	"npm":  true,
+	"yarn": true,
+}
```
BenElferink, line 20: Should there be `node` as well?

```suggestion
var nodeExecutables = map[string]bool{
	"node":  true,
	"npm":  true,
	"yarn": true,
}
```

RonFed (approved):

PR #2388, opened 2025-02-04
Add Operator

damemi:
This adds a very basic operator for installing, uninstalling, and upgrading Odigos. It is based on the same functions used by the CLI, which should be refactored into a common directory at some point in the future.

## Important files:

* `operator/api/v1alpha1/odigos_types.go`: This creates the `operator.odigos.io` self-contained API. The CRD in this file represents an installation of Odigos, and supports most of the current config fields. This API can be versioned separately from the main Odigos API fairly easily because it is self-contained, allowing us to support future breaking changes with conversion webhooks.
* `operator/internal/controller/odigos_controller.go`: This is the main reconciler for `Odigos` CRs. It handles new installations (by creating a CR), modifying an existing installation (modifying the CR), uninstalling (deleting the CR), or upgrades (by upgrading the operator, see below)

Almost everything else is generated by operator-sdk and shouldn't need much, if any review.

## Usage

The Operator is installed in any namespace the user wants (such as `default` or `odigos-operator-system`). The Operator then watches only its own namespace for `Odigos` objects.

When an `Odigos` object is created, the Operator will install Odigos in its own namespace. So the operator and Odigos run in the same namespace. Only one `Odigos` object can exist.

If the `Odigos` object is modified, the Operator will re-run the install, patching any changes.

If the `Odigos` object is deleted, the Operator will uninstall Odigos. The `Odigos` object uses a finalizer to allow resources to be cleaned up.

OpenShift's Operator Lifecycle Manager will handle upgrading the Operator pod from one version to the next. When the new operator pod starts, it will pick up the existing Odigos config and attempt an upgrade.

## Important notes

The version of Odigos is pinned by the `VERSION` variable, defined in `operator/Makefile`. This uses kustomize to define a configmap that the operator uses to know what version of Odigos to pull.

Because the operator code depends on the current CLI code, the operator should stay pinned to the same version as Odigos for now. This means that upgrades can be done by simply upgrading the operator, which will find the existing Odigos config, and perform the same upgrade logic as the `upgrade` cli command.

The operator does depend on a broad range of RBAC rules. It requires every permission that Odigos has (to be able to grant Odigos those permissions), plus the ability to create and delete all of Odigos's resources (deployments, serviceaccounts, etc). We might be able to scope the latter permissions to just the operator's namespace.

## Screenshots

Here's how it looks to install with OpenShift:

Odigos listed in Operators view

Screenshot 2025-02-07 at 3 39 29 PM

Odigos Operator config page

Screenshot 2025-02-07 at 3 39 39 PM

Creating an `Odigos` object (installing Odigos)

Screenshot 2025-02-07 at 3 40 05 PM

Odigos object status post-install/upgrade

Screenshot 2025-02-07 at 3 40 30 PM


operator/config/crd/kustomization.yaml
```diff
@@ -0,0 +1,22 @@
+# This kustomization.yaml is not intended to be run by itself,
```
blumamir, line 1: Are these kubsomize things relevant to odigos? do we need to keep them?

operator/cmd/main.go
```diff
@@ -0,0 +1,218 @@
...
+		// can access the metrics endpoint. The RBAC are configured in 'config/rbac/kustomization.yaml'. More info:
+		// https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.19.0/pkg/metrics/filters#WithAuthenticationAndAuthorization
+		metricsServerOptions.FilterProvider = filters.WithAuthenticationAndAuthorization
+	}
+
+	currentNamespace := env.GetCurrentNamespace()
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme:                 scheme,
+		Metrics:                metricsServerOptions,
+		WebhookServer:          webhookServer,
```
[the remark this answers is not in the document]
damemi: this is scaffolded code from kubebuilder/operator-sdk, I don't think we are technically using it right now. But even if we don't set it, controller-runtime [still sets it to a default empty Server anyway](https://github.com/kubernetes-sigs/controller-runtime/blob/6f611112490242f9e8d5b0ccdc0f3bb8c86c85ef/pkg/manager/manager.go#L551-L553)

I'd rather not modify the default scaffolded code without really understanding it, and this seems like something that could cause headaches if we do add a webhook later on

operator/internal/controller/odigos_controller.go
```diff
@@ -0,0 +1,345 @@
...
+	"github.com/odigos-io/odigos/cli/pkg/kube"
+
+	"github.com/odigos-io/odigos/api/k8sconsts"
+	"github.com/odigos-io/odigos/common"
+	operatorv1alpha1 "github.com/odigos-io/odigos/operator/api/v1alpha1"
+)
+
+const (
+	odigosVersionConfigMap = "odigos-version"
+	odigosVersionKey       = "ODIGOS_VERSION"
```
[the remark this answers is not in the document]
damemi: done

operator/internal/controller/odigos_controller.go
```diff
@@ -0,0 +1,345 @@
...
+
+	odigosInstalledCondition = "OdigosInstalled"
+	odigosUpgradeCondition   = "OdigosUpgraded"
+)
+
+// OdigosReconciler reconciles a Odigos object
+type OdigosReconciler struct {
+	client.Client
+	KubeClient *kube.Client
+	Scheme     *runtime.Scheme
```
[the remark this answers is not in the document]
damemi: something else that's scaffolded by kubebuilder, and I don't know enough of how it's used to say whether it can be removed or not

operator/internal/controller/odigos_controller.go
```diff
@@ -0,0 +1,345 @@
...
+	if len(version) > 0 {
+		if string(version[0]) != "v" {
+			version = "v" + version
+		}
+	} else {
+		meta.SetStatusCondition(&odigos.Status.Conditions, metav1.Condition{
+			Type:               odigosInstalledCondition,
+			Status:             metav1.ConditionFalse,
+			Reason:             "InvalidOdigosVersion",
+			Message:            "could not determine Odigos version from odigos-version configmap or ODIGOS_VERSION environment variable",
```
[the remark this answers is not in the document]
damemi: I'm using a configmap generated by kustomize to pin the Odigos version in the operator bundle. That configmap gets mounted to the environment variable value.

I wasn't really a fan of that, at first I tried to just set an environment variable directly, but kustomize doesn't seem to support that. So instead, we [generate a configmap](https://github.com/damemi/odigos/blob/e6e21ca0298b7e04c5ad8c262a0b06c5f2df610c/operator/config/manager/kustomization.yaml#L9-L12) which lets us [overwrite that value when the bundle is built](https://github.com/damemi/odigos/blob/e6e21ca0298b7e04c5ad8c262a0b06c5f2df610c/operator/Makefile#L274)

The intent is to automate operator releases that are automatically pinned to the same tag as the Odigos images they deploy:

1. Release Odigos v1.0.XYZ
2. Run `make bundle VERSION=1.0.XYZ`
3. Kustomize updates manifests to set ODIGOS_VERSION=1.0.XYZ
4. Operator bundle contains generated, pinned manifests
5. Submit update Operator bundle to red hat

operator/config/crd/kustomization.yaml
damemi: yeah, this is how the ready-to-deploy operator manifests are built and packaged into the bundle with operator-sdk/kubebuilder

operator/internal/controller/odigos_controller.go
```diff
@@ -0,0 +1,345 @@
...
+	for _, o := range odigosList.Items {
+		if o.GetName() != odigos.GetName() {
+			meta.SetStatusCondition(&odigos.Status.Conditions, metav1.Condition{
+				Type:               odigosInstalledCondition,
+				Status:             metav1.ConditionFalse,
+				Reason:             "AlreadyInstalled",
+				Message:            "odigos is already installed in namespace " + ns,
+				ObservedGeneration: odigos.GetGeneration(),
+			})
+			return ctrl.Result{Requeue: false}, utilerrors.NewAggregate(
```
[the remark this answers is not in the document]
damemi: thanks, I think I'll just log the error and let the `status.Update` return an error (and retry) if it fails because I do want that coming through

operator/internal/controller/odigos_controller.go
```diff
@@ -0,0 +1,345 @@
...
+					r.Status().Update(ctx, odigos),
+				})
+		}
+	} else if odigos.Spec.OnPremToken != "" {
+		odigosTier = common.OnPremOdigosTier
+		odigosProToken = odigos.Spec.OnPremToken
+	}
+
+	// validate user input profiles against available profiles
+	cmd.ValidateUserInputProfiles(odigosTier)
```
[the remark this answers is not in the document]
damemi: good point refactored to return an error (and updated to handle in cli)

operator/internal/controller/odigos_controller.go
```diff
@@ -0,0 +1,345 @@
...
+				[]error{errors.New(fmt.Sprintf("error getting current odigos config: %+v", err)),
+					r.Status().Update(ctx, odigos),
+				})
+		}
+
+		odigosConfig = common.OdigosConfiguration{
+			ConfigVersion: 1,
+		}
+	} else {
+		odigosConfig = *config
```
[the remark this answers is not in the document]
damemi: I don't think we have that for `comon.OdigosConfiguration`, looks like it's just a struct, not generated to a runtime object anywhere? https://github.com/odigos-io/odigos/blob/b60c0e4be3366a3f709c4e4f58119679290d504f/common/odigos_config.go#L97

operator/internal/controller/odigos_controller.go
```diff
@@ -0,0 +1,345 @@
...
+	odigosConfig.InstrumentorImage = "keyval/odigos-instrumentor"
+
+	if odigos.Spec.OpenShiftEnabled {
+		if odigos.Spec.ImagePrefix == "" {
+			odigosConfig.ImagePrefix = k8sconsts.RedHatImagePrefix
+		}
+		odigosConfig.OdigletImage = k8sconsts.OdigletImageUBI9
+		odigosConfig.InstrumentorImage = k8sconsts.InstrumentorImageUBI9
+		odigosConfig.AutoscalerImage = k8sconsts.AutoScalerImageUBI9
+	}
```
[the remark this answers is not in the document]
damemi: They parse it in their own cli install functions based on the odigos config: https://github.com/odigos-io/odigos/blob/b60c0e4be3366a3f709c4e4f58119679290d504f/cli/cmd/resources/scheduler.go#L319-L322

since we want to remove odiglet/instrumentor/autoscaler flags anyway, we should probably update those components to just check the config too

cli/cmd/resources/datacollection.go
```diff
@@ -107,11 +107,11 @@ func NewDataCollectionResourceManager(client *kube.Client, ns string, config *co
...
 
-func (a *dataCollectionResourceManager) InstallFromScratch(ctx context.Context) error {
+func (a *dataCollectionResourceManager) InstallFromScratch(ctx context.Context, ownerReferences []metav1.OwnerReference) error {
 	resources := []kube.Object{
 		NewDataCollectionServiceAccount(a.ns),
 		NewDataCollectionClusterRole(a.config.Psp),
 		NewDataCollectionClusterRoleBinding(a.ns),
 	}
-	return a.client.ApplyResources(ctx, a.config.ConfigVersion, resources)
+	return a.client.ApplyResources(ctx, a.config.ConfigVersion, resources, ownerReferences)
```
damemi, line 116: note: think about passing ownerreferences through the kube.Client

cli/cmd/resources/instrumentor.go
```diff
@@ -178,6 +178,21 @@ func NewInstrumentorClusterRole(ownerPermissionEnforcement bool) *rbacv1.Cluster
 				Resources: []string{"statefulsets"},
 				Verbs:     []string{"get", "list", "watch", "update", "patch"},
 			},
+			{
+				// Required for OwnerReferencesPermissionEnforcement (on by default in OpenShift)
+				// When we create an InstrumentationConfig, we set the OwnerReference to the related workload.
+				// Controller-runtime sets BlockDeletion: true. So with this Admission Plugin we need permission to
+				// update finalizers on the workloads so that they can block deletion.
+				// seehttps://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement
+				APIGroups: []string{"apps"},
```
damemi, line 187: wrap these in the openshift flag

cli/cmd/resources/instrumentor.go
```diff
@@ -178,6 +178,21 @@ func NewInstrumentorClusterRole(ownerPermissionEnforcement bool) *rbacv1.Cluster
 				Resources: []string{"statefulsets"},
 				Verbs:     []string{"get", "list", "watch", "update", "patch"},
 			},
+			{
+				// Required for OwnerReferencesPermissionEnforcement (on by default in OpenShift)
+				// When we create an InstrumentationConfig, we set the OwnerReference to the related workload.
+				// Controller-runtime sets BlockDeletion: true. So with this Admission Plugin we need permission to
+				// update finalizers on the workloads so that they can block deletion.
+				// seehttps://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement
```
blumamir, line 186:
```suggestion
				// see https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement
```

cli/cmd/resources/managers.go
```diff
@@ -10,12 +10,12 @@ import (
 // set apiKey to nil for no-op.
 // set to empty string for "no api key" (non odigos cloud mode).
 // set to a valid api key for odigos cloud mode.
-func CreateResourceManagers(client *kube.Client, odigosNs string, odigosTier common.OdigosTier, proTierToken *string, config *common.OdigosConfiguration, odigosVersion string) []resourcemanager.ResourceManager {
+func CreateResourceManagers(client *kube.Client, odigosNs string, odigosTier common.OdigosTier, proTierToken *string, config *common.OdigosConfiguration, odigosVersion, installationMethod string) []resourcemanager.ResourceManager {
```
blumamir, line 13: nit: consider using `K8sInstallationMethod` as the type for new argument to communicate that this is an enum with close set of values, and only convert to string when populating the config map

```suggestion
func CreateResourceManagers(client *kube.Client, odigosNs string, odigosTier common.OdigosTier, proTierToken *string, config *common.OdigosConfiguration, odigosVersion, installationMethod K8sInstallationMethod) []resourcemanager.ResourceManager {
```

docs/setup/installation.mdx
```diff
@@ -9,11 +9,12 @@ sidebarTitle: "Installation"
 
 Checkout the [installation options](/setup/installation-options) to customize your installation.
 
-There are 2 ways to install odigos in your kubernetes cluster:
+There are 3 ways to install odigos in your kubernetes cluster:
 - [Odigos CLI](#odigos-cli)
 - [Helm Chart](#helm-chart)
+- [OpenShift Operator](/setup/installation-options#openshift-installation)
```
blumamir, line 15: nit: can we have the markdown link consistent for all 3 options?

helm/odigos/templates/instrumentor/clusterrole.yaml
```diff
@@ -67,6 +67,12 @@ rules:
     verbs:
       - update
 {{- end }}
+  - apiGroups:
+      - operator.odigos.io
+    resources:
+      - odigos/finalizers
+    verbs:
+      - update
```
blumamir, line 75: if we install with helm, I guess we don't really need the operator permissions, right? will we even have the `operator.odigos.io` api group and `odigos` CRD in helm?

We can have this for competence if that was the motivation, but this is not actually needed to my understanding right?

operator/README.md
blumamir, line 1: nit: remove / update auto generated code from this file so it's not confusing in the future :)

operator/api/v1alpha1/odigos_types.go
```diff
@@ -0,0 +1,98 @@
...
+package v1alpha1
+
+import (
+	"github.com/odigos-io/odigos/common"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// OdigosSpec defines the desired state of Odigos
+type OdigosSpec struct {
```
blumamir, line 26: we can consider adding the `mount-method` here as well:

https://docs.odigos.io/instrumentations/configuration/mount-method#mount-method

so users can change the instrumentation device if they observe issues with this method. it was added lately to the config as well

blumamir (approved):
Thanks again @damemi 

Added few nits but this looks great

PR #2389, opened 2025-02-04
fix: sperate cli namespace flag for sources and install

blumamir:
The new `sources` cli command was reusing the `namespaceFlag` of the install command, setting it to default value `default`.
This leads to `odigos install` trying to install odigos into default ns if run without setting this flag.

### DIFF
--- a/cli/cmd/sources.go
+++ b/cli/cmd/sources.go
@@ -19,7 +19,8 @@ import (
 var (
 	sourceFlags *pflag.FlagSet
 
-	namespaceFlagName = "namespace"
+	namespaceFlagName   = "namespace"
+	sourceNamespaceFlag string
 
 	allNamespacesFlagName = "all-namespaces"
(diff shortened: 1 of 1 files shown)


cli/cmd/sources.go
```diff
@@ -230,7 +231,7 @@ func parseSourceLabelFlags() (string, string, string, labels.Set) {
 
 func init() {
 	sourceFlags = pflag.NewFlagSet("sourceFlags", pflag.ContinueOnError)
-	sourceFlags.StringVarP(&namespaceFlag, namespaceFlagName, "n", "default", "Kubernetes Namespace for Source")
```
blumamir, line 233: reviewer note: the `namespaceFlag` is defined in `install` command and this line overrides it's default value `odigos-system` with `default` which leaks into the `install`.

BenElferink (approved):

PR #2391, opened 2025-02-05
 Process and performance improvements in runtime inspection logic 

AvihuHenya:
This PR introduces process improvements in runtime inspections, enhances language detection accuracy, and optimizes the matching logic to reduce false positives.

Instead of iterating through all available inspections for language identification, we now introduce a two-stage analysis approach:

Quick Scan – A lightweight, fast check to determine the possible language.
Deep Scan – A more thorough analysis performed only if necessary.
Additionally, improvements have been made in how each language is checked for accurate identification, reducing the chances of false positives.

Key Improvements
1. Optimized Language Detection – Eliminates unnecessary iterations by introducing a structured two-stage approach.
2. Refined Matching Logic – Improves how each language is detected, reducing false positives.
3. Performance Enhancement – Speeds up runtime inspections by avoiding redundant evaluations.

Why This Change?
Previously, language detection involved iterating through all available checks, which was inefficient and sometimes led to false positives. With this PR:

The Quick Scan provides an early indication, reducing the need for exhaustive analysis.
The Deep Scan is only performed when necessary, reducing overhead.
Refined language checks ensure better accuracy in detection, minimizing false positives.


procdiscovery/pkg/process/process.go
```diff
@@ -186,3 +189,12 @@ func getRelevantEnvVars(pid int) ProcessEnvs {
 
 	return envs
 }
+
+func GetProcessExeFile(pid int) (*os.File, error) {
+	path := fmt.Sprintf("/proc/%d/exe", pid)
+	fileData, err := os.Open(path)
```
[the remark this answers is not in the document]
AvihuHenya: sure,
please check whenever i close the file in appropriate location.

odiglet/pkg/ebpf/distribution_matcher.go
```diff
@@ -27,6 +28,13 @@ func (dm *podDeviceDistributionMatcher) Distribution(ctx context.Context, e K8sP
 		Environments: process.ProcessEnvs{
 			DetailedEnvs: e.procEvent.ExecDetails.Environments,
 		},
+		Exefile: func() *os.File {
```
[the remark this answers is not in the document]
AvihuHenya: As discussed, there is no option to avoid creating the ExeFIle, and the nil checks are already in the Inpect's functions.

odiglet/pkg/kube/runtime_details/inspection.go
```diff
@@ -52,6 +52,7 @@ func runtimeInspection(ctx context.Context, pods []corev1.Pod, criClient *criwra
 					inspectProc = &proc
 					break
 				}
+				defer proc.Exefile.Close()
```
RonFed, line 55: I think this won't catch all cases.

Maybe we should do something like this above:
```Go
processes, err := process.FindAllInContainer(string(pod.UID), container.Name)
if err != nil {
	log.Logger.Error(err, "failed to find processes in pod container", "pod", pod.Name, "container", container.Name, "namespace", pod.Namespace)
	return nil, err
}

defer func() {
    for _, proc := range processes {
        if proc.Exefile != nil {
            proc.Exefile.Close()
        }
    }
} ()
```

procdiscovery/pkg/inspectors/golang/golang.go
```diff
@@ -18,8 +17,11 @@ const GolangVersionRegex = `go(\d+\.\d+\.\d+)`
 var re = regexp.MustCompile(GolangVersionRegex)
 
 func (g *GolangInspector) Inspect(p *process.Details) (common.ProgrammingLanguage, bool) {
-	file := fmt.Sprintf("/proc/%d/exe", p.ProcessID)
-	_, err := buildinfo.ReadFile(file)
+	if p.Exefile == nil {
+		return "", false
```
RonFed, line 21: As we talked, if the file is nil for some reason, we should fallback to the old way

procdiscovery/pkg/inspectors/golang/golang.go
```diff
@@ -28,8 +30,7 @@ func (g *GolangInspector) Inspect(p *process.Details) (common.ProgrammingLanguag
 }
 
 func (g *GolangInspector) GetRuntimeVersion(p *process.Details, containerURL string) *version.Version {
-	file := fmt.Sprintf("/proc/%d/exe", p.ProcessID)
-	buildInfo, err := buildinfo.ReadFile(file)
+	buildInfo, err := buildinfo.Read(p.Exefile)
```
RonFed, line 33: Same as above - add nil check for the file and fallback to the old way if its nil

procdiscovery/pkg/inspectors/python/python.go
```diff
@@ -40,14 +38,11 @@ func (p *PythonInspector) GetRuntimeVersion(proc *process.Details, containerURL
 }
 
 func (p *PythonInspector) isLibPythonLinked(proc *process.Details) bool {
-	f := fmt.Sprintf("/proc/%d/exe", proc.ProcessID)
-	file, err := os.Open(f)
-	if err != nil {
+	if proc.Exefile == nil {
```
RonFed, line 41: if the file is mil, fallback to the old way

odiglet/pkg/ebpf/distribution_matcher.go
```diff
@@ -27,6 +28,13 @@ func (dm *podDeviceDistributionMatcher) Distribution(ctx context.Context, e K8sP
...
+			file, err := process.GetProcessExeFile(e.procEvent.PID)
```
RonFed, line 32: We need to close the file in a defer

procdiscovery/pkg/inspectors/nodejs/nodejs.go
```diff
@@ -12,23 +12,32 @@ import (
 
 type NodejsInspector struct{}
 
-var v8Regex = regexp.MustCompile(`^(?:.*/)?node(\d+)?$`)
+// Matches any file path ending with:
```
blumamir, line 15: update the comment. 
we are not comparing file path, but exe name

blumamir (commented):
Left few nit comments.

Please add description for this PR, stating what it fixes and how.

procdiscovery/pkg/inspectors/langdetect.go
```diff
@@ -24,62 +24,131 @@ func (e ErrLanguageDetectionConflict) Error() string {
...
+		}
+	}()
+
+	// Step 2: Set up default language detection result
+	detectedLanguageDetails := common.ProgramLanguageDetails{
+		Language: common.UnknownProgrammingLanguage,
+	}
+
+	// Step 3: Define a reusable function to inspect the process
+	runInspection := func(selectInspectionMethod func(Inspector) InspectFunc) (common.ProgramLanguageDetails, error) {
```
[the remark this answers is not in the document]
AvihuHenya: It can be, but since runInspection is only used within the current context and doesn't need to be accessed elsewhere i thought its better to declare it inline.

procdiscovery/pkg/process/process.go
```diff
@@ -38,6 +38,73 @@ type Details struct {
 	Environments ProcessEnvs
 }
 
+type ProcessContext struct {
+	Details
+
+	ExeFile  *os.File
+	MapsFile *os.File
```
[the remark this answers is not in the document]
AvihuHenya: you're right. changed.

procdiscovery/pkg/inspectors/java/java.go
```diff
@@ -12,6 +13,10 @@ import (
 
 type JavaInspector struct{}
 
+// libjvmRegex is a regular expression that matches any path containing "libjvm.so",
+// ensuring that we correctly detect the presence of the JVM shared library.
+var libjvmRegex = regexp.MustCompile(`.*/libjvm\.so`)
```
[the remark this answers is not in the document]
AvihuHenya: good point. 
i replace both regex in Quick and Deep scan with string operations, now its even more efficient

procdiscovery/pkg/inspectors/nodejs/nodejs.go
AvihuHenya: here also i have change my code and avoid using regex

AvihuHenya:
after approve this PR, i adds test case for checking process detection of java app that doesnt has "java" keyword in the exec command. now its not pass the workload lifecycle tests (as expected). 
https://github.com/odigos-io/odigos/pull/2523

procdiscovery/go.mod
```diff
@@ -7,8 +7,32 @@ require (
...
+	golang.org/x/text v0.22.0 // indirect
+	google.golang.org/genproto/googleapis/api v0.0.0-20250127172529-29210b9bc287 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20250127172529-29210b9bc287 // indirect
+	google.golang.org/grpc v1.70.0 // indirect
+	google.golang.org/protobuf v1.36.5 // indirect
+)
+
 require (
 	github.com/go-logr/logr v1.4.2 // indirect
+	github.com/odigos-io/odigos/odiglet v1.0.160
```
RonFed, line 35: pass the logger from odiglet - to remove this dependency

procdiscovery/pkg/process/process.go
```diff
@@ -38,6 +38,73 @@ type Details struct {
...
+	}
+
+	if pcx.mapsFile != nil {
+		if err := pcx.mapsFile.Close(); err != nil && firstErr == nil {
+			firstErr = err
+		}
+		pcx.mapsFile = nil
+	}
+
+	return firstErr
```
RonFed, line 72:
```suggestion
	var err error

	if pcx.exeFile != nil {
		err = errors.Join(err, pcx.exeFile.Close())
		pcx.exeFile = nil
	}

	if pcx.mapsFile != nil {
	        err = errors.Join(err, pcx.mapsFile.Close())
		pcx.mapsFile = nil
	}

	return err
```

procdiscovery/pkg/process/process.go
```diff
@@ -38,6 +38,73 @@ type Details struct {
 	Environments ProcessEnvs
 }
 
+type ProcessContext struct {
+	Details
+
+	exeFile  *os.File
+	mapsFile *os.File
+}
```
RonFed, line 46:
```suggestion
type ProcessFile interface {
       io.ReadSeekCloser
       io.ReaderAt
}

type ProcessContext struct {
	Details

        exeFile    ProcessFile
        mapsFile   ProcessFile
}
```

procdiscovery/pkg/inspectors/langdetect.go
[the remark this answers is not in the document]
RonFed: I agree with Amir

procdiscovery/pkg/inspectors/langdetect.go
```diff
@@ -24,62 +25,94 @@ func (e ErrLanguageDetectionConflict) Error() string {
 	return fmt.Sprintf("language detection conflict between %v and %v", e.languages[0], e.languages[1])
 }
 
-type LanguageInspector interface {
-	Inspect(process *process.Details) (common.ProgrammingLanguage, bool)
+type InspectFunc func(pcx *process.ProcessContext) (common.ProgrammingLanguage, bool)
+
+// Inspector holds two kinds of checks as well as an optional runtime version getter.
```
RonFed, line 30: The version getter is a different interface, so maybe worth updating this comment

procdiscovery/pkg/inspectors/langdetect.go
```diff
@@ -24,62 +25,94 @@ func (e ErrLanguageDetectionConflict) Error() string {
...
 }
 
-func VerifyLanguage(process process.Details, lang common.ProgrammingLanguage) bool {
-	inspector, ok := inspectorsMap[lang]
+func VerifyLanguage(proc process.Details, lang common.ProgrammingLanguage) bool {
+	inspector, ok := inspectorsByLanguage[lang]
 	if !ok {
 		return false
 	}
+	procContext := process.NewProcessContext(proc)
```
RonFed, line 110: need to add a close here in defer (we can avoid logging the close error in here I think)

procdiscovery/pkg/inspectors/nodejs/nodejs.go
```diff
@@ -12,23 +13,39 @@ import (
...
+	// - "node" followed by digits (e.g., "node8", "node10", etc.)
+	// - One of the recognized package managers: "npm" or "yarn"
+	//
+	// The check:
+	// - `strings.HasPrefix(baseExe, "node")` ensures it starts with "node".
+	// - `len(baseExe) == 4` allows "node" as a standalone executable.
+	// - `unicode.IsDigit(rune(baseExe[4]))` ensures that if there’s an extra character, it's a number (rejecting cases like "nodejs").
+	if strings.HasPrefix(baseExe, "node") &&
+		(len(baseExe) == 4 || unicode.IsDigit(rune(baseExe[4]))) ||
+		nodeExecutables[baseExe] {
```
RonFed, line 36: since we are checking one character here `baseExe[4]` I'm not sure `node14` will pass. Or alternatively strings like `node1k` can pass as well. 
If that is the case, maybe worth to re-consider the regex.

RonFed:
- [ ] before merging, check on enterprise with kv-mall

procdiscovery/pkg/inspectors/langdetect.go
AvihuHenya: i added close in defer and also the log on error (I think we do need to log error here)

procdiscovery/pkg/inspectors/nodejs/nodejs.go
AvihuHenya: `unicode.IsDigit(rune(baseExe[4])) ` check if the character at index 4 (i.e., the 5th character) is a digit, so node14 and node1k will pass.  only exe like `nodeX` wont pass

procdiscovery/pkg/inspectors/python/python.go
```diff
@@ -19,35 +18,53 @@ const (
...
 )
 
-func (p *PythonInspector) Inspect(proc *process.Details) (common.ProgrammingLanguage, bool) {
-	if strings.Contains(proc.ExePath, pythonProcessName) || strings.Contains(proc.CmdLine, pythonProcessName) {
+func (p *PythonInspector) QuickScan(pcx *process.ProcessContext) (common.ProgrammingLanguage, bool) {
+	proc := pcx.Details
+	baseExe := filepath.Base(proc.ExePath)
+
+	// Check if baseExe starts with "python"
+	if len(baseExe) >= 6 && baseExe[:6] == "python" {
```
RonFed, line 26: this will miss `python3.12` which I think is something we should capture.

maybe just using `^python(?:\d+(?:\.\d+)*)?$` will be the simplest here

procdiscovery/pkg/inspectors/dotnet/dotnet.go
```diff
@@ -15,22 +13,28 @@ type DotnetInspector struct{}
 
 const processName = "dotnet"
 
-func (d *DotnetInspector) Inspect(p *process.Details) (common.ProgrammingLanguage, bool) {
-	if strings.HasPrefix(filepath.Base(p.ExePath), processName) {
+func (d *DotnetInspector) QuickScan(pcx *process.ProcessContext) (common.ProgrammingLanguage, bool) {
+	baseExe := filepath.Base(pcx.ExePath)
+
+	// Only allow exact match for "dotnet"
+	if baseExe == processName {
```
RonFed, line 20: similar to other languages, I think requiring an exact equality here is too strict

RonFed (commented):
looks great, just the dotnet quick scan can be improved a bit

procdiscovery/pkg/inspectors/dotnet/dotnet.go
AvihuHenya: from what i read about dotnet i know that in the official .NET SDK and runtime installation `dotnet` is the single executable, and the version management is handled inside the dotnet command, not in the executable name (like we saw this week with python).  you think i still need to change the dotnet QuickScan logic?

RonFed (approved):
🎉

blumamir (approved):

PR #2394, opened 2025-02-05
Add RHEL dockerfiles to regular build CI

damemi:


RonFed (approved):

BenElferink (approved):

edeNFed:
Are you sure it's not too much to build both on every PR? What do you think about instead adding a check that checks the diff between Dockerfile and the rhel Docketfile to avoid drifting

damemi:
@edeNFed it is a lot, but there are differences between the two files that we want (the base image and labels, but we could probably ignore those). But changes that aren't related to the dockerfile could break too, like https://github.com/odigos-io/odigos/pull/2393

blumamir (approved):
I too am not sure if we need to run all that in each PR, but suggest we merge it in so we have coverage for now, know of issues early and better estimate it's behavior. 

We can always review what we want to test in each PR and move some things to a nightly / before release, so we test thoroughly less frequently?

I guess running all 3 k8s versions and all the e2e tests with each PR might be something we also want to reconsider.

PR #2400, opened 2025-02-06
Change odiglet base image to distroless

edeNFed:


blumamir (approved):

RonFed (approved):

RonFed:
@damemi Do we need to do something similar also for the RHEL images?

PR #2406, opened 2025-02-07
fix: requeue if failed to trigger an un-instrument rollout

RonFed:
* Reduce flakiness in the `sources` e2e test by waiting 1 second between rollouts. This cause flakiness since we are now using a timestamp annotation to trigger a rollout. This timestamp has a resolution of seconds, hence if the test progresses through 2 steps in less than a second we'll have a missing rollout.
* If a rollout fails when we want to un-instrument a workload, add requeue.

### DIFF
--- a/tests/e2e/source/chainsaw-test.yaml
+++ b/tests/e2e/source/chainsaw-test.yaml
@@ -322,6 +322,16 @@ spec:
       try:
         - apply:
             file: ../../common/apply/instrument-default-ns.yaml
+    - name: Wait for workloads to roll out new revisions
+      try:
+        - script:
+            timeout: 70s
+            content: |
+              kubectl rollout status deployment -l app=coupon
+              kubectl rollout status deployment -l app=frontend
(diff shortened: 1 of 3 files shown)


tests/e2e/source/chainsaw-test.yaml
```diff
@@ -94,12 +94,7 @@ spec:
             file: 01-workloads.yaml
         - script:
             timeout: 70s
-            content: |
-              kubectl rollout status deployment -l app=coupon
-              kubectl rollout status deployment -l app=frontend
-              kubectl rollout status deployment -l app=inventory
-              kubectl rollout status deployment -l app=pricing
-              kubectl rollout status deployment -l app=membership
+            content: ./wait_for_rollout.sh
```
blumamir, line 97: at this point, we don't yet know if the rollout has started or not, right?
can it lead to race conditions?

blumamir (approved):

PR #2408, opened 2025-02-07
feat: add a status condition to ic for instrumentation details

blumamir:
This adds a condition that describes why the workload is instrumented (which kind of source), along with it's name for better troubleshooting and traceability.

Examples:

image

The function in k8sutils now returns more info about the decision, which can  later be used for odigos describe and possibly be also persisted with ic delete if we ever need finalizers for it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -337,6 +337,28 @@ dev-tests-setup: dev-tests-kind-cluster cli-build build-images load-to-kind
 dev-tests-setup-no-build: TAG := e2e-test
 dev-tests-setup-no-build: dev-tests-kind-cluster load-to-kind
 
+# When you run an e2e test locally, and want a grafana instance to troubleshoot the results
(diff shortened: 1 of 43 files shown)


BenElferink (approved):

RonFed:
I'm not sure if a condition is the best approach for this.
Since it only describes a reason, and does not describe an event in time -
i.e the existence of an instrumentationConfig implies we decided to instrument, so this condition feels redundant. 

We could add a dedicated field for it in the status such as `creationReason` or `instrumentationReason` - I think that will probably describe it better.

Regarding the reasons for not instrumenting, I think we could add those to the CR if and when we'll implement a finalizer for it. Right now the instrumentationConfig should not exist if we decided not to instrument. Those reasons are great for `describe` - but not sure they should be part of the API - at least not for now.

RonFed (approved):

PR #2409, opened 2025-02-08
fix: recreate autoscaler resources if deleted or modified

blumamir:
If for any reason, the data-collection daemonset, gateway deployment, or one of the config maps is deleted, then it's not re-created at the moment.

This PR makes it so they are.

We had a bug where the collectors group would get deleted, and it also deleted the daemonset. then if the collectors group is re-created immediately, the daemonset would stay deleted and odigos would not work

### DIFF
--- a/autoscaler/controllers/collectorsgroup_controller.go
+++ b/autoscaler/controllers/collectorsgroup_controller.go
@@ -26,6 +26,8 @@ import (
 
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/version"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


damemi:
great fix, do you think it would be worth adding a test? Something that just creates a pipeline, deletes the pod, and checks to see that it comes back

blumamir:
> great fix, do you think it would be worth adding a test? Something that just creates a pipeline, deletes the pod, and checks to see that it comes back

While it's great to have many tests, I feel this behavior is really niche and the value we will get from asserting it is low compared to the time and ongoing maintenance we would  have to invest in the test.

But I'm open to be convinced otherwise if you feel it's important.

BenElferink (approved):

autoscaler/controllers/collectorsgroup_controller.go
```diff
@@ -61,6 +63,9 @@ func (r *CollectorsGroupReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 func (r *CollectorsGroupReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&odigosv1.CollectorsGroup{}).
+		Owns(&appsv1.DaemonSet{}).  // in case the ds is deleted or modified for any reason, this will reconcile and recreate it
+		Owns(&appsv1.Deployment{}). // in case the deployment is deleted or modified for any reason, this will reconcile and recreate it
+		Owns(&corev1.ConfigMap{}).  // in case the configmap is deleted or modified for any reason, this will reconcile and recreate it
```
RonFed, line 68: I think we should update the event filter. 
This will add a lot of events which we can easily filter by namespace.

blumamir: Not sure exactly how it's implemented, but i've tested it to work, and I believe the reconcile will be called only if the resource has an owner of this type which we do.

Anyhow, the cache pulls in only the relevant objects, so this will not trigger on every daemonset or deployment in the cluster, just those in cache

RonFed (approved):

PR #2413, opened 2025-02-09
Run collectors as nonroot user

edeNFed:
* Remove `odigosresource` processor
* Set log access via init container
* Limit `hostmetrics` scope by mounting less directories


blumamir (approved):
LGTM.

1. do we still need [this](https://github.com/odigos-io/odigos/blob/95104530d059b1a528c0179ea8f270146caa357a/cli/cmd/resources/datacollection.go#L67)?
2. wondering if we need to test this in openshift as well?

edeNFed:
@blumamir I rolled back the changes to `hostmetrics`.
will do the rest of the comments on a separate PR if that's fine with you 🙏

collector/Dockerfile
```diff
@@ -5,6 +5,8 @@ WORKDIR /go/src/
 ARG TARGETARCH
 RUN GOOS=linux GOARCH=$TARGETARCH make build-odigoscol
 
-FROM gcr.io/distroless/base:latest
+# Choose a minimal image with sh + acl utils
+FROM registry.access.redhat.com/ubi9:9.5-1736404036
```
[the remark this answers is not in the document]
damemi: We shouldn't need that

autoscaler/controllers/datacollection/daemonset.go
```diff
@@ -244,26 +245,66 @@ func getDesiredDaemonSet(datacollection *odigosv1.CollectorsGroup,
...
+								},
+							},
+						},
+					},
+					InitContainers: []corev1.Container{
+						{
+							Name:  "set-logs-acls",
+							Image: commonconfig.ControllerConfig.CollectorImage,
+							Command: []string{
+								"/bin/sh",
```
damemi, line 293: I'm not sure, but I'll test if we'll need similar selinux changes for openshift (like in odiglet: https://github.com/odigos-io/odigos/blob/1872e358980c16440f65fcaece2a9ea214db5ec9/odiglet/pkg/instrumentation/fs/agents.go#L149)

autoscaler/controllers/datacollection/daemonset.go
```diff
@@ -279,23 +320,33 @@ func getDesiredDaemonSet(datacollection *odigosv1.CollectorsGroup,
...
+									ReadOnly:  true,
+								},
+								{
+									Name:      "host-sys",
+									MountPath: "/hostfs/sys",
+									ReadOnly:  true,
+								},
+								{
+									Name:      "host-var",
+									MountPath: "/hostfs/var",
```
RonFed, line 349: Can you please write above each mount why is it required?

PR #2422, opened 2025-02-10
Add workflow to publish images to artifact registry

damemi:
This adds a workflow to publish our components to Artifact Registry. It will currently run separate/alongside our normal release pipeline to verify that everything works before we make the switch.


RonFed (approved):

.github/workflows/publish-modules-artifactregistry.yml
```diff
@@ -0,0 +1,93 @@
...
+          file: >-
+            ${{ matrix.service == 'odiglet' && format('odiglet/{0}', matrix.dockerfile) ||
+                matrix.service == 'collector' && format('collector/{0}', matrix.dockerfile) ||
+                matrix.service == 'ui' && format('frontend/{0}', matrix.dockerfile) ||
+                matrix.dockerfile }}
+          context: >-
+            ${{
+              (matrix.service == 'collector' && 'collector') ||
+              '.'
+            }}
```
blumamir, line 93: we can send here an notification to our slack release channel to notify on success/errors so we know to fix them

blumamir (approved):
This looks great.

we also need to publish the cli image to the registry since it's being used by helm today: https://github.com/odigos-io/odigos/blob/8dbd659a5466332089c6569a2609511afd8a45fc/helm/odigos/values.yaml#L164

The cli image is being built and published here:
https://github.com/odigos-io/odigos/blob/8dbd659a5466332089c6569a2609511afd8a45fc/.github/workflows/release.yml#L94

Also, before we release the cli we make sure all images are published successfully here: 
https://github.com/odigos-io/odigos/blob/8dbd659a5466332089c6569a2609511afd8a45fc/.github/workflows/release.yml#L60
 This is so that we don't release cli and helm and then the install fails due to unpublished image. we can do something similar now or later for our private repo

.github/workflows/publish-modules-artifactregistry.yml
damemi: I didn't want to spam the slack if this fails since we're not actually depending on these images yet. When we know it works, I will merge this workflow into the existing publish action that notifies slack. For now, I kept it to just what's needed to build. I'll be watching it closely anyway :)

PR #2425, opened 2025-02-11
feat: write agents deployment hash to pods as label

blumamir:


RonFed (changes requested):
Deployment is a confusing word in this context to put in our API,
I think we should rename this

PR #2427, opened 2025-02-11
Move python out from device

tamirdavid1:


odiglet/pkg/instrumentation/instrumentlang/python.go
RonFed, line 1: Do we need to keep this file?

RonFed (approved):

api/k8sconsts/consts.go
```diff
@@ -26,3 +26,9 @@ var (
 	DefaultIgnoredNamespaces = []string{"kube-system", "local-path-storage", "istio-system", "linkerd", "kube-node-lease"}
 	DefaultIgnoredContainers = []string{"istio-proxy", "vault-agent", "filebeat", "linkerd-proxy", "fluentd", "akeyless-init"}
 )
+
+// Agents related consts
+var (
+	OpampServerHostEnvName = "ODIGOS_OPAMP_SERVER_HOST"
+	OpAMPPort              = 4320
```
blumamir, line 33: As this is not a k8s specific value, we can consider hosting it in `common` instead of `api`

odiglet/pkg/instrumentation/instrumentlang/python.go
```diff
@@ -1,58 +1,13 @@
...
-
-	if _, ok := uniqueDestinationSignals[common.MetricsObservabilitySignal]; ok {
-		metricsExporter = "otlp"
-	}
-	if _, ok := uniqueDestinationSignals[common.TracesObservabilitySignal]; ok {
-		tracesExporter = "otlp"
-	}
 
 	return &v1beta1.ContainerAllocateResponse{
-		Envs: map[string]string{
```
blumamir, line 42: please keep the envs for the device until we sunset it.
So that we don't need to think about edge race conditions while migrating, and can quickly revert back to using the device if needed.

instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -179,7 +179,8 @@ func (p *PodsWebhook) injectOdigosInstrumentation(ctx context.Context, pod *core
 		if (runtimeDetails.Language == common.JavascriptProgrammingLanguage && otelSdk == common.OtelSdkEbpfEnterprise) ||
 			(runtimeDetails.Language == common.GoProgrammingLanguage && otelSdk == common.OtelSdkEbpfCommunity) ||
 			(runtimeDetails.Language == common.JavaProgrammingLanguage && otelSdk == common.OtelSdkEbpfEnterprise) ||
-			(runtimeDetails.Language == common.MySQLProgrammingLanguage && otelSdk == common.OtelSdkEbpfEnterprise) {
+			(runtimeDetails.Language == common.MySQLProgrammingLanguage && otelSdk == common.OtelSdkEbpfEnterprise) ||
+			(runtimeDetails.Language == common.PythonProgrammingLanguage && otelSdk == common.OtelSdkEbpfCommunity) {
```
blumamir, line 183:
```suggestion
			(runtimeDetails.Language == common.PythonProgrammingLanguage && otelSdk == common.OtelSdkNativeCommunity) {
```

blumamir (approved):
Approved but added few minors

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -139,3 +147,28 @@ func getContainerEnvVarPointer(containerEnv *[]corev1.EnvVar, envVarName string)
...
+			Name: "NODE_IP",
+			ValueFrom: &corev1.EnvVarSource{
+				FieldRef: &corev1.ObjectFieldSelector{
+					FieldPath: "status.hostIP",
+				},
+			},
+		},
+		corev1.EnvVar{
+			Name:  commonconsts.OpampServerHostEnvName,
+			Value: fmt.Sprintf("$(NODE_IP):%d", commonconsts.OpAMPPort),
```
blumamir, line 163: WDYT about making this environment variable only host name, this way we don't need any formatting.

blumamir (approved):

PR #2432, opened 2025-02-11
fix: avoid adding the mount paths multiple times

blumamir:
In case the webhook is triggered more than once

### DIFF
--- a/instrumentor/controllers/agentenabled/podswebhook/mount.go
+++ b/instrumentor/controllers/agentenabled/podswebhook/mount.go
@@ -8,22 +8,17 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-func checkIfMountDirectoryExists(containerSpec *corev1.Container, dir string) bool {
-	for _, volumeMount := range containerSpec.VolumeMounts {
-		if volumeMount.SubPath == dir {
-			return true
-		}
-	}
-	return false
-}
-
 func MountDirectory(containerSpec *corev1.Container, dir string) {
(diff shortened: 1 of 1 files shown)


instrumentor/controllers/agentenabled/podswebhook/mount.go
```diff
@@ -8,10 +8,24 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+func checkIfMountDirectoryExists(containerSpec *corev1.Container, dir string) bool {
```
damemi, line 11: I would just inline these functions with a comment about why you're doing this loop

instrumentor/controllers/agentenabled/podswebhook/mount.go
```diff
@@ -21,7 +35,22 @@ func MountDirectory(containerSpec *corev1.Container, dir string) {
 	})
 }
 
+func checkIfVolumExists(pod *corev1.Pod) bool {
```
damemi, line 38: same

damemi (approved):

instrumentor/controllers/agentenabled/podswebhook/mount.go
blumamir: done

instrumentor/controllers/agentenabled/podswebhook/mount.go
blumamir: done

BenElferink (approved):
🏆

PR #2434, opened 2025-02-12
Add readiness probe for instrumentor

RonFed:


instrumentor/main.go
```diff
@@ -294,6 +292,19 @@ func main() {
 	}
 }
 
+func addHealthAndReadyChecks(mgr ctrl.Manager) error {
```
damemi, line 295: nit: does this need to be its own function?

damemi (approved):

PR #2440, opened 2025-02-13
Create cmd folder in instrumentor, scheduler and autoscaler

RonFed:


blumamir (approved):

autoscaler/migration.go
```diff
@@ -1,4 +1,4 @@
-package main
+package autosacler
```
damemi, line 1: opened https://github.com/odigos-io/odigos/pull/2442 to fix this typo (looks like I missed it before this merged)

PR #2445, opened 2025-02-14
refactor: move init code from main to instrumentor package

RonFed:
This PR mainly refactor the instrumentor main function, and extracts it's code to the `instrumentor` package.
* `instrumentor.New` initlizes the manager and all its controllers and webhooks.
* `instrumentor.Run` launches the runnables including the manager itself.

The telemetry reporter was adjusted to accept a context so we can cancel it once the instrumentor shuts down and exit gracefully.


instrumentor/controllers/startlangdetection/manager.go
```diff
@@ -83,5 +83,19 @@ func SetupWithManager(mgr ctrl.Manager) error {
 		return err
 	}
 
+	err = builder.
```
damemi, line 86: this webhook is used by startlangdetection and deleteinstrumentationconfig controllers, so it can be set on the main instrumentor manager like it originally was

blumamir (approved):

PR #2447, opened 2025-02-16
feat: add generic device only for mounts

blumamir:


tamirdavid1 (approved):

distros/yamls/java-ebpf-instrumentations.yaml
```diff
@@ -23,4 +23,4 @@ spec:
   runtimeAgent:
     directoryNames:
       # - "{{ODIGOS_AGENTS_DIR}}/java-ebpf"
-    device: 'instrumentation.odigos.io/java-ebpf-enterprise'
+    device: 'instrumentation.odigos.io/agentmount'
```
edeNFed, line 26: Maybe better name than `aegntmount`? how about `odigos.io/instrumentation` ?

edeNFed: or if we need to keep the same prefix I would do
`instrumentation.odigos.io/generic`

PR #2450, opened 2025-02-16
refactor: pass distros provider as configuration for the instrumentor

RonFed:


distros/yamls/render.go
```diff
@@ -12,45 +10,42 @@ import (
...
 		return nil, err
 	}
 
-	otelDistro := distroResource{}
-	err = yaml.Unmarshal(yamlBytes, &otelDistro)
-	if err != nil {
-		return nil, err
+	distrosByName := make(map[string]*distro.OtelDistro)
+
+	for _, file := range files {
```
blumamir, line 30: should this be removed? looks like it's duplicated now with `oteldistributions.go`

instrumentor/cmd/main.go
```diff
@@ -20,8 +20,13 @@ import (
 	"flag"
 	"os"
 
+	"github.com/odigos-io/odigos/common"
+	"github.com/odigos-io/odigos/distros"
+	"github.com/odigos-io/odigos/distros/yamls"
```
blumamir, line 25: Do we want consumers to pick such internal directories of the distros module?
I guess it should be `internal` in this package, but let's not use it here.

blumamir (approved):

distros/yamls/render.go
RonFed: Removed

instrumentor/cmd/main.go
RonFed: Done, the yamls package is not being imported

PR #2474, opened 2025-02-18
Add OpenShift docs note on exposing UI in-cluster with Routes

damemi:
This is an openshift-specific setting, and it can be different depending on the cluster, so there isn't much we can guide users on. But making them aware of the right service to expose is helpful.

In my AWS OpenShift cluster, for example, I needed to set `tls.termination: edge` manually on the Route (it doesn't seem to be a cli setting in `oc expose`).


BenElferink (approved):
🏆

PR #2477, opened 2025-02-19
Fix UI performance issues

BenElferink:
Performance fixes:

- Killed `InstrumentationInstance` watcher
- Edit `InstrumentationConfig` watcher - modified events to send single events
- On "modified events", fetch single entity (instead of re-fetching all)
- Moved filters into UI containers (instead of filtering on-fetch): [see commit here](https://github.com/odigos-io/ui-containers/commit/d89df2c098374f483bd62831a18044414fe55dfb)
- Increased metrics polling interval from 3 seconds to 5 seconds

Bug fixes:
- Stop polling `Describe Odigos` to fix update-tokens in CLI drawer: [see commit here](https://github.com/odigos-io/ui-containers/commit/67822f8ef82528937ad9a6d388c273b3adf3264d)


frontend/kube/watchers/instrumentation_config_watcher.go
```diff
@@ -39,8 +39,11 @@ func StartInstrumentationConfigWatcher(ctx context.Context, namespace string) er
 
 	instrumentationConfigModifiedEventBatcher = NewEventBatcher(
 		EventBatcherConfig{
-			MinBatchSize: 1,
-			Duration:     5 * time.Second,
+			// By applying 100 every 1 second, we ensure that the event is always an individual event - no batching.
+			// We need to do this because we have to get an event with the target ID, which is not possible with batching.
+			// We need the target ID to fetch the individual entity, instead of fetching all entities.
+			MinBatchSize: 100,
+			Duration:     1 * time.Second,
```
RonFed, line 46: This will increase the amount of events sent to the UI
which will increase the amount of queries to k8s api server.
I understand that the queries will be cheeper now, but I'm not sure this will improve the end result.

frontend/kube/watchers/instrumentation_instance_watcher.go
RonFed, line 1: Why is this removed? 
This watcher provided events about "condition" update of the `x/y instances are healthy".
This condition is currently not written on the instrumentationConfig itself.

frontend/kube/watchers/instrumentation_config_watcher.go
BenElferink: The problem with batches is that we cannot identify the workloads, which means we cannot fetch them individually.
If anything I should remove the batcher completely and send single events.

frontend/kube/watchers/instrumentation_instance_watcher.go
BenElferink: Because we do not display anything about the instance in the UI, all this watcher did was show a toast notification - without any info on which instances failed, followed by a full re-fetch...

Don't we show failed instances in describe?

frontend/gqlgen.yml
```diff
@@ -46,12 +46,14 @@ models:
     fields:
       apiTokens:
         resolver: true
-      k8sActualNamespace:
-        resolver: true
       k8sActualNamespaces:
         resolver: true
+      k8sActualNamespace:
```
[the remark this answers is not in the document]
BenElferink: Yes, when selecting a namespace, we use this to get it's sources for instrumentation.

frontend/graph/schema.resolvers.go
```diff
@@ -150,6 +150,31 @@ func (r *computePlatformResolver) Sources(ctx context.Context, obj *model.Comput
...
+	kind := sourceID.Kind
+	name := sourceID.Name
+
+	list, err := kube.DefaultClient.OdigosClient.InstrumentationConfigs(ns).List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list InstrumentationConfigs in namespace %s: %w", ns, err)
+	}
+
+	// Loop with early exit
+	for _, ic := range list.Items {
```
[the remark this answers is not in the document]
BenElferink: We cannot perform "get" because instru configs don't have any labels to fetch with.

This means I have to "list" all instructions configs in that namespace, and loop over them until 1 matches all conditions (ns, name, kind).

frontend/graph/schema.resolvers.go
```diff
@@ -150,6 +150,31 @@ func (r *computePlatformResolver) Sources(ctx context.Context, obj *model.Comput
...
 }
 
+// Source is the resolver for the source field.
+func (r *computePlatformResolver) Source(ctx context.Context, obj *model.ComputePlatform, sourceID model.K8sSourceID) (*model.K8sActualSource, error) {
+	ns := sourceID.Namespace
+	kind := sourceID.Kind
+	name := sourceID.Name
+
+	list, err := kube.DefaultClient.OdigosClient.InstrumentationConfigs(ns).List(ctx, metav1.ListOptions{})
+	if err != nil {
```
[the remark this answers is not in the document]
BenElferink: We cannot perform "get" because instru configs don't have any labels to fetch with.

frontend/kube/watchers/instrumentation_instance_watcher.go
RonFed: We show the `x/y healthy condition` which is calculated based on the instances.
Removing this watcher means the condition won't get updated

frontend/kube/watchers/instrumentation_config_watcher.go
RonFed: I understand, I'm not sure it's better

frontend/graph/schema.resolvers.go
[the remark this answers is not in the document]
alonkeyval: you can get it by name
image

something like:
```
ic, err := kube.DefaultClient.OdigosClient.InstrumentationConfigs(ns).Get(ctx, name, metav1.GetOptions{})
if err != nil {
    return nil, fmt.Errorf("failed to get InstrumentationConfig %s in namespace %s: %w", name, ns, err)
}
```

frontend/kube/watchers/instrumentation_instance_watcher.go
BenElferink: But we handle the when we get the instru configs:
```go
// Convert each InstrumentationConfig to the K8sActualSource type
for _, ic := range list.Items {
	src := instrumentationConfigToActualSource(ic)
	services.AddHealthyInstrumentationInstancesCondition(ctx, &ic, src)
	actualSources = append(actualSources, src)
}
```

frontend/graph/schema.resolvers.go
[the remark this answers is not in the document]
BenElferink: done 👍

frontend/graph/schema.resolvers.go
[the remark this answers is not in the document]
BenElferink: done 👍

PR #2478, opened 2025-02-19
Minimal java inspection improvement

AvihuHenya:


procdiscovery/pkg/inspectors/java/java.go
```diff
@@ -18,7 +18,7 @@ const JavaVersionRegex = `\d+\.\d+\.\d+\+\d+`
 var re = regexp.MustCompile(JavaVersionRegex)
 
 func (j *JavaInspector) Inspect(proc *process.Details) (common.ProgrammingLanguage, bool) {
-	if strings.Contains(proc.ExePath, processName) || strings.Contains(proc.CmdLine, processName) {
+	if filepath.Base(proc.ExePath) == processName {
```
RonFed, line 21: Do we need to do a contains check or a regex here instead of equality? I'm not sure if some systems use java8 for example as the exe name.

procdiscovery/pkg/inspectors/java/java.go
```diff
@@ -12,13 +11,17 @@ import (
...
 
-const processName = "java"
 const JavaVersionRegex = `\d+\.\d+\.\d+\+\d+`
 
-var re = regexp.MustCompile(JavaVersionRegex)
+// Matches any file path ending with:
+//   - "java" (e.g., /usr/bin/java)
+//   - "javaw" (though less common on Linux)
+//   - "java" / "javaw" followed by version digits (e.g., java8, java11, java17).
+var exeRegex = regexp.MustCompile(`.*/javaw?(?:\d+)?$`)
```
RonFed, line 20: Why do we need the `/` ?

procdiscovery/pkg/inspectors/java/java.go
```diff
@@ -12,13 +11,17 @@ import (
...
+var versionRegex = regexp.MustCompile(JavaVersionRegex)
 
 func (j *JavaInspector) Inspect(proc *process.Details) (common.ProgrammingLanguage, bool) {
-	if strings.Contains(proc.ExePath, processName) || strings.Contains(proc.CmdLine, processName) {
+	if exeRegex.MatchString(proc.ExePath) {
```
RonFed, line 24: should we take `filepath.Base` of the path here?
similar to how we did for node
In that case we'd definitly want to remove the `/` from the regex

procdiscovery/pkg/inspectors/java/java.go
AvihuHenya: some versioning mistake. i fix it in the new commit.

RonFed (approved):

PR #2480, opened 2025-02-20
Align cross cloud provider tests to new registry

tamirdavid1:


.github/workflows/cross-cloud-tests.yaml
```diff
@@ -203,10 +204,10 @@ jobs:
...
         continue-on-error: true
 
-      - name: Destroy Resources
-        if: always()
-        run: |
-          tofu -chdir=$TF_DIR destroy -auto-approve
+      # - name: Destroy Resources
+      #   if: always()
+      #   run: |
+      #     tofu -chdir=$TF_DIR destroy -auto-approve
```
RonFed, line 210: Should this be commented?

RonFed (approved):

.github/workflows/cross-cloud-tests.yaml
tamirdavid1: Thanks for checking
This is just test, you can see that i've created this PR from upstream.
i'll close this and send the origin one soon

PR #2483, opened 2025-02-20
Reduce fp mysql inspection

AvihuHenya:


RonFed:
Can you please include only the mysql change in this PR?

PR #2489, opened 2025-02-23
fix: data-flow flickering on re-renders

BenElferink:
This PR solves the flickering on re-renders for data-flow.
1. By [separating scroll handler from change handler](https://github.com/odigos-io/ui-containers/commit/409eebcbf45c1c59c7e0d5e46ee19aecd64842fd)
2. By [memoizing nodes & edges](https://github.com/odigos-io/ui-containers/commit/25d7b2ea4e1d18ee64d71aeaf7df6c424228fbe3)


frontend/graph/schema.resolvers.go
```diff
@@ -145,6 +147,8 @@ func (r *computePlatformResolver) Sources(ctx context.Context, obj *model.Comput
 		actualSources = append(actualSources, src)
 	}
 
+	r.Logger.Info("resolved sources query", "count", len(list.Items), "duration", time.Since(startTime).String())
```
alonkeyval, line 150: we dont need this log

BenElferink: Amir and I added this so we can debug clients with slow fetch time

alonkeyval (approved):

PR #2495, opened 2025-02-24
Move workload lifecycle to ecr

tamirdavid1:


damemi (commented):
@tamirdavid1 would you like to push these to the new registry? Happy to help with that setup if you'd like, or if ecr is easier then no worries

tamirdavid1:
> @tamirdavid1 would you like to push these to the new registry? Happy to help with that setup if you'd like, or if ecr is easier then no worries
Hi @damemi, I thought about this i'm familiar more with ecr and prefer putting it there.
If you find this critical lets discuss and I will move it once back

BenElferink (approved):

RonFed:
I think we should push to the new registry for consistency.

RonFed (approved):

PR #2504, opened 2025-02-26
first phase of inspection improvements

AvihuHenya:


procdiscovery/pkg/inspectors/nginx/nginx.go
```diff
@@ -26,10 +27,19 @@ const (
...
 	}
 
+	cmdline := strings.Split(string(p.CmdLine), "\x00") // Split arguments
+
+	// Iterate over cmdline arguments and check the basename
+	for _, arg := range cmdline {
+		if filepath.Base(arg) == "nginx" {
+			return common.NginxProgrammingLanguage, true
+		}
+	}
```
RonFed, line 41: Why is this required? I think we should stick with the above check alone

procdiscovery/pkg/inspectors/dotnet/dotnet.go
```diff
@@ -12,7 +13,13 @@ import (
 
 type DotnetInspector struct{}
 
+const processName = "dotnet"
+
 func (d *DotnetInspector) Inspect(p *process.Details) (common.ProgrammingLanguage, bool) {
+	if filepath.Base(p.ExePath) == processName {
```
RonFed, line 19: Is the equality necessary here?
can we have `dotent6 ...` ?

RonFed (approved):

PR #2508, opened 2025-02-27
UI - fix filter crash + minor performance tweaks

BenElferink:


blumamir (approved):

frontend/services/utils.go
```diff
@@ -1,22 +1,53 @@
...
-const cdnUrl = "https://d15jtxgb40qetw.cloudfront.net"
+const (
+	cdnUrl = "https://d15jtxgb40qetw.cloudfront.net"
+)
 
 func GetImageURL(image string) string {
 	return path.Join(cdnUrl, image)
 }
 
+func GetPageLimit(ctx context.Context) (int, error) {
```
alonkeyval, line 28: I think this func more suitable to config.go or in the sources.go

BenElferink: I'm not sure about `config.go`, as there's no relation between pagination and the UI config.
And if we move to `sources.go`, we'll have to move it back when adding pagination to other resources in the future.

Yes, it's used only by sources for now, but pagination can be applied to any resource we desire. Which is why I think `utils.go` is the best place for it.

PR #2513, opened 2025-02-28
Refactor workload-lifecycle e2e to be more fault tolerant

damemi:
The workload-lifecycel e2e can flake because it only runs the traffic generation job once, then loops checking for traces. If auto-instrumentation isn't actually running for a service before the job is generated, the loop is doomed to fail.

This changes how the `workload-lifecycle` e2e generates test traces and counts them:

* Instead of creating the traffic job once, create it on every loop that checks for traces
* When checking for traces, use a new `jq` filter to count the unique occurance of each service name

Unlike the source e2e, which can just check for a [minimum number of spans](https://github.com/odigos-io/odigos/blob/40adcb13abebb2b0a813394562ef431935a592dc/tests/e2e/source/tracesql/wait-for-trace.yaml), we have to aggregate the unique services in these checks. That's because in the source e2e, all of the services are tied in a single trace.  In this test, the services are separate, so the loop could generate multiple traces for the same service while waiting for others to be ready. This means we can't rely on just checking `minimum`, because we might hit that minimum even if all services haven't sent traces yet.

We might be able to eventually drop the `custom_jq` field and just make that the default, but because other jobs might be using the `traceql_runner.sh` script I'm not doing that yet to not break anything else.


BenElferink (dismissed):
Very nice! 🏆
Do you think this loop can be implemented for `cli-upgrade` as well?

RonFed:
The reason we are generating the traffic once is since (at leas in theory) we should have made all the necessary asserts that make sure that all the services are instrumented and Odigos is ready. If that is not the case (which is probably what happens today) - it means we either missing an assert or we have bugs.
Those asserts also reflect what we present to the users - if all the `InstrumentationInstances` are healthy etc' they should expect traces.
This change, although reducing the flakiness - might hide some of the bugs from us IMO.

PR #2520, opened 2025-03-03
perf: separate instrumentation-instances from paginated-sources query

BenElferink:
Previously, when paginating sources in the UI, we would list the `InstrumentationConfigs`, and for each config we would get the `InstrumentationInstances` (just to add a "health| condition message to the config).

Unfortunately, this approach costs a lot of time in some production clusters. When fetching a batch of 10 apps, it would take less than 20ms to list the configs, and it would take up to 20s to get the instances.

---

This PR improves the UI load-time, by separating instrumentation-instances from the paginated-sources-query.

The new approach paginates sources much quicker, displaying them in the UI - and allowing the user to work with them. Once all the sources are there, we’ll fetch & update each source in the list with it's instances-health condition message.

So the UI will have all sources - the sources will keep updating their conditions, until we processed the last source in list, then we stop the loop completely and listen to Server Sent Events (any modified event, we refetch that source by ID - getting the updated data, including it’s instances).

### DIFF
--- a/frontend/graph/schema.resolvers.go
+++ b/frontend/graph/schema.resolvers.go
@@ -9,7 +9,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
-	"time"
 
 	"github.com/odigos-io/odigos/api/k8sconsts"
 	"github.com/odigos-io/odigos/api/odigos/v1alpha1"
@@ -24,7 +23,6 @@ import (
 	"github.com/odigos-io/odigos/k8sutils/pkg/env"
 	"github.com/odigos-io/odigos/k8sutils/pkg/pro"
 	"github.com/odigos-io/odigos/k8sutils/pkg/workload"
-	"golang.org/x/sync/errgroup"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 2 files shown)


frontend/webapp/hooks/sources/useSourceCRUD.ts
```diff
@@ -34,9 +34,38 @@ export const useSourceCRUD = (): UseSourceCrud => {
...
+
+    if (isLastItemInArray) setPaginating(ENTITY_TYPES.SOURCE, false);
+  };
+
+  const fetchExtendedSources = async () => {
+    const items = usePaginatedStore.getState().sources;
+
+    for (let i = 0; i < items.length; i++) {
+      const { namespace, name, kind } = items[i];
+      fetchSourceById({ namespace, name, kind }, true, i === items.length - 1);
```
alonkeyval, line 60: move those variable to var with descriptive name for better readability__

frontend/webapp/hooks/sources/useSourceCRUD.ts
```diff
@@ -34,9 +34,38 @@ export const useSourceCRUD = (): UseSourceCrud => {
   const [fetchPaginated, { loading: isFetching }] = useLazyQuery<{ computePlatform: { sources: PaginatedData<FetchedSource> } }>(GET_SOURCES);
   const [fetchById, { loading: isFetchingById }] = useLazyQuery<{ computePlatform: { source: FetchedSource } }, { sourceId: WorkloadId }>(GET_SOURCE);
 
+  const fetchSourceById = async (id: WorkloadId, bypassPaginationLoader: boolean = false, isLastItemInArray: boolean = false) => {
```
alonkeyval, line 37: UseSourceCrud interface needs to be modify as well

frontend/webapp/hooks/sources/useSourceCRUD.ts
```diff
@@ -34,9 +34,38 @@ export const useSourceCRUD = (): UseSourceCrud => {
...
+      notifyUser(NOTIFICATION_TYPE.ERROR, error.name || CRUD.READ, error.cause?.message || error.message);
+    } else if (!!data?.computePlatform.source) {
+      addPaginated(ENTITY_TYPES.SOURCE, [data.computePlatform.source]);
+    }
+
+    if (isLastItemInArray) setPaginating(ENTITY_TYPES.SOURCE, false);
+  };
+
+  const fetchExtendedSources = async () => {
+    const items = usePaginatedStore.getState().sources;
```
alonkeyval, line 56: second thought about this, maybe we do needs to restrict the API call cycle. in the current state, it loops on all the sources, which can be a lot and call fetchSourceById, maybe we do can batch the calls with Promise.allSettled or await for each result

frontend/webapp/hooks/sources/useSourceCRUD.ts
```diff
@@ -34,9 +34,38 @@ export const useSourceCRUD = (): UseSourceCrud => {
   const [fetchPaginated, { loading: isFetching }] = useLazyQuery<{ computePlatform: { sources: PaginatedData<FetchedSource> } }>(GET_SOURCES);
   const [fetchById, { loading: isFetchingById }] = useLazyQuery<{ computePlatform: { source: FetchedSource } }, { sourceId: WorkloadId }>(GET_SOURCE);
 
+  const fetchSourceById = async (id: WorkloadId, bypassPaginationLoader: boolean = false, isLastItemInArray: boolean = false) => {
+    // We should not fetch while sources are being instrumented.
+    if (useInstrumentStore.getState().isAwaitingInstrumentation) return;
```
alonkeyval, line 39: cant we use all the store state as global instead of getting them each call?

frontend/webapp/hooks/sources/useSourceCRUD.ts
alonkeyval: fetchSourceById use inner function of fetchById, i think we needs to improve readabilltity here

frontend/webapp/hooks/sources/useSourceCRUD.ts
BenElferink: Unfortunately not in the functions, because the function scopes are not in-sync with hooks/state/store until the processes have finished completely.

frontend/graph/schema.resolvers.go
```diff
@@ -118,61 +116,32 @@ func (r *computePlatformResolver) K8sActualNamespace(ctx context.Context, obj *m
 
 // Sources is the resolver for the sources field.
 func (r *computePlatformResolver) Sources(ctx context.Context, obj *model.ComputePlatform, nextPage string) (*model.PaginatedSources, error) {
-	startTime := time.Now()
 	limit, _ := services.GetPageLimit(ctx)
```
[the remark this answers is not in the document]
BenElferink: How?

frontend/webapp/hooks/sources/useSourceCRUD.ts
```diff
@@ -34,9 +34,38 @@ export const useSourceCRUD = (): UseSourceCrud => {
...
+  const fetchExtendedSources = async () => {
+    const items = usePaginatedStore.getState().sources;
+
+    for (let i = 0; i < items.length; i++) {
+      const { namespace, name, kind } = items[i];
+      fetchSourceById({ namespace, name, kind }, true, i === items.length - 1);
+    }
+  };
+
   const fetchSources = async (getAll: boolean = true, page: string = '') => {
```
alonkeyval, line 65: I think we can change this (fetchSources) to startStartPaginateingSources or something

frontend/webapp/hooks/sources/useSourceCRUD.ts
```diff
@@ -34,9 +34,38 @@ export const useSourceCRUD = (): UseSourceCrud => {
...
+    if (!!error) {
+      notifyUser(NOTIFICATION_TYPE.ERROR, error.name || CRUD.READ, error.cause?.message || error.message);
+    } else if (!!data?.computePlatform.source) {
+      addPaginated(ENTITY_TYPES.SOURCE, [data.computePlatform.source]);
+    }
+
+    if (isLastItemInArray) setPaginating(ENTITY_TYPES.SOURCE, false);
+  };
+
+  const fetchExtendedSources = async () => {
```
alonkeyval, line 53: fetchExtendedSources is bisaclly fetch sources by id without any extended data

alonkeyval (approved):

PR #2536, opened 2025-03-06
GKE detect workaround

RonFed:
## Description

Create a workaround to handle https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/1026
figure out whether we're running on GKE at the startup of the autoscaler (with a timeout of 2 seconds) - this should be removed once the issue above is resolved and the collector dependency is updated.
In addition, the `resourcedetection` processor is updated to have a timeout of 2 seconds.

## How Has This Been Tested?

- [x] Manual Testing
Tested both on GKE in which the resource detection returns fast (milliseconds), and in non GKE (kind) in which we have the 2 second timeout.

## User Facing Changes

Not expected, users running on GKE should still see the resource attributes.


autoscaler/controllers/datacollection/configmap.go
```diff
@@ -151,7 +151,17 @@ func calculateConfigMapData(nodeCG *odigosv1.CollectorsGroup, sources *odigosv1.
...
 	}
-	processorsCfg["resourcedetection"] = config.GenericMap{"detectors": []string{"ec2", "gcp", "azure"}}
+	resourceDetectionProcessor := config.GenericMap{
+		"detectors": []string{"ec2", "azure"},
+		"timeout":   "2s",
+	}
+	// This is a workaround to avoid adding the gcp detector if not running on a gke environment
+	// once https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/1026 is resolved, we can always put the gcp detector
+	if commonconf.ControllerConfig.OnGKE {
+		resourceDetectionProcessor["detectors"] = append(resourceDetectionProcessor["detectors"].([]string), "gcp")
```
edeNFed, line 161: I know this cant happen but maybe check that "detectors" exists before doing a cast? this has the potential to panic

damemi (approved):

PR #2538, opened 2025-03-06
fix: use vpid in e2e test assert after agent merge

blumamir:
Updating the tests after merging https://github.com/odigos-io/opentelemetry-node/pull/18


damemi (approved):

BenElferink (approved):
🙌

blumamir:
cli-upgrade is failing because it assert the new value (`process.vpid`) with the value emitted by the previous agent before upgrading (`process.pid`)

PR #2540, opened 2025-03-06
Change label->Source migration to only take effect if a Source doesn't already exist

damemi:
## Description

Currently, the controller that migrates the deprecated `odigos-instrumentation` labels to Sources still supports controlling instrumentation through the label. That means it will both 1) create a Source for a labeled workload and 2) update the created source if the label changes

This creates problems because our code is fully migrated to rely on Sources, not labels. So, for example:

1. User has an existing workload with `odigos-instrumentation: disabled`
2. User updates to a Source-based version of Odigos
3. Odigos migrates the labeled workload to a Source with `disableInstrumentation: true`
4. User then enables the workload through the UI
5. UI only updates the Source to `disableInstrumentation: false` (not the label)
6. New rollout of workload happens
7. Because the workload still has `odigos-instrumentation: disabled`, migration controller finds the existing Source and updates `disableInstrumentation: true`
8. Workload remains uninstrumented

This was provided as a convenience for migrating users (who may still have scripts or automation that rely on the labels). However,

* It has now been announced for several releases that the label is deprecated
* We provide full Source manipulation through k8s labels and the Odigos CLI
* The above hotloop is difficult to solve for a deprecated feature

Because of that, this changes the migration behavior to only create a Source for a labeled workload **if a Source does not already exist**.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [x] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

Added the following tests to the Source e2e:

1. Label a workload with `odigos-instrumentation: enabled` creates an Enabled Source
2. Changing the workload to `odigos-instrumentation: disabled` has no effect (existing Enabled Source overrides label)
3. Label a workload with `odigos-instrumentation: disabled` creates a Disabled Source
4. Changing the workload to `odigos-instrumentation: enabled` has no effect (existing Disabled Source overrides label)
5. Label a workload with `odigos-instrumentation: disabled` to create a Disabled Source, then update the Source to enable instrumentation (Source overrides `disabled` label)
6. Label a workload with `odigos-instrumentation: enabled` to create an Enabled Source, then update the Source to disable instrumentation (Source overrides `enabled` label)

## Kubernetes Checklist

Odigos no longer updates existing Source objects based on workload labels

## User Facing Changes

- [x] Users need to take action before upgrading

Users will no longer be able to change Workload instrumentation settings through the deprecated labels.


docs/pipeline/sources/update.mdx
```diff
@@ -68,3 +68,14 @@ After we have the desired Source we can patch it:
...
+If you don't know the name of a Source object (such as when a Source
+is automatically created by instrumentation done via the UI), you can
+combine the use of Source labels with the `kubectl patch` command. For
+example:
+
+```shell
+kubectl patch source \
+$(kubectl get source -l 'odigos.io/workload-name=<workloadName>,odigos.io/workload-kind=<Deployment/StatefulSet/DaemonSet>' --no-headers -o custom-columns=":metadata.name") \
+--type=merge \
+-p '{"spec":{"otelServiceName":"<serviceName>"}}'
```
RonFed, line 80: Do we support doing it with the odigos cli?

RonFed (approved):
🥇

blumamir (approved):

PR #2541, opened 2025-03-07
test: assert less attributes with cli-upgrade before upgrading

blumamir:
## Description

Attempt to fix the cli-upgrade test failure introduced with #2538 .

This PR makes it so that we don't assert instrumentation attributes when installing odigos latest version as the first step for upgrading.

When we have breaking changes in agents or the attributes which we are asserting, the test that uses the same assert file, will always break on the latest published version. 

I separated the files, so one yaml only checks if the applications are instrumented, and the second one also assert the full attributes and internal data which might break. cli-upgrade uses the first one for "light" assertions only.

It will require that we apply any future updates to both files.

## How Has This Been Tested?

Attempting to make the cli-upgrade tests pass.

- [x] Updated e2e Tests

## Kubernetes Checklist

Only Tests

## User Facing Changes

Only Tests


tests/common/assert/simple-demo-instrumented-full.yaml
```diff
@@ -0,0 +1,390 @@
+### Simple-demo Deployments ###
+apiVersion: apps/v1
```
damemi, line 2: what do you think about doing something like this:

* A single `simple-demo.yaml` file for just the deployments
* A `simple-demo-instrumented.yaml` file that just checks for the InstrumentationConfigs and InstrumentationInstances, without the fields that cause problems
* A `simple-demo-instrumented-full.yaml` file that just checks for the InstrumentationConfigs and InstrumentationInstances, but includes the full details

It's an extra file, but it reduces the duplication to just the IC and Instances instead of also requiring changes to the deployments across files too

I also wonder if Kyverno supports some sort of conditional in its templating. There are [bindings](https://kyverno.github.io/chainsaw/0.2.3/quick-start/bindings/), which seem like a good start, but I don't know if that would let us do something like saw `if full==true { include full attributes }`

blumamir: Thanks, I also played with it a bit and tried to optimize this part.
I think there are many opportunities for improvements, once the cli-upgrade will be stable, I will fix the code and apply some cleanups.
Currently mostly playing around trying things to stabilize the test

damemi: Yeah we can keep iterating on it, deflaking the cli-upgrade test right now is more important

damemi (approved):

tests/common/assert/simple-demo-instrumented-full.yaml
blumamir: removed the deployments part, as it was not observed to help with flaky tests issues

tests/e2e/cli-upgrade/chainsaw-test.yaml
```diff
@@ -140,6 +136,11 @@ spec:
       try:
         - assert:
             file:  ../../common/assert/pipeline-ready.yaml
+        - script:
+            content: |
+              # Make sure there are no leftover gateway pods from pre-upgrade
+              kubectl rollout status deployment -n odigos-test odigos-gateway --timeout=60s
```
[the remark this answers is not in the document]
blumamir: removed this part, as it was not related to the issues we observed

PR #2543, opened 2025-03-07
Add suggested-namespace and supported arch labels to Operator CSV

damemi:
## Description

Add the following labels to the Operator's base CSV to define a default namespace and supported architectures:

```
    operatorframework.io/suggested-namespace: "odigos-operator-system"
    operatorframework.io/os.linux: supported
    operatorframework.io/arch.amd64: supported
    operatorframework.io/arch.arm64: supported
```

See:

* https://docs.openshift.com/container-platform/4.18/operators/operator_sdk/osdk-generating-csvs.html#olm-enabling-operator-for-multi-arch_osdk-generating-csvs
* https://docs.openshift.com/container-platform/4.18/operators/operator_sdk/osdk-generating-csvs.html#osdk-suggested-namespace_osdk-generating-csvs

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/operator/config/manifests/bases/odigos-operator.clusterserviceversion.yaml
+++ b/operator/config/manifests/bases/odigos-operator.clusterserviceversion.yaml
@@ -16,6 +16,10 @@ metadata:
     features.operators.openshift.io/token-auth-aws: "false"
     features.operators.openshift.io/token-auth-azure: "false"
     features.operators.openshift.io/token-auth-gcp: "false"
+    operatorframework.io/suggested-namespace: "odigos-operator-system"
+    operatorframework.io/os.linux: supported
+    operatorframework.io/arch.amd64: supported
+    operatorframework.io/arch.arm64: supported
     operators.openshift.io/valid-subscription: Odigos Enterprise subscription (for
       enterprise features)
     support: Odigos


edeNFed (approved):

operator/config/manifests/bases/odigos-operator.clusterserviceversion.yaml
```diff
@@ -16,6 +16,10 @@ metadata:
     features.operators.openshift.io/token-auth-aws: "false"
     features.operators.openshift.io/token-auth-azure: "false"
     features.operators.openshift.io/token-auth-gcp: "false"
+    operatorframework.io/suggested-namespace: "odigos-operator-system"
```
blumamir, line 19: why not `odigos-system`?
this is the same namespace where odigos itself will be deployed, right?

damemi: `x-operator-system` is common for operators, and in this case the operator is installed in the namespace before Odigos is installed and managed by the operator

PR #2558, opened 2025-03-10
Handle otel agent starting

tamirdavid1:
## Description
This PR fixes two issues related to instrumentation instance updates:

Clientset Throttling – When multiple sources and instances were instrumented in parallel, the Odiglet clientset was throttled, causing delays in fetching pods. This led to failed client requests as the OpAMP server couldn't retrieve instance data in time.

Blocking Response While Saving – Previously, saving instances blocked the request handler, delaying responses to clients. Now, updates are sent to a channel, and a worker goroutine processes them asynchronously. This ensures immediate responses while also preventing race conditions.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests

- [X] Manual Testing
- [X] Manual Load Test
**Created cluster with 20 python application each of them running a 10 workers applications**

## Kubernetes Checklist

- [X] Introduces additional calls to the API Server (potential performance impact)

## User Facing Changes


odiglet/odiglet.go
```diff
@@ -126,10 +126,19 @@ func (o *Odiglet) Run(ctx context.Context) {
...
 	})
 
+	// Create a buffered channel for instrumentation instances updates
+	updateChannel := make(chan server.InstrumentationUpdateTask, 300)
+
+	// Start the worker goroutine to process instrumentation instances updates sequentially
+	g.Go(func() error {
+		server.ProcessInstrumentationUpdates(groupCtx, updateChannel, log.Logger)
+		return nil
+	})
```
RonFed, line 136: I think this part can be moved into the `StartOpAmpServer` function - the opamp module will handle this channel - as it is the only one using it,

opampserver/pkg/server/server.go
```diff
@@ -176,3 +180,30 @@ func StartOpAmpServer(ctx context.Context, logger logr.Logger, mgr ctrl.Manager,
...
+
+func ProcessInstrumentationUpdates(ctx context.Context, updateChannel chan InstrumentationUpdateTask, logger logr.Logger) {
+	logger.Info("Starting instrumentation update worker")
+
+	for {
+		select {
+		case <-ctx.Done():
+			logger.Info("Shutting down instrumentation update worker")
+			return
+		case task := <-updateChannel:
```
RonFed, line 199: I think the channel is not being closed before the OpAmp server exits, we should add the closure of the channel once the server is about to exit, this will also signal this goroutine to exit.

opampserver/pkg/server/server.go
```diff
@@ -89,11 +92,14 @@ func StartOpAmpServer(ctx context.Context, logger logr.Logger, mgr ctrl.Manager,
...
-			err = handlers.UpdateInstrumentationInstanceStatus(ctx, &agentToServer, connectionInfo)
-			if err != nil {
-				logger.Error(err, "Failed to persist instrumentation device status")
-				// still return the opamp response
+
+		// Only update the InstrumentationInstance if the message contains the relevant data
+		// This is to avoid unnecessary updates when the message is a heartbeat
+		if connectionInfo != nil && (agentToServer.AgentDescription != nil || agentToServer.Health != nil) {
+			select {
+			case updateChannel <- InstrumentationUpdateTask{ctx, handlers, &agentToServer, connectionInfo}:
```
RonFed, line 100: The `handlers` is the same for all events I think, we can remove it from the event sent on the channel and pas it to the `ProcessInstrumentationUpdates` function

opampserver/pkg/server/server.go
```diff
@@ -176,3 +193,33 @@ func StartOpAmpServer(ctx context.Context, logger logr.Logger, mgr ctrl.Manager,
...
+
+type InstrumentationUpdateTask struct {
+	ctx            context.Context
+	handlers       *ConnectionHandlers
+	agentToServer  *protobufs.AgentToServer
+	connectionInfo *connection.ConnectionInfo
+}
+
+func ProcessInstrumentationUpdates(ctx context.Context, updateChannel chan InstrumentationUpdateTask, logger logr.Logger) {
+	logger.Info("Starting instrumentation update worker")
```
RonFed, line 205:
```suggestion
	logger.Info("Starting instrumentation instance update worker")
```

opampserver/pkg/server/server.go
```diff
@@ -176,3 +193,33 @@ func StartOpAmpServer(ctx context.Context, logger logr.Logger, mgr ctrl.Manager,
...
+				return
+			}
+
+			err := task.handlers.UpdateInstrumentationInstanceStatus(task.ctx, task.agentToServer, task.connectionInfo)
+
+			if err != nil {
+				logger.Error(err, "Failed to update instrumentation instance")
+			}
+		}
+	}
```
RonFed, line 224: option to consider

```suggestion
	for task := range(updateChannel) {
		err := task.handlers.UpdateInstrumentationInstanceStatus(task.ctx, task.agentToServer, task.connectionInfo)

		if err != nil {
			logger.Error(err, "Failed to update instrumentation instance")
		}
	}
```

RonFed (approved):
🚀 🚀

PR #2560, opened 2025-03-11
Fix waiting for runtime detection for 0 pods

tamirdavid1:
## Description

Fixes an issue where instrumenting a new workload (e.g., a deployment) with no running pods gets stuck in an infinite loading state in the UI.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [X] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


instrumentor/controllers/startlangdetection/workload_controllers.go
```diff
@@ -85,7 +85,12 @@ func reconcileWorkload(ctx context.Context, k8sClient client.Client, objKind k8s
 
 	markedForInstChanged := meta.SetStatusCondition(&ic.Status.Conditions, markedForInstrumentationCondition)
 	runtimeDetailsChanged := initiateRuntimeDetailsConditionIfMissing(ic, workloadObj)
-	agentEnabledChanged := initiateAgentEnabledConditionIfMissing(ic)
+
+	// Set agent enabled condition only if replicas are available to prevent infinite UI loading.
+	agentEnabledChanged := false
+	if workloadObj.AvailableReplicas() > 0 {
```
damemi, line 91: We might need to update our event filters, otherwise I think the agentEnabledCondition might never get set to true.

If the workload was just instrumented (either by creating/updating a Source, or the workload was created) then this function will run. If the workload has 0 available replicas at that time, we don't set this condition.

Then, if the workload later changes to >1 available replicas, this function won't run again because we only [watch for creation events on workloads](https://github.com/odigos-io/odigos/blob/0321201632679fdb4ec47722c94b627d80428140/instrumentor/controllers/startlangdetection/manager.go#L26)

This filter used to check if the number of available replicas had changed, but we dropped that because it was no longer needed when we switched to Sources. If it is needed now, maybe we should bring back that additional check in the filter, or see if there is another way to fix this issue.

I don't really like checking if the number of available replicas has changed in the event filter but if there's no other way to get these changes picked up, then that's fine.

damemi (commented):
heads up this will conflict with https://github.com/odigos-io/odigos/pull/2506

blumamir (approved):

damemi:
@tamirdavid1 the new location for this is `syncWorkload()` in [sourceinstrumentation/common.go](https://github.com/odigos-io/odigos/blob/64dd0599f94cb3701c37f55d7f13b6089c452284/instrumentor/controllers/sourceinstrumentation/common.go#L83)

PR #2561, opened 2025-03-11
docs: add resource requirements to system requirments

blumamir:
Adds a section to system requirements in docs, describing the amount of CPU/memory odigos requires from the cluster / node


blumamir:
> Should you update this section too? https://docs.odigos.io/benchmarks#other-odigos-components

Sorry @damemi saw that just as I was merging. will update there as well

PR #2562, opened 2025-03-11
operator+cli: Change how component image references are set

damemi:
## Description

This changes how component image references are defined. The goal of this is for the operator to install components by their pinned SHA256 value (instead of the usual tag), which is a requirement for OpenShift certification (to ensure operands refer to their actual certified image, and it supports disconnected environments)

The main changes are in the cli. Instead of initializing each component's image within each component's individual manager, it is done once during the install command. The image references are persisted to the odigos config, which is passed to each resource manager, where they can consume the image names themselves.

This centralizes and simplifies the handling of image name cases (default, enterprise, and ubi9).

This then allows the operator to set all of these image names by environment variables (using the new config struct) which are generated by `operator-sdk` to refer directly to the pinned digest for the specified tags.

To handle this in each component, the `GetImageName` function is updated to check if the image is a pinned SHA, and if so it does not append the tag.

## How Has This Been Tested?

- [x] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


common/odigos_config.go
```diff
@@ -114,4 +111,15 @@ type OdigosConfiguration struct {
 	UiPaginationLimit         int                            `json:"uiPaginationLimit,omitempty"`
 	CentralBackendURL         string                         `json:"centralBackendURL,omitempty"`
 	MountMethod               *MountMethod                   `json:"mountMethod,omitempty"`
+	ImageReferences           ImageReferences                `json:"imageReferences,omitempty"`
+}
+
+type ImageReferences struct {
```
RonFed, line 117: Do we actually need this? I'm asking because:
1. The old image options were not used. 
2. They were removed from helm values in https://github.com/odigos-io/odigos/pull/2531
3. The operator seems to use env vars to pass the images, so this is still not used?

damemi: This is different from the old image options and helm values (this PR actually removes those too)

Each component still needs to define its image, right now that's done individually within each resource manager. That includes handling each possible combination between default/ubi9 and community/enterprise.

This is used to set all of the component images in one place and pass that once to the CreateResourceManagers function.

The operator reads the environment variables and uses this struct to pass those values in a single place (CreateResourceManagers). Otherwise, we would need a way to pass each image to each resource manager, which spreads that out into a lot of places.

RonFed: I agree with making the image name decision centralized. My point is (and I'm not sure if that is the case) these images should not be allowed to be configured externally via helm values or via the odigosConfig config map. 
We have a closed set of installation options that make use of this, and I don't think we should allow users setting any image here by editing the config map.

damemi: Ahh, I understand you now. Yes, this struct is only used internally and its values should only be set by us, not the user.

I made it so that the values in the configmap aren't actually consumed anywhere. Instead, whenever we get the odigos configmap, I'm calling `GetImageReferences` to initialize these values from scratch. This function is meant to initialize the image values from scratch using the closed options you mentioned, and overwrite whatever is in the configmap.

If there's anywhere I missed that would be using these directly, we should update that with a fresh call to GetImageReferences. We can also try just not writing these values to the odigos config map, but like I said above it is convenient to have all the image references in one place, but not necessary.

damemi: @RonFed I pushed a new commit moving these image references to the `kube.Client` struct, like what we did for OwnerReferences

Thinking about it, we can't strictly enforce that the values from the configmap will *never* get blindly applied, so it's risky to expose that at all. This kubeClient struct is entirely internal, so it should address that. Let me know what you think

RonFed: Thanks @damemi.
We could just have the struct as a standalone one and passing it around to the relevant components in the CLI?
Having it in the Client struct feels a bit not related.

damemi: @RonFed we can do that too, it's just a little more wiring up to pass it everywhere. So I thought either the Config or Client structs since they are already passed to each resource.

That's originally what I had for the OwnerReferences, and I think maybe @blumamir suggested putting them in the Client struct for that reason. 

Maybe this could be a chance to put both image references and owner references into an options struct that gets passed to each resource manager. What do you think about doing that?

damemi: @RonFed I pushed a new commit making a separate options struct for both OwnerReferences and ImageReferences, and this is a new spot to be used for non-client, internal options for the resource managers. ptal thanks

cli/cmd/resources/resourcemanager/resourcemanager.go
```diff
@@ -2,8 +2,25 @@ package resourcemanager
...
 import (
 	"context"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type ManagerOpts struct {
+	ImageReferences ImageReferences
+	OwnerReferences []metav1.OwnerReference
+}
```
RonFed, line 12: Can you please add comments here about how these fields are used?

operator/internal/controller/odigos_controller.go
```diff
@@ -62,6 +63,32 @@ type OdigosReconciler struct {
...
+	k8sconsts.InstrumentorEnterpriseImage:     "RELATED_IMAGE_ENTERPRISE_INSTRUMENTOR",
+	k8sconsts.InstrumentorEnterpriseImageUBI9: "RELATED_IMAGE_ENTERPRISE_INSTRUMENTOR",
+
+	k8sconsts.UIImage:     "RELATED_IMAGE_FRONTEND",
+	k8sconsts.UIImageUBI9: "RELATED_IMAGE_FRONTEND",
+
+	k8sconsts.OdigletImageName:           "RELATED_IMAGE_ODIGLET",
+	k8sconsts.OdigletImageUBI9:           "RELATED_IMAGE_ODIGLET",
+	k8sconsts.OdigletEnterpriseImageName: "RELATED_IMAGE_ODIGLET",
+	k8sconsts.OdigletEnterpriseImageUBI9: "RELATED_IMAGE_ODIGLET",
```
RonFed, line 86: not sure if that's a typo or not (since I saw that these ENTERPRISE envs are defined)

```suggestion
	k8sconsts.OdigletEnterpriseImageName: "RELATED_IMAGE_ENTERPRISE_ODIGLET",
	k8sconsts.OdigletEnterpriseImageUBI9: "RELATED_IMAGE_ENTERPRISE_ODIGLET",
```

RonFed (approved):
looks great - maybe one typo in the image name map in the operator

operator/internal/controller/odigos_controller.go
damemi: great catch thank you

PR #2564, opened 2025-03-12
GitHub workflows grooming

ErikNaz-odigos:
## Description
Created a single nighty workflow.
Moved image creation logic out of cloud provider tests
Added a chaos test

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/.github/workflows/build-dev-images.yml
+++ b/.github/workflows/build-dev-images.yml
@@ -1,0 +1,85 @@
+name: Build and push images
+
+on:
+  workflow_call: # Workflow trigger
+    inputs:
+      target_tag:
+        description: 'The tag to use when pushing the images to the registry'
+        required: false
+        type: string
+      registry:
+        description: 'The registry to push images to'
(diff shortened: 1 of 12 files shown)


damemi (approved):

.github/workflows/build-dev-images.yml
RonFed, line 1: nit: this can be done with matrix similarly to what we have in other build workflows.

ideally we could have a single build workflow that can have inputs such as:
- whether we want to push at all (for PRs we would just want to build local images)
- where to push to
- where to login to (aws/gcp) before pushing

having this we could use this single workflow in PRs/releas/nightly etc' and passing different inputs

not necessary for this PR, just an idea

RonFed (approved):
💯

PR #2569, opened 2025-03-13
cli: Add simple enable/disable Source commands

damemi:
## Description

This adds new commands to the CLI for enabling/disabling Sources: `odigos sources <enable/disable> <type> <name>`

These commands are essentially shorthand for the more explicit (and less user-friendly) `odigos sources create/update` commands.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

- [x] Changes UI, CLI, or K8s Manifests aspects in a way that users need to be aware of

### DIFF
--- a/cli/cmd/sources.go
+++ b/cli/cmd/sources.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"text/tabwriter"
@@ -10,7 +11,10 @@ import (
 	"github.com/odigos-io/odigos/cli/cmd/sources_utils"
 	cmdcontext "github.com/odigos-io/odigos/cli/pkg/cmd_context"
 	"github.com/odigos-io/odigos/cli/pkg/confirm"
+	"github.com/odigos-io/odigos/cli/pkg/kube"
+	"github.com/odigos-io/odigos/k8sutils/pkg/workload"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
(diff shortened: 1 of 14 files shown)


cli/cmd/sources.go
```diff
@@ -330,6 +374,128 @@ var sourceStatusCmd = &cobra.Command{
...
+		// alert the user that these workloads won't be affected by the command
+		selector := labels.NewSelector()
+		workloadKindRequirement, err := labels.NewRequirement(k8sconsts.WorkloadKindLabel, selection.NotIn, []string{string(k8sconsts.WorkloadKindNamespace)})
+		if err != nil {
+			return nil, err
+		}
+		selector.Add(*workloadKindRequirement)
+		sources, err := client.OdigosClient.Sources(sourceNamespace).List(ctx, metav1.ListOptions{LabelSelector: selector.String()})
+		if len(sources.Items) > 0 {
+			fmt.Printf("NOTE: Configured Namespace Source, but the following Workload Sources will not be affected:\n")
```
blumamir, line 489: nit: consider adding more info about why they will not be affected (existing workload specific configuration that take precedent over the namespace settings).

I can imagine this message to be cryptic for new users without enough context

cli/cmd/sources.go
```diff
@@ -330,6 +374,128 @@ var sourceStatusCmd = &cobra.Command{
...
+			for _, source := range sources.Items {
+				fmt.Printf("Source: %s (Workload=%s, Kind=%s)\n", source.GetName(), source.Spec.Workload.Name, source.Spec.Workload.Kind)
```
blumamir, line 491: should we show here all sources? or just those that have the opposite "disableInstrumentation" value than the one the namespace is setting? 

e.g. the ones which are truly affected by this change

cli/cmd/sources.go
```diff
@@ -372,6 +538,19 @@ func init() {
 	sourcesCmd.AddCommand(sourceUpdateCmd)
 	sourcesCmd.AddCommand(sourceStatusCmd)
 
+	sourcesCmd.AddCommand(sourceEnableCmd)
+	sourceEnableCmd.PersistentFlags().StringVarP(&sourceNamespaceFlag, namespaceFlagName, "n", "default", "Kubernetes Namespace for Source")
```
blumamir, line 542: will we see the `-n` also for `odigos sources enable namespace`? if I understand correctly it is not used for this subcommand.

do you think we can add it only to the relevant commands to reduce chance of mistakes?

cli/cmd/sources.go
```diff
@@ -330,6 +374,128 @@ var sourceStatusCmd = &cobra.Command{
...
+					Kind:      workloadKind,
+					Name:      objName,
+					Namespace: objNamespace,
+				},
+			},
+		}
+
+	}
+
+	source.Spec.DisableInstrumentation = disableInstrumentation
```
blumamir, line 467: I wonder if it's easy and useful to denote a "no-op" operation (e.g. the source was effectively uninstrumented and this command keeps it as is, or source was intrumented and is now uninstrumented).

so if user runs a command that effectively does not change the instrumentation state, it might be a mistake and the user should not expect any change in odigos.

Not sure it's easy to achieve for all the possible states, but if it does, can be nice-to-have it :)

cli/cmd/sources.go
```diff
@@ -330,6 +374,128 @@ var sourceStatusCmd = &cobra.Command{
...
+	if len(sources.Items) > 0 {
+		source, err = client.OdigosClient.Sources(sourceNamespace).Update(ctx, source, v1.UpdateOptions{})
+	} else {
+		source, err = client.OdigosClient.Sources(sourceNamespace).Create(ctx, source, v1.CreateOptions{})
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	if workloadKind == k8sconsts.WorkloadKindNamespace {
```
blumamir, line 478: nit: for a workload kinds, if the current instrumentation state is different than the one of the namespace, we can consider mentioning that here.

something like "NOTE: disabled instrumentation for workload Source, which exclude it from being instrumented by the namespace 'default' Source"

cli/cmd/sources.go
```diff
@@ -67,6 +72,45 @@ odigos sources delete --group mygroup --all-namespaces
...
+# Disable statefulset "foo" in namespace "bar"
+odigos sources disable statefulset foo -n bar
+`,
+}
+
+var sourceEnableCmd = &cobra.Command{
+	Use:   "enable [workload type] [workload name] [flags]",
+	Short: "Enable a source for Odigos instrumentation.",
+	Long:  "This command enables the given workload for Odigos instrumentation. It will create a Source object (if one does not already exist)",
+	Example: `
```
blumamir, line 102: Not sure if it works in cobra for non-leaf commands, but wondering if we can have an `Alias` here to also react to `instrument` (and `uninstrument` for disable).

cli/cmd/sources.go
```diff
@@ -67,6 +72,45 @@ odigos sources delete --group mygroup --all-namespaces
...
+# Enable deployment "foo" in namespace "default"
+odigos sources enable deployment foo
+
+# Enable namespace "bar" in namespace "default"
```
blumamir, line 106: The namespace is "bar". default is not relevant in this command, right?

```suggestion
# Enable workloads in namespace "bar"
```

blumamir (approved):
Thanks @damemi 

added few optional suggestions (only if they make sense to you and are easy to achieve). please have a look before merging and consider if you want to add any.

One thing that I would really like to have, is in [this docs page](https://docs.odigos.io/pipeline/sources/delete) at the top - the CLI command users should run if they need a quick reference on how to "uninstrument" or remove a source. something that is straight forward for unexperienced or new users to follow and execute.

Maybe?

```
To remove a source from odigos, run "odigos sources disable deployment <deployment-name> -n <deployment-namespace>"

This command will rollout existing pods to restart without any odigos additions.

It can be undone (re-instrumented) by running "odigos sources enable deployment <deployment-name> -n <deployment-namespace>"
```

cli/cmd/sources.go
blumamir: also the "disable" command should be updated

damemi:
Great suggestions @blumamir, I think I got everything. Please take another look thanks!

PR #2570, opened 2025-03-13
fix: dependabot yarn -> npm

BenElferink:
According to the [docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#package-ecosystem-), `yarn` should be defined with `npm`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -49,7 +49,7 @@ updates:
       day: sunday
       interval: weekly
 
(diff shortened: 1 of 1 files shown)


blumamir (approved):
nice catch!
how did you find it?

BenElferink:
> nice catch!
> 
> how did you find it?

I realized dependabot never made a PR for package.json's - so I looked for the documentation and there it was 😄

PR #2598, opened 2025-03-17
feat: inject some distro env vars from manifest instead of code branching

blumamir:
## Description

We want to retire some old code responsible for injecting some env vars, to make the logic sit in the distro manifest instead of being set in code.

This has the benefit of consolidating all the relevant per-distro configurations into one file, and removing some of the tier-aware code which should be migrated to enterprise instrumentor.

## How Has This Been Tested?

- [ ] Updated e2e Tests
- [X] Manual Testing - tested this in my local kind cluster. checked the values in the manifest after rollout with new instrumentor, and made sure all services are running, healthy, and emitting traces.

## Kubernetes Checklist

Just refactor of the code in the pods webhook. no changes in k8s api.

It's notable to mention that the new implementation checks for the presence of an env var before setting it's value in the manifest.

## User Facing Changes

Internal refactors only. no user face change is expected.

### DIFF
--- a/instrumentor/controllers/agentenabled/podswebhook/env.go
+++ b/instrumentor/controllers/agentenabled/podswebhook/env.go
@@ -5,14 +5,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-func InjectOdigosK8sEnvVars(existingEnvNames *map[string]struct{}, container *corev1.Container, distroName string, ns string) {
-
-	injectEnvVarToPodContainer(existingEnvNames, container, k8sconsts.OdigosEnvVarContainerName, container.Name)
-	injectEnvVarToPodContainer(existingEnvNames, container, k8sconsts.OdigosEnvVarDistroName, distroName)
-	injectEnvVarObjectFieldRefToPodContainer(existingEnvNames, container, k8sconsts.OdigosEnvVarPodName, "metadata.name")
-	injectEnvVarToPodContainer(existingEnvNames, container, k8sconsts.OdigosEnvVarNamespace, ns)
-}
-
 func injectEnvVarObjectFieldRefToPodContainer(existingEnvNames *map[string]struct{}, container *corev1.Container, envVarName, envVarRef string) {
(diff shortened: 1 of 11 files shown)


BenElferink (approved):

instrumentor/controllers/agentenabled/podswebhook/env.go
```diff
@@ -42,3 +38,30 @@ func injectEnvVarToPodContainer(existingEnvNames *map[string]struct{}, container
...
+	injectNodeIpEnvVar(existingEnvNames, container)
+	opAmpServerHost := fmt.Sprintf("$(NODE_IP):%d", commonconsts.OpAMPPort)
+	injectEnvVarToPodContainer(existingEnvNames, container, commonconsts.OpampServerHostEnvName, opAmpServerHost)
+}
+
+func InjectOtlpHttpEndpointEnvVar(existingEnvNames *map[string]struct{}, container *corev1.Container) {
+	injectNodeIpEnvVar(existingEnvNames, container)
+	otlpHttpEndpoint := service.LocalTrafficOTLPHttpDataCollectionEndpoint("$(NODE_IP)")
+	injectEnvVarToPodContainer(existingEnvNames, container, commonconsts.OtelExporterEndpointEnvName, otlpHttpEndpoint)
+}
```
RonFed, line 67: Map types are reference types - no need to pass them as pointers here
nit:
instead of passing the same pointers for all these functions we can transform them to instance methods:
```Go
type ContainerEnvs struct {
    envKeys map[string]struct{}
    container *corev1.Container
}

func (ce *ContainerEnvs) InjectOtlpHttpEndpointEnvVar() {
	ce.injectNodeIpEnvVar()
	otlpHttpEndpoint := service.LocalTrafficOTLPHttpDataCollectionEndpoint("$(NODE_IP)")
	ce.injectEnvVarToPodContainer(commonconsts.OtelExporterEndpointEnvName, otlpHttpEndpoint)
}

```

RonFed (approved):
🥇 
left a small improvement suggestion

PR #2605, opened 2025-03-19
feat: derive rule-type in UI backend

BenElferink:
## Description

Instead of fetching rules, then mapping them on the client-side with rule-types...
We now map the rule-types in the backend, preventing an array-map on the client-side.

## How Has This Been Tested?

- [x] Manual Testing


frontend/services/instrumentationrule.go
```diff
@@ -15,6 +15,58 @@ import (
...
+		result["Namespace"] = codeAttributes.Namespace
+	}
+	if codeAttributes.Stacktrace != nil {
+		result["Stacktrace"] = codeAttributes.Stacktrace
+	}
+	return result
+}
+
+func deriveTypeFromRule(rule *model.InstrumentationRule) model.InstrumentationRuleType {
+	if rule.PayloadCollection != nil && !areAllKeysNull(convertPayloadCollectionToMap(rule.PayloadCollection)) {
```
alonbraymok, line 61: PayloadCollection is not support all keys as null?

frontend/services/instrumentationrule.go
```diff
@@ -25,44 +77,56 @@ func ListInstrumentationRules(ctx context.Context) ([]*model.InstrumentationRule
...
-			PayloadCollection:        convertPayloadCollection(rule.Spec.PayloadCollection),
-			CodeAttributes:           (*model.CodeAttributes)(rule.Spec.CodeAttributes),
-		})
+			Workloads:                convertWorkloads(r.Spec.Workloads),
+			InstrumentationLibraries: convertInstrumentationLibraries(r.Spec.InstrumentationLibraries),
+			PayloadCollection:        convertPayloadCollection(r.Spec.PayloadCollection),
+			CodeAttributes:           (*model.CodeAttributes)(r.Spec.CodeAttributes),
+		}
+
+		rule.Type = deriveTypeFromRule(rule)
```
alonbraymok, line 98: why this not inside the constructor?

frontend/services/instrumentationrule.go
```diff
@@ -15,6 +15,58 @@ import (
...
+		result["DbQuery"] = payload.DbQuery
+	}
+	if payload.Messaging != nil {
+		result["Messaging"] = payload.Messaging
+	}
+	return result
+}
+
+// Converts CodeAttributes to a map[string]interface{} for areAllKeysNull
+func convertCodeAttributesToMap(codeAttributes *model.CodeAttributes) map[string]interface{} {
```
alonbraymok, line 37: this need to retrun typed value insted of  map[string]interface{}

frontend/services/instrumentationrule.go
```diff
@@ -15,6 +15,58 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// Converts PayloadCollection to a map[string]interface{} for areAllKeysNull
+func convertPayloadCollectionToMap(payload *model.PayloadCollection) map[string]interface{} {
```
alonbraymok, line 19: this need to retrun typed value insted of  map[string]interface{}

frontend/services/instrumentationrule.go
```diff
@@ -25,44 +77,56 @@ func ListInstrumentationRules(ctx context.Context) ([]*model.InstrumentationRule
...
+		Notes:                    &r.Spec.Notes,
+		Disabled:                 &r.Spec.Disabled,
+		Mutable:                  mutable,
+		ProfileName:              profileName,
+		Workloads:                convertWorkloads(r.Spec.Workloads),
+		InstrumentationLibraries: convertInstrumentationLibraries(r.Spec.InstrumentationLibraries),
+		PayloadCollection:        convertPayloadCollection(r.Spec.PayloadCollection),
+	}
+
+	rule.Type = deriveTypeFromRule(rule)
```
alonbraymok, line 128: same

frontend/webapp/hooks/instrumentation-rules/useInstrumentationRuleCRUD.ts
```diff
@@ -48,33 +39,30 @@ export const useInstrumentationRuleCRUD = (): UseInstrumentationRuleCrud => {
...
-        const rule = res.updateInstrumentationRule;
-        const type = deriveTypeFromRule(rule);
-        addEntities(ENTITY_TYPES.INSTRUMENTATION_RULE, mapFetched([rule]));
-        notifyUser(STATUS_TYPE.SUCCESS, CRUD.UPDATE, `Successfully updated "${type}" rule`, rule.ruleId);
-      },
+  const [mutateUpdate, uState] = useMutation<{ updateInstrumentationRule: InstrumentationRule }, { ruleId: string; instrumentationRule: InstrumentationRuleFormData }>(UPDATE_INSTRUMENTATION_RULE, {
+    onError: (error) => notifyUser(STATUS_TYPE.ERROR, error.name || CRUD.UPDATE, error.cause?.message || error.message),
+    onCompleted: (res) => {
+      const rule = res.updateInstrumentationRule;
+      const type = deriveTypeFromRule(rule);
```
alonbraymok, line 61: we can use rule.type now right?
```suggestion
      const type = rule.type
```

frontend/services/instrumentationrule.go
BenElferink: If this rule is `PayloadCollection`, then the keys contain empty objects.
Null was true for client-side, and still is for server-side.

frontend/services/instrumentationrule.go
BenElferink: Because you don't have the `PayloadCollection` and `CodeAttribute` keys until after the constructor.

frontend/services/instrumentationrule.go
BenElferink: Changed to boolean.

alonbraymok (approved):

PR #2606, opened 2025-03-19
Remove odigosnameprocessor

edeNFed:
## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


collector/odigosotelcol/go.mod
```diff
@@ -4,6 +4,8 @@ module odigos.io/opentelemetry-collector/cmd/odigosotelcol
 
 go 1.23.0
 
+toolchain go1.23.4
```
blumamir, line 7: should we add this?

blumamir (approved):
LGTM

collector/odigosotelcol/go.mod
BenElferink: That is auto-added every time we run `cd collector && make genodigoscol generate`.
And then we have to remove it manually because the CI `verify-collector-ocb` will fail again.

blumamir:
> @edeNFed are there any permissions we are currently requesting from k8s to enable this? which we can now cleanup as well?

I think we need to cleanup this as well:
https://github.com/odigos-io/odigos/blob/164ba3a18e72a8c5d98cef3aa60d12413e22b6e3/autoscaler/controllers/datacollection/daemonset.go#L257

PR #2608, opened 2025-03-19
Create sqllite in UI pod

tamirdavid1:
## Description

This is the first PR in a series aimed at removing the InstrumentationInstance object, which represents an instrumented process. Right now, the UI will run this database but won’t use it, allowing us to update the schema as needed. Once all changes are made, the UI will have the state of all processes stored in the database, so it can query directly instead of fetching all InstrumentationInstance objects from the cluster.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [X] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


BenElferink (approved):

frontend/services/db/models.go
```diff
@@ -0,0 +1,101 @@
...
+// - `k8s_pod_name` → The pod in which the process is running (pods can have multiple containers).
+// - `k8s_namespace_name` → Ensures uniqueness across namespaces (pods can have the same name in different namespaces).
+// - `k8s_container_name` → A pod may contain multiple containers, each with its own PID space.
+// - `created_at` → Ensures uniqueness even if the same PID is reused after a pod restart.
+
+type InstrumentedProcess struct {
+	OdigletName       string    `gorm:"column:odiglet_name; type:text; not null"`                           // Agent/collector name
+	NodeName          string    `gorm:"column:k8s_node_name; type:text; not null"`                          // Kubernetes node
+	WorkloadName      string    `gorm:"column:workload_name; type:text; not null"`                          // Kubernetes workload name
+	WorkloadKind      string    `gorm:"column:workload_kind; type:text; not null"`                          // Kubernetes workload kind [sts/ds/deploy]
```
RonFed, line 23: Can we define it as an enum in the db schema and in go?

frontend/services/db/models.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// InstrumentedProcessError stores processes that are marked as unhealthy (`healthy = false`)
+// along with their `healthy_reason`. This table automatically gets updated via a database
+// trigger whenever a new healthy process is inserted into `InstrumentedProcesses`.
+// It also has a foreign key constraint with `OnDelete:CASCADE`, ensuring that
+// errors are automatically removed when the corresponding process is deleted.
+
+type InstrumentedProcessError struct {
+	PID           int       `gorm:"primaryKey; column:process_pid; not null"`
```
RonFed, line 46:
```suggestion
	ContainerPID           int       `gorm:"primaryKey; column:container_pid; not null"`
```

frontend/services/db/models.go
```diff
@@ -0,0 +1,101 @@
...
+	PodName       string    `gorm:"primaryKey; column:k8s_pod_name; type:text; not null"`
+	Namespace     string    `gorm:"primaryKey; column:k8s_namespace_name; type:text; not null"`
+	ContainerName string    `gorm:"primaryKey; column:k8s_container_name; type:text; not null"`
+	CreatedAt     time.Time `gorm:"primaryKey; column:created_at; not null; default:CURRENT_TIMESTAMP"`
+	HealthyReason string    `gorm:"column:healthy_reason; type:text; not null"`
+
+	// Foreign Key Reference (Auto-delete errors when the process is removed)
+	InstrumentedProcess InstrumentedProcess `gorm:"constraint:OnDelete:CASCADE;foreignKey:PID,PodName,Namespace,ContainerName,CreatedAt"`
```
RonFed, line 54: This will not add more fields to the table, right?

frontend/services/db/models.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (InstrumentedProcessError) TableName() string {
+	return "instrumented_processes_errors"
+}
+
+func AutoMigrate(db *gorm.DB) {
```
RonFed, line 61: What is the goal of this function? what are we migrating from?

frontend/services/db/models.go
tamirdavid1: It wont, just a foreign key

frontend/services/db/models.go
tamirdavid1: We dont, just an easy way to create the schema using the go-orm, for future uses maybe this ORM will be necessary

tamirdavid1: i'll change it to init schema

RonFed (approved):

PR #2619, opened 2025-03-23
Add new samplers to sampling processor

ErikNaz-odigos:
## Description

Added two new samplers for the `odigossamplingprocessor`
1. servicename sampler - samples only traces that contain spans from service X
2. spanattribute sampler - samples traces based on an attribute name, value and an operation [exists, equals, not equals]
## How Has This Been Tested?

- [x] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


autoscaler/controllers/actions/sampling/servicenamesampler.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (h *ServiceNameSamplerHandler) GetActionScope(action metav1.Object) string {
+	return "global" // or "service" if you want it more scoped
+}
+
+func (cfg *ServiceNameConfig) Validate() error {
+	if cfg.ServiceName == "" {
+		return errors.New("service_name cannot be empty")
+	}
+	if cfg.FallbackSamplingRatio < 0 || cfg.FallbackSamplingRatio > 100 {
```
tamirdavid1, line 75: I wonder if ServiceName sampler should have fallback which will make setting 1 rule of it potentially blocked all other services spans (if fallback ratio == 0).
I guess that most of the users will use this to sample X amount of spans from a given Service.
WDYT?

autoscaler/controllers/actions/sampling/servicenamesampler.go
```diff
@@ -0,0 +1,79 @@
...
+	return rules
+}
+
+func (h *ServiceNameSamplerHandler) GetActionReference(action metav1.Object) metav1.OwnerReference {
+	a := action.(*actionv1.ServiceNameSampler)
+	return metav1.OwnerReference{APIVersion: a.APIVersion, Kind: a.Kind, Name: a.Name, UID: a.UID}
+}
+
+func (h *ServiceNameSamplerHandler) GetActionScope(action metav1.Object) string {
+	return "global" // or "service" if you want it more scoped
```
tamirdavid1, line 68: following the next comment, maybe consider moving it to service

autoscaler/controllers/actions/sampling/spanattributesampler.go
```diff
@@ -0,0 +1,89 @@
...
+	if cfg.AttributeKey == "" {
+		return errors.New("attribute_key cannot be empty")
+	}
+	if cfg.Condition != "exists" && cfg.Condition != "equals" && cfg.Condition != "not_equals" {
+		return errors.New("condition must be one of: exists, equals, not_equals")
+	}
+	if (cfg.Condition == "equals" || cfg.Condition == "not_equals") && cfg.ExpectedValue == "" {
+		return errors.New("expected_value is required for 'equals' and 'not_equals'")
+	}
+	if cfg.FallbackSamplingRatio < 0 || cfg.FallbackSamplingRatio > 100 {
```
tamirdavid1, line 85: same thoughts as above

collector/processors/odigossamplingprocessor/internal/sampling/servicename.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (s *ServiceNameRule) Validate() error {
+	if s.ServiceName == "" {
+		return errors.New("service name cannot be empty")
+	}
+	return nil
+}
+
+func (s *ServiceNameRule) Evaluate(td ptrace.Traces) (filterMatch bool, conditionMatch bool, fallbackRatio float64) {
+	logger, _ := zap.NewDevelopment()
```
tamirdavid1, line 26: I dont think that we should log for every match it will be too verbose

collector/processors/odigossamplingprocessor/internal/sampling/spanattribute.go
```diff
@@ -0,0 +1,74 @@
...
+func (s *SpanAttributeRule) Evaluate(td ptrace.Traces) (filterMatch bool, conditionMatch bool, fallbackRatio float64) {
+	rs := td.ResourceSpans()
+	for i := 0; i < rs.Len(); i++ {
+		scopeSpans := rs.At(i).ScopeSpans()
+		for j := 0; j < scopeSpans.Len(); j++ {
+			spans := scopeSpans.At(j).Spans()
+			for k := 0; k < spans.Len(); k++ {
+				attrs := spans.At(k).Attributes()
+				val, found := attrs.Get(s.AttributeKey)
+				if !found || val.Type() != pcommon.ValueTypeStr {
```
tamirdavid1, line 54: what will be the val.Type of status_code value for example? 
in case it is non-string, i think maybe we should cast it to string?
e.g someone want to filter for status_code == 504 we use ExpectedValue as a string and it never match

tamirdavid1: maybe worth adding a test for it

collector/processors/odigossamplingprocessor/rule_engine.go
```diff
@@ -51,7 +51,7 @@ func (re *RuleEngine) ShouldSample(td ptrace.Traces) bool {
 			return decision
 		}
 	}
-	return true
+	return false
```
tamirdavid1, line 54: This means that the default in case non of the rules satisfied we will not sample right ?

docs/api-reference/actions.odigos.io.v1alpha1.mdx
```diff
@@ -11,6 +11,8 @@
 - [PiiMasking](#PiiMasking)
 - [ProbabilisticSampler](#ProbabilisticSampler)
 - [RenameAttribute](#RenameAttribute)
+- [ServiceNameSampler](#ServiceNameSampler)
```
tamirdavid1, line 14: You should also add this on: `docs/mint.json`, currently it is not available on the mainpage bar

docs/pipeline/actions/sampling/servicenamesampler.mdx
```diff
@@ -0,0 +1,100 @@
+---
+title: "Service Name Sampler"
+description: "This action is an [Endpoint Action](/pipeline/actions/introduction#scope-categories) that samples traces based on the presence of specified services within a trace."
```
tamirdavid1, line 3: You set this as Global

docs/pipeline/actions/sampling/servicenamesampler.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+import BetaWarning from '/snippets/beta-warning.mdx';
+import AssumeNoMeaning from '/snippets/assume-no-meaning.mdx';
+
+<BetaWarning />
+
+## Considerations
+
+<Warning>
+  Before enabling **service name sampler**, please note the following:
```
tamirdavid1, line 16: any chance you can edit error sampling document if you already touch docs
Written there > Before enabling `probabilistic sampler` (which suppose to be error sampler)

docs/pipeline/actions/sampling/spanattributesampler.mdx
```diff
@@ -0,0 +1,111 @@
...
+  </Accordion>
+</AccordionGroup>
+
+<Info>
+  Traces containing at least one span matching the attribute filters will always be retained. If no span matches, traces will be sampled according to the fallback ratio.
+</Info>
+
+## Basic Example
+
+The following example retains all traces with spans containing an attribute `error` set to `true`. Additionally, it retains 10% of all other traces.
```
tamirdavid1, line 82: nit: i think better to take other example then error = true (because we have already error sampler). install the kv-mall suite it has some interesting spans communicating with brokers etc.

docs/pipeline/actions/sampling/servicenamesampler.mdx
```diff
@@ -0,0 +1,100 @@
...
+  <Accordion title="disabled">
+    **disabled** `boolean` : Allows you to temporarily disable the action, but keep it saved for future use.
+    - This field is *optional*, defaults to `false`
+  </Accordion>
+  <Accordion title="signals *">
+    **signals** `string[]` : An array with the signals that the processor will act on.
+    - This field is *required*
+    - Supported values: `TRACES`
+  </Accordion>
+  <Accordion title="service_names *">
```
tamirdavid1, line 55: I think this example wont work, the key for services is now services and not service_names

tamirdavid1: it should be something like
services:
    - service_name: x
       fallbacksamplingRatio: y

docs/pipeline/actions/sampling/spanattributesampler.mdx
```diff
@@ -0,0 +1,111 @@
...
+  <Accordion title="signals *">
+    **signals** `string[]` : An array with the signals that the processor will act on.
+    - This field is *required*
+    - Supported values: `TRACES`
+  </Accordion>
+  <Accordion title="attribute_filters *">
+    **attribute_filters** `object[]` : Specifies which span attributes should trigger trace sampling.
+    - This field is *required*
+
+    <AccordionGroup>
```
tamirdavid1, line 59: where is Condition [required] / expected value [semi required ]

docs/pipeline/actions/sampling/spanattributesampler.mdx
```diff
@@ -0,0 +1,111 @@
...
+</Info>
+
+## Basic Example
+
+The following example retains all traces with spans containing an attribute `error` set to `true`. Additionally, it retains 10% of all other traces.
+
+<Steps>
+  <Step>
+    Create a YAML file with the following content:
+
```
tamirdavid1, line 87: once you fix the example
I think it is best to give 3 example:
`equals, not equals, exists`
and explain what will be the implications setting them on this service and globally on the environment

docs/pipeline/actions/sampling/servicenamesampler.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+- Reducing the number of collected traces based on essential services optimizes storage and analysis costs.
+
+## Configuration Options
+
+<AccordionGroup>
+  <Accordion title="actionName">
+    **actionName** `string` : Allows you to attach a meaningful name to the action for convenience.
+    - This field is *optional*
+    - <AssumeNoMeaning />
```
tamirdavid1, line 39: Seems like assume no review is not looking good actually on the UI
image

autoscaler/controllers/actions/sampling/servicenamesampler.go
ErikNaz-odigos: WDYM? I don't really understand this comment.
Currently the filter supports fallback sample ration to sample other services.
But we can add an option to sample X% from a specific service

autoscaler/controllers/actions/sampling/spanattributesampler.go
ErikNaz-odigos: Or even better, support putting filters (All or most of them), on different scopes in order to allow more fine grained filtering

collector/processors/odigossamplingprocessor/internal/sampling/servicename.go
ErikNaz-odigos: Good catch, That's a debug message i've left inside

collector/processors/odigossamplingprocessor/internal/sampling/spanattribute.go
ErikNaz-odigos: I don't want to starting casting stuff to strings and assuming things.
If we want to support other types we can change the implementation to allow multiple types and operations.
Currently we have `equals`, `not equals` and  `exists` for string types.
And a typical definition looks like this:
`
{
	"attribute_key": "some.attr"
	"condition": "equals"
	"expected_value": "prod"
	"fallback_sampling_ratio": 40
}
`

We can do something alone the lines of
`
{
	"attribute_key": "some.attr"
	"wanted": {
		"type": "int", # Support all types in openTelemetry spans
		"condition": "equals" # For example support equals, less, above, lessEquals, aboveEquals
		"expected_value": 4
	}
	"fallback_sampling_ratio": 40
}
`

collector/processors/odigossamplingprocessor/rule_engine.go
ErikNaz-odigos: Yeah, don't sample on default

ErikNaz-odigos: All rules and also the fallback ratio didn't catch

api/actions/v1alpha1/servicenamesampler_types.go
```diff
@@ -0,0 +1,58 @@
...
+	Signals    []common.ObservabilitySignal `json:"signals"`
+
+	// List of services to sample based on presence in the trace
+	// +kubebuilder:validation:Required
+	Services []ServiceNameFilter `json:"services"`
+}
+
+type ServiceNameFilter struct {
+	// +kubebuilder:validation:Required
+	ServiceName string `json:"service_name"`
```
alonkeyval, line 22: it will be great to align all the json var to camelCase instead of snake_case: service_name => serviceName

current state make as map same of the action in the ui

alonkeyval: take a look how @blumamir implement cluster info for example: addclusterinfo_types.go

docs/pipeline/actions/sampling/latencysampler.mdx
```diff
@@ -1,6 +1,6 @@
 ---
 title: "Latency Sampler"
-description: "This action is an [Endpoint Action](/pipeline/actions/introduction#scope-categories) that samples traces based on their duration for a specific service and endpoint (HTTP route) filter."
+description: "This action is an [Service Action](/pipeline/actions/introduction#scope-categories) that samples traces based on span attributes values."
```
tamirdavid1, line 3: Service action?

docs/pipeline/actions/sampling/probabilisticsampler.mdx
```diff
@@ -8,7 +8,9 @@ icon: "tag"
 import BetaWarning from '/snippets/beta-warning.mdx';
 import AssumeNoMeaning from '/snippets/assume-no-meaning.mdx';
 
-<BetaWarning />
+<Note>
+  <strong>Deprecation Notice:</strong> This feature is deprecated and will be removed in version 1.0.170.
+</Note>
```
tamirdavid1, line 13: Lets bump it to 1.0.180, it will be released on 1.0.171 and still exists

api/actions/v1alpha1/spanattributesampler_types.go
```diff
@@ -0,0 +1,238 @@
...
+	Operation string `json:"operation"`
+
+	// JsonPath is required for:
+	//   - jsonpath_exists
+	//   - key_equals
+	//   - key_not_equals
+	//
+	// It is a JSONPath expression used to navigate the JSON structure.
+	// Example: $.user.role or $.events[0].code
+	JsonPath string `json:"json_path,omitempty"`
```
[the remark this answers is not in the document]
ErikNaz-odigos: Yeah, `jsonPath` is required only when accessing attributes inside of the json (Inside of the attribute)

collector/processors/odigossamplingprocessor/internal/sampling/spanattribute.go
```diff
@@ -0,0 +1,323 @@
...
+			"jsonpath_exists":  true,
+			"contains_key":     true,
+			"not_contains_key": true,
+			"key_equals":       true,
+			"key_not_equals":   true,
+		}
+		if !validOps[s.Operation] {
+			return errors.New("invalid json operation")
+		}
+		// For all JSON operations, a jsonPath is required.
```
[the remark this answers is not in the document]
ErikNaz-odigos: Not for all, only tested for some
But i've updated the comment

collector/processors/odigossamplingprocessor/internal/sampling/spanattribute.go
```diff
@@ -0,0 +1,323 @@
...
+						}
+					}
+					expectedBool, err := strconv.ParseBool(s.ExpectedValue)
+					if err != nil || attr.Type() != pcommon.ValueTypeBool {
+						continue
+					}
+					if s.Operation == "equals" && attr.Bool() == expectedBool {
+						return true, true, fallbackRatio
+					}
+				case TypeJSON:
```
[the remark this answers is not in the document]
ErikNaz-odigos: You're correct I've kept only `is_valid_json`, `is_invalid_json`, `contains_key` and `not_contains_key`

collector/processors/odigossamplingprocessor/rule_engine_test.go
```diff
@@ -199,5 +180,53 @@ func TestRuleEngine_MultipleEndpointRules_OneSatisfied(t *testing.T) {
...
+	// Trace with an actual error → global rule satisfied → takes precedence
+	trace := testutil.NewTrace().
+		AddResource("any-service").
+		AddSpan("bad op", testutil.WithStatus(ptrace.StatusCodeError)).
+		Done().Build()
+
 	assert.True(t, engine.ShouldSample(trace))
 }
+
+// TestRuleEngine_EmptyRules ensures engine handles config with no rules
```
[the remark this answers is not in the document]
ErikNaz-odigos: I disagree with you, an empty rule_engine is an invalid state and I would like to stop all reporting in that case
wdyt?

tamirdavid1 (approved):

PR #2621, opened 2025-03-23
test: simple-demo use simple-trace-db instead of tempo

blumamir:
This PR migrates the e2e tests from tempo to simple-trace-db.

Tempo introduce quite a lot of complexity which cases flakiness in the tests due to queries with wrong results


BenElferink (approved):

tamirdavid1 (approved):

tests/e2e/source/queries/resource-attributes.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: e2e.tests.odigos.io/v1
+kind: TraceTest
+description: |
+  This test check the following resource attributes:
+  A. odigos.version attribute exists on all spans, since it's an upgrade, we can have version of both the old and new odigos deployments.
```
RonFed, line 5: this is specific to upgrade test? if so, it should not be under `sources` I guess or the comment should be modified

RonFed (approved):

tests/e2e/source/queries/resource-attributes.yaml
blumamir: fixed

PR #2651, opened 2025-03-27
cli: add update-offsets command

damemi:
## Description

Depends on:

* https://github.com/odigos-io/enterprise-go-instrumentation/pull/1168
* https://github.com/odigos-io/odigos-enterprise/pull/748/

This adds a basic command to pull the latest published offsets data, update a configmap with the data, and patch the Odiglet to mount the data.

Use latest offsets with `odigos update-offsets`

Revert changes with `odigos update-offsets --default`

This also updates Odiglet install/upgrade to preserve custom offsets (for when user upgrades offsets before upgrading odigos, see comment in-line)

## How Has This Been Tested?

- [ ] Added Unit Tests
- [x] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

- [x] New Query/feature supported in all the k8s versions supported by Odigos

## User Facing Changes


cli/cmd/offsets.go
```diff
@@ -0,0 +1,235 @@
...
+odigos update-offsets --default
+`,
+	Run: func(cmd *cobra.Command, args []string) {
+		ctx := cmd.Context()
+		client := cmdcontext.KubeClientFromContextOrExit(ctx)
+		ns := cmd.Flag("namespace").Value.String()
+
+		resp, err := http.Get(consts.GoOffsetsPublicURL)
+		if err != nil {
+			fmt.Println(fmt.Sprintf("\033[31mERROR\033[0m Cannot get latest offsets: %s", err))
```
blumamir, line 54: can we check here if the error might be no internet access and add a note about it? so users are aware this requires internet and cannot work offline.

cli/cmd/offsets.go
```diff
@@ -0,0 +1,235 @@
...
+
+# Revert to using the default offsets data shipped with Odigos
+odigos update-offsets --default
+`,
+	Run: func(cmd *cobra.Command, args []string) {
+		ctx := cmd.Context()
+		client := cmdcontext.KubeClientFromContextOrExit(ctx)
+		ns := cmd.Flag("namespace").Value.String()
+
+		resp, err := http.Get(consts.GoOffsetsPublicURL)
```
blumamir, line 52: suggestion: before calling this function, let's check the current tier and avoid updating for community version, as this feature is not expected to work with community odiglet and might mislead users.

common/consts/consts.go
```diff
@@ -47,6 +47,12 @@ const (
 	InstrumentationConfig   = "InstrumentationConfig"
 	InstrumentationInstance = "InstrumentationInstance"
 	Destination             = "Destination"
+
+	// ConfigMap used to store custom/updated Go instrumentation offsets
+	GoOffsetsPublicURL = "https://storage.googleapis.com/odigos-cloud/offset_results.json"
+	GoOffsetsConfigMap = "odigos-go-offsets"
+	GoOffsetsFileName  = "offset_results.json"
+	GoOffsetsEnvVar    = "OTEL_GO_OFFSETS_FILE"
```
blumamir, line 55: as these are k8s specific, they belong in the api module `api/k8sconsts`, the `common` module is shared with vmagent.

cli/cmd/resources/odiglet.go
```diff
@@ -505,6 +510,84 @@ func NewOdigletDaemonSet(ns string, version string, imagePrefix string, imageNam
...
+	// This also means a user could be still using old offsets, and might want to revert that to pull in latest Odigos changes
+	// TODO: Detect superceded offsets automatically so user doesn't need to manually revert
+	_, err := client.Clientset.CoreV1().ConfigMaps(ns).Get(ctx, consts.GoOffsetsConfigMap, metav1.GetOptions{})
+	if err == nil {
+		// Add offsets volume (if not already exist)
+		volumes := ds.Spec.Template.Spec.Volumes
+		if volumes == nil {
+			volumes = make([]v1.Volume, 0)
+		}
+		addVolume := true
```
blumamir, line 530: suggestion: this function is the same as the one in `offsets.go` right? can we extract them into a common util and use from both places? 
can also be more readable so it's just a function call here and not 50 lines of code :)

cli/cmd/resources/odiglet.go
```diff
@@ -588,7 +671,7 @@ func (a *odigletResourceManager) InstallFromScratch(ctx context.Context) error {
 
 	// before creating the daemonset, we need to create the service account, cluster role and cluster role binding
 	resources = append(resources,
-		NewOdigletDaemonSet(a.ns, a.odigosVersion, a.config.ImagePrefix, a.managerOpts.ImageReferences.OdigletImage, a.odigosTier, a.config.OpenshiftEnabled,
+		NewOdigletDaemonSet(ctx, a.client, a.ns, a.odigosVersion, a.config.ImagePrefix, a.managerOpts.ImageReferences.OdigletImage, a.odigosTier, a.config.OpenshiftEnabled,
```
blumamir, line 674: the `client` here is actually meant to be used for `ApplyResources`.
The `NewOdigletDaemonSet` function only cares if the file is present or not, so you can consider hoisting this query to a prior stage and just pass a bool to `odigletResourceManager` and indicate if offset file should be injected or not.

This will help keep the resource manager `New...` functions as pure functions, that cannot have errors or side effect, and will not `os.Exit` the process.

cli/cmd/resources/odiglet.go
```diff
@@ -505,6 +510,84 @@ func NewOdigletDaemonSet(ns string, version string, imagePrefix string, imageNam
...
+			}
+		}
+		if addEnvVar {
+			envVars = append(envVars, v1.EnvVar{Name: consts.GoOffsetsEnvVar, Value: offsetFileMountPath + "/" + consts.GoOffsetsFileName})
+			ds.Spec.Template.Spec.Containers[0].Env = envVars
+		}
+	} else {
+		if !apierrors.IsNotFound(err) {
+			fmt.Println(fmt.Sprintf("\033[31mERROR\033[0m Unable to get Go offsets ConfigMap: %s", err))
+			os.Exit(1)
```
blumamir, line 586: as mentioned in another comment, I think we should only `os.Exit` at the top, so it will not surprise anyone in the future. especially since these are used in install, and upgrade which are more sensitive commands

cli/cmd/resources/odiglet.go
```diff
@@ -505,6 +510,84 @@ func NewOdigletDaemonSet(ns string, version string, imagePrefix string, imageNam
...
+	_, err := client.Clientset.CoreV1().ConfigMaps(ns).Get(ctx, consts.GoOffsetsConfigMap, metav1.GetOptions{})
+	if err == nil {
+		// Add offsets volume (if not already exist)
+		volumes := ds.Spec.Template.Spec.Volumes
+		if volumes == nil {
+			volumes = make([]v1.Volume, 0)
+		}
+		addVolume := true
+		for _, vol := range volumes {
+			if vol.Name == consts.GoOffsetsConfigMap {
```
blumamir, line 532: can this really happen? the volumes is the array that we created right above which will never contain this config map.

consider instead of mutating the ds right after creating it, to have `extraVolume` array with 0 or 1 elements which you create at the start of this function, and then append it with `...` spread right into the volumes in the manifest. and also for the rest of the changes here.

I think it will be more maintainable this way and will align better with the other existing implementations in this package

cli/cmd/resources/odiglet.go
```diff
@@ -505,6 +510,84 @@ func NewOdigletDaemonSet(ns string, version string, imagePrefix string, imageNam
...
+	// 1) User updates offsets to support Go 1.24
+	// 2) Odigos releases support for Go 1.24
+	// 3) User updates offsets to support Go 1.24.1
+	// 4) User updates Odigos (no default support for 1.24.1)
+	// This also means a user could be still using old offsets, and might want to revert that to pull in latest Odigos changes
+	// TODO: Detect superceded offsets automatically so user doesn't need to manually revert
+	_, err := client.Clientset.CoreV1().ConfigMaps(ns).Get(ctx, consts.GoOffsetsConfigMap, metav1.GetOptions{})
+	if err == nil {
+		// Add offsets volume (if not already exist)
+		volumes := ds.Spec.Template.Spec.Volumes
```
blumamir, line 526: also need to support it in helm

docs/cli/odigos_update-offsets.mdx
```diff
@@ -0,0 +1,55 @@
+---
+title: "odigos update-offsets"
+sidebarTitle: "odigos update-offsets"
+---
+## odigos update-offsets
```
blumamir, line 5: add a link from https://docs.odigos.io/instrumentations/golang/ebpf#about-go-offsets to this CLI command so users are aware how to fix the offset issue if they encounter it

cli/cmd/offsets.go
```diff
@@ -0,0 +1,235 @@
...
+	"github.com/odigos-io/odigos/common/consts"
+)
+
+var (
+	offsetFileMountPath = "/offsets"
+	useDefault          bool
+)
+
+var offsetsCmd = &cobra.Command{
+	Use:   "update-offsets",
```
blumamir, line 29: WDYT about adding it as subcommand to `odigos pro` so it's clear this feature is not expected to work for community. this will also keep the number of top level cli commands from growing.

blumamir: https://github.com/odigos-io/odigos/blob/8ca0480a57e4152a961d162bae095f217804b081/cli/cmd/pro.go#L22

cli/cmd/offsets.go
```diff
@@ -0,0 +1,235 @@
...
+			envVars = append(envVars, v1.EnvVar{Name: consts.GoOffsetsEnvVar, Value: offsetFileMountPath + "/" + consts.GoOffsetsFileName})
+			ds.Spec.Template.Spec.Containers[0].Env = envVars
+		}
+	}
+	return ds
+}
+
+func init() {
+	rootCmd.AddCommand(offsetsCmd)
+	offsetsCmd.Flags().StringVarP(&namespaceFlag, "namespace", "n", consts.DefaultOdigosNamespace, "target k8s namespace for Odigos installation")
```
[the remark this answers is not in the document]
damemi: The intent here is to keep the configmap in the installed odigos namespace, so we do actually want to use the same flag right?

cli/cmd/offsets.go
damemi: sgtm!

cli/cmd/resources/odiglet.go
damemi: I created a new resource function `NewOdigletGoOffsetsConfigMap`, which does still check for an existing map (to not overwrite the data), and handles errors better through the current InstallFromScratch flow. Let me know what you think

api/k8sconsts/odiglet.go
```diff
@@ -17,6 +17,12 @@ const (
 	OdigletOSSInstalledLabel        = "odigos.io/odiglet-oss-installed"
 	OdigletEnterpriseInstalledLabel = "odigos.io/odiglet-enterprise-installed"
 	OdigletInstalledLabelValue      = "true"
+
+	// ConfigMap used to store custom/updated Go instrumentation offsets
+	GoOffsetsPublicURL = "https://storage.googleapis.com/odigos-cloud/offset_results_min.json"
+	GoOffsetsConfigMap = "odigos-go-offsets"
+	GoOffsetsFileName  = "offset_results.json"
```
blumamir, line 24: should we have the word "go" in this file name to make it obvious that these offsets are only for golang? so in the future if we add additional offsets things are better structured.

tests/e2e/cli-upgrade/chainsaw-test.yaml
```diff
@@ -109,13 +109,20 @@ spec:
         - assert:
             file: ../../common/assert/simple-demo-instrumented.yaml
 
+    - name: Use custom Go offsets file to make sure it works with upgrade
+      try:
+        - script:
+            content: ../../../cli/odigos update-offsets --namespace odigos-test
```
blumamir, line 115:
```suggestion
            content: ../../../cli/odigos pro update-offsets --namespace odigos-test
```

tests/e2e/cli-upgrade/chainsaw-test.yaml
```diff
@@ -109,13 +109,20 @@ spec:
         - assert:
             file: ../../common/assert/simple-demo-instrumented.yaml
 
+    - name: Use custom Go offsets file to make sure it works with upgrade
```
blumamir, line 112: why is offsets a special case in regards to upgrade?

I think this test should be added somewhere else and not in `cli-upgrade`

tests/e2e/cli-upgrade/chainsaw-test.yaml
```diff
@@ -189,3 +196,11 @@ spec:
...
             name: odiglet
             namespace: odigos-test
+
+    - name: Revert custom Go offsets
+      try:
+        - script:
+            content: ../../../cli/odigos update-offsets --default --namespace odigos-test
+        - assert:
+            file: assert-odigos-upgraded.yaml
+            timeout: 3m
```
blumamir, line 206: do we really need `3m`?? it's expected to take just few seconds at most, no?

cli/cmd/offsets.go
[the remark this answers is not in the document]
damemi: > but not to reuse the same go variable as this might lead to unintended side-effects like https://github.com/odigos-io/odigos/pull/2389 .

I'm saying I think we actually do want to make sure we're reusing the same Go variable, because this should always be acting on the same namespace as `install.go`. In #2389, the problem was that the `namespace` flag could be applied to *any* namespace. Here, the configmap is always in the `odigos-system` namespace (or wherever it was installed)

cli/cmd/pro.go
```diff
@@ -92,10 +104,189 @@ func executeRemoteUpdateToken(ctx context.Context, client *kube.Client, namespac
...
+	},
+}
+
+func getLatestOffsets(revert bool) []byte {
+	if revert {
+		return []byte{}
+	}
+
+	resp, err := http.Get(k8sconsts.GoOffsetsPublicURL)
+	if err != nil {
```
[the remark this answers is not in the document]
damemi: I don't think there is a way to definitively say if the error is due to internet connection? For example an error like `network unreachable` could be due to more than just that the user doesn't have internet, and even in that case I think the http error itself should be enough info. wdyt?

We could add a "possibly check your internet connection" warning but it seems a little redundant if the error itself indicates that as a possibility

cli/cmd/pro.go
```diff
@@ -92,10 +104,189 @@ func executeRemoteUpdateToken(ctx context.Context, client *kube.Client, namespac
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      k8sconsts.GoOffsetsConfigMap,
+					Namespace: ns,
+				},
+				Data: map[string]string{
+					k8sconsts.GoOffsetsFileName: string(data),
+				},
+			}
+
+			cm, err = client.Clientset.CoreV1().ConfigMaps(ns).Create(ctx, cm, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
damemi: I added this to support the current version (and for testing against the current version). So, the user could start using the new Go offsets in the next release of the CLI *without* needing to update their installed Odigos.

If you think that's unnecessary, and we can just say this is only supported in the next version of Odigos, we can get rid of this

cli/cmd/pro.go
```diff
@@ -92,10 +104,189 @@ func executeRemoteUpdateToken(ctx context.Context, client *kube.Client, namespac
...
+
+	data, err := io.ReadAll(resp.Body)
+	if err != nil {
+		fmt.Println(fmt.Sprintf("\033[31mERROR\033[0m Unable to read response body: %s", err))
+		os.Exit(1)
+	}
+	return data
+}
+
+func updateOdigletDaemonSet(ctx context.Context, client *kube.Client, ns string, revert bool) *appsv1.DaemonSet {
```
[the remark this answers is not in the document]
damemi: Same reasoning as https://github.com/odigos-io/odigos/pull/2651#discussion_r2027510354

tests/e2e/cli-upgrade/chainsaw-test.yaml
damemi: Want to make sure upgrade doesn't overwrite existing custom offsets. For example:

1. User is on odigos 1.0.167
2. User updates app to go 1.24.X
3. Users pulls custom offsets for go 1.24.X
4. Odigos 1.0.168 releases support for go 1.24.X, but user does not yet update Odigos
5. User updates their app to go 1.24.X+1
6. User updates Odigos

Since this update flow is async, and there could be a delay between the user updating their own apps vs updating Odigos, the latest Odigos could actually be *behind* what the user is expecting to run.

This does have the reverse effect too, though, and the user could end up behind what Odigos supports. Maybe we should add more advanced logic to check if the default Odigos offsets are ahead of what's currently in the cluster and use whichever is latest

api/k8sconsts/odiglet.go
damemi: Sure, we should probably update the published file name too but we can do that

cli/cmd/pro.go
```diff
@@ -92,10 +103,190 @@ func executeRemoteUpdateToken(ctx context.Context, client *kube.Client, namespac
...
+# Pull the latest offsets and restart Odiglet
+odigos pro update-offsets
+
+# Revert to using the default offsets data shipped with Odigos
+odigos pro update-offsets --default
+`,
+	Run: func(cmd *cobra.Command, args []string) {
+		ctx := cmd.Context()
+		client := cmdcontext.KubeClientFromContextOrExit(ctx)
+		ns := cmd.Flag("namespace").Value.String()
```
blumamir, line 129: consider using the existing function [GetOdigosNamespace](https://github.com/odigos-io/odigos/blob/f7408eb106b832a77cfb937a2c538be24a762b1d/cli/cmd/resources/namespace.go#L25)

and remove the namespace from the arguments. this is what we do in all other commands except for install

helm/odigos/templates/odiglet/daemonset.yaml
```diff
@@ -88,6 +88,8 @@ spec:
                   fieldPath: metadata.namespace
             - name: OTEL_LOG_LEVEL
               value: info
+            - name: OTEL_GO_OFFSETS_FILE
+              value: /offsets/offset_results.json
```
blumamir, line 92: WDYT about wrapping the offsets related sections in helm with `{{- if .Values.onPremToken }}`?

cli/cmd/resources/odiglet.go
```diff
@@ -292,9 +298,25 @@ func NewOdigletDaemonSet(ns string, version string, imagePrefix string, imageNam
 				Name:  k8sconsts.CustomContainerRuntimeSocketEnvVar,
 				Value: customContainerRuntimeSocketPath})
 	}
-
-	additionalVolumes := append(customContainerRuntimeSocketVolumes, odigosSeLinuxHostVolumes...)
-	additionalVolumeMounts := append(customContainerRunetimeSocketVolumeMounts, odigosSeLinuxHostVolumeMounts...)
+	additionalVolumes = append(additionalVolumes, customContainerRuntimeSocketVolumes...)
+	additionalVolumeMounts = append(additionalVolumeMounts, customContainerRunetimeSocketVolumeMounts...)
+
+	goOffsetsVolume := corev1.Volume{
```
blumamir, line 304: consider skipping the offsets manifest section for tier community

cli/cmd/pro.go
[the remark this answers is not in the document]
blumamir: If someone uses current version, then it will still not work, right? as odiglet's are not yet ready to process it.

I can't think of a real scenario where this solves a problem, but don't mind keeping it if you prefer

cli/cmd/pro.go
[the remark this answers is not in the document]
blumamir: For simplicity, I would suggest to remove this part and let the resource managers take care of this part

cli/cmd/pro.go
```diff
@@ -92,10 +103,190 @@ func executeRemoteUpdateToken(ctx context.Context, client *kube.Client, namespac
...
+
 func init() {
 	rootCmd.AddCommand(proCmd)
 
 	proCmd.Flags().String("onprem-token", "", "On-prem token for Odigos")
 	proCmd.MarkFlagRequired("onprem-token")
 	proCmd.PersistentFlags().BoolVarP(&updateRemoteFlag, "remote", "r", false, "use odigos ui service in the cluster to update the onprem token")
+
+	proCmd.AddCommand(offsetsCmd)
+	offsetsCmd.Flags().StringVarP(&namespaceFlag, "namespace", "n", consts.DefaultOdigosNamespace, "target k8s namespace for Odigos installation")
```
blumamir, line 290: as mentioned in another comment, I support auto detecting ns like other commands.
If we choose to keep it, update the cmd text (copy-paste from install) :)

blumamir (approved):

damemi:
@blumamir thanks for the review, I applied your suggestions. This is just blocked on https://github.com/odigos-io/odigos-enterprise/pull/825 then this can merge

PR #2653, opened 2025-03-30
[GEN-2127]: add Seq destination

BenElferink:
## Description

Add [Seq](https://datalust.co/seq) as destination.

## How Has This Been Tested?

- [x] Manual Testing

---

Proof of data:

Screenshot 2025-03-30 at 17 19 48
Screenshot 2025-03-30 at 17 19 55


destinations/data/seq.yaml
```diff
@@ -0,0 +1,30 @@
...
+      supported: false
+    logs:
+      supported: true
+  fields:
+    - name: SEQ_ENDPOINT
+      displayName: OTLP HTTP Endpoint
+      componentType: input
+      componentProps:
+        required: true
+        tooltip: 'The format is `host:port`, `host` is required and equals the cluster internal DNS record of the Seq instance, `port` is required and equals any of the exposed ports of the Seq instance (defaults to 5341), do not add the rest of the path `/ingest/otlp/v1/*`.'
```
blumamir, line 22: `host:port` is how gRPC endpoint usually looks like.

For http endpoint, it's a convention to also include the scheme (`http` / `https`).

destinations/data/seq.yaml
```diff
@@ -0,0 +1,30 @@
...
+    - name: SEQ_API_KEY
+      displayName: API Key
+      componentType: input
+      secret: true
+      componentProps:
+        type: password
+        required: false
```
blumamir, line 29: can we send to `seq` without api key?

destinations/data/seq.yaml
blumamir, line 22: if port is required then why do we need to specify a default.

common/config/seq.go
```diff
@@ -0,0 +1,62 @@
...
+	endpoint, exists := config[SEQ_ENDPOINT]
+	if !exists {
+		return nil, errorMissingKey(SEQ_ENDPOINT)
+	}
+	endpoint, err := parseOtlpHttpEndpoint(endpoint)
+	if err != nil {
+		return nil, err
+	}
+	if !urlHostContainsPort(endpoint) {
+		endpoint += ":5341/ingest/otlp"
```
blumamir, line 32: should we assert that the path is empty before appending to it? users might get confused and paste the entire http endpoint here in which case there will be no error and the collector will fail to start.

Consider extending `parseOtlpHttpEndpoint` to also verify path, or make another assertion for that

blumamir (approved):
left few nits

PR #2665, opened 2025-03-31
[GEN-1948]: add Bonree ONE destination

BenElferink:
## Description

Add [Bonree ONE](https://www.bonree.com) as destination.


destinations/data/bonree.yaml
```diff
@@ -0,0 +1,39 @@
...
+        required: true
+        tooltip: Obtained from the probe deployment page.
+    - name: BONREE_ENVIRONMENT_ID
+      displayName: Environment ID
+      componentType: input
+      secret: true
+      componentProps:
+        type: password
+        required: false
+        tooltip: The default environment does not need to be configured.
```
blumamir, line 39: What do you mean by that?

BenElferink: No input will define a default name like in AWS EMF

PR #2667, opened 2025-03-31
feat: set only 1 envvar in java in case one of the options defined + …

tamirdavid1:
…create test for it

## Description

This PR addresses the issue observed in `appsDynamic` and improves the overall reliability of environment variable injection.

Specifically:

If either `JAVA_OPTS` or `JAVA_TOOL_OPTIONS` is already defined (in the manifest or at runtime), we avoid setting the other to prevent conflicts.

If neither is defined, we set both environment variables with the appropriate Odigos values.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [X] Updated e2e Tests
- [X] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/tests/e2e/env-injection/01-assert-env-vars.yaml
+++ b/tests/e2e/env-injection/01-assert-env-vars.yaml
@@ -1,0 +1,42 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    app: java-latest-version
+  namespace: default
+spec:
+  containers:
+    - image: public.ecr.aws/odigos/java-latest-version:v0.0.1
+      name: java-latest-version
+      (env[?name=='JAVA_OPTS']):
+       - value: "-javaagent:/var/odigos/java/javaagent.jar"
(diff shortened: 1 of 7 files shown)


tests/e2e/env-injection/01-assert-env-vars.yaml
```diff
@@ -0,0 +1,42 @@
...
+  labels:
+    app: java-supported-manifest-env
+  namespace: default
+spec:
+  containers:
+    - image: public.ecr.aws/odigos/java-supported-manifest-env:v0.0.1
+      name: java-supported-manifest-env
+      (env[?name=='JAVA_OPTS']):
+      - value: "-Dnot.work=true -javaagent:/var/odigos/java/javaagent.jar"
+      (env[?name=='JAVA_TOOL_OPTIONS']): []
```
blumamir, line 28: nit: comment here why this is the expected output (as it's not trivial without all the context)

blumamir (approved):

PR #2671, opened 2025-04-02
consolidate release CLI

RonFed:
More steps to clean to release flow:

* Move the Helm release to the registry CLI workflow.
* Move CLI binaries GH publish (using go-releaser) to the registry CLI workflow.
* Remove duplicated Slack notifications and fix wrong messages.
* Add `odigos-enterprise-instrumentor` to be checked before publishing the CLI.


.github/workflows/release-artifactregistry.yml
```diff
@@ -52,7 +52,7 @@ jobs:
 
       - name: Verify Components Image Ready
         run: |
-          declare -a REPOS=("odigos-autoscaler" "odigos-scheduler" "odigos-instrumentor" "odigos-odiglet" "odigos-collector" "odigos-enterprise-odiglet" "odigos-ui")
+          declare -a REPOS=("odigos-autoscaler" "odigos-scheduler" "odigos-instrumentor" "odigos-odiglet" "odigos-collector" "odigos-enterprise-odiglet" "odigos-ui" "odigos-enterprise-instrumentor")
```
blumamir, line 55: should we also verify ubi images were published successfully here?

blumamir, line 55: "odigos-enterprise-instrumentor" is not in `release.yml`. 

It's because we plan to remove it soon so no need to maintain it right?

.github/workflows/release-artifactregistry.yml
```diff
@@ -90,6 +90,15 @@ jobs:
         with:
           node-version: 18
 
+      - uses: goreleaser/goreleaser-action@v5
```
blumamir, line 93: there seems to already be v6 for this action.
probably better to bump in a different PR later on.

blumamir (approved):

.github/workflows/release-artifactregistry.yml
RonFed: right

damemi: I'd say yes, because this job triggers the openshift certification (which needs the ubi images): https://github.com/odigos-io/odigos/blob/5f33a76c774eb9e6ec1d9e7e2ec48665572339aa/.github/workflows/release-artifactregistry.yml#L130

Alternatively, maybe just the openshift certification job should check for the ubi images. We will need to make sure it can be idempotent with this job though, for example:

1. Regular check+cli push succeeds
2. Get to openshift job
3. UBI images aren't available

Then we re-trigger once the images are available:

1. Will regular check+cli push be a no-op, or a failure since it already pushed?

Maybe we should instead move the preflight submit to its own worfklow, with both a repository_dispatch and workflow_dispatch. Give it a check similar to this but for ubi images, and then that isolates it for idempotency

damemi (approved):

PR #2673, opened 2025-04-02
[GEN-1930]: add GreptimeDB destination

BenElferink:
## Description

Add add [GreptimeDB](https://greptime.com) as destination.

## How Has This Been Tested?

- [x] Manual Testing

__

Proof of data:

Screenshot 2025-04-02 at 10 48 05


common/config/seq.go
```diff
@@ -28,6 +28,10 @@ func (j *Seq) ModifyConfig(dest ExporterConfigurer, currentConfig *Config) ([]st
 	if err != nil {
 		return nil, err
 	}
+	err = urlHostShouldNotContainPath(endpoint)
```
blumamir, line 31: We already execute url parsing above in `parseOtlpHttpEndpoint`.
What do you think to add this check into the existing function which we call anyway in all cases

destinations/data/greptime.yaml
```diff
@@ -0,0 +1,45 @@
...
+      supported: false
+  fields:
+    - name: GREPTIME_ENDPOINT
+      displayName: OTLP HTTP Endpoint
+      componentType: input
+      componentProps:
+        type: text
+        required: true
+        placeholder: https://<unique_id>.<aws_region>.aws.greptime.cloud
+        tooltip: Do not include the path `/v1/otlp/v1/metrics`.
```
blumamir, line 24: nice to have: in some other destinations, we check for this path and know to ignore it. so if the user enters it by mistake, things still work

blumamir (approved):

common/config/checkly.go
```diff
@@ -28,6 +28,10 @@ func (j *Checkly) ModifyConfig(dest ExporterConfigurer, cfg *Config) ([]string,
 	if err != nil {
 		return nil, err
 	}
+	err = urlHostShouldNotContainPath(endpoint)
+	if err != nil {
+		return nil, err
+	}
```
blumamir, line 34: seems `parseOtlpGrpcUrl` already checks for the path. this can be removed

common/config/utils.go
```diff
@@ -85,6 +85,38 @@ func parseOtlpGrpcUrl(rawURL string, encrypted bool) (grpcUrl string, err error)
...
+		return "", fmt.Errorf("failed to parse otlp http endpoint: %w", err)
+	}
+
+	if parsedUrl.Scheme != "http" && parsedUrl.Scheme != "https" {
+		return "", fmt.Errorf("invalid otlp http endpoint scheme: %s", parsedUrl.Scheme)
+	}
+
+	// A path is required
+	if requiredPath != "" {
+		err = urlHostShouldNotContainPath(noWhiteSpaces)
```
blumamir, line 101: consider simply checking if `Path` == "" or `Path` == "requiredPath"

blumamir (approved):

blumamir (approved):

PR #2677, opened 2025-04-03
Support env valuefrom config

tamirdavid1:
## Description

This PR adds support for environment variables whose values are sourced from a ConfigMap, using the following format:

```
env:
  valueFrom:
    configMapKeyRef:
      key: ...
      name: ...
```
To achieve this, the original environment variable is renamed to ORIGINAL_<key>, and a new environment variable is defined as <key>=$(ORIGINAL_<key>) with additional modifications added by Odigos.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [X] Updated e2e Tests
- [X] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -228,3 +238,16 @@ func setOtelSignalsExporterEnvVars(ctx context.Context, logger logr.Logger,
 		corev1.EnvVar{Name: commonconsts.OtelTracesExporter, Value: tracesExporter},
 	)
 }
+func isValueFromConfigmap(envVar *corev1.EnvVar) bool {
+	return envVar.ValueFrom.ConfigMapKeyRef != nil
+}
+func handleValueFromEnvVar(container *corev1.Container, envVar *corev1.EnvVar, originalName, odigosValue string) {
+	originalNewKey := "ORIGINAL_" + envVar.Name
+	envVar.Name = originalNewKey
```
blumamir, line 246: nit: those should go together with `append(container.Env`, so that we never update one and not the other. should be moved into the `if` below

Makefile
```diff
@@ -428,17 +428,17 @@ build-tag-push-eks-image/%:
...
-	make -j 3 build-tag-push-eks-image/scheduler SUMMARY="Scheduler for Odigos" DESCRIPTION="Scheduler manages the installation of OpenTelemetry Collectors with Odigos." TAG=$(TAG) ORG=$(ORG) IMG_SUFFIX=$(IMG_SUFFIX)
-	make -j 3 build-tag-push-eks-image/collector DOCKERFILE=collector/$(DOCKERFILE) BUILD_DIR=collector SUMMARY="Odigos Collector" DESCRIPTION="The Odigos build of the OpenTelemetry Collector." TAG=$(TAG) ORG=$(ORG) IMG_SUFFIX=$(IMG_SUFFIX)
-	make -j 3 build-tag-push-eks-image/ui DOCKERFILE=frontend/$(DOCKERFILE) SUMMARY="UI for Odigos" DESCRIPTION="UI provides the frontend webapp for managing an Odigos installation." TAG=$(TAG) ORG=$(ORG) IMG_SUFFIX=$(IMG_SUFFIX)
+	make -j 3 build-tag-push-ecr-image/odiglet DOCKERFILE=odiglet/$(DOCKERFILE) SUMMARY="Odiglet for Odigos" DESCRIPTION="Odiglet is the core component of Odigos managing auto-instrumentation. This container requires a root user to run and manage eBPF programs." TAG=$(TAG) ORG=$(ORG) IMG_SUFFIX=$(IMG_SUFFIX)
+	make -j 3 build-tag-push-ecr-image/autoscaler SUMMARY="Autoscaler for Odigos" DESCRIPTION="Autoscaler manages the installation of Odigos components." TAG=$(TAG) ORG=$(ORG) IMG_SUFFIX=$(IMG_SUFFIX)
```
ErikNaz-odigos, line 439: Maybe move all of the `make -j 3` to `make -j $(nproc)` to utilize the maximal available resources for the build

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -228,3 +238,16 @@ func setOtelSignalsExporterEnvVars(ctx context.Context, logger logr.Logger,
 		corev1.EnvVar{Name: commonconsts.OtelTracesExporter, Value: tracesExporter},
 	)
 }
+func isValueFromConfigmap(envVar *corev1.EnvVar) bool {
+	return envVar.ValueFrom.ConfigMapKeyRef != nil
```
blumamir, line 242: The `ValueFrom` can also be `SecretKeyRef`, `ResourceFieldRef`, `FieldRef`.

What will happen in these cases?

blumamir (approved):

tests/e2e/env-injection/01-assert-env-vars.yaml
```diff
@@ -45,4 +45,44 @@ spec:
...
+    app: inventory
+  namespace: default
+spec:
+  containers:
+    - image: registry.odigos.io/odigos-demo-inventory:v0.1.14
+      name: inventory
+      (env[?name=='PYTHONPATH']):
+      - value: $(ORIGINAL_PYTHONPATH):/var/odigos/python:/var/odigos/python/opentelemetry/instrumentation/auto_instrumentation
+    # JAVA_TOOL_OPTIONS is not set in the Dockerfile of java-supported-docker-env,
+    # so we need to ensure it is also not set in the pod spec.
```
RonFed, line 63: this comment seems not related? since its a Python app

tests/e2e/env-injection/01-assert-env-vars.yaml
```diff
@@ -45,4 +45,44 @@ spec:
...
+    app: frontend
+  namespace: default
+spec:
+  containers:
+    - image: registry.odigos.io/odigos-demo-frontend:v0.1.14
+      name: frontend
+      (env[?name=='JAVA_TOOL_OPTIONS']):
+      - value: $(ORIGINAL_JAVA_TOOL_OPTIONS) -javaagent:/var/odigos/java/javaagent.jar
+    # JAVA_TOOL_OPTIONS is not set in the Dockerfile of java-supported-docker-env,
+    # so we need to ensure it is also not set in the pod spec.
```
RonFed, line 83: This comment seems not related? - it is not sed in the Dockerfile - it is from the config map

k8sutils/pkg/cri/criwrapper.go
```diff
@@ -34,6 +34,8 @@ var defaultRuntimeEndpoints = []string{
 	"unix:///run/containerd/containerd.sock",
 	"unix:///run/crio/crio.sock",
 	"unix:///var/run/cri-dockerd.sock",
+	"unix:///run/k3s/containerd/containerd.sock",
+	"unix:///var/lib/rancher/rke2/agent/containerd/containerd.sock",
```
RonFed, line 38: Looking at the odiglet mounts - the `/var/lib` looks like it won't be avaialbe?
```
Volumes: append([]corev1.Volume{
	{
		Name: "run-dir",
		VolumeSource: corev1.VolumeSource{
			HostPath: &corev1.HostPathVolumeSource{
				Path: "/run",
			},
		},
	},
	{
		Name: "device-plugins-dir",
		VolumeSource: corev1.VolumeSource{
			HostPath: &corev1.HostPathVolumeSource{
				Path: "/var/lib/kubelet/device-plugins",
			},
		},
	},
	{
		Name: "odigos",
		VolumeSource: corev1.VolumeSource{
			HostPath: &corev1.HostPathVolumeSource{
				Path: k8sconsts.OdigosAgentsDirectory,
			},
		},
	},
	{
		Name: "kernel-debug",
		VolumeSource: corev1.VolumeSource{
			HostPath: &corev1.HostPathVolumeSource{
				Path: "/sys/kernel/debug",
			},
		},
	},
}, odigosSeLinuxHostVolumes...),
```

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
tamirdavid1: Change to check ValueFrom, it should support also this.

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
tamirdavid1: changed, thanks!

Makefile
tamirdavid1: Np, i will do that in another PR

k8sutils/pkg/cri/criwrapper.go
tamirdavid1: fixed

RonFed (approved):

helm/odigos/templates/odiglet/daemonset.yaml
```diff
@@ -128,6 +130,9 @@ spec:
         - name: run-dir
           hostPath:
             path: /run
+        - name: rancher
```
damemi, line 133: what's this? Should it check if the cluster is actually running rancher?

tamirdavid1: @damemi, it we define it as this (without type) it wont do nothing in if not exists

damemi: @tamirdavid1 what is it for though? And it will look odd in non-rancher clusters

cli/cmd/config.go
```diff
@@ -34,6 +34,7 @@ var configCmd = &cobra.Command{
 	- "ignored-namespaces": List of namespaces to be ignored.
 	- "ignored-containers": List of containers to be ignored.
 	- "mount-method": Determines how Odigos agent files are mounted into the pod's container filesystem. Options include k8s-host-path (direct hostPath mount) and k8s-virtual-device (virtual device-based injection).
+	- "container-runtime-socket-path": Path to the custom container runtime socket (e.g /var/lib/rancher).
```
RonFed, line 37: Do we need to add this to `values.yaml`?

helm/odigos/templates/odiglet/daemonset.yaml
RonFed, line 133: This should be taken from the new flag/value, right?

cli/cmd/config.go
tamirdavid1: added

helm/odigos/templates/odiglet/daemonset.yaml
tamirdavid1: @damemi Thanks for the feedback.
I've added an option to configure customContainerRuntimeSocketPath, to support this case and also more cases that we can potentially observ in future installations.

RonFed (approved):

tests/e2e/env-injection/01-install-test-apps.yaml
```diff
@@ -116,4 +116,111 @@ spec:
...
+    metadata:
+      labels:
+        app: inventory
+    spec:
+      containers:
+        - name: inventory
+          env:
+          - name: PYTHONPATH
+            valueFrom:
+              configMapKeyRef:
```
blumamir, line 151: like on adding test for that 👍

tests/e2e/env-injection/01-assert-env-vars.yaml
```diff
@@ -45,4 +45,42 @@ spec:
...
+    - image: registry.odigos.io/odigos-demo-frontend:v0.1.14
+      name: frontend
+      # We are checking that in case valueFrom exists we're appending Odigos additions to the ORIGINAL_{envar}
+      (env[?name=='JAVA_TOOL_OPTIONS']):
+      - value: $(ORIGINAL_JAVA_TOOL_OPTIONS) -javaagent:/var/odigos/java/javaagent.jar
+      (env[?name=='ORIGINAL_JAVA_TOOL_OPTIONS']):
+      - valueFrom:
+          configMapKeyRef:
+            key: JAVA_TOOL_OPTIONS
+            name: env-configmap
```
blumamir, line 86: nit (not suggesting to do it now):

As we discovered in the previous debug sessions, the order here is important. e.g. if someone will set first `JAVA_TOOL_OPTIONS` which references `ORIGINAL_JAVA_TOOL_OPTIONS`, this will not work (if I remember correctly).

The tests here will assert a value in the manifest, but not that the app can start successfully after we modified the envs. might be worth to extend the tests one day to also make sure apps can run,

blumamir (approved):
Looks great!

Left small suggestions, but it's good to go whenever you want

tests/e2e/env-injection/01-assert-env-vars.yaml
tamirdavid1: thats great feedback!

PR #2679, opened 2025-04-03
[GEN-2472]: add VictoriaMetrics destination

BenElferink:
## Description

Add [VictoriaMetrics](https://victoriametrics.com) as destination.

## How Has This Been Tested?

- [x] Manual Testing

---

Proof of data:

Screenshot 2025-04-03 at 15 22 32
Screenshot 2025-04-03 at 15 22 22


destinations/data/victoriametrics.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: internal.odigos.io/v1beta1
+kind: Destination
+metadata:
+  type: victoriametrics
```
blumamir, line 4: Since there is also OSS victoria metrics, perhaps we should call this one "victoriametricscloud" so that we can add the self-hosted destination later and not have collision?

common/config/victoriametricscloud.go
```diff
@@ -0,0 +1,59 @@
+package config
+
+import (
+	"github.com/odigos-io/odigos/common"
+)
+
+const (
+	VICTORIA_METRICS_ENDPOINT = "VICTORIA_METRICS_ENDPOINT"
+	VICTORIA_METRICS_TOKEN    = "VICTORIA_METRICS_TOKEN"
```
blumamir, line 9: Should we also add 'CLOUD` to these? e.g. `VICTORIA_METRICS_CLOUD_ENDPOINT`.

Changing this in the future is a breaking change

common/dests.go
```diff
@@ -66,4 +66,5 @@ const (
 	TingyunDestinationType                DestinationType = "tingyun"
 	TraceloopDestinationType              DestinationType = "traceloop"
 	UptraceDestinationType                DestinationType = "uptrace"
+	VictoriaMetricsDestinationType        DestinationType = "victoriametrics"
```
blumamir, line 69: This should also mention "cloud" so we don't have a conflict once we introduce the self-hosted victoriametrics destination

common/config/victoriametricscloud.go
```diff
@@ -0,0 +1,59 @@
...
+
+type VictoriaMetricsCloud struct{}
+
+func (j *VictoriaMetricsCloud) DestType() common.DestinationType {
+	return common.VictoriaMetricsCloudDestinationType
+}
+
+func (j *VictoriaMetricsCloud) ModifyConfig(dest ExporterConfigurer, cfg *Config) ([]string, error) {
+	config := dest.GetConfig()
+	uniqueUri := "victoriametrics-" + dest.GetID()
```
blumamir, line 20: also here - should be "cloud" so not to conflict with future destination

blumamir (approved):

PR #2681, opened 2025-04-06
ci: attempt to improve flake in UI tests

BenElferink:
## Description

This PR addresses 2 things:
- Brings back the await for "updating" toast, this ensures it is closed and not covering any elements.
- Attempt to work-around the warning modal flake.

## How Has This Been Tested?

- [x] Updated e2e Tests


frontend/webapp/app/(overview)/layout.tsx
```diff
@@ -13,7 +13,9 @@ import { DataFlowActionsMenu, NavIconIds, SideNav, ToastList } from '@odigos/ui-
 const PageContent = styled(FlexColumn)`
   width: 100%;
   height: 100vh;
-  background-color: ${({ theme }) => theme.colors.primary};
+  background-color: ${({ theme }) =>
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    (theme as any).colors.primary};
```
alonkeyval, line 18: you need to use DefaultTheme instead of any

alonkeyval: why we need it in first place?

BenElferink: Something in the latest version of `styled-components` or `ui-kit` made that any break the build.
This is temporary, I need to export the theme type `ITheme` from the kit, which is different from `DefaultTheme`.

alonkeyval (approved):

PR #2708, opened 2025-04-10
Inject env vars by adding configuration in .values (helm)

AvihuHenya:
This feature enables injecting environment variables defined in the Helm .values file into the relevant pods. It includes validation to skip any env var already present in the container spec to avoid duplication. 
Additionally, an "enabled" flag has been added, allowing users to define env vars once and toggle injection on or off as needed. The structure is open to redesign if a more intuitive approach is preferred.
Screenshot 2025-04-10 at 18 20 55


instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -130,6 +128,33 @@ func (p *PodsWebhook) Default(ctx context.Context, obj runtime.Object) error {
...
+			existingEnvNames := getEnvVarNamesSet(container)
+
+			for envName, envValue := range langConfig.EnvVars {
+				existingEnvNames = podswebhook.InjectEnvVarToPodContainer(
+					existingEnvNames,
+					container,
+					envName,
+					envValue,
+				)
+			}
```
RonFed, line 154: I think we can move this to the `podswebhook` package where all the other env stuff is done.
This mean we can also keep the `InjectEnvVarToPodContainer` un-exported.

RonFed (commented):
Overall looks great,
I think we should add an equivalent way to set these env vars from the CLI.

helm/odigos/values.yaml
```diff
@@ -18,6 +18,25 @@ ignoredNamespaces:
 # you can add additional container names to ignore by adding them to the list
 ignoredContainers:
 
+# Instrumentation settings for language-specific configurations
+# additionalInstrumentationEnvs:
```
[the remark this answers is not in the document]
AvihuHenya: i fixed it and also add usage example in the comment

helm/odigos/values.yaml
```diff
@@ -18,8 +18,30 @@ ignoredNamespaces:
 # you can add additional container names to ignore by adding them to the list
 ignoredContainers:
 
-# Name of the cluster, will be used to identify this cluster in the centralized backend
-clusterName: ''
+# Instrumentation settings for language-specific configurations
```
RonFed, line 21: Maybe worth to add more explanations here, something like:
"This is for OTEL agents configuration, see the OTEL docs for language specific configuration,
For eBPF distributions, these values can't be used to configure exporting and batching since these are handled by the Odiglet.
This is an advanced feature, only use this if you know what you are doing"

instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -128,6 +127,10 @@ func (p *PodsWebhook) Default(ctx context.Context, obj runtime.Object) error {
 	// we can pull only our pods into cache, and follow the lifecycle of the instrumentation process.
 	pod.Labels[k8sconsts.OdigosAgentsMetaHashLabel] = ic.Spec.AgentsMetaHash
 
+	if odigosConfig.AdditionalInstrumentationEnvs != nil {
+		podswebhook.InjectEnvsBasedOnLanguage(&odigosConfig, pod, &ic)
+	}
+
```
RonFed, line 133: nit: I think this should be before the label above - the label should be the last thing that we do.

instrumentor/controllers/agentenabled/podswebhook/env.go
```diff
@@ -71,3 +73,50 @@ func InjectOtlpHttpEndpointEnvVar(existingEnvNames EnvVarNamesMap, container *co
 	existingEnvNames = injectEnvVarToPodContainer(existingEnvNames, container, commonconsts.OtelExporterEndpointEnvName, otlpHttpEndpoint)
 	return existingEnvNames
 }
+
+func InjectEnvsBasedOnLanguage(odigosConfig *common.OdigosConfiguration, pod *corev1.Pod, ic *odigosv1alpha1.InstrumentationConfig) {
```
RonFed, line 77: Maybe a better name might be:
`AddUserDefinedEnvForLang`
`AddUserConfiguredEnvForLang`

something like this might better explain what this is doing

RonFed (commented):
still think we need the CLI implementation here,
maybe we can use the `odigos config set` command?

helm/odigos/values.yaml
AvihuHenya: Thanks for the example, look at the updated comment

helm/odigos/values.yaml
```diff
@@ -21,6 +21,40 @@ ignoredContainers:
...
+  #     enabled: true
+  #     env:
+  #       OTEL_INSTRUMENTATION_COMMON_EXPERIMENTAL_VIEW_TELEMETRY_ENABLED: "true"
+  # Note: For eBPF-based distributions, exporting and batching cannot be configured here, as they are managed by the Odiglet.
+  # Warning: This is an advanced feature. Only modify these settings if you are familiar with OTEL and its implications.
+  languages:
+    java:
+      enabled: true
+      env: 
+        OTEL_INSTRUMENTATION_COMMON_EXPERIMENTAL_VIEW_TELEMETRY_ENABLED: "true"
```
RonFed, line 41: Do we want this to be the default value? it looks super specific - and I think the user requested this should set he's required values

RonFed (approved):

PR #2710, opened 2025-04-10
feat: url-template processor

blumamir:
## Description

Add collector processor that tries to heuristically guess the templated url value when it's missing based on the `url.path`, `http.target`, `url.full` or `http.url` attributes.

It will insert the relevant templated attributes (`http.route` for server spans and `url.template` for client spans) and update span name to add the template path (`GET` -> `GET /users/:id`)

## How Has This Been Tested?

- [x] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

Does not affect kubernetes

## User Facing Changes

New functionality currently as processor but without action ATM


collector/processors/odigosurltemplateprocessor/processor.go
```diff
@@ -0,0 +1,181 @@
...
+	if !found {
+		// we only enhance http spans, so if there is no http.method attribute, we can skip it
+		return
+	}
+
+	switch span.Kind() {
+
+	case ptrace.SpanKindClient:
+		// client spans write the url templated value in "url.template" attribute.
+		enhanceSpan(span, httpMethod, semconv.AttributeURLTemplate)
```
RonFed, line 172: Do we want to pass a slice of attributes to be able to handle multiple semantic conventions versions? not sure if these attributes were changed, but we could have a logic to find the one that is in the span.

collector/processors/odigosurltemplateprocessor/processor.go
```diff
@@ -0,0 +1,181 @@
...
+	newSpanName := fmt.Sprintf("%s %s", httpMethod, templatedUrl)
+	span.SetName(newSpanName)
+}
+
+func enhanceSpan(span ptrace.Span, httpMethod string, targetAttribute string) {
+
+	attr := span.Attributes()
+
+	if _, found := attr.Get(targetAttribute); found {
+		// avoid overriding the attribute if it is already set
```
RonFed, line 143: nit: It should not really happen but, if the attribute is present - do we want to make sure the span name contains it and maybe add it to the span name if the name is only the method?
For semantic conventions compliment instrumentations - this should not happen - but we can add this check here as well.

collector/processors/odigosurltemplateprocessor/processor.go
```diff
@@ -0,0 +1,181 @@
...
+	// prefer the updated semantic convention "url.path" if available
+	if urlPath, found := attr.Get(semconv.AttributeURLPath); found {
+		return urlPath.AsString(), true
+	}
+
+	// fallback to the old "http.target" attribute which might still be used
+	// by some instrumentations.
+	// TODO: remove this fallback in the future when all instrumentations are aligned with
+	// update semantic conventions and no longer report "http.target"
+	if httpTarget, found := attr.Get("http.target"); found {
```
RonFed, line 66: nit: we can import semconv_v<older version> - and take the key string from there instead of having it hard coded.

collector/processors/odigosurltemplateprocessor/processor.go
```diff
@@ -0,0 +1,181 @@
...
+		// the "http.target" attribute might contain a query string, so we need to
+		// split it and only use the path part.
+		// for example: "/user?id=123" => "/user"
+		path := strings.SplitN(httpTarget.AsString(), "?", 2)[0]
```
RonFed, line 70: the `net/url` package contains some functions for parsing URLs - which probably handle some edge cases. Maybe worth to consider using them

collector/processors/odigosurltemplateprocessor/processor.go
```diff
@@ -0,0 +1,181 @@
...
+func getFullUrl(attr pcommon.Map) (string, bool) {
+	// prefer the updated semantic convention "url.full" if available
+	if fullUrl, found := attr.Get(semconv.AttributeURLFull); found {
+		return fullUrl.AsString(), true
+	}
+	// fallback to the old "http.url" attribute which might still be used
+	// by some instrumentations.
+	// TODO: remove this fallback in the future when all instrumentations are aligned with
+	// update semantic conventions and no longer report "http.url"
+	if httpUrl, found := attr.Get("http.url"); found {
```
RonFed, line 85: same as above comment - can take this from the older semconv

collector/processors/odigosurltemplateprocessor/templatize.go
```diff
@@ -0,0 +1,27 @@
...
+	"regexp"
+	"strings"
+)
+
+var (
+	uuidRegex   = regexp.MustCompile(`^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$`)
+	numberRegex = regexp.MustCompile(`^\d+$`)
+)
+
+func templatizeURLPath(path string) string {
```
RonFed, line 13: I think this function should be an instance function of the processor.
This will have to be the case when we allow users to pass configuration for the processor (do we plan to have configurable regexes as part of this PR or not).

I think we at least need to be able to get configuration for the template place holders - i.e maybe user went to call  it `:userId` or anything else.
If they have multiple UIUDs in a path probably the first and the second have different meanings.

We can decide not to include this in this first PR, but I think we should at least write the processor in a way that would allow adding the configurations easily in the future.

collector/processors/odigosurltemplateprocessor/templatize.go
```diff
@@ -0,0 +1,27 @@
...
+	modified := false
+	segments := strings.Split(path, "/")
+	for i, segment := range segments {
+		if uuidRegex.MatchString(segment) || numberRegex.MatchString(segment) {
+			segments[i] = ":id"
```
RonFed, line 18: nit: why `:id` and not `{id}` - I don't really have a preference - but I think I saw the latter one more in use

collector/processors/odigosurltemplateprocessor/templatize.go
```diff
@@ -0,0 +1,27 @@
...
+var (
+	uuidRegex   = regexp.MustCompile(`^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$`)
+	numberRegex = regexp.MustCompile(`^\d+$`)
+)
+
+func templatizeURLPath(path string) string {
+	modified := false
+	segments := strings.Split(path, "/")
+	for i, segment := range segments {
+		if uuidRegex.MatchString(segment) || numberRegex.MatchString(segment) {
```
RonFed, line 17: not sure if it has performance impact but - 
we could unify the 2 regexes to 1 with an or between them.
It is of course equivalent - just wondering it can help in performance a bit in this hot path.

collector/processors/odigosurltemplateprocessor/processor.go
blumamir: not sure what you mean by that. I pass the `span` that has attributes inside which i use

collector/processors/odigosurltemplateprocessor/processor.go
blumamir: this can be a nice enhancement if we discover those issue really occur and cannot be fixed in source (instrumentation).
don't want to add more complexity and ways things can go wrong in this code unless it addresses a real problem

collector/processors/odigosurltemplateprocessor/processor.go
blumamir: nice. added (also for `http.url` and `http.method`)

collector/processors/odigosurltemplateprocessor/templatize.go
blumamir: added the templatization rules as mentioned with configuration and made this a member funcion

collector/processors/odigosurltemplateprocessor/templatize.go
blumamir: I think it really depends on the language and framework, there are a lot of conventions. perhaps wrapping in "{}" is more common.

collector/processors/odigosurltemplateprocessor/templatize.go
blumamir: I don't really know, but I think this is readable and guessing the performance impact is quite minimal between the 2 options.

collector/processors/odigosurltemplateprocessor/templatize.go
```diff
@@ -0,0 +1,147 @@
...
+
+type RulePathSegment struct {
+
+	// if this rule path segment is a static string (e.g. "users"), this value will be non-empty
+	StaticString string
+
+	// it this rule segment path is replaced with templated name, the TemplateName will be non-empty
+	TemplateName string
+	// for templated segment names, a user can also include an optional regexp
+	// which mush match for the rule to be applied.
```
RonFed, line 22:
```suggestion
	// which must match for the rule to be applied.
```

collector/processors/odigosurltemplateprocessor/templatize.go
```diff
@@ -0,0 +1,147 @@
...
+	RegexpPattern *regexp.Regexp
+}
+
+type TemplatizationRule []RulePathSegment
+
+func parseRuleTemplateString(ruleTemplateString string) (string, *regexp.Regexp, error) {
+	// template string is in the format name:optional-regexp.
+	// for example: {userId:[0-9]+} or {userId}
```
RonFed, line 30: is it valid to provide just a regex? e.g {[0-9]}
if that is the case this code will assume the regex is the template name, right?

collector/processors/odigosurltemplateprocessor/templatize.go
```diff
@@ -0,0 +1,147 @@
...
+	var regexpPattern *regexp.Regexp
+	if len(parts) == 2 {
+		regexpString := strings.TrimSpace(parts[1])
+		if regexpString == "" {
+			return "", nil, errors.New("invalid rule template string. regexp is empty")
+		}
+		var err error
+		regexpPattern, err = regexp.Compile(regexpString)
+		if err != nil {
+			return "", nil, err
```
RonFed, line 50: nit: maybe worth to wrap this in our error to better understand where this is coming from

collector/processors/odigosurltemplateprocessor/processor.go
RonFed: we are passing here the `url.template` attribute to identify if the span already has low cardinality template - but this attribute might have a different name, depending on the semconv, right? my intention was that maybe we need to look also for the older attributes to decide

collector/processors/odigosurltemplateprocessor/templatize.go
```diff
@@ -0,0 +1,147 @@
...
+			// if the static string does not match, we can't use this rule
+			return "", false
+		}
+
+		if ruleSegment.TemplateName != "" && ruleSegment.RegexpPattern != nil {
+			// if this segment is a templated segment, it must match the regexp pattern
+			if !ruleSegment.RegexpPattern.MatchString(pathSegment) {
+				// if the regexp pattern does not match, we can't use this rule
+				return "", false
+			}
```
RonFed, line 112: for a rule like `/{id}/{email}/{op}` - will it match every 3 segment path? is this the expected behavior?
just an idea for the future - we might want to have service names associated to rules

collector/processors/odigosurltemplateprocessor/processor_test.go
```diff
@@ -0,0 +1,415 @@
...
+			expectedName:      "GET /user/{id}", // fallback to name "id" when missing
+			expectedHttpRoute: "/user/{id}",
+		},
+		{
+			name:              "segment-rule-with-spaces",
+			rules:             []string{"/user/{user-name : [a-zA-Z]+}"},
+			path:              "/user/John",
+			expectedName:      "GET /user/{user-name}",
+			expectedHttpRoute: "/user/{user-name}",
+		},
```
RonFed, line 392: cam we add a test case for a regex only segment like {[0-9]}?
or this should be part of the config validation test?

collector/processors/odigosurltemplateprocessor/processor.go
blumamir: I think we can do it with the new names, and if there is feedback that this is needed, we can add it later on. want to keep it simple and get feedback for potential improvments

collector/processors/odigosurltemplateprocessor/templatize.go
blumamir: the template name is "userId" and the regexp must match for it to be picked up

collector/processors/odigosurltemplateprocessor/templatize.go
blumamir: yes, any "{}" segment in the url will be templated if it matches.
regarding service names, total. let's see what issue are reported and iterate to improve

collector/processors/odigosurltemplateprocessor/processor_test.go
blumamir: it should be like this `{:[0-9]+}`. so before the `:` are the segment template name, and after the `:` is the regexp

collector/processors/odigosurltemplateprocessor/templatize.go
blumamir: done

collector/processors/odigosurltemplateprocessor/templatize.go
blumamir: `{[0-9]}` is not a valid rule, as it will be interpreted as the name of the template

RonFed (approved):

PR #2723, opened 2025-04-14
feat: calculate url template action

blumamir:
## Description

This PR adds a new action to auto calculate url template low-cardinality value for http spans when possible.

## How Has This Been Tested?

- [x] Manual Testing

## Kubernetes Checklist

- [x] Changes how Odigos interacts with Kubernetes - adds a new CRD and RBAC permissions
- [x] Introduces additional calls to the API Server (potential performance impact) - watch the new CRD in autoscaler
- [x] New Query/feature supported in all the k8s versions supported by Odigos

## User Facing Changes

User can add new action. docs been updated


autoscaler/controllers/actions/action_controller.go
```diff
@@ -0,0 +1,172 @@
...
+	// templatization segments like this: "/{name:regex}" and are used to match and replace the segment with the name.
+	// e.g. "/v1/{foo:regex}/bar/{baz}" will match "/v1/123/bar/456" and will replace it with "/v1/:foo/bar/:baz"
+	// if regex is not used, the segment will always match and replaced with the name.
+	// if regex is used, and does not match, the segment will be skipped and will not take effect.
+	TemplatizationRules []string `json:"templatization_rules"`
+}
+
+// giving an action, return it's specific processor details
+func actionProcessorDetails(action *odigosv1.Action) (string, int, any, error) {
+	if action.Spec.CalculateUrlTemplate != nil {
```
damemi, line 39: what about a type switch, so the `action` can just have one spec field?

PR #2731, opened 2025-04-16
operator: Update API field documentation

damemi:
## Description

This updates the Operator CRD fields to have clearer explanations of what they are for, which ones are optional, and what are the defaults.

This also changes the `UIMode` in the operator CR to use the UIMode type (instead of string), and adds an `enum` to the UIMode type for API validation.

The API fields are re-ordered to be presented in a general level of importance.

The Operator API is added to the main docs.odigos.io site

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/common/odigos_config.go
+++ b/common/odigos_config.go
@@ -1,6 +1,8 @@
 package common
 
 type ProfileName string
+
+// +kubebuilder:validation:Enum=normal;readonly
 type UiMode string
 
 const (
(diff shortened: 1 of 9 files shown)


common/odigos_config.go
```diff
@@ -1,6 +1,8 @@
 package common
 
 type ProfileName string
+
+// +kubebuilder:validation:Enum=normal;readonly
```
BenElferink, line 5: ♥️

BenElferink (approved):

Issue #1436, opened 2024-08-09
Add PHP support


mikkeschiren:
Are there any plans to add support for PHP?

RonFed:
Hey, it is in our roadmap. Currently with low priority

mikkeschiren:
As PHP is the most common language for web applications it would make sense to higher the priority :)

mikkeschiren:
I still think taht this should be a prio :)

mikkeschiren:
Still think the same :)

Fgruntjes:
Same here, "PHP is used by 75.2% of all the websites whose server-side programming language we know." (source: https://w3techs.com/technologies/details/pl-php)

gourav-rcrm:
Is there any release schedule for PHP?

BenElferink:
Working on it, 1st PR is open: https://github.com/odigos-io/odigos/pull/2727

BenElferink:
Following up with the 2nd PR: https://github.com/odigos-io/odigos/pull/2732
OTel injection succeeded, telemetry data not ready yet.

BenElferink:
Updated the same PR: https://github.com/odigos-io/odigos/pull/2732
Telemetry data is available 🚀

BenElferink:
All merged and released :tada:
Please install Odigos `v1.0.175` or higher to instrument PHP applications.

BenElferink:
Support for multi-arch (amu & arm) has been added recently too, will be released with `v1.0.180` (likely today).
I'm closing this issue for now, if you have further questions, feel free to create another issue or join our Slack community :)

PR #2732, opened 2025-04-17
feat: auto-instrument PHP versions 8+

BenElferink:
## Description

Resolves issue #1436

This PR injects and enables the [OpenTelemetry PHP extension](https://github.com/open-telemetry/opentelemetry-php-instrumentation) into PHP-detected pods.

- The `opentelemetry.so` file is compiled for each PHP version (see [PHP agents here](https://github.com/odigos-io/opentelemetry-php))
- Each PHP version is assigned it's own agent

###  Proof of injection (from `kubectl exec` into the PHP-detected pod):

Screenshot 2025-04-17 at 9 38 32

### Proof of trace:

Screenshot 2025-04-18 at 22 30 43
Screenshot 2025-04-20 at 23 41 59
Screenshot 2025-04-21 at 17 55 00

## How Has This Been Tested?

- [x] Manual Testing

---

TODO
- Write demo PHP app, and include it in E2E tests
- Update docs


odiglet/Dockerfile
```diff
@@ -85,6 +82,50 @@ RUN ARCH_SUFFIX=$(cat /tmp/arch_suffix) && \
     wget https://github.com/odigos-io/opentelemetry-dotnet-instrumentation/releases/download/${DOTNET_OTEL_VERSION}/OpenTelemetry.AutoInstrumentation.Native-${ARCH_SUFFIX}.so && \
     mv OpenTelemetry.AutoInstrumentation.Native-${ARCH_SUFFIX}.so linux-glibc-${ARCH_SUFFIX}/OpenTelemetry.AutoInstrumentation.Native.so
 
+
+# PHP: clone agents repo
+FROM --platform=$BUILDPLATFORM php:8-cli AS php-agents
+WORKDIR /php-agents
+RUN apt-get update && apt-get upgrade -y && apt-get install -y git
+RUN git clone https://github.com/odigos-io/opentelemetry-php
```
edeNFed, line 90: I suggest pinning the PHP instrumentation to a specific version like we do in the other languages, and then let dependabot update it.
This will avoid us being surprised by new versions automatically pushed to Odigos

BenElferink: The OTel extension version is controlled by the agents repo (see [Makefile here](https://github.com/odigos-io/opentelemetry-php/blob/main/Makefile)):
```env
PHP_OTEL_VERSION=1.1.2
PHP_VERSIONS=8.0 8.1 8.2 8.3 8.4
```

---

However, I did change the Odiglet Dockerfile to checkout at "main" like with the Node.js agent 😃

odiglet/Dockerfile
```diff
@@ -120,6 +129,13 @@ COPY --from=nodejs-agent-native-community-builder /repos/odigos/agents/nodejs-na
 # .NET
 COPY --from=dotnet-builder /dotnet-instrumentation /instrumentations/dotnet
 
+# PHP
+COPY --from=php-agents /php-agents/opentelemetry-php/8.0 /instrumentations/php/8.0
+COPY --from=php-agents /php-agents/opentelemetry-php/8.1 /instrumentations/php/8.1
+COPY --from=php-agents /php-agents/opentelemetry-php/8.2 /instrumentations/php/8.2
+COPY --from=php-agents /php-agents/opentelemetry-php/8.3 /instrumentations/php/8.3
+COPY --from=php-agents /php-agents/opentelemetry-php/8.4 /instrumentations/php/8.4
```
RonFed, line 137: what is the size of those folders? more importantly - what is the diff in the odiglet image size?

odiglet/Dockerfile
```diff
@@ -85,6 +82,18 @@ RUN ARCH_SUFFIX=$(cat /tmp/arch_suffix) && \
...
 
+
+# PHP: Clone agents repo (contains pre-compiled binaries and libraries for each PHP version)
+FROM --platform=$BUILDPLATFORM php:8-cli AS php-agents
+WORKDIR /php-agents
+ARG PHP_AGENT_VERSION=main
+RUN apt-get update && apt-get upgrade -y && apt-get install -y git
+RUN git clone https://github.com/odigos-io/opentelemetry-php \
+    && cd opentelemetry-php \
+    && git checkout $PHP_AGENT_VERSION
```
RonFed, line 93: I agree with Eden, I think we should pin the tag of the version, and update this via depndabot like we have for Java and .NET.

distros/yamls/php-community.yaml
```diff
@@ -0,0 +1,43 @@
...
+      # and ignore /etc/php.d/*.ini (default)
+      #
+      # RUN PHP_INI_SCAN_DIR=:/usr/local/etc/php.d php
+      # will 1st load all files in /etc/php.d/*.ini (default),
+      # then load /usr/local/etc/php.d/*.ini (custom)
+      #
+      # RUN PHP_INI_SCAN_DIR=/usr/local/etc/php.d: php
+      # will 1st load all files in /usr/local/etc/php.d/*.ini (custom),
+      # then load /etc/php.d/*.ini (default)
+      - envName: PHP_INI_SCAN_DIR
```
RonFed, line 34: Will we need to add this to the whole env appending logic? i.e - if what happens if the use also have a value for this env var?

RonFed (commented):
🚀

odiglet/Dockerfile
BenElferink: The Odiglet image size from v1.0.172 release is `439MB`.
The Odiglet image size from development build is `541MB`.

So the diff is `102MB` for all 5 agents together.

odiglet/Dockerfile
BenElferink: As previously mentioned;
The OTel extension version is controlled (pinned) by the agents repo (see [Makefile here](https://github.com/odigos-io/opentelemetry-php/blob/main/Makefile)):
```env
PHP_OTEL_VERSION=1.1.2
PHP_VERSIONS=8.0 8.1 8.2 8.3 8.4
```

distros/yamls/php-community.yaml
BenElferink: In theory the user should not use this env.

The user would normally include all configuration files in a default directory for that app, otherwise they would have to define their custom file using `PHPRC` env.

I might be wrong though, all info can be found [here](https://www.php.net/manual/en/configuration.file.php#configuration.file.scan)

RonFed: It looks similar to other languages I think.
Probably most users won't use this, but we have to correctly handle the case where they do have it defined - otherwise we are risking crashing the instrumented process.

instrumentor/controllers/agentenabled/podswebhook/env.go
```diff
@@ -28,11 +32,17 @@ func injectEnvVarObjectFieldRefToPodContainer(existingEnvNames EnvVarNamesMap, c
...
 
-func injectEnvVarToPodContainer(existingEnvNames EnvVarNamesMap, container *corev1.Container, envVarName, envVarValue string) EnvVarNamesMap {
+func injectEnvVarToPodContainer(existingEnvNames EnvVarNamesMap, container *corev1.Container, envVarName, envVarValue string, runtimeDetails *odigosv1.RuntimeDetailsByContainer) EnvVarNamesMap {
 	if _, exists := existingEnvNames[envVarName]; exists {
 		return existingEnvNames
 	}
 
+	if runtimeDetails != nil {
+		// This was added to support dynamic versions for PHP agents
+		majorMinor := common.MajorMinorStringOnly(common.GetVersion(runtimeDetails.RuntimeVersion))
```
blumamir, line 42: check if `runtimeDetails.RuntimeVersion` is missing and avoid injecting any envs in this case

blumamir (approved):
Don't forget to apply these updates also for odigos enterprise 🚀

instrumentor/controllers/agentenabled/podswebhook/env.go
BenElferink: Good catch, done

odiglet/Dockerfile
BenElferink: I tagged the agents repo and changed the cloning process 👍

odiglet/Dockerfile
BenElferink: I tagged the agents repo and changed the cloning process 👍

PR #2738, opened 2025-04-18
feat: run cluster collector before destinations are added

blumamir:
## Description

At the moment, cluster collector is only started once a destination is added. This causes unnecessary delay in demos and ramp-ups.

With this change, scheduler will always create the cluster collector group right after effective config is calculated. then autoscaler will pick this up and deploy the cluster collector.

By the time the user opens the UI and adds its sources, we only need to create node collector which will be faster.

In the future we might want to add a mode for odigos that only sets up instrumentation but allows users to send to their own collector and not deploy our cluster collector. in this case this logic should be modified.

## How Has This Been Tested?

- [x] Manual Testing

## Kubernetes Checklist

Removed RBAC permissions for scheduler as it's no longer used after this change 🚀

## User Facing Changes

After installing odigos, an additional pod for cluster collector will run even before adding any destination. this should not be dramatic as it's just one more deployment to the 4 that we already deploy and the ds. plus it's only for a short while as the collector will start anyway once the user setup odigos in ui or gitops


edeNFed (approved):

damemi (commented):
just curious does the collector not crashloop when its config doesn't have an exporter/destination set? I know the collector can be finnicky if the config isn't perfect so wondering about that

blumamir:
> just curious does the collector not crashloop when its config doesn't have an exporter/destination set? I know the collector can be finnicky if the config isn't perfect so wondering about that

good point. checking 👍

blumamir:
> just curious does the collector not crashloop when its config doesn't have an exporter/destination set? I know the collector can be finnicky if the config isn't perfect so wondering about that

@damemi  It's working because we always have the self-metrics pipeline:

```
service:
  pipelines:
    metrics/otelcol:
      receivers:
      - prometheus/self-metrics
      processors:
      - resource/pod-name
      exporters:
      - otlp/odigos-own-telemetry-ui
  telemetry:
    metrics:
      readers:
      - pull:
          exporter:
            prometheus:
              host: 0.0.0.0
              port: 8888
    resource: {}
```

damemi:
> > just curious does the collector not crashloop when its config doesn't have an exporter/destination set? I know the collector can be finnicky if the config isn't perfect so wondering about that
> 
> It's working because we always have the self-telemetry pipeline

ahh nice! 👍

PR #2748, opened 2025-04-20
feat: allow to set k8s node logs directory for logs collection

blumamir:
## Description

Logs are collected from `/var/log` on each k8s host.
However, some clusters can have these logs in a different directory (for example `/mnt/var/logs`).

This PR adds an option to configure this path

## How Has This Been Tested?

- [x] Manual Testing

## Kubernetes Checklist

No

## User Facing Changes

Option to set a new config value in cli/helm


blumamir:
replaced with #2750

PR #2759, opened 2025-04-21
feat: add support for LD_PRELOAD odigos-loader

RonFed:
## Description

Using LD_PRELOAD env var as a way of adding the instrumentation relevant environment variables.
This feature will be controlled by a configuration flag. A new configuration called `envVarInjectionMethod` is added which have the values
1. `pod-manifest` indicating we should only try to add env vars through the pod manifest (what we do today). This is the default value.
2. `loader-fallback-to-pod-manifest` - first try using the loader - if that is not possible, fallback to the pod-manifest approach. Reasons for that are: a) the user have an already defined LD_PRELOAD value - and currently we want to avoid manipulating that value. b) we detect secure execution mode in the container.
3. `loader` - only trying the loader method. If we can't add it for one of the reasons mentioned we won't instrument the pod.

## How Has This Been Tested?

- [x] Added Unit Tests
The loader repo contains extensive set of unit tests and e2e tests specifically for testing the loader.
- [x] Updated e2e Test
Updated the `env-injection` e2e to include all the 3 methods and assert the env vars accordingly.
- [x] Manual Testing
Manually tested the integration with simple-demo, verified the env vars are added correctly and traces are emitted.

## Kubernetes Checklist

When this method is used the Pods mutating webhook doesn't add the PYTHONPATH, NODE_OPTIONS  and JAVA_TOOL_OPTIONS.
No.
NA
the new env injection method config option is added and addressed in CLI and helm.
-No.


odiglet/Dockerfile.rhel
```diff
@@ -148,6 +148,9 @@ COPY --from=php-agents /php-agents/opentelemetry-php/8.2 /instrumentations/php/8
 COPY --from=php-agents /php-agents/opentelemetry-php/8.3 /instrumentations/php/8.3
 COPY --from=php-agents /php-agents/opentelemetry-php/8.4 /instrumentations/php/8.4
 
+ARG ODIGOS_LOADER_VERSION=v0.0.2
```
blumamir, line 151:
```suggestion
ARG ODIGOS_LOADER_VERSION=v0.0.3
```

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -38,6 +40,45 @@ func InjectOdigosAgentEnvVars(ctx context.Context, logger logr.Logger, podWorklo
...
+			})
+			return
+		}
+
+		// the LD_PRELOAD env var is preset. for now, we don't attempt to append our value to the user defined one.
+		if *injectionMethod == common.LoaderEnvInjectionMethod {
+			// we're specifically requested to use the loader env var injection method
+			// and the user defined LD_PRELOAD env var is already present.
+			// so we avoid the fallback to pod manifest env var injection method
+			return
```
blumamir, line 74: In this case, we will have no instrumentation and no log or indication for it anywhere, right?

We will mark the pod as instrumented with the label but will not actually inject any env var to do the instrumentation

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -74,6 +115,15 @@ func InjectOdigosAgentEnvVars(ctx context.Context, logger logr.Logger, podWorklo
 	}
 }
 
+func getEnvVarFromRuntimeDetails(runtimeDetails *odigosv1.RuntimeDetailsByContainer, envVarName string) (string, bool) {
+	for _, envVar := range runtimeDetails.EnvVars {
```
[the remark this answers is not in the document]
RonFed: My understanding it the `EnvVars` are taken from `/proc` and should include more cases than `EnvFromContainerRuntime` (e.g scripts)?

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -38,6 +40,45 @@ func InjectOdigosAgentEnvVars(ctx context.Context, logger logr.Logger, podWorklo
...
+		return
+	}
+
+	// check if odigos loader should be used
+	if *injectionMethod == common.LoaderEnvInjectionMethod || *injectionMethod == common.LoaderFallbackToPodManifestInjectionMethod {
+		odigosLoaderPath := filepath.Join(k8sconsts.OdigosAgentsDirectory, commonconsts.OdigosLoaderName)
+
+		manifestValExits := getContainerEnvVarPointer(&container.Env, commonconsts.LdPreloadEnvVarName) != nil
+		runtimeDetailsVal, foundInInspection := getEnvVarFromRuntimeDetails(runtimeDetails, commonconsts.LdPreloadEnvVarName)
+		runtimeUnsetOrExpected := !foundInInspection || runtimeDetailsVal == odigosLoaderPath
```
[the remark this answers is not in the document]
RonFed: see other comment: I was thinking using the `EnvVars` can detect more cases (scripts)

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -38,6 +40,45 @@ func InjectOdigosAgentEnvVars(ctx context.Context, logger logr.Logger, podWorklo
...
+
+		if !manifestValExits && runtimeUnsetOrExpected {
+			// adding to the pod manifest env var:
+			// if the env var is not already present in the manifest and the runtime details env var is not set or set to the odigos loader path.
+			// the odigos loader path may be detected in the runtime details from previous installations or from terminating pods.
+			container.Env = append(container.Env, corev1.EnvVar{
```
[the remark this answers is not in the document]
RonFed: I think here: https://github.com/odigos-io/odigos/blob/f44dbdfe0548f0d3c6598475c829bc372ee43acb/instrumentor/internal/webhook_env_injector/webhook_env_injector.go#L36-L39

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
RonFed: Good point. Do you think I should return error from this function?

tests/common/assert/simple-demo-instrumented-full.yaml
```diff
@@ -44,10 +44,10 @@ spec:
...
       (env[?name == 'OTEL_RESOURCE_ATTRIBUTES']):
         - value: 'k8s.pod.name=$(ODIGOS_POD_NAME),k8s.container.name=frontend,k8s.namespace.name=default,k8s.deployment.name=frontend'
-      (env[?name == 'JAVA_OPTS']):
-        - value: '-javaagent:/var/odigos/java/javaagent.jar'
-      (env[?name == 'JAVA_TOOL_OPTIONS']):
-        - value: '-javaagent:/var/odigos/java/javaagent.jar'
+      # (env[?name == 'JAVA_OPTS']):
+      #   - value: '-javaagent:/var/odigos/java/javaagent.jar'
+      # (env[?name == 'JAVA_TOOL_OPTIONS']):
+      #   - value: '-javaagent:/var/odigos/java/javaagent.jar'
```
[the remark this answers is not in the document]
RonFed: I reverted this - since the default is going to be pod-manifest env injection - this assertion is still valid.

edeNFed:
Looks great!

BenElferink (approved):

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
tamirdavid1: return an error will just log it but it will still be marked as Instrumented

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -38,6 +40,53 @@ func InjectOdigosAgentEnvVars(ctx context.Context, logger logr.Logger, podWorklo
...
+		return
+	}
+
+	// check if odigos loader should be used
+	if *injectionMethod == common.LoaderEnvInjectionMethod || *injectionMethod == common.LoaderFallbackToPodManifestInjectionMethod {
+		odigosLoaderPath := filepath.Join(k8sconsts.OdigosAgentsDirectory, commonconsts.OdigosLoaderName)
+
+		manifestValExits := getContainerEnvVarPointer(&container.Env, commonconsts.LdPreloadEnvVarName) != nil
+		runtimeDetailsVal, foundInInspection := getEnvVarFromRuntimeDetails(runtimeDetails, commonconsts.LdPreloadEnvVarName)
+		ldPreloadUnsetOrExpected := !foundInInspection || runtimeDetailsVal == odigosLoaderPath
```
blumamir, line 56: should we use `strings.Contains()` instead of equals to cover cases where our value is added to some other value?

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
blumamir: I think yes, and we should avoid any modifications to the pod manifest (especially setting the label that indicates that it is instrumented).

might be tricky to achieve

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -38,6 +40,53 @@ func InjectOdigosAgentEnvVars(ctx context.Context, logger logr.Logger, podWorklo
...
+				Name:  commonconsts.LdPreloadEnvVarName,
+				Value: odigosLoaderPath,
+			})
+			return
+		}
+
+		// the LD_PRELOAD env var is preset. for now, we don't attempt to append our value to the user defined one.
+		if *injectionMethod == common.LoaderEnvInjectionMethod {
+			// we're specifically requested to use the loader env var injection method
+			// and the user defined LD_PRELOAD env var is already present.
```
tamirdavid1, line 73:
```suggestion
			// and the user defined LD_PRELOAD env var is already present or running in a secure execution mode.
```

blumamir (approved):

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -38,6 +40,53 @@ func InjectOdigosAgentEnvVars(ctx context.Context, logger logr.Logger, podWorklo
...
+	}
+
+	// check if odigos loader should be used
+	if *injectionMethod == common.LoaderEnvInjectionMethod || *injectionMethod == common.LoaderFallbackToPodManifestInjectionMethod {
+		odigosLoaderPath := filepath.Join(k8sconsts.OdigosAgentsDirectory, commonconsts.OdigosLoaderName)
+
+		manifestValExits := getContainerEnvVarPointer(&container.Env, commonconsts.LdPreloadEnvVarName) != nil
+		runtimeDetailsVal, foundInInspection := getEnvVarFromRuntimeDetails(runtimeDetails, commonconsts.LdPreloadEnvVarName)
+		ldPreloadUnsetOrExpected := !foundInInspection || runtimeDetailsVal == odigosLoaderPath
+		notSecureExecution := runtimeDetails.SecureExecutionMode != nil && !*runtimeDetails.SecureExecutionMode
```
tamirdavid1, line 57: I think better check for SecureExecution (without Not), it will simplify the `case !notSecureExecution:` later in the code

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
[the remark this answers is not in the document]
tamirdavid1: indeed EnvVars is from proc which include merge from all env sources

PR #2749, opened 2025-04-20
Add compression and headers fields to the OTLP/HTTP destination

AvihuHenya:
Screenshot 2025-04-20 at 16 10 04
Screenshot 2025-04-20 at 14 37 12
Screenshot 2025-04-20 at 14 36 38
## How Has This Been Tested?
I used a New Relic account to test the destination configuration (New Relic URL and API key as a header) against the kv-mall applications. I was able to visualize that logs and metrics were successfully sent to that destination.

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [X] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

- [ ] Changes how Odigos interacts with Kubernetes
- [ ] Introduces additional calls to the API Server (potential performance impact)
- [ ] New Query/feature supported in all the k8s versions supported by Odigos
- [ ] Modifies Odigos manifests (addressed in both CLI and Helm)
- [ ] Changes RBAC permissions

PR #2763, opened 2025-04-22
Add support for optional fields in otlpgrpc destination configuration

AvihuHenya:
Similar to [this PR](https://github.com/odigos-io/odigos/pull/2749), but targeting the generic-otlp (gRPC) destination. It introduces an additional optional field, insecure_skip_verify, which is relevant only when TLS is enabled.
Screenshot 2025-04-22 at 17 50 47
Screenshot 2025-04-22 at 17 49 44


common/config/genericotlp.go
```diff
@@ -17,23 +23,63 @@ func (g *GenericOTLP) DestType() common.DestinationType {
 }
 
 func (g *GenericOTLP) ModifyConfig(dest ExporterConfigurer, currentConfig *Config) ([]string, error) {
-	url, exists := dest.GetConfig()[genericOtlpUrlKey]
+	config := dest.GetConfig()
+
+	url, exists := config[genericOtlpUrlKey]
 	if !exists {
 		return nil, errors.New("Generic OTLP gRPC endpoint not specified, gateway will not be configured for otlp")
 	}
```
BenElferink, line 31: nit: we can simplify this error, see utils here: https://github.com/odigos-io/odigos/blob/54b624935cb2398a8761b4c0bdc61f73dcc93506/common/config/utils.go#L145C6-L145C21

```suggestion
	url, exists := config[genericOtlpUrlKey]
	if !exists {
		return nil, errorMissingKey(genericOtlpUrlKey)
	}
```

common/config/genericotlp.go
```diff
@@ -17,23 +23,63 @@ func (g *GenericOTLP) DestType() common.DestinationType {
...
+		"insecure": !tlsEnabled,
+	}
+	caPem, caExists := config[genericOtlpCaPemKey]
+	if caExists && caPem != "" {
+		tlsConfig["ca_pem"] = caPem
+	}
+	insecureSkipVerify, skipExists := config[genericOtlpInsecureSkipVerify]
+	if skipExists && insecureSkipVerify != "" {
+		tlsConfig["insecure_skip_verify"] = insecureSkipVerify
+	}
```
BenElferink, line 51: If `insecure_skip_verify` is expecting type boolean, then it might fail.
Consider using `parseBool` from utils here: https://github.com/odigos-io/odigos/blob/54b624935cb2398a8761b4c0bdc61f73dcc93506/common/config/utils.go#L129C6-L129C15

```suggestion
	if skipExists && insecureSkipVerify != "" {
		tlsConfig["insecure_skip_verify"] = parseBool(insecureSkipVerify)
	}
```

BenElferink (approved):
🏆🚀

Don't forget to generate destination docs:
```
cd docs
python3 sync-dest-doc.py
```

common/config/genericotlp.go
AvihuHenya: From what i check the config value is of type `type GenericMap map[string]interface{}`, so it can hold both types. 
wdyt?

PR #2764, opened 2025-04-23
Support setting a NodeSelector on Odigos components

damemi:
## Description

This adds the option to the Operator to support installing Odigos only on specific nodes. This can be extended to CLI and Helm too, if desired.

This may seem odd, because it means that Odigos will only be able to instrument applications that are scheduled onto the same node(s) that Odigos is installed on.  Because of that, this likely won't get used often. 

But we have seen a user that is running in a shared, but restricted, environment where they have control over their own applications, but not the underlying node infrastructure. They also have a very restricted quota in their namespace.

In this case, the Odiglet Daemonset is consuming their quota because it is attempting to schedule an odiglet on every (shared) node, even though they only need one. In this case, it is up to the user to ensure their application is also scheduled on the proper node, but this gives them the option to limit how many Odiglets are scheduled.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


RonFed:
Maybe I missed it. We should also apply the node selector to the data-collection daemonset, right?

autoscaler/controllers/clustercollector/deployment.go
```diff
@@ -37,14 +37,20 @@ func syncDeployment(dests *odigosv1.DestinationList, gateway *odigosv1.Collector
...
+
 	secretsVersionHash, err := destinationsSecretsVersionsHash(ctx, c, dests)
 	if err != nil {
 		return nil, errors.Join(err, errors.New("failed to get secrets hash"))
 	}
 
 	// Use the hash of the secrets  to make sure the gateway will restart when the secrets (mounted as environment variables) changes
 	configDataHash := common.Sha256Hash(secretsVersionHash)
-	desiredDeployment, err := getDesiredDeployment(dests, configDataHash, gateway, scheme, imagePullSecrets, odigosVersion)
+	desiredDeployment, err := getDesiredDeployment(dests, configDataHash, gateway, scheme, imagePullSecrets, odigosVersion, odigletNodeSelector)
```
RonFed, line 53: It is more important to add the node selector to the node collector I guess.

autoscaler/controllers/nodecollector/daemonset.go
```diff
@@ -155,7 +155,7 @@ func syncDaemonSet(ctx context.Context, datacollection *odigosv1.CollectorsGroup
 func getOdigletDaemonsetPodSpec(ctx context.Context, c client.Client, namespace string) (*corev1.PodSpec, error) {
 	odigletDaemonset := &appsv1.DaemonSet{}
 
-	if err := c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: odigletDaemonSetName}, odigletDaemonset); err != nil {
+	if err := c.Get(ctx, client.ObjectKey{Namespace: namespace, Name: k8sconsts.OdigletDaemonSetName}, odigletDaemonset); err != nil {
```
RonFed, line 158: if we're using the global const, we can remove the local one

cli/cmd/install.go
```diff
@@ -187,6 +195,22 @@ func installOdigos(ctx context.Context, client *kube.Client, ns string, config *
 
 }
 
+func parseNodeSelectorFlag() (map[string]string, error) {
+	nodeSelector := make(map[string]string)
+	if len(nodeSelector) == 0 {
+		return nodeSelector, nil
+	}
```
RonFed, line 202: this is creating an empty map, and check if that map is empty

```suggestion
	nodeSelector := make(map[string]string)
	if len(nodeSelectorFlag) == 0 {
		return nodeSelector, nil
	}
```

helm/odigos/values.yaml
```diff
@@ -202,3 +202,5 @@ gke:
   enabled: false
 
 profiles: []
+
+nodeSelector: {}
```
RonFed, line 206: Can you please add a short comment above, explaining this?

helm/odigos/templates/ui/deployment.yaml
```diff
@@ -17,6 +17,10 @@ spec:
       labels:
         app: odigos-ui
     spec:
+      {{- if .Values.nodeSelector }}
+      nodeSelector:
+      {{- toYaml .Values.nodeSelector | nindent 8 }}
```
RonFed, line 22: I'm not familiar with this syntax,
do we have a way to test that it compiles to a valid yaml in the CI?

autoscaler/controllers/clustercollector/deployment.go
damemi: the node collector already copies the Odiglet's node selector: https://github.com/odigos-io/odigos/blob/0cb5130a3244e645bca3ce06f59259d35790dd13/autoscaler/controllers/nodecollector/daemonset.go#L211

Since this setting applies to all components, this was conveniently already here

helm/odigos/templates/ui/deployment.yaml
damemi: We do have a test that lints the helm charts https://github.com/odigos-io/odigos/blob/0cb5130a3244e645bca3ce06f59259d35790dd13/.github/workflows/helm-lint.yml#L9

As for valid yaml, I could add another test case to the helm test that sets a nodeSelector value

damemi: see change here https://github.com/odigos-io/odigos/pull/2764/files#diff-e44f2077117d0cdb342309ca1b353df7892f943cce6e3cf8e863216f4b8dd2de

should be a no-op (since odiglets already have the os=linux selector) but will make sure that the helm chart can be applied when this value is set

RonFed (approved):
We can also add this to the `odigos config set` command which will allow us some more flexibility to modify this value in an existing installation.

damemi:
> We can also add this to the `odigos config set` command which will allow us some more flexibility to modify this value in an existing installation.

done!

PR #2765, opened 2025-04-23
ci: add PHP e2e tests (and support multi-arch)

BenElferink:
## Description

- Added simple-demo-app (named "currency") in E2E tests to support PHP testing
- Moved all duplicated yamls from queries dir to common/queries
- Bumped  PHP agent to support `amd64` and `arm64`

## How Has This Been Tested?

- [x] Updated e2e Tests

NOTE:
> CLI Upgrade will keep on failing until release because the current release contains PHP for 'arm64' only.
> Support for 'amd64' was added in this PR and won't be seen by CLI upgrade until release.


BenElferink:
CLI Upgrade will keep on failing until release because the current release contains PHP for 'arm64' only.

Support for 'amd64' was added in this PR and won't be seen by CLI upgrade until release.

tests/common/apply/install-simple-demo.yaml
```diff
@@ -16,11 +16,11 @@ spec:
     spec:
       containers:
         - name: coupon
-          image: registry.odigos.io/odigos-demo-coupon:v0.1
-          imagePullPolicy: IfNotPresent
+          image: registry.odigos.io/odigos-demo-coupon:v0.1.15
+          imagePullPolicy: Always
```
RonFed, line 20: for CI speed, we should prefer to not alway pull.
Not sure if we do pull every time today (probably we are) - but keeping this `IfNotPresent` will allow us to improve that.

tests/common/assert/odigos-upgraded.yaml
RonFed, line 1: this is used only in the cli-upgrade, so we don't have to move it to the `common` I guess

tests/common/queries/context-propagation.yaml
```diff
@@ -2,6 +2,14 @@ apiVersion: e2e.tests.odigos.io/v1
 kind: TraceTest
 description: This test checks if the context propagation is working correctly between different languages
 query: |
+  length(
+    [?
+      span.resourceAttributes."telemetry.sdk.language" == 'php' && span.serviceName == 'currency' && span.kind == 'server'
+        && 
+      parent.span.serviceName == 'frontend' && parent.span.resourceAttributes."telemetry.sdk.language" == 'java'
+    ]
+  ) > `0` &&
```
RonFed, line 11: big like

tests/e2e/helm-chart/chainsaw-test.yaml
```diff
@@ -71,36 +71,20 @@ spec:
...
-                kind load docker-image registry.odigos.io/odigos-demo-pricing:v0.1
-                kind load docker-image registry.odigos.io/odigos-demo-frontend:v0.2
-              else
-                echo "Skipping docker pull and kind load for cross-cloud-tests mode"
-              fi
         - apply:
             file: ../../common/apply/install-simple-demo.yaml
         - script:
-            timeout: 70s
+            timeout: 2m
```
RonFed, line 80: Is this increased because of the PHP service? do we have a way to make it go up faster?

tests/e2e/cli-upgrade/chainsaw-test.yaml
```diff
@@ -135,37 +132,38 @@ spec:
...
-
-              # Wait for the job to complete
-              job_name=$(kubectl get -f ../../common/apply/generate-traffic-job.yaml -o=jsonpath='{.metadata.name}')
-              kubectl wait --for=condition=complete job/$job_name
-
-              # Delete the job
-              kubectl delete -f ../../common/apply/generate-traffic-job.yaml
+              kubectl wait --for=condition=complete job/$(kubectl get -f ../../common/apply/generate-traffic-job.yaml -o=jsonpath='{.metadata.name}')
+        - delete:
+            file: ../../common/apply/generate-traffic-job.yaml
```
RonFed, line 142: Not sure if the order matters here.
We first need to create the job and than wait for its completion
The way it's written it looks the other way around.

RonFed (approved):

tests/common/apply/install-simple-demo.yaml
BenElferink: You're right I'll revert this, it was important during draft-testing.

tests/common/assert/odigos-upgraded.yaml
BenElferink: My goal was to clear the queries directory from each individual test. Hopefully I can take care of the source and workload lifecycle tests soon too so we have a single directory for all apply/assert/query yaml files.

tests/e2e/helm-chart/chainsaw-test.yaml
BenElferink: It doesn't need the extra time, it was just "synced" with the timeouts defined in the other tests

tests/e2e/cli-upgrade/chainsaw-test.yaml
BenElferink: It goes by order: assert -> await -> delete

tests/e2e/cli-upgrade/chainsaw-test.yaml
```diff
@@ -58,23 +58,20 @@ spec:
         - assert:
             timeout: 1m
             file: ../../common/assert/simple-trace-db-running.yaml
-    - name: Install Demo App
+
+    - name: Install - Simple Demo Apps
       try:
-        - script:
```
blumamir, line 63: awesome to have this removed 🎉

tests/e2e/helm-chart/chainsaw-test.yaml
```diff
@@ -133,37 +117,38 @@ spec:
...
                 if [ $? -eq 0 ]; then
                   break
                 fi
+                sleep 1
               done
+      catch:
+        - script:
+            content: |
+              java_pod_name=$(kubectl get pods -n default -o jsonpath="{.items[*].metadata.name}" | tr ' ' '\n' | grep ^frontend)
+              kubectl logs $java_pod_name -n default
```
blumamir, line 145: we do we print specifically java here?

blumamir (approved):
So many improvements, big like!

tests/e2e/helm-chart/chainsaw-test.yaml
BenElferink: frontend generates the trace, if there's an issue it can log it for us

PR #2777, opened 2025-04-24
ci: notify slack on success / failure of offset check pre release

blumamir:
When we push a new tag for release, and the offset PR in enterprise repo is not merge, the release stops but there isn't any message in slack for that.

This PR adds a success / error messages so we can be aware + a link to the offsets PR for quick reaction with minimal toll.

It's mostly chatGPT and not trivial to test, so we need to keep an eye in the next few releases.

We also need to think if it's easy to re-trigger the release in OSS after the offsets PR is merged (in case one stopped due to missing offsets) but that's a task for future work.

### DIFF
--- a/.github/workflows/publish-modules.yml
+++ b/.github/workflows/publish-modules.yml
@@ -20,23 +20,26 @@ jobs:
       - name: Verify no open offsets PRs
         id: verify
         run: |
+          # Fetch open PRs and filter by "offsets" label
           result=$(curl -s -H "Accept: application/vnd.github+json" \
             -H "Authorization: Bearer ${{ secrets.RELEASE_BOT_TOKEN }}" \
             "https://api.github.com/repos/odigos-io/${{ matrix.repo }}/pulls?state=open&per_page=100")
 
-          pr_links=$(echo "$result" | jq '[.[] | select(.labels | any(.name == "offsets")) | .html_url]')
+          pr_links=$(echo "$result" \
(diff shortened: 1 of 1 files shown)


BenElferink (approved):

.github/workflows/publish-modules.yml
```diff
@@ -18,12 +18,52 @@ jobs:
...
       - name: Verify no open offsets PRs
+        id: verify
         run: |
-          curl -s -H "Accept: application/vnd.github+json" \
-          -H "Authorization: Bearer ${{ secrets.RELEASE_BOT_TOKEN }}" \
-          "https://api.github.com/repos/odigos-io/${{ matrix.repo }}/pulls?state=open&per_page=100" | \
-          jq '[.[] | select(.labels | any(.name == "offsets"))]' | \
-          jq 'if length > 0 then error("❌ Error: Open PRs with label \"offsets\" found!") else empty end'
+          # Fetch open PRs and filter by "offsets" label
+          result=$(curl -s -H "Accept: application/vnd.github+json" \
```
RonFed, line 24: I think doing it in one step will be shorter and simpler

RonFed (approved):

.github/workflows/publish-modules.yml
blumamir: I wanted it to be the same as all the other places where we do it. so it's more manageable

damemi (approved):

PR #2781, opened 2025-04-25
feat: honeycomb test connection and docs improvments

blumamir:
## Description

Improving docs for honeycomb, replacing some out-of-date images and instructions.
add "test connection" for honeycomb with auth api

## How Has This Been Tested?

- [x] Manual Testing

## Kubernetes Checklist

No

## User Facing Changes

"Test Connection" option added for honeycomb destination in UI


RonFed (changes requested):
The test connection can fully simulate the OTLP exported config, not sure why we need the exception function

blumamir:
> The test connection can fully simulate the OTLP exported config, not sure why we need the exception function

When sending empty otlp I get:

```
{
    "data": {
        "testConnectionForDestination": {
            "succeeded": false,
            "statusCode": 500,
            "destinationType": "honeycomb",
            "message": "Permanent error: rpc error: code = InvalidArgument desc = request body should not be empty",
            "reason": "failed to connect"
        }
    }
}
```

The auth endpoint is the reliable option to test connection with honeycomb

blumamir:
reference: https://api-docs.honeycomb.io/api/auth/getauth

edeNFed (approved):
Nice! We can probably design the test connection better to not having to do if else per unique destination but thats not critical for now

frontend/services/test_connection/test_connection.go
```diff
@@ -69,8 +69,47 @@ func getConnectionTester(exporterID string) ExporterConnectionTester {
...
+
+	client := &http.Client{}
+
+	honeycombEndpoint, found := dest.GetConfig()["HONEYCOMB_ENDPOINT"]
+	if !found {
+		return TestConnectionResult{Succeeded: false, Message: "HONEYCOMB_ENDPOINT not found in config", Reason: FailedToConnect, DestinationType: dest.GetType(), StatusCode: http.StatusInternalServerError}
+	}
+	authEndpoint := fmt.Sprintf("https://%s/1/auth", honeycombEndpoint)
+
+	req, err := http.NewRequest("GET", authEndpoint, nil)
```
RonFed, line 85: Please use http.NewRequestWithContext and pass the context from the resolver

RonFed (approved):

PR #2782, opened 2025-04-25
fix: Align helm to CLI by setting default imagePrefix to our registry

RonFed:
This PR fix a bug in which we didn't set the `imagePrefix` config option when installing with helm.
This cause weird behaviors for example:
When installing via helm and then running `odigos config set ...` the config command would read the current config - see an empty image prefix and would return an image like "odigos-ui/v1.0.177" - which is invalid and will result in `ImagePullBackOff`.

Adding the util imagePrefix helm function, allowing us to use it in:
1. When setting the `odigos-config` values.
2. When calling the util `imageName` which is now using the `imagePrefix` util.

Writing the imagePrefix correctly to the config map - allows compatibility with the CLI commands.


edeNFed (approved):

helm/odigos/values.yaml
```diff
@@ -5,7 +5,7 @@ imagePullSecrets: []
 # By default, images are pulled from odigos registry at `registry.odigos.io`
 # If you use custom or internal registry to serve in your cluster, you can set the imagePrefix to your registry.
 # For example, if you set imagePrefix to `myregistry.io/odigos`, the images will be pulled from `myregistry.io/odigos/odigos-<component>:<tag>`
-imagePrefix:
+imagePrefix: 'registry.odigos.io'
```
blumamir, line 8: didn't try to run it myself, but I think if we populate the imagePrefix in the `values.yaml`, then we can conflict with the logic for openshift [here](https://github.com/odigos-io/odigos/blob/de1554d717c8a0d83049007f9d65c6057c843a84/helm/odigos/templates/_helpers.tpl#L4).

blumamir (approved):

PR #2783, opened 2025-04-25
Add generic Action CRD

damemi:
## Description

This adds a generic `Action` CRD that can be used to configure any Action. This PR implements it for the following basic existing Actions:

* AddClusterInfo
* DeleteAttribute
* PiiMasking
* RenameAttribute

It re-defines the non-common fields of each Action as its own `<Action>Config` object. These objects implement the `ActionConfig` interface which currently provides 2 common functions used by all Actions:

* `ProcessorType()`
* `OrderHint()`

This simplifies common codepaths, centralizes declaring these variables into easy-to-find spots (right with the API type for the action's config), and enforces a level of compile time compatibility.

The existing Action controllers have been updated to share a function that generates the actual YAML config for the action alongside the new controller.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


autoscaler/controllers/actions/action_controller.go
```diff
@@ -0,0 +1,186 @@
...
+
+	err = r.Patch(ctx, processor, client.Apply, client.FieldOwner(action.Name), client.ForceOwnership)
+	if err != nil {
+		statusErr := r.reportReconciledToProcessorFailed(ctx, action, odigosv1.ActionTransformedToProcessorReasonFailedToCreateProcessor, err)
+		if statusErr == nil {
+			return ctrl.Result{}, err // return original error on success
+		} else {
+			logger := ctrl.LoggerFrom(ctx)
+			logger.Error(statusErr, "Failed to set status on action")
+			return ctrl.Result{}, err // return original error on success
```
blumamir, line 180:
```suggestion
			return ctrl.Result{}, err // return original error for the Patch fail
```

blumamir (approved):

api/actions/v1alpha1/addclusterinfo_types.go
```diff
@@ -31,6 +31,18 @@ type OtelAttributeWithValue struct {
...
+func (AddClusterInfoConfig) ProcessorType() string {
+	return "resource"
+}
+
+func (AddClusterInfoConfig) OrderHint() int {
+	return 1
+}
+
 // AddClusterInfoSpec defines the desired state of AddClusterInfo action
 type AddClusterInfoSpec struct {
```
RonFed, line 47: Maybe we should mark all the old structs ad deprecated

api/odigos/v1alpha1/action_types.go
```diff
@@ -0,0 +1,103 @@
...
+	// Which signals should this action operate on.
+	Signals []common.ObservabilitySignal `json:"signals"`
+
+	AddClusterInfo *actionsv1.AddClusterInfoConfig `json:"addClusterInfo,omitempty"`
+
+	DeleteAttribute *actionsv1.DeleteAttributeConfig `json:"deleteAttribute,omitempty"`
+
+	PiiMasking *actionsv1.PiiMaskingConfig `json:"piiMasking,omitempty"`
+
+	RenameAttribute *actionsv1.RenameAttributeConfig `json:"renameAttribute,omitempty"`
```
RonFed, line 66: Do we consider a state in which more than one of these is not nil to be invalid?
if so, do we want to enforce that?

api/actions/v1alpha1/addclusterinfo_types.go
blumamir: I would suggest that we complete the migration for all actions, change in UI, profiles and docs to use the new CRD and then add a deprecation on old CRDs

api/odigos/v1alpha1/action_types.go
blumamir: Do you mean as a webhook?

BenElferink:
Very cool change!
Will be looking forward to this on the UI side 😃

---

- One action was not listed though, that's `K8sAttributesResolver`.
- And also, our Samplers (latency, probability, error and a couple new ones) are defined as actions, think we can include them as well? Or should we separate samplers from actions?

autoscaler/controllers/actions/root.go
```diff
@@ -1,13 +1,23 @@
...
+
+	v1 "github.com/odigos-io/odigos/api/actions/v1alpha1"
+	odigosv1 "github.com/odigos-io/odigos/api/odigos/v1alpha1"
 )
 
 func SetupWithManager(mgr ctrl.Manager) error {
-
 	err := ctrl.NewControllerManagedBy(mgr).
+		For(&odigosv1.Action{}).
+		Complete(&ActionReconciler{
```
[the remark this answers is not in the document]
damemi: GenerationChangedPredicate alone should work, that only triggers for updates to `spec`, not status

api/odigos/v1alpha1/action_types.go
damemi: Yeah this needs to be enforced somehow. A validation webhook is good, but also maybe add the same check in the controller itself in case the webhook goes down

This is one spot where having a generic interface field that can just be type-switched would be helpful if kubebuilder allowed it (https://github.com/kubernetes-sigs/controller-tools/issues/636)

tests/e2e/action-webhooks/chainsaw-test.yaml
RonFed, line 1: I think we should not do this as an e2e.
Our e2e takes time and overhead and in general not stable.
This kind of test can be done in a controller suit I think and completely in Go.
It is also true for the `sources` and `sources-webhook` tests, but since this is a new feature I don't think we should add another e2e.

damemi: I agree that we're overloading the e2es a bit. Unfortunately, the controller-suite isn't well suited for testing webhooks (you can unit test the validation functions, but actually setting up the endpoint and testing that API objects actually get validated is outside the scope of the suite tooling, this is why I had to do the Source Webhook e2e)

I also think it is important that we have some cases to make sure we're installing the webhook correctly, configuring the service on the pods, generating the right TLS cert, etc, which we need an e2e for.

You raise a good point that this doesn't need to be its own full e2e test, so I combined it with the existing Source Webhook e2e to reduce the runtime and risk of flaking (setting up Odigos is what usually causes flakes and takes the longest). I'm going to be working on refactoring a lot of the e2es this/next week into unit tests so I'll take another look at this then too.

RonFed: I think the validation logic should be tested in go.
I also saw this as an example for testing a webhook: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/webhook/server_test.go
maybe it can help.
The e2e that da installation and instrumentation are implicitly testing that the webhooks are set up correctly.

PR #2784, opened 2025-04-26
fix: remove sources in uninstall and wait for the pods to roll out cleanly

RonFed:
## Description

Fixing a bug in which the instrumented pods were not rolled out during uninstall to remove the instrumentation.
There was an existing code that removes the Sources, but that was done after the instrumentor deployment was removed - hence, the controller was missing and the pods remained "instrumented".

This PR adds the sources deletion as the first step of the uninstall process, and potentially waits for all the pods to roll out before continuing to delete the rest of the Odigos resources.

Waiting for pods to roll out without odigos modifications can take time in larger clusters, hence a `--no-wait` flag is added which skips this waiting phase.

On the other hand, when deleting the Sources we have to wait for the `instrumentor` to remove them. Not doing that can lead to garbage sources being left behind, and even the Source CRD can remain after an uninstall (since once a CRD is deleted, a finalizer is added to only remove it once all its instances are deleted).

```
➜  cli git:(fix_uninstall) ✗ go run -tags=embed_manifests . uninstall --yes          
Removing Odigos Sources...                             ✔
Waiting for pods to rollout without instrumentation...
        Waiting for 12 pods to rollout without instrumentation...
        Waiting for 3 pods to rollout without instrumentation...
        Waiting for 3 pods to rollout without instrumentation...
        Waiting for 2 pods to rollout without instrumentation...
All pods rolled out without instrumentation            ✔
Uninstalling Odigos Deployments                        ✔
Uninstalling Odigos DaemonSets                         ✔
Uninstalling Odigos ConfigMaps                         ✔
Uninstalling Odigos Services                           ✔
Uninstalling Odigos RBAC                               ✔
Uninstalling Odigos Secrets                            ✔
Uninstalling Odigos MutatingWebhookConfigurations      ✔
Uninstalling Odigos ValidatingWebhookConfigurations    ✔
Uninstalling Namespace odigos-system                   ✔
Waiting for namespace to be deleted                    ✔
Cleaning up Odigos node labels                         ✔
Rolling back odigos changes to pods                    ✔
Rolling back odigos changes to namespaces              ✔
Uninstalling Odigos CRDs                               ✔

SUCCESS: Odigos uninstalled.
```

## How Has This Been Tested?

- [x] Manual Testing
Tested manually uninstall via CLI and via Helm and made sure the pods are rolled out.

## Kubernetes Checklist

## User Facing Changes


cli/cmd/uninstall.go
```diff
@@ -143,6 +151,25 @@ func UninstallClusterResources(ctx context.Context, client *kube.Client, ns stri
 
 }
 
+func waitForPodsToRolloutWithoutInstrumentation(ctx context.Context, client *kube.Client) {
+	l := log.Print("Rolling back odigos changes to pods...")
+	instrumentedPodReq, _ := k8slabels.NewRequirement(k8sconsts.OdigosAgentsMetaHashLabel, selection.Exists, []string{})
+	wait.PollImmediate(1*time.Second, 5*time.Minute, func() (bool, error) {
```
blumamir, line 157: looks like [it's deprecated](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#PollImmediate) in favor of `PollUntilContextTimeout`:

```
Deprecated: This method does not return errors from context, use PollUntilContextTimeout. Note that the new method will no longer return ErrWaitTimeout and instead return errors defined by the context package. Will be removed in a future release.
```

cli/cmd/uninstall.go
```diff
@@ -143,6 +151,25 @@ func UninstallClusterResources(ctx context.Context, client *kube.Client, ns stri
...
+		if err != nil {
+			return false, err
+		}
+		if len(pods.Items) == 0 {
+			l.Success()
+			return true, nil
+		}
+		return false, nil
+	})
+}
```
blumamir, line 171: suggestion: print if all pods are rolled-out successfully or if we timed out and implications

cli/cmd/uninstall.go
```diff
@@ -143,6 +151,25 @@ func UninstallClusterResources(ctx context.Context, client *kube.Client, ns stri
 
 }
 
+func waitForPodsToRolloutWithoutInstrumentation(ctx context.Context, client *kube.Client) {
+	l := log.Print("Rolling back odigos changes to pods...")
```
blumamir, line 155: suggestion: we are going to preform an operation that can take 5 min.
Let's say something about it in the message so user knows cli is not stuck

BenElferink:
Big like 👍
Wish we could find an alternative to rollouts though, this could be a long wait on large clusters

BenElferink (approved):

damemi (approved):

cli/cmd/uninstall.go
```diff
@@ -429,21 +473,54 @@ func uninstallConfigMaps(ctx context.Context, client *kube.Client, ns, _ string)
...
+	if deleteErr != nil {
+		return deleteErr
+	}
+
+	// make sure all sources are deleted, this is required regardless of the --no-wait flag,
+	// in order to make sure the Source CRD can be deleted later in the uninstall process.
+	// failing to remove all the sources may cause the Source CRD to not get removed - since kubernetes
+	// has a finalizer on a CRD, waiting for all the CR instances to be deleted before removing the CRD.
+	err = wait.PollUntilContextTimeout(ctx, 5*time.Second, 1*time.Minute, true, func(innerCtx context.Context) (bool, error) {
+		sources, err := client.OdigosClient.Sources("").List(innerCtx, metav1.ListOptions{})
```
blumamir, line 500: nit: limit to 1 so not to overload the api server? especially on polling

cli/cmd/uninstall.go
```diff
@@ -429,21 +473,54 @@ func uninstallConfigMaps(ctx context.Context, client *kube.Client, ns, _ string)
...
+			l.Success()
+			return true, nil
+		}
+		return false, nil
+	})
+
+	if err != nil {
+		if errors.Is(err, context.DeadlineExceeded) {
+			fmt.Printf("\033[33m!\tWARN\033[0m deadline exceeded for waiting sources to be deleted\n")
+		} else if errors.Is(err, context.Canceled) {
```
blumamir, line 514: What will cause the context to get canceled?

cli/cmd/uninstall.go
```diff
@@ -429,21 +473,54 @@ func uninstallConfigMaps(ctx context.Context, client *kube.Client, ns, _ string)
...
+			return false, err
+		}
+		if len(sources.Items) == 0 {
+			l.Success()
+			return true, nil
+		}
+		return false, nil
+	})
+
+	if err != nil {
```
blumamir, line 511: Why some errors are `fmt.Printf` in this function, and others are returned from the function and are printed in the `Run`?
Should we keep it consistent?

cli/cmd/uninstall.go
```diff
@@ -57,6 +60,17 @@ and rollback any metadata changes made to your objects.`,
 				}
 			}
 
+			// delete all sources, and wait for the pods to rollout without instrumentation
+			// this is done before the instrumentor is removed, to ensure that the instrumentation is removed
+			err := removeAllSources(ctx, client)
+			if err != nil {
+				fmt.Printf("\033[31mERROR\033[0m Failed to remove all sources: %s\n", err)
+				os.Exit(1)
```
blumamir, line 68: `removeAllSources` handles some errors internally, in this case the installation will continue and remove instrumentor which might lead to stale sources, no?

cli/cmd/uninstall.go
```diff
@@ -143,6 +157,36 @@ func UninstallClusterResources(ctx context.Context, client *kube.Client, ns stri
...
+			l.Success()
+			return true, nil
+		}
+		log.Print(fmt.Sprintf("\tWaiting for %d pods to rollout without instrumentation...\n", len(pods.Items)))
+		return false, nil
+	})
+
+	if pollErr != nil {
+		if errors.Is(pollErr, context.DeadlineExceeded) {
+			fmt.Printf("\033[33m!\tWARN\033[0m deadline exceeded for waiting pods to roll out cleanly\n")
```
blumamir, line 182: This is good for the users to know, but what should they expect and do with this info? (which I guess is going to be quite common)

cli/cmd/uninstall.go
```diff
@@ -143,6 +157,36 @@ func UninstallClusterResources(ctx context.Context, client *kube.Client, ns stri
...
+		}
+		if errors.Is(pollErr, context.Canceled) {
+			fmt.Printf("\033[33m!\tWARN\033[0m context canceled for waiting pods to roll out cleanly\n")
```
blumamir, line 185: "context canceled" is an implementation detail of golang, and not a message we should show users of CLI in my opinion

cli/cmd/uninstall.go
```diff
@@ -143,6 +157,36 @@ func UninstallClusterResources(ctx context.Context, client *kube.Client, ns stri
 
 }
 
+func waitForPodsToRolloutWithoutInstrumentation(ctx context.Context, client *kube.Client) {
+	instrumentedPodReq, _ := k8slabels.NewRequirement(k8sconsts.OdigosAgentsMetaHashLabel, selection.Exists, []string{})
+	fmt.Printf("Waiting for pods to rollout without instrumentation...\n")
```
blumamir, line 162:
```suggestion
	fmt.Printf("Waiting for pods to rollout without instrumentation... This might take a while\n")
```

docs/cli/odigos_uninstall.mdx
```diff
@@ -34,8 +34,9 @@ odigos install
 ### Options
 
 ```
-  -h, --help   help for uninstall
-      --yes    skip the confirmation prompt
+  -h, --help      help for uninstall
+      --no-wait   skip waiting for pods to rollout without instrumentation
```
blumamir, line 38: nit: consider adding example to the above section

cli/cmd/uninstall.go
RonFed: I was think that doing CTRL+C would cause it but it looks like it did not.
I don't mind removing this or keeping it for future use when we'll have handling for it

cli/cmd/uninstall.go
RonFed: good point

cli/cmd/uninstall.go
RonFed: updated to return the error and print it in Run

blumamir (approved):

PR #2785, opened 2025-04-27
feat: include/exclude filters urltemplating

blumamir:
## Description

For url templating processor, add option to set include/exclude for specifying which workloads to act on

## How Has This Been Tested?

- [x] Added Unit Tests
- [x] Manual Testing

## Kubernetes Checklist

No

## User Facing Changes

More config options

### DIFF
--- a/collector/processors/odigosurltemplateprocessor/README.md
+++ b/collector/processors/odigosurltemplateprocessor/README.md
@@ -9,31 +9,106 @@ HTTP span names SHOULD be {method} {target} if there is a (low-cardinality) targ
 ```
 
 The target should be a templated string (e.g. not `/user/1234` but `/user/{id}`).
-The templated value is sometimes available to instrumentations in server spans where the framework and instrumentation supports such feature, but it is never available in client spans.
+The templated value is sometimes available to instrumentations in server spans where the framework and instrumentation supports such feature, but it is almost never available in client spans.
 
 To work around this, this processor will attempt to heuristically "guess" a templated value, and fill it in the span name and relevant attribute.
 
 ## Mechanism
 
-## Relevant Spans
+### Relevant Spans
 
 The following conditions must be met for a span to be considered relevant for this processor:
 
+0. the span matches any processor "include" or "exclude" filters which can limit the spans to be processed.
 1. an http span - contains `http.request.method` or `http.method` attribute.
 2. the attribute is not already set by instrumentation. e.g. no `http.route` for server spans and no `url.template` for client spans.
 3. the url path is recorded on a relevant attribute in the span (`url.path` / `url.full`) or the deprecated attributes (`http.target` / `http.url`).
 4. path can be parsed from the relevant attributes.
 
-## Templated Route Attribute
+### Templated Route Attribute
 
 For spans that match the above constraints, the processor will calculate the templated url and set it in the relevant attributes:
 
 - `url.template` - for client spans.
 - `http.route` - for server spans.
 
-## Span Name
+### Span Name
 
 If the span name equals the method (e.g. "GET"), and the processor is able to calculate a templated route, the span name will be set to `{method} {target}`. Otherwise, the span name will not be modified.
+
+## Configuration
+
+Example configuration: (see more details for each option below)
+
+```yaml
+processors:
+  odigosurltemplateprocessor:
+
+    # when include is set, the span must match at least one of the properties to be processed.
+    include:
+      k8s_workloads:
+        - namespace: "default"
+          kind: "deployment" ## or "daemonset" or "statefulset"
+          name: "myapp1"
+        - namespace: "default"
+          kind: "deployment"
+          name: "myapp2"
+
+    # when exclude is set, a span that matches the filter properties will be excluded from processing.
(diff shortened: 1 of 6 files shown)


collector/processors/odigosurltemplateprocessor/README.md
```diff
@@ -9,32 +9,107 @@ HTTP span names SHOULD be {method} {target} if there is a (low-cardinality) targ
...
 To work around this, this processor will attempt to heuristically "guess" a templated value, and fill it in the span name and relevant attribute.
 
 ## Mechanism
 
-## Relevant Spans
+### Relevant Spans
 
 The following conditions must be met for a span to be considered relevant for this processor:
 
+0. the span matches any processor "include" or "exclude" filters which can limit the spans to be processed.
```
RonFed, line 22: maybe worth to point to the section about include/exclude rules below, e.g:
"see section below about include/exclude rules."

collector/processors/odigosurltemplateprocessor/README.md
```diff
@@ -9,32 +9,107 @@ HTTP span names SHOULD be {method} {target} if there is a (low-cardinality) targ
...
+          name: "noisyapp"
+
+    # This option allows fine-tuning for specific paths to customize what to templatize and what not.
+    # The rule looks like this: "/v1/{foo:regex}/bar/{baz}".
+    # Each segment part in "{}" denote templatization, and all other segments should match the text exactly.
+    # Inside the "{}" you can optionally set the template name and matching regex.
+    # The template name is the name that will be used in the span name and attributes (e.g. "/users/{userId}").
+    # The regex is optional, and if provided, it will be used to match the segment.
+    # If the regex does not match, the rule will be skipped and other rules and templatization will be evaluated.
+    # Example: "/v1/{foo:\d+}" will match "/v1/123" producing "/v1/foo", but not with "/v1/abc".
```
RonFed, line 72:
```suggestion
    # Example: "/v1/{foo:\d+}" will match "/v1/123" producing "/v1/{foo}", but not with "/v1/abc".
```

collector/processors/odigosurltemplateprocessor/README.md
```diff
@@ -9,32 +9,107 @@ HTTP span names SHOULD be {method} {target} if there is a (low-cardinality) targ
...
+
+    # list of additional regex patterns that will be used to match and templated matching path segment.
+    # It allows users to define their own regex patterns for custom id formats used/observed in their applications.
+    # Note that this regexp should catch ids, but avoid catching other static unrelated strings.
+    # For example, if you have ids in the system like "ap123" then a regexp that matches "^ap\d+" would be good,
+    # but regexp like "^ap" is too permissive and will also catch "/api".
+    # compatible with golang regexp module https://pkg.go.dev/regexp
+    # for performance reasons, avoid using compute-intensive expressions or adding too many values here.
+    custom_ids_regexp:
+      - "^inc_\d+$"
```
RonFed, line 86: I think we should add a comment here clarifying the difference between these 2 fields and their use cases.

collector/processors/odigosurltemplateprocessor/filtermatcher.go
```diff
@@ -0,0 +1,108 @@
...
+type workloadStringRepresentation string
+
+func k8sWorkloadToStringRepresentation(workload K8sWorkload) workloadStringRepresentation {
+	lowercaseKind := strings.ToLower(workload.Kind)
+	return workloadStringRepresentation(fmt.Sprintf("%s/%s/%s", workload.Namespace, lowercaseKind, workload.Name))
+}
+
+func resourceToWorkloadStringRepresentation(resource pcommon.Resource) (workloadStringRepresentation, error) {
+
+	ns, ok := resource.Attributes().Get(string(semconv.K8SNamespaceNameKey))
```
RonFed, line 24: kind of a nit:
if we want to maximize performance here we can:
1. call resource.Attributes() once instead of re-creating the `Map` each time.
2. Use `resource.Attributes().Range(...)` - this will allow us to range over the elements once and save all that we need to make a decision.

RonFed (approved):

collector/processors/odigosurltemplateprocessor/README.md
blumamir: It's in the README bellow

collector/processors/odigosurltemplateprocessor/filtermatcher.go
blumamir: It's only run once per matcher and there should be 1 matcher

RonFed:
```
ns, ok := resource.Attributes().Get(string(semconv.K8SNamespaceNameKey))
...
deployment, ok := resource.Attributes().Get(string(semconv.K8SDeploymentNameKey))
...
statefulset, ok := resource.Attributes().Get(string(semconv.K8SStatefulSetNameKey))
...
daemonset, ok := resource.Attributes().Get(string(semconv.K8SDaemonSetNameKey))
```

so worst case 4 iterations on all the resources for every span

PR #2805, opened 2025-04-30
openshift: Check if SEManage rule already exists

damemi:
## Description

SELinux doesn't seem to consistently handle re-applying existing rules with `semanage`. It can sometimes result in an error that currently prevents the Odiglet from starting up. This PR does 2 things for RHEL hosts:

1. First check if an selinux rule exists for Odigos. If it does, don't both trying to re-apply
2. If we do hit some (unknown) error anyway, log the error and don't crash the Odiglet. This means we can still start up as a best effort in error cases, with the tradeoff that Odigos may not work if there was a different issue.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/odiglet/odiglet.go
+++ b/odiglet/odiglet.go
@@ -202,7 +202,6 @@ func OdigletInitPhase(clientset *kubernetes.Clientset) {
 	// executing selinux commands to make agents readable by pods.
 	if err := fs.ApplyOpenShiftSELinuxSettings(); err != nil {
 		log.Logger.Error(err, "Failed to apply SELinux settings on RHEL host")
-		os.Exit(-1)
 	}
 
 	os.Exit(0)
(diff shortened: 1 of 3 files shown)


odiglet/odiglet.go
```diff
@@ -202,7 +202,6 @@ func OdigletInitPhase(clientset *kubernetes.Clientset) {
 	// executing selinux commands to make agents readable by pods.
 	if err := fs.ApplyOpenShiftSELinuxSettings(); err != nil {
 		log.Logger.Error(err, "Failed to apply SELinux settings on RHEL host")
-		os.Exit(-1)
```
RonFed, line 205: maybe we should comment this out, and add here a comment about this being temporary.

RonFed (approved):

PR #2809, opened 2025-05-02
feat: add secure-execution mode detection

RonFed:
## Description

As a preparation for #2759, adding secure-execution mode detection to our runtime detection.
More details on what secure-execution mode is: https://man7.org/linux/man-pages/man8/ld.so.8.html.
We get the info from the auxiliary vector of the instrumented process:
-  https://www.man7.org/linux/man-pages/man5/proc_pid_auxv.5.html
-  https://www.man7.org/linux/man-pages/man3/getauxval.3.html

## How Has This Been Tested?

- [x] Added Unit Tests
For the `auxv` parsing function.
- [x] Updated e2e Tests
`workload-lifecycle` is updated to assert the secure execution mode is false on all the workloads except the `python-alpine` workload which is updated to run in secure-execution mode - and we assert that accordingly.
- [x] Manual Testing
To test the case where secure-execution is true, I used [ingress-nginx](https://github.com/kubernetes/ingress-nginx/tree/main) and verified we successfully identified it. (The reason for secure-execution there is adding extra capabilities to the non-root running process).

## Kubernetes Checklist

Added field to `InstrumentationConfig` in a backward compatible way.

## User Facing Changes

- [x] Users need to take action before upgrading
Existing `InstrumentationConfigs` will remain compatible - effectively having a `nil` secure-execution - which means we don't fully know it. This is a result of the way we do runtime detection today -  once per workload.


procdiscovery/pkg/process/process.go
```diff
@@ -166,12 +173,18 @@ func GetPidDetails(pid int) Details {
 	exePath := getExePath(pid)
 	cmdLine := getCommandLine(pid)
 	envVars := getRelevantEnvVars(pid)
+	secureExecutionMode, err := isSecureExecutionMode(pid)
+	secureExecutionModePtr := &secureExecutionMode
+	if err != nil {
+		secureExecutionModePtr = nil
```
edeNFed, line 179: Should we log something?

edeNFed (approved):

procdiscovery/pkg/process/process.go
RonFed: looking at the code, it looks we ignore all the errors from the different `/proc` operations.
I prefer to add this error reporting for all functions in a follow up PR.

procdiscovery/pkg/process/process.go
```diff
@@ -262,3 +275,31 @@ func getRelevantEnvVars(pid int) ProcessEnvs {
...
+	path := fmt.Sprintf("/proc/%d/auxv", pid)
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return false, fmt.Errorf("failed to read auxv: %w", err)
+	}
+
+	// https://www.man7.org/linux/man-pages/man5/proc_pid_auxv.5.html
+	for i := 0; i+16 <= len(data); i += 16 {
+		typ := binary.NativeEndian.Uint64(data[i : i+8])
+		val := binary.NativeEndian.Uint64(data[i+8 : i+16])
```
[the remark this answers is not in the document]
RonFed: done

tamirdavid1 (approved):
Approved,
can you just verify the sql spans that we discussed post sqlite deletion?
Thanks

PR #2813, opened 2025-05-04
Fix central prod install

alonkeyval:
This PR introduces a Service resource for the Central Backend component, enabling network communication between the Central UI and the Central Backend within the cluster.

Added a ClusterIP Service (default) for the Central Backend under the same namespace (typically odigos-central).

The service exposes port 8081, matching the backend's exposed port.

Ensures the Central UI can now reach the backend using internal DNS (e.g., http://central-backend.odigos-central.svc.cluster.local:8081/graphql).


cli/cmd/resources/centralodigos/centralui.go
```diff
@@ -60,6 +61,16 @@ func NewCentralUIDeployment(ns, imagePrefix, imageName, version string) *appsv1.
...
 							Name:  k8sconsts.CentralUI,
 							Image: containers.GetImageName(imagePrefix, imageName, version),
+							Env: []corev1.EnvVar{
+								{
+									Name:  k8sconsts.EnvNextPublicBackendHTTPURL,
+									Value: fmt.Sprintf("http://%s.%s.svc.cluster.local:8081/graphql", k8sconsts.CentralBackendName, ns),
+								},
+								{
+									Name:  k8sconsts.EnvNextPublicBackendWSURL,
+									Value: fmt.Sprintf("ws://%s.%s.svc.cluster.local:8081/graphql", k8sconsts.CentralBackendName, ns),
```
RonFed, line 71: the central UI can calculate these URL on his own
why do we want to pass it as an env var?

alonbraymok: You're right, and rethink it, it will not work when we will port forward the UI to the local user machine

PR #2821, opened 2025-05-05
instrumentor: Move Source e2e tests to controller suite

damemi:
## Description

This creates Source controller suite tests for the following cases:

**Workload Instrumentation**

* Creating a Workload Source results in an InstrumentationConfig for that workload
* Deleting a Workload Source removes the InstrumentationConfig for that workload
* Disabling a Workload Source removes the InstrumentationConfig for that workload

**Namespace Instrumentation**

* Creating a Namespace Source results in an InstrumentationConfig for workloads in that namespace
* Creating a Namespace Source results in an InstrumentationConfig for workloads in that namespace which are not already instrumented
* Creating a Namespace Source does not result in an InstrumentationConfig for explicitly excluded workloads in that namespace
* Deleting an Excluded Workload Source in an instrumented namespace results in an InstrumentationConfig for that workload
* Updating an Excluded Workload Source in an instrumented namespace to `disableInstrumentation=false` results in an InstrumentationConfig for that workload

**Namespace Uninstrumentation**

* Deleting a Namespace Source removes the InstrumentationConfig for each workload in that namespace
* Deleting a Namespace Source does not remove the InstrumentationConfig for explicitly enabled workloads in that namespace

Accordingly, the now-redundant and flaky e2e test cases for the above are removed. The e2e flow for Source instrumentation now only tests that instrumentation, propagation, and correlation are successful. It also still tests the label->source migration webhooks, which will be removed when the legacy labels are completely deprecated.

## How Has This Been Tested?

- [x] Added Unit Tests
- [x] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


instrumentor/controllers/sourceinstrumentation/source_controller_test.go
```diff
@@ -0,0 +1,195 @@
...
+				source = testutil.NewMockSource(deployment, false)
+				Expect(k8sClient.Create(ctx, source)).Should(Succeed())
+				testutil.AssertInstrumentationConfigCreated(ctx, k8sClient, instrumentationConfig)
+			})
+
+		})
+
+		When("Sources are uninstrumented", func() {
+			It("Deletes the InstrumentationConfig for the uninstrumented workload", func() {
+				Expect(k8sClient.Delete(ctx, source)).Should(Succeed())
```
RonFed, line 44: Do we need to first create the source in this block before deleting it?
Or do all the `It` blocks on the same level executed sequentially? (I think that's the case just want to make sure)

instrumentor/controllers/sourceinstrumentation/source_controller_test.go
```diff
@@ -0,0 +1,195 @@
...
+				source = testutil.NewMockSource(namespace, false)
+				Expect(k8sClient.Create(ctx, source)).Should(Succeed())
+				testutil.AssertInstrumentationConfigCreated(ctx, k8sClient, instrumentationConfig)
+			})
+
+			It("Creates an InstrumentationConfig for workloads in the namespace which are not already instrumented", func() {
+				deployment2 := testutil.NewMockTestDeployment(namespace, "test-deployment-2")
+				Expect(k8sClient.Create(ctx, deployment2)).Should(Succeed())
+				source2 := testutil.NewMockSource(deployment2, false)
+				Expect(k8sClient.Create(ctx, source2)).Should(Succeed())
```
RonFed, line 83: should we assert here that a corresponding instrumentation config was created for source2?
Or is the goal here is to create a new deployment in the instrumented ns, and see the instrumentation config getting created? If that is the case we should not create source2

instrumentor/controllers/sourceinstrumentation/source_controller_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+				source = testutil.NewMockSource(namespace, false)
+				Expect(k8sClient.Create(ctx, source)).Should(Succeed())
+				testutil.AssertInstrumentationConfigCreated(ctx, k8sClient, instrumentationConfig)
```
RonFed, line 87: source was already created in the previous It block and asserted there - should we only assert here for source2?

instrumentor/controllers/sourceinstrumentation/source_controller_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+			It("Creates an InstrumentationConfig when an Excluded Workload Source is deleted", func() {
+				deployment2 := testutil.NewMockTestDeployment(namespace, "test-deployment-2")
+				Expect(k8sClient.Create(ctx, deployment2)).Should(Succeed())
+				source2 := testutil.NewMockSource(deployment2, true)
+				Expect(k8sClient.Create(ctx, source2)).Should(Succeed())
+				instrumentationConfig2 := testutil.NewMockInstrumentationConfig(deployment2)
+
+				source = testutil.NewMockSource(namespace, false)
+				Expect(k8sClient.Create(ctx, source)).Should(Succeed())
```
RonFed, line 111: maybe this should be in the `BeforeEach`?

RonFed (approved):
🚀

instrumentor/controllers/sourceinstrumentation/source_controller_test.go
```diff
@@ -0,0 +1,195 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("Source controller", func() {
+	ctx := context.Background()
+	var namespace *corev1.Namespace
+	var deployment *appsv1.Deployment
+	var source *odigosv1.Source
+	var instrumentationConfig *odigosv1.InstrumentationConfig
```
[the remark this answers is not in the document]
damemi: Good question, we run `go test -p 1` which I believe should run them all sequentially

instrumentor/controllers/sourceinstrumentation/source_controller_test.go
damemi: I don't think resources are persisted across It blocks like that, is that the case? Otherwise there would be an error trying to create the Source again saying it already exists

instrumentor/controllers/sourceinstrumentation/source_controller_test.go
damemi: In some of these tests I'm testing what happens when a Source is created after another Source, so the order is important in those

PR #2823, opened 2025-05-06
feat: odigos central UI

alonkeyval:
closes #2509

new central command for portforward the central backend + UI to the user's local machine


cli/cmd/pro.go
```diff
@@ -31,6 +31,11 @@ var (
 	useDefault       bool
 )
 
+const (
+	CentralBackendPort = "8081"
+	CentralUIPort      = "3000"
+)
```
RonFed, line 37: Maybe we should put these consts in `k8sutils` since they will be used by the UI and central backend as wrell

cli/cmd/pro.go
```diff
@@ -271,6 +276,57 @@ func installCentralBackendAndUI(ctx context.Context, client *kube.Client, ns str
...
+		ctx := cmd.Context()
+		client := cmdcontext.KubeClientFromContextOrExit(ctx)
+
+		backendPod, err := findPodWithAppLabel(ctx, client, proNamespaceFlag, k8sconsts.CentralBackendAppName)
+		if err != nil {
+			fmt.Printf("\033[31mERROR\033[0m Cannot find backend pod: %v\n", err)
+			os.Exit(1)
+		}
+		go func() {
+			err := kube.PortForwardWithContext(ctx, backendPod, client, CentralBackendPort, "localhost")
```
RonFed, line 293: why do we need to port forward the backend?

alonkeyval: We're port-forwarding the backend because, unlike the open source version where the UI and backend are bundled together and served from a single pod, Odigos Central separates them into two distinct pods: one for the backend and one for the UI.

In the open source setup, the UI build files are served directly by the backend (e.g., from a Go HTTP server), so we only need to port-forward the backend to access both frontend and backend.

But in Odigos Central, the frontend and backend run independently, so to enable full communication (e.g., API calls from the UI to the backend), we must expose both services locally by port-forwarding both pods.

This ensures the UI can reach the backend on the expected port (typically 8081), and everything works as intended when running locally.

cli/pkg/kube/portforward.go
```diff
@@ -0,0 +1,78 @@
...
+		select {
+		case <-signals:
+		case <-returnCtx.Done():
+		}
+		close(stopChannel)
+	}()
+
+	fmt.Printf("Odigos UI is available at: http://%s:%s\n\n", localAddress, localPort)
+	fmt.Printf("Port-forwarding from %s/%s\n", pod.Namespace, pod.Name)
+	fmt.Printf("Press Ctrl+C to stop\n")
```
RonFed, line 38: since we're going to call 2 port forwards this will be printed twice,
maybe we should have the caller to this function print accordingly

cli/cmd/pro.go
```diff
@@ -271,6 +276,57 @@ func installCentralBackendAndUI(ctx context.Context, client *kube.Client, ns str
...
+		if err != nil {
+			fmt.Printf("\033[31mERROR\033[0m Cannot find UI pod: %v\n", err)
+			os.Exit(1)
+		}
+
+		err = kube.PortForwardWithContext(ctx, uiPod, client, CentralUIPort, "localhost")
+		if err != nil {
+			fmt.Printf("\033[31mERROR\033[0m UI port-forward failed: %v\n", err)
+			os.Exit(1)
+		}
```
RonFed, line 309: I think before we're exiting (when the user do ctrl+c), we need to wait for the other port forward in the goroutine to exit

cli/pkg/kube/portforward.go
```diff
@@ -0,0 +1,74 @@
...
+
+func PortForwardWithContext(ctx context.Context, pod *corev1.Pod, client *Client, localPort, localAddress string) error {
+	stopChannel := make(chan struct{}, 1)
+	readyChannel := make(chan struct{})
+	signals := make(chan os.Signal, 1)
+	signal.Notify(signals, os.Interrupt)
+	defer signal.Stop(signals)
+
+	returnCtx, returnCtxCancel := context.WithCancel(ctx)
+	defer returnCtxCancel()
```
RonFed, line 26: since now this function can be called multiple times - I think this part should be done by the caller once - i.e setting up the context and cancel for signlas should be done by the caller - and passing the context to this function

cli/cmd/resources/centralodigos/centralbackend.go
```diff
@@ -86,3 +88,35 @@ func NewCentralBackendDeployment(ns, imagePrefix, imageName, version string) *ap
...
+			},
+		},
+		Spec: corev1.ServiceSpec{
+			Type: corev1.ServiceTypeClusterIP,
+			Selector: map[string]string{
+				"app": k8sconsts.CentralBackendAppName,
+			},
+			Ports: []corev1.ServicePort{
+				{
+					Port:       8081,
```
RonFed, line 112: use the const of the port

RonFed (approved):
left one comment

PR #2825, opened 2025-05-06
Enhance runtime detection

tamirdavid1:
## Description

This PR introduce two changes 
1. Use `ImageStatus.Info` instead of `GetContainerInfo` - include values from Dockerfile
2. Merge RuntimedetailsByContainer in case ld_preload added/ secureexecution identified/runtime version updated

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


odiglet/pkg/kube/runtime_details/inspection.go
```diff
@@ -275,3 +295,57 @@ func GetRuntimeDetails(ctx context.Context, kubeClient client.Client, podWorkloa
...
+					found = true
+					break
+				}
+			}
+			if !found {
+				existing.EnvVars = append(existing.EnvVars, newEnv)
+				updated = true
+			}
+		}
+	}
```
RonFed, line 317: I don't think this should be a nested loop - since once we enter the inner loop it does not make sense to continue in the outer one, right?
maybe we should have
```Go
foundInExisting := false
for _, newEnv := range existing.EnvVars {
       if existingEnv.Name == consts.LdPreloadEnvVarName {
		foundInExisting = true
		break
	}
}

if !foundInExisting {
   for _, newEnv := range new.EnvVars {
        if newEnv.Name == consts.LdPreloadEnvVarName  {
           existing.EnvVars = append(existing.EnvVars, newEnv)
	   updated = true
        }
   }
}
```

odiglet/pkg/kube/runtime_details/inspection.go
```diff
@@ -241,11 +240,32 @@ func persistRuntimeDetailsToInstrumentationConfig(ctx context.Context, kubeclien
...
+	// 2. LD_PRELOAD is identified in EnvFromContainerRuntime [DockerFile]
+	// 3. SecureExecutionMode is set to true.
+	// 4. RuntimeVersion changes
 	if len(currentConfig.Status.RuntimeDetailsByContainer) > 0 {
-		return nil
+		updated := false
+		for _, newDetail := range newRuntimeDetials {
+			for j := range currentConfig.Status.RuntimeDetailsByContainer {
+				existingDetail := &currentConfig.Status.RuntimeDetailsByContainer[j]
+				if newDetail.ContainerName == existingDetail.ContainerName {
```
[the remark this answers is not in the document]
tamirdavid1: Sounds like a very niche issue, in case lang change we will need better solution of marking runtime detection as not valid and force complete runtime detection.
I think it is a good feature but need to handle in different PR

odiglet/pkg/kube/runtime_details/inspection.go
```diff
@@ -275,3 +295,55 @@ func GetRuntimeDetails(ctx context.Context, kubeClient client.Client, podWorkloa
...
+
+	// 1. Merge LD_PRELOAD from EnvVars [/proc/pid/environ]
+	odigosStr := "odigos"
+	mergeLdPreloadEnvVars(new.EnvVars, &existing.EnvVars, &odigosStr, &updated)
+
+	// 2. Merge LD_PRELOAD from EnvFromContainerRuntime [DockerFile]
+	mergeLdPreloadEnvVars(new.EnvFromContainerRuntime, &existing.EnvFromContainerRuntime, nil, &updated)
+
+	// 3. Update SecureExecutionMode if needed
+	if new.SecureExecutionMode != nil && *new.SecureExecutionMode {
```
blumamir, line 310: Why do we require that `*new.SecureExecutionMode` is true?
Don't we want to always update it to the correct value if it's not nil?

odiglet/pkg/kube/runtime_details/inspection.go
```diff
@@ -275,3 +295,55 @@ func GetRuntimeDetails(ctx context.Context, kubeClient client.Client, podWorkloa
...
+		if existing.SecureExecutionMode == nil || !*existing.SecureExecutionMode {
+			existing.SecureExecutionMode = new.SecureExecutionMode
+			updated = true
+		}
+	}
+
+	// 4. Update RuntimeVersion if different
+	if new.RuntimeVersion != "" && new.RuntimeVersion != existing.RuntimeVersion {
+		existing.RuntimeVersion = new.RuntimeVersion
```
blumamir, line 319: 🤩

odiglet/pkg/kube/runtime_details/inspection.go
```diff
@@ -275,3 +295,55 @@ func GetRuntimeDetails(ctx context.Context, kubeClient client.Client, podWorkloa
...
+	}
+
+	return updated
+}
+
+func mergeLdPreloadEnvVars(
+	newEnvs []odigosv1.EnvVar,
+	existingEnvs *[]odigosv1.EnvVar,
+	skipIfContains *string,
+	updated *bool,
```
blumamir, line 330: nit: consider returning the `updated` instead of changing it via side effect on a pointer

blumamir (approved):

odiglet/pkg/kube/runtime_details/inspection.go
tamirdavid1: Great comment, Thanks!

odiglet/pkg/kube/runtime_details/inspection.go
RonFed: Since we have this value per container, and there can be multiple processes in a container - I think we should be extra safe here - and avoid possible wrong updates - especially the transition from secure to not secure.
We can have a non deterministic behavior for containers that have multiple processes and and initialization scripts (since it depends when the runtime detection runs).

Making the secure execution flag stick if we found it to be true is not ideal, but accidentally updating it to false can have a bad impact on the instrumented container.

odiglet/pkg/kube/runtime_details/inspection.go
```diff
@@ -275,3 +295,52 @@ func GetRuntimeDetails(ctx context.Context, kubeClient client.Client, podWorkloa
...
+
+	// 1. Merge LD_PRELOAD from EnvVars [/proc/pid/environ]
+	odigosStr := "odigos"
+	updated = mergeLdPreloadEnvVars(new.EnvVars, &existing.EnvVars, &odigosStr)
+
+	// 2. Merge LD_PRELOAD from EnvFromContainerRuntime [DockerFile]
+	updated = mergeLdPreloadEnvVars(new.EnvFromContainerRuntime, &existing.EnvFromContainerRuntime, nil)
+
+	// 3. Update SecureExecutionMode if needed
+	if new.SecureExecutionMode != nil && existing.SecureExecutionMode != nil && *new.SecureExecutionMode != *existing.SecureExecutionMode {
```
RonFed, line 310: if the existing is nil and the new is not nil, we want ti update, right?

odiglet/pkg/kube/runtime_details/inspection.go
tamirdavid1: I think that changing it from nil should allowed both for true and false. but avoid changing false > true or true > false.
@blumamir @RonFed is this make sense for both of u?

k8sutils/pkg/cri/criwrapper.go
```diff
@@ -199,3 +135,74 @@ func extractContainerID(containerUri string) string {
...
+	imageRef := statusResp.GetStatus().GetImage().GetImage()
+	if imageRef == "" {
+		return nil, fmt.Errorf("image ref is empty for container %s", containerID)
+	}
+
+	// Step 2: Use CRI image service to inspect image env vars
+	return rc.GetImageEnvVarsFromCRI(ctx, imageRef)
+}
+
+func (rc *CriClient) GetImageEnvVarsFromCRI(ctx context.Context, imageRef string) (map[string]string, error) {
```
RonFed, line 167: nit: this function don't need to be exported as it is only used locally

RonFed (approved):

blumamir (approved):

PR #2826, opened 2025-05-06
Instrumentation rule for HTTP Headers

edeNFed:
The changes enable users to define rules for collecting specific HTTP headers. The changes include updates to the documentation and autogenerated code to reflect the new functionality.


api/odigos/v1alpha1/instrumentationrules/http_headers.go
```diff
@@ -0,0 +1,9 @@
+package instrumentationrules
+
+// +kubebuilder:object:generate=true
+// +kubebuilder:deepcopy-gen=true
+type HttpHeadersCollection struct {
+
+	// Limit payload collection to specific header keys.
+	HeaderKeys *[]string `json:"headerKeys,omitempty"`
```
RonFed, line 8: don't think this has to be a pointer

blumamir, line 8: need to also add it in docs:

https://docs.odigos.io/pipeline/rules/introduction

api/odigos/v1alpha1/instrumentationrules/http_headers.go
```diff
@@ -0,0 +1,9 @@
+package instrumentationrules
+
+// +kubebuilder:object:generate=true
+// +kubebuilder:deepcopy-gen=true
+type HttpHeadersCollection struct {
+
+	// Limit payload collection to specific header keys.
```
blumamir, line 7: nil means collect all headers and empty list is collect no headers?

RonFed (approved):

PR #2827, opened 2025-05-06
e2e: remove multi-apps test + fix frontend source metrics names

damemi:
## Description

This test is redundant with the Source e2e, except for:

1. Verify node odiglet label added
2. Verify collector metrics in the UI

These 2 can also probably be de-duplicated but this is a good first step

Update: This also fixes a bug in the frontend's source metrics service which stores metrics based on their _workload_ name, but retrieves them based on their _service_ name, which means when a source has a custom service name, its throughput metrics won't get retrieved by the graphql query. This was caught when moving the collector metrics UI case because the Source e2e sets a custom service name.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [x] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


tests/e2e/source/chainsaw-test.yaml
```diff
@@ -139,6 +145,19 @@ spec:
             name: odiglet
             namespace: odigos-test
 
+    - name: '[2 - Workload Instrumentation] Verify collector metrics are collected by the UI'
```
BenElferink, line 148: Wouldn't it be better to have this in `tests/e2e/ui/chainsaw-test.yaml`?

damemi: Probably, but I don't want to add traffic generation to that test right now since it doesn't need any traffic for its other cases as far as I can tell

BenElferink: You're right, I didn't think about that

frontend/services/collector_metrics/node_collector.go
```diff
@@ -177,23 +177,22 @@ func (sourcesMetrics *sourcesMetrics) sourcesMetrics() map[common.SourceID]traff
...
-	}
-
 	ns, ok := attrs.Get(K8SNamespaceNameKey)
 	if !ok {
 		return common.SourceID{}, errors.New("namespace not found")
 	}
 
 	var kind k8sconsts.WorkloadKind
-	if _, ok := attrs.Get(K8SDeploymentNameKey); ok {
+	var name pcommon.Value
```
BenElferink, line 186: ♥️🙏

RonFed (commented):
We should also remove the multi-apps test from the `Required` list in GitHub to avoid PRs looking stuck

blumamir (approved):

frontend/services/collector_metrics/node_collector.go
```diff
@@ -177,23 +177,22 @@ func (sourcesMetrics *sourcesMetrics) sourcesMetrics() map[common.SourceID]traff
 }
 
 func metricAttributesToSourceID(attrs pcommon.Map) (common.SourceID, error) {
-	name, ok := attrs.Get(ServiceNameKey)
```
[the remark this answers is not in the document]
BenElferink: From the UI side of things - serviceName is a "display only" value, we do not use it for identifying workloads for any CRUD operations.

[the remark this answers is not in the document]
damemi: yeah this const isn't used anywhere else, removed

PR #2832, opened 2025-05-07
Change cache access

TheErikNaz:
## Description
Change the cache access for a repo scope

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -18,8 +18,8 @@ jobs:
         uses: actions/cache@v4
         with:
           path: .lycheecache
-          key: cache-lychee-${{ github.repository_id }}
-          restore-keys: cache-lychee-${{ github.repository_id }}
+          key: cache-lychee-${{ github.event.pull_request.number }}
+          restore-keys: cache-lychee-${{ github.event.pull_request.number }}
 
       - name: Link Checker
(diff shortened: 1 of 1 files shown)


damemi (approved):

RonFed:
This means that each PR will have its own cache, right?
It's not in the repo scope as the PR headline says I guess.

.github/workflows/check-links.yml
```diff
@@ -42,4 +42,4 @@ jobs:
         if: always()
         with:
           path: .lycheecache
-          key: ${{ steps.restore-cache.outputs.cache-primary-key }}
+          key: cache-lychee-${{ github.event.pull_request.number }}
```
BenElferink, line 45: I think we can remove this entire segment.
We don't have to save the cache, that's an optional step and TBH I think it's interfering with the default/internal save:
```
Failed to save: Unable to reserve cache with key cache-lychee-2832, another job may be creating this cache.
Warning: Cache save failed.
```

TheErikNaz: Yeah 
I've made some more changes, now the cache is on a PR level
and also, only the first commit in the PR is cached

PR #2837, opened 2025-05-07
Add dependencies-syncer-bot PR check to release workflow

damemi:
## Description

Similar to https://github.com/odigos-io/odigos/pull/2680, this adds a pre-release check for open enterprise dependency sync PRs.

Currently, the release flow will create a release PR in the enterprise repo that is effectively a dependency sync. However, if there are currently failing tests with the current dependency sync, it would be good to catch these before we start a release flow instead of having to debug it mid release.

Making this a separate job so it can depend on the verify-offsets check, which itself will trigger a dependency bump if there are offsets changes to be merged

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/.github/workflows/publish-modules.yml
+++ b/.github/workflows/publish-modules.yml
@@ -65,8 +65,63 @@ jobs:
             \"tag\": \"verify-offsets-failure\"
           }" $SLACK_WEBHOOK_URL      
 
-  print-tag:
+  verify-dependencies-sync:
     needs: verify-offsets
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


damemi:
> I believe the release PR in Enterprise already inherits auto-syncer dependencies

Yeah, my goal here though is that if there are already failures with the dependency sync to catch that before we push a new tag to all our modules. So this forces us to fix it without starting a release

RonFed (approved):

PR #2843, opened 2025-05-08
destination: Add Splunk OTLPHTTP exporter

damemi:
## Description

According to the upstream docs, the [Splunk SAPM exporter is deprecated](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a0044fba6fb76f034be52fe7bb466f72b2c2314a/exporter/sapmexporter/README.md#deprecated) in favor of using an OTLPHTTP exporter configured for Splunk.

For users, we can add a preconfigured OTLP exporter that accepts the necessary Splunk config based on Splunk docs: https://docs.splunk.com/observability/en/gdi/opentelemetry/components/otlphttp-exporter.html#configuration-examples

This PR doesn't add every OTLPHTTP option, but just the ones we currently support:

* TLS
* Compression
* Splunk Realm
* Splunk Access Token

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/cli/cmd/pro.go
+++ b/cli/cmd/pro.go
@@ -11,6 +11,7 @@ import (
 	"os/signal"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/odigos-io/odigos/api/k8sconsts"
 	"github.com/odigos-io/odigos/cli/cmd/resources"
@@ -27,6 +28,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 var (
@@ -180,7 +182,13 @@ odigos pro update-offsets --default
 			os.Exit(1)
 		}
 
-		fmt.Println("Updated Go Offsets.")
+		fmt.Println("Updated Go Offsets, restarting Odiglet to use the new offsets.")
+		err = restartOdiglet(ctx, client, ns)
+		if err != nil {
+			fmt.Println(fmt.Sprintf("\033[31mERROR\033[0m Unable to restart Odiglet: %s", err))
+			os.Exit(1)
+		}
+		fmt.Println("Odiglet restarted successfully.")
 	},
 }
 
@@ -402,6 +410,22 @@ func findPodWithAppLabel(ctx context.Context, client *kube.Client, ns, appLabel 
 		return nil, fmt.Errorf("pod %s is not running", pod.Name)
 	}
 	return pod, nil
+}
+
+func restartOdiglet(ctx context.Context, client *kube.Client, ns string) error {
+	// Create patch to add/update the restartedAt annotation
+	patch := fmt.Sprintf(`{"spec":{"template":{"metadata":{"annotations":{"kubectl.kubernetes.io/restartedAt":"%s"}}}}}`,
+		time.Now().Format(time.RFC3339))
+
+	// Patch the Odiglet daemonset
+	_, err := client.AppsV1().DaemonSets(ns).Patch(
+		ctx,
+		k8sconsts.OdigletDaemonSetName,
+		types.StrategicMergePatchType,
+		[]byte(patch),
+		metav1.PatchOptions{},
+	)
+	return err
 }
 
 func init() {


damemi:
@BenElferink do you know what I need to update here to make the logo show up right for both destinations?
Screenshot 2025-05-08 at 3 26 58 PM

destinations/data/splunksapm.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: internal.odigos.io/v1beta1
 kind: Destination
 metadata:
-  type: splunk
-  displayName: Splunk
+  type: splunksapm
```
BenElferink, line 4: This will break the pipelines/data-streams of those who've already setup Splunk as a destination. An existing dest type cannot be changed without a migration script in cli/helm upgrade.

BenElferink (changes requested):

BenElferink:
> @BenElferink do you know what I need to update here to make the logo show up right for both destinations? Screenshot 2025-05-08 at 3 26 58 PM

Once merged to main, any new SVGs added should be uploaded to S3.
Any new dest icon should be added to the UI Kit for offline availability, but there's a fallback using the S3 bucket in-case the kit wasn't updated yet.

So typically just add the SVG, and reference it in the YAML. It'll be visible after merge.

damemi:
> > @BenElferink do you know what I need to update here to make the logo show up right for both destinations? Screenshot 2025-05-08 at 3 26 58 PM
> 
> Once merged to main, any new SVGs added should be uploaded to S3. Any new dest icon should be added to the UI Kit for offline availability, but there's a fallback using the S3 bucket in-case the kit wasn't updated yet.
> 
> So typically just add the SVG, and reference it in the YAML. It'll be visible after merge.

but, both of these are just using the existing `splunk.svg` that is already uploaded, shouldn't that work?

destinations/data/splunksapm.yaml
damemi: Ah I didn't know that, I assumed the autoscaler would do a re-sync on upgrade with a fresh cache since it's a new pod

BenElferink:
> > > @BenElferink do you know what I need to update here to make the logo show up right for both destinations? Screenshot 2025-05-08 at 3 26 58 PM
> > 
> > 
> > Once merged to main, any new SVGs added should be uploaded to S3. Any new dest icon should be added to the UI Kit for offline availability, but there's a fallback using the S3 bucket in-case the kit wasn't updated yet.
> > So typically just add the SVG, and reference it in the YAML. It'll be visible after merge.
> 
> but, both of these are just using the existing `splunk.svg` that is already uploaded, shouldn't that work?

Hmm might be a bug then, would have to check the UI Kit to be sure.

damemi:
Hi @BenElferink just wanted to check if you had any time to take a look at the svg issue, thanks!

BenElferink:
> Hi @BenElferink just wanted to check if you had any time to take a look at the svg issue, thanks!

I didn't find any bugs.
It's just the destination types (`splunkotlp` `splunksapm`), they are not recognized by the UI Kit for a local SVG import, and the types are not aligned with the name of the cloud SVG file, so it results with no image:

```ts
const logo = LOGOS[type];
if (logo && !forceFromUrl) return { icon: logo };

// Fallback to fetched-icon (in-case new destinations were added, and we haven't added the icon to the kit yet)
return { iconSrc: `https://d15jtxgb40qetw.cloudfront.net/${type}.svg` };
```

---

I'd just have to add the relevant type to the list of enums.
Are you planning on keeping the new names? Or are you reverting one of them to `splunk`?

Note: keeping both new names will require a migration script for existing setups!

---

Edit: UI-Kit PR is open:
https://github.com/odigos-io/ui-kit/pull/161

BenElferink:
@damemi almost there, just need to merge in OSS: https://github.com/odigos-io/odigos/pull/2982

---

Edit: merged, just pull and you should see the icons

damemi:
@BenElferink oh thanks! so when adding a new destination type, they also need to be added to the ui-kit for local svg support?

Also, I see you already merged it but as you pointed out I don't want to have to migrate the existing splunk destination (now `splunksapm` in this PR). So, I will change `splunksapm` back to just `splunk`

BenElferink:
> @BenElferink oh thanks! so when adding a new destination type, they also need to be added to the ui-kit for local svg support?
> 
> Also, I see you already merged it but as you pointed out I don't want to have to migrate the existing splunk destination (now `splunksapm` in this PR). So, I will change `splunksapm` back to just `splunk`

For new destination types, yes they need to be handled in the UI Kit for local SVG imports (offline). Otherwise, as long as the type matches the name of the SVG file, we should fallback to an SVG pulled from S3 (e.g. destination type == `jaeger` and SVG name == `jaeger.svg`).

No worries, the Kit supports all 3 types now, so feel free to use any of: `splunk`, `splunksapm`, or `splunksapm`.

damemi:
Thanks @BenElferink, updated ptal!

PR #2846, opened 2025-05-09
go-offsets: Fix Helm chart upgrade and use correct file name

damemi:
## Description

Depends on https://github.com/odigos-io/odigos-enterprise/pull/918

This fixes an error where the Helm chart template would properly set an empty `offsets` configmap value on install, but on `helm upgrade` the existing configmap would cause the template to render nothing, leading to this error:

```
{"level":"error","ts":1746817250.7225494,"caller":"enterprise/main.go:65","msg":"failed to set up ebpf factories","error":"stat /offsets/offset_results.json: no such file or directory","stacktrace":"main.main\n\t/go/src/github.com/odigos-io/odigos-enterprise/odiglet/cmd/enterprise/main.go:65\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:272"}
```

This makes sure the value is always present, set or not. Doing so requires the JSON offsets value to be escaped (if it exists). https://github.com/odigos-io/odigos-enterprise/pull/918 updates the enterprise odiglet to parse that encoded JSON value

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


RonFed:
we should add a `helm-upgrade` test similar to the one we have for cli, I guess that would have caught this

RonFed (approved):

BenElferink (approved):

PR #2847, opened 2025-05-10
feat: allow hexencoded url template path with captical letters

blumamir:
This PR address minor improvments to url templatization following user feedbacks:

- support hex encoded strings with capital letters.
- fix configuration example in the README
- minor text improvments in README
- add warning at beginning of README about high cardinality 

Add unit test

### DIFF
--- a/collector/processors/odigosurltemplateprocessor/templatize.go
+++ b/collector/processors/odigosurltemplateprocessor/templatize.go
@@ -30,7 +30,7 @@ var (
 	// - An even number of characters
 	//
 	// It is considered safe as:
-	// - letters are only limited to lowercase a-f, which any real word with 16 chars or more will fail.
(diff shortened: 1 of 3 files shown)


collector/processors/odigosurltemplateprocessor/templatize.go
```diff
@@ -36,11 +36,11 @@ var (
 	//
 	// Explanation (ChatGPT):
 	// - (?:...) — A non-capturing group.
-	// - [0-9a-f]{2} — Matches exactly two hexadecimal characters.
+	// - [0-9a-fA-F]{2} — Matches exactly two hexadecimal characters.
```
RonFed, line 39: The comment above mentions only lower case is supported, so it should be updated.

RonFed (approved):

PR #2848, opened 2025-05-10
chore: update to go1.24

RonFed:


damemi (commented):
@RonFed do you think we should wait until opentelemetry updates to 1.24? https://github.com/open-telemetry/opentelemetry-collector/blob/4ca0f1829e0a348738caab8799237d395a975888/go.mod#L11

RonFed:
@damemi hmm that's a good point. It does not seem to break anything upgrading to 1.24 and the collector packages on 1.23. 
I saw that the k8s packages bumps require 1.24 hence I created this PR.

damemi:
@RonFed ah good point, I didn't know k8s was up to 1.24 already. We should probably go with that then, I don't think it will hurt the collector packages

BenElferink (approved):

PR #2849, opened 2025-05-11
Karpenter use startup taints

tamirdavid1:
## Description

This PR addresses a customer issue when installing Karpenter alongside Odigos.

Currently, when odiglet is scheduled on a node, it adds a label indicating that it's installed. Separately, the instrumentor webhook sets a node affinity on instrumented pods to ensure they are only scheduled on nodes where odiglet has already been deployed.

The problem arises because the label added by odiglet is applied dynamically after scheduling, and Karpenter is unaware of it — it's not part of the NodePool's provisioning template. As a result, Karpenter does not provision a new node, since it sees no way to satisfy the pod's node affinity.

To solve this, Karpenter provides a feature called startup taints. These taints are applied when a node is provisioned and are meant to be removed by services like odiglet once the node is fully initialized. Since Karpenter ignores these taints when scheduling pods, it can still provision the node even though no pods match it yet. Once odiglet finishes preparing the node, it removes the startupTaint, allowing instrumented pods to be scheduled.

I've documented the full setup process for enabling this integration. Please review it carefully — the flow can be a bit unintuitive, so your feedback or suggestions for clarification would be valuable.
## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [X] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

- [X] Users need to take action before upgrading


TheErikNaz (approved):

k8sutils/pkg/node/node.go
```diff
@@ -33,3 +38,57 @@ func DetermineNodeOdigletInstalledLabelByTier() string {
...
+	const (
+		startupTaintKey    = consts.KarpenterStartupTaintKey
+		startupTaintEffect = v1.TaintEffectNoSchedule
+		maxRetries         = 5
+		retryInterval      = 1 * time.Second
+	)
+
+	ctx := context.Background()
+
+	for attempt := 1; attempt <= maxRetries; attempt++ {
```
RonFed, line 52: nit: `client-go` has the builtin retry functions - that we use for the instrumentation instance updates.
maybe better to also use it here

cli/cmd/resources/odiglet.go
```diff
@@ -123,7 +129,7 @@ func NewOdigletClusterRole(psp, ownerPermissionEnforcement bool) *rbacv1.Cluster
 			{ // Needed for virtual device registration
 				APIGroups: []string{""},
 				Resources: []string{"nodes"},
-				Verbs:     []string{"get", "list", "watch", "patch"},
+				Verbs:     []string{"get", "list", "watch", "patch", "update"},
```
RonFed, line 132: please update the comment above with the reason

helm/odigos/templates/odigos-config-cm.yaml
```diff
@@ -135,4 +135,7 @@ data:
     {{- if .Values.nodeSelector }}
     nodeSelector:
       {{- toYaml .Values.nodeSelector | nindent 6 }}
+    {{- if .Values.karpenterEnabled }}
+    karpenterEnabled: {{ .Values.karpenterEnabled }}
+    {{- end }}
```
RonFed, line 140: from the `values.yaml` it looks like it should be
`.Values.karpenter.enabled`
or we should change the values.yaml
I prefer the `karpenter.enabled` one

tamirdavid1: Thanks for the comment

odiglet/odiglet.go
```diff
@@ -189,13 +191,37 @@ func OdigletInitPhase(clientset *kubernetes.Clientset) {
...
 	}
 
-	odigletInstalledLabel := k8snode.DetermineNodeOdigletInstalledLabelByTier()
+	ctx := context.Background()
 
-	log.Logger.V(0).Info("Adding Label to Node", "odigletLabel", odigletInstalledLabel)
+	// Get Odigos configuration to check if Karpenter is enabled
+	odigosConfig, err := k8sutils.GetCurrentOdigosConfigWithClientset(ctx, clientset)
+	if err != nil {
+		if errors.Is(err, k8sutils.ErrOdigosEffectiveConfigNotFound) {
```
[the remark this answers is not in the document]
tamirdavid1: Added retry mechism

RonFed (approved):

odiglet/odiglet.go
```diff
@@ -189,13 +192,45 @@ func OdigletInitPhase(clientset *kubernetes.Clientset) {
...
+	if odigosConfig.KarpenterEnabled != nil && *odigosConfig.KarpenterEnabled {
+		log.Logger.Info("KarpenterEnabled is true, attempting to remove startup taint")
+		if err := k8snode.RemoveStartupTaint(clientset, nn); err != nil {
+			log.Logger.Error(err, "Failed to remove startup taint from node")
+			os.Exit(-1)
+		} else {
+			log.Logger.Info("Successfully removed startup taint from node")
+		}
+	} else {
+		// Karpenter is not enabled, add the odiglet installed label to the node <Default behavior>
```
blumamir, line 226: can we also add this label with Karpetner?
will it do any harm, givin that the webhook will not set any taints for it?

helm/odigos/values.yaml
```diff
@@ -249,3 +249,10 @@ avoidInjectingJavaOptsEnvVar: null
 # Optional NodeSelector to apply to all Odigos components
 # Note: Odigos will only be able to instrument workloads on the same nodes.
 nodeSelector: {}
+
+
+# Karpenter settings, before changing to true please modify the karpenter-node-template.yaml file
+# to add the odigos-agent startupTaints configuration.
+# https://karpenter.sh/docs/concepts/nodepools/
```
blumamir, line 256: consider giving the link to the section you've added to docs with specific odigos instructions

instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -86,8 +86,10 @@ func (p *PodsWebhook) Default(ctx context.Context, obj runtime.Object) error {
 		return nil
 	}
 
-	// Add odiglet installed node-affinity to the pod
-	podutils.AddOdigletInstalledAffinity(pod)
+	// Add odiglet installed node-affinity to the pod, for non Karpenter installations
+	if odigosConfig.KarpenterEnabled == nil || !*odigosConfig.KarpenterEnabled {
+		podutils.AddOdigletInstalledAffinity(pod)
+	}
```
blumamir, line 92: else - we should add affinity to never be scheduled on nodes with `odigos.io/needs-init` so that we don't attempt to start an instrumented pod on a node that has no `/mnt/var/log` setup

odiglet/odiglet.go
```diff
@@ -189,13 +192,45 @@ func OdigletInitPhase(clientset *kubernetes.Clientset) {
...
+		} else {
+			log.Logger.Error(err, "Failed to fetch Odigos configuration")
+			os.Exit(-1)
+		}
+	}
+
+	// If Karpenter is enabled, remove the startup taint from the node
+	if odigosConfig.KarpenterEnabled != nil && *odigosConfig.KarpenterEnabled {
+		log.Logger.Info("KarpenterEnabled is true, attempting to remove startup taint")
+		if err := k8snode.RemoveStartupTaint(clientset, nn); err != nil {
```
blumamir, line 219: consider always attempting to remove the taint if found

blumamir (approved):

instrumentor/controllers/agentenabled/pods_webhook.go
tamirdavid1: As discussed, this is happen because of the taint added to the node by Karpenter with the effect of NoSchedule

odiglet/odiglet.go
tamirdavid1: done, really simplify the code! Thanks

helm/odigos/values.yaml
tamirdavid1: added

PR #2858, opened 2025-05-12
Migrate to docker bake

TheErikNaz:
## Description
Migrate our build system to use docker bake
Cleanup and optimize Makefile

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -350,7 +350,6 @@ helm-install-central:
 		--namespace odigos-central \
 		--set image.tag=$(ODIGOS_CLI_VERSION) \
 		--set onPremToken=$(ONPREM_TOKEN) \
-	kubectl label namespace odigos-central odigos.io/central-system-object="true" --overwrite
 
 helm-uninstall:
 	@echo "Uninstalling odigos using helm"
--- a/Makefile
+++ b/Makefile
@@ -38,20 +38,21 @@ print-tag:
 	@echo $(TAG)
 define bake-load
 	@TAG=$(TAG) ORG=$(ORG) IMG_SUFFIX=$(IMG_SUFFIX) \
-	docker buildx bake $(1) --pull --load \
+	docker buildx bake $(1) --load \
 	$(if $(PLATFORMS),--set '*.platform=$(PLATFORMS)',)
 endef
 
 define bake-push
 	@TAG=$(TAG) ORG=$(ORG) IMG_SUFFIX=$(IMG_SUFFIX) \
-	docker buildx bake $(1) --pull --push \
+	docker buildx bake $(1) --push \
 	$(if $(PLATFORMS),--set '*.platform=$(PLATFORMS)',)
 endef
 
 # Pattern rules for every service image
-.PHONY: build-% push-%
-build-%: ; $(call bake-load,$*)
-push-%:  ; $(call bake-push,$*)
+build-%: FORCE ; $(call bake-load,$*)
+push-%: FORCE ; $(call bake-push,$*)
+
+FORCE:
 
 # Convenience groups matching the HCL
 .PHONY: build-images push-images build-images-rhel push-images-rhel build-cli build-cli-rhel
@@ -75,11 +76,11 @@ push-cli-rhel: push-cli
 # ──────────────────────────────────────────────
 # Kind helpers
 # ──────────────────────────────────────────────
-.PHONY: load-to-kind-% load-to-kind
-load-to-kind-%:
+.PHONY:
+load-to-kind-%: FORCE
 	kind load docker-image $(ORG)/odigos-$*$(IMG_SUFFIX):$(TAG)
 
-load-to-kind:
+load-to-kind: FORCE
 	$(MAKE) -j $(nproc) load-to-kind-instrumentor load-to-kind-autoscaler \
 		load-to-kind-scheduler load-to-kind-odiglet \
 		load-to-kind-collector load-to-kind-ui load-to-kind-cli \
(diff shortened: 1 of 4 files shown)


RonFed:
still need to review,
it will be awsome to also have the cli image in this infra
see https://github.com/odigos-io/odigos/blob/98f8ab4f2445827e4e23fe04e298bf81a59b18cd/.github/workflows/release-artifactregistry.yml#L119

Makefile
```diff
@@ -122,131 +148,150 @@ verify-nodejs-agent:
...
-push-scheduler:
-	$(MAKE) push-image/scheduler SUMMARY="Scheduler for Odigos" DESCRIPTION="Scheduler manages the installation of OpenTelemetry Collectors with Odigos." TAG=$(TAG) ORG=$(ORG) IMG_SUFFIX=$(IMG_SUFFIX)
 
-.PHONY: push-collector
-push-collector:
-	$(MAKE) push-image/collector DOCKERFILE=collector/$(DOCKERFILE) BUILD_DIR=collector SUMMARY="Odigos Collector" DESCRIPTION="The Odigos build of the OpenTelemetry Collector." TAG=$(TAG) ORG=$(ORG) IMG_SUFFIX=$(IMG_SUFFIX)
+# ──────────────────────────────────────────────
+# Deploy helpers
+# ──────────────────────────────────────────────
+.PHONY: deploy-% deploy
```
RonFed, line 167: I think PHONY does not know to process patterns like that

Makefile
```diff
@@ -122,131 +148,150 @@ verify-nodejs-agent:
...
-deploy:
-	make deploy-odiglet && make deploy-autoscaler && make deploy-collector && make deploy-instrumentor && make deploy-scheduler && make deploy-ui
+# ──────────────────────────────────────────────
+# Helm install helper
+# ──────────────────────────────────────────────
+.PHONY: helm-install
+helm-install:
+	@echo "Installing odigos using helm"
+	helm upgrade --install odigos ./helm/odigos \
+		--create-namespace --namespace odigos-system \
```
RonFed, line 238: the namespace can also be an env var with the default value of `odigos-system`

Makefile
```diff
@@ -122,131 +148,150 @@ verify-nodejs-agent:
...
+		--set image.tag=$(ODIGOS_CLI_VERSION) \
+		--set clusterName=$(CLUSTER_NAME) \
+		--set centralProxy.centralBackendURL=$(CENTRAL_BACKEND_URL) \
+		--set onPremToken=$(ONPREM_TOKEN) \
```
RonFed, line 242: will this work well if the values are empty?

docker-bake.hcl
```diff
@@ -0,0 +1,145 @@
...
+    DESCRIPTION  = "Kubernetes Operator for Odigos installs Odigos"
+  }
+  tags = ["${ORG}/odigos-operator${IMG_SUFFIX}:${TAG}"]
+}
+
+target "odiglet" {
+  inherits   = ["common"]
+  context    = "."
+  dockerfile = "odiglet/Dockerfile"
+  args = {
```
RonFed, line 41: we also have `LD_FLAGS` as an argument to all the images

Makefile
```diff
@@ -1,20 +1,74 @@
...
+IMG_SUFFIX ?=
+BUILD_DIR  ?= .
+
+
+# ──────────────────────────────────────────────
+# Build / push helpers
+# ──────────────────────────────────────────────
+define bake-load
+	@TAG=$(TAG) ORG=$(ORG) IMG_SUFFIX=$(IMG_SUFFIX) \
+	docker buildx bake $(1) --load --pull \
```
RonFed, line 32: From the docs pull flag indicates `Always pull image` - which I think we don't want

Makefile
```diff
@@ -1,20 +1,74 @@
...
+# Core variables
+# ──────────────────────────────────────────────
 TAG ?= $(shell odigos version --cluster)
 ODIGOS_CLI_VERSION ?= $(shell odigos version --cli)
 CLUSTER_NAME ?= local-dev-cluster
-CENTRAL_BACKEND_URL ?= 
+CENTRAL_BACKEND_URL ?=
 ORG ?= registry.odigos.io
+PARALLEL ?= $(shell nproc)
+PLATFORMS ?=
```
RonFed, line 10: what will be the default value here?
ideally it should only build for arm64 if running on arm64

Makefile
TheErikNaz: Yeah, everything has a default value

Makefile
TheErikNaz: Good idea, Adding it.

cli/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+COPY api/        ./api/
+COPY autoscaler/ ./autoscaler/
+COPY common/     ./common/
+COPY k8sutils/   ./k8sutils/
+COPY profiles/   ./profiles/
+COPY cli/        ./cli/
+
+WORKDIR /src/cli
+RUN CGO_ENABLED=0 GOOS=linux \
+    go build -tags=embed_manifests -o odigos .
```
RonFed, line 31: we should add support for LD_FLAGS similar to other containers we build.
this will allow us to strip the binary and redyce the image size

TheErikNaz: I've planned to do a second PR where I merge the rhel and regular images, Let's do it there

RonFed (approved):
🥇

PR #2864, opened 2025-05-14
fix: handle empty path in url templating

blumamir:
## Description

When the url path is empty string like in "http://example.com", the span names ends up like "GET " (extra space at the end), which:
- creates 2 different values in UIs (both "GET" and "GET " with space), which is confusing to users.
- non intuitive what it actually means.

This PR fixes the behavior so that empty values in http span names are always set to "/" to denote root path.

## How Has This Been Tested?

- [x] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

No

## User Facing Changes

Fix bug in processor which user will notice


collector/processors/odigosurltemplateprocessor/processor_test.go
```diff
@@ -339,6 +339,61 @@ func TestProcessor_Traces(t *testing.T) {
...
+		},
+		{
+			name:          "client span with empty template",
+			spanKind:      ptrace.SpanKindClient,
+			inputSpanName: "GET",
+			inputSpanAttrs: map[string]any{
+				"http.request.method": "GET",
+				"url.template":        "",
+			},
+			expectedSpanName: "GET /",
```
RonFed, line 386: maybe worth to add test fro some unknown high cardinality path - and assert we get the method name as span name

RonFed (approved):

collector/processors/odigosurltemplateprocessor/processor_test.go
blumamir: added test for invalid url 👍

PR #2865, opened 2025-05-14
cli(upgrade): support OSS → Enterprise migration via --onprem-token 

AvihuHenya:
## Description
This PR adds support for migrating an existing Odigos OSS installation to an Enterprise tier using the `activate` command.

### Key Change
  - If `--onprem-token` is passed during `odigos pro activate`, the CLI triggers a migration to the Enterprise (on-prem) tier.
  - The `OnPremOdigosTier` is applied and relevant enterprise resources are deployed.
  - This enables seamless transition for users moving from OSS to a managed on-prem experience without uninstalling/reinstalling.

Migrate an OSS installation to enterprise tier without version change: `odigos pro activate --onprem-token <token>`

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [X] Manual Testing
- [ ] Manual Load Test


cli/cmd/upgrade.go
```diff
@@ -93,42 +109,37 @@ and apply any required migrations and adaptations.`,
...
-		}
+		onPremTokenSet := cmd.Flag("onprem-token").Changed
+		var currentTier common.OdigosTier
 
-		currentTier, err := odigospro.GetCurrentOdigosTier(ctx, client, ns)
-		if err != nil {
-			fmt.Println("Odigos cloud login failed - unable to read the current Odigos tier.")
-			os.Exit(1)
+		if onPremTokenSet {
+			currentTier = common.OnPremOdigosTier
```
BenElferink, line 119: Shouldn't there be some sort of token validation, and save it as k8s secret?

AvihuHenya: Currently, when using `--onprem-token` the token is stored in a secret, and validation occurs when the odiglet starts. I do think it's an important feature, but it's not within the scope of this task.

BenElferink (approved):

cli/cmd/upgrade.go
```diff
@@ -93,42 +109,37 @@ and apply any required migrations and adaptations.`,
...
+			}
 		}
 
 		managerOpts := resourcemanager.ManagerOpts{
 			ImageReferences: GetImageReferences(currentTier, openshiftEnabled),
 		}
 
-		resourceManagers := resources.CreateResourceManagers(client, ns, currentTier, nil, config, versionFlag, installationmethod.K8sInstallationMethodOdigosCli, managerOpts)
+		resourceManagers := resources.CreateResourceManagers(
+			client, ns, currentTier, &odigosOnPremToken, config, versionFlag,
```
blumamir, line 133: before this PR we would only call `CreateResourceManagers` with `onPremToken` equals `nil` to indicate we don't want to modify it.

now we will always set the 4th parameter to `&odigosOnPremToken` to possible empty string. this will be changed for `community` tiers as well, and also for `odigos update` without a new token.

Please follow the code to make sure this change will not affect any other command. specifically, I am worried about [this section](https://github.com/odigos-io/odigos/blob/fd56e0f070c1e29d63b8c3667028646fe1752208/cli/cmd/resources/odigospro/manager.go#L35) which checks for nil but not for empty strings

cli/cmd/upgrade.go
```diff
@@ -93,42 +109,37 @@ and apply any required migrations and adaptations.`,
...
-		currentTier, err := odigospro.GetCurrentOdigosTier(ctx, client, ns)
-		if err != nil {
-			fmt.Println("Odigos cloud login failed - unable to read the current Odigos tier.")
-			os.Exit(1)
+		if onPremTokenSet {
+			currentTier = common.OnPremOdigosTier
+		} else {
+			currentTier, err = odigospro.GetCurrentOdigosTier(ctx, client, ns)
+			if err != nil {
+				fmt.Println("Odigos cloud login failed - unable to read the current Odigos tier.")
```
blumamir, line 123: I think it's an existing copy-paste mistake 😅.

this is not `odigos cloud login` command so we can update the text while at it :)

blumamir (commented):
I think the main part that we were not sure about are the agent files that odiglet copies in it's init phase.

Recommend to research, test, and document the following:

- How do odiglet node labels behaves when we update from OSS odiglet to enterprise?
- karpetner uses a different label system, which has no separation of OSS and enterprise. I suspect this change can cause user apps pods to crash if they are scheduled on a node which has not completed odiglet enterprise setup and the enterprise files are not found in `/var/odigos` mount.
- there are certain eBPF files with special delicate handling on when they should be modified and when not. please talk with @tamirdavid1 or @RonFed to make sure this logic is expected to work on upgrades.
- if we have an option to migrate from OSS -> enterprise, should we also add the reverse way? e.g. enterprise -> OSS downgrade

tamirdavid1:
> I think the main part that we were not sure about are the agent files that odiglet copies in it's init phase.
> 
> Recommend to research, test, and document the following:
> 
> * How do odiglet node labels behaves when we update from OSS odiglet to enterprise?
> * karpetner uses a different label system, which has no separation of OSS and enterprise. I suspect this change can cause user apps pods to crash if they are scheduled on a node which has not completed odiglet enterprise setup and the enterprise files are not found in `/var/odigos` mount.
> * there are certain eBPF files with special delicate handling on when they should be modified and when not. please talk with @tamirdavid1 or @RonFed to make sure this logic is expected to work on upgrades.
> * if we have an option to migrate from OSS -> enterprise, should we also add the reverse way? e.g. enterprise -> OSS downgrade

i'll give me two cents on this.
1. **About karpenter** as describes [here](https://docs.odigos.io/setup/odigos-with-karpenter) and as Amir mentioned Karepnter installations are using different labling system. we ask the user to configure in Karpenter startup taints.
And then during the pod reconcile we're checking if we're in the case of karpenter installation:
https://github.com/odigos-io/odigos/blob/main/instrumentor/controllers/agentenabled/pods_webhook.go#L92
This introduce potential risk for users that uses karpenter and upgrade from oss to enterprise.
I think that we can for first step to not allow upgrading from oss to enterprise in case this config is set.
2. **About ebpf file** - as i see it, in case karpenter not enabled there should'nt be any problem with the ebpf and enterprise equivalent files because each workload that will be instrumented post this change will be set with the enterprise node affinity and the new odiglets (enterprise ones) will deploy and also it on the node label.

@RonFed would like to hear also your opinion here

cli/cmd/pro.go
```diff
@@ -232,6 +233,74 @@ var centralInstallCmd = &cobra.Command{
...
+	Short: "Migrate Odigos from Community to Enterprise tier",
+	Run: func(cmd *cobra.Command, args []string) {
+		ctx := cmd.Context()
+		client := cmdcontext.KubeClientFromContextOrExit(ctx)
+
+		onPremToken := cmd.Flag("onprem-token").Value.String()
+		if onPremToken == "" {
+			fmt.Println("\033[31mERROR\033[0m onprem-token is required")
+			os.Exit(1)
+		}
```
blumamir: since we set it with `MarkFlagRequired`, can it be empty?

cli/cmd/pro.go
```diff
@@ -232,6 +233,74 @@ var centralInstallCmd = &cobra.Command{
...
+		} else if err != nil {
+			fmt.Printf("\033[31mERROR\033[0m Failed to check if Odigos is already installed: %s\n", err)
+			os.Exit(1)
+		}
+
+		fmt.Println("Starting migration from Community to Enterprise tier...")
+
+		odigosConfig, err := resources.GetCurrentConfig(ctx, client, ns)
+		if err != nil {
+			fmt.Printf("Error getting config maps: %v\n", err)
```
blumamir:
```suggestion
			fmt.Printf("Error reading odigos configuration: %v\n", err)
```

cli/cmd/pro.go
```diff
@@ -232,6 +233,74 @@ var centralInstallCmd = &cobra.Command{
...
+			return
+		}
+		if odigosConfig.KarpenterEnabled != nil && *odigosConfig.KarpenterEnabled {
+			fmt.Println("\033[31mERROR\033[0m Migration is not supported when Karpenter is installed.")
```
blumamir: nit: add comment why this is not supported for future us looking at this code :)

blumamir:
```suggestion
			fmt.Println("\033[31mERROR\033[0m Migration is not supported when odigos is installed with 'KarpenterEnabled' option. uninstall odigos community and reinstall odigos with enterprise onprem token")
```

cli/cmd/pro.go
```diff
@@ -232,6 +233,74 @@ var centralInstallCmd = &cobra.Command{
...
+			os.Exit(1)
+		}
+
+		managerOpts := resourcemanager.ManagerOpts{
+			ImageReferences: GetImageReferences(common.OnPremOdigosTier, openshiftEnabled),
+		}
+		config, err := resources.GetCurrentConfig(ctx, client, ns)
```
blumamir: we already have the config at this point. no need to re-fetch it

cli/cmd/pro.go
```diff
@@ -232,6 +233,74 @@ var centralInstallCmd = &cobra.Command{
...
+		if err != nil {
+			fmt.Println("Odigos pro migrate failed - unable to read current config.")
+			os.Exit(1)
+		}
+
+		cm, err := client.CoreV1().ConfigMaps(ns).Get(ctx, k8sconsts.OdigosDeploymentConfigMapName, metav1.GetOptions{})
+		if err != nil {
+			fmt.Println("Odigos pro migrate failed - unable to get version ConfigMap.")
```
blumamir:
```suggestion
			fmt.Println("Odigos pro migrate failed - unable to get odigos deployment ConfigMap.")
```

cli/cmd/pro.go
```diff
@@ -232,6 +233,74 @@ var centralInstallCmd = &cobra.Command{
...
+			fmt.Printf("\033[31mERROR\033[0m Failed to check if Odigos is already installed: %s\n", err)
+			os.Exit(1)
+		}
+
+		fmt.Println("Starting migration from Community to Enterprise tier...")
+
+		odigosConfig, err := resources.GetCurrentConfig(ctx, client, ns)
+		if err != nil {
+			fmt.Printf("Error getting config maps: %v\n", err)
+			return
```
blumamir: nit: for consistency with other checks here

```suggestion
			os.Exit(1)
```

cli/cmd/pro.go
```diff
@@ -232,6 +233,74 @@ var centralInstallCmd = &cobra.Command{
...
+			client, ns, common.OnPremOdigosTier, &onPremToken, config, odigosVersion,
+			installationmethod.K8sInstallationMethodOdigosCli, managerOpts)
+
+		err = resources.ApplyResourceManagers(ctx, client, resourceManagers, "Synching")
+		if err != nil {
+			fmt.Println("Odigos pro migrate failed - unable to apply resources.")
+			os.Exit(1)
+		}
+
+		fmt.Println("Migration completed successfully.")
```
blumamir:
```suggestion
		fmt.Println("Migration completed successfully. Odigos is upgraded to enterprise tier")
```

cli/cmd/pro.go
```diff
@@ -232,6 +233,74 @@ var centralInstallCmd = &cobra.Command{
 	},
 }
 
+var migrateCmd = &cobra.Command{
+	Use:   "migrate",
```
blumamir: how about:

```suggestion
	Use:   "activate",
```

I think "migrate" can be a bit less clear in this context.

cli/cmd/pro.go
AvihuHenya: great catch. sure
needs to be updated for `central install` command also

cli/cmd/pro.go
AvihuHenya: get it. you thing that it should also populate to the logs got print?
like `fmt.Println("Odigos pro migrate failed - unable to apply resources.")` needs to become `fmt.Println("Odigos pro activate failed - unable to apply resources.")` etc?

cli/cmd/pro.go
```diff
@@ -232,6 +228,69 @@ var centralInstallCmd = &cobra.Command{
 	},
 }
 
+// Replaced 'migrate' command with 'activate' to reflect activation of Enterprise tier
```
blumamir: this comment should not be committed to the code :)

blumamir (approved):

PR #2867, opened 2025-05-14
cli: detect Odigos namespace via labeled ConfigMap instead of namespace label

AvihuHenya:
## Description
Change how we detect where Odigos is installed.

This resolves the issue where users must manually run: `kubectl label namespace odigos-system odigos.io/system-object="true"` after installing Odigos via Helm.

## Short summary of the changes 
Instead of filtering all namespaces for the odigosSystem label, we now look for a ConfigMap with the expected label and extract its namespace. This avoids the need for extra steps post-installation.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [X] Manual Testing
- [ ] Manual Load Test


damemi (approved):
👍

cli/cmd/resources/namespace.go
```diff
@@ -23,23 +22,19 @@ func NewNamespace(name string) *v1.Namespace {
 }
 
 func GetOdigosNamespace(client *kube.Client, ctx context.Context) (string, error) {
-	namespaces, err := client.CoreV1().Namespaces().List(ctx, metav1.ListOptions{
+	configMaps, err := client.CoreV1().ConfigMaps("").List(ctx, metav1.ListOptions{
```
blumamir, line 25: For the migrartion period, let's keep them both.
first - try to search for namespace with label, and if not found, try with configmap.

If we make this change at once, new cli will have problem working with odigos version that has not yet migrated to the new lables.

helm/README.md
```diff
@@ -30,7 +30,6 @@ helm repo add odigos https://odigos-io.github.io/odigos
 ```sh
 helm repo update
 helm upgrade --install odigos odigos/odigos --namespace odigos-system --create-namespace
-kubectl label namespace odigos-system odigos.io/system-object="true"
```
blumamir, line 33: need to also update [the docs](https://github.com/odigos-io/odigos/blob/main/docs/setup/installation.mdx#install-odigos-1)

and the [Makefile](https://github.com/odigos-io/odigos/blob/15a2d6fae279b1967f30dee081db76446e801ad4/Makefile#L347)

helm/odigos/templates/odigos-deployment.yaml
```diff
@@ -3,6 +3,9 @@ kind: ConfigMap
 metadata:
   name: odigos-deployment
   namespace: '{{ .Release.Namespace }}'
+  labels:
+    odigos.io/config: '1'
```
blumamir, line 7: let's avoid adding this label from helm (and possibly remove it from other places where we use it)

blumamir (changes requested):
Blocking since I think we should not remove the namespace checking as it will create compatibility issues when cli is updated but odigos version in cluster is still old.

Please test and make sure it works:

- install odigos with helm in current version
- access `odigos ui` and other cli commands with cli after this change.

We should add a comment to remove the namespace code sometime in the far future and explain why it's still needed

helm/odigos/templates/odigos-deployment.yaml
AvihuHenya: God point, I’ve updated it in all three Helm ConfigMap templates.

BenElferink (approved):
Looks good to me

cli/cmd/resources/namespace.go
blumamir: Cool, `odigos-deployment` is added in this PR, but `odigos-config` has the label and should work 👍

cli/cmd/resources/namespace.go
```diff
@@ -23,23 +24,19 @@ func NewNamespace(name string) *v1.Namespace {
...
-
-	if len(namespaces.Items) == 0 {
+	if len(configMap.Items) == 0 {
 		return "", errNoOdigosNamespaceFound
-	} else if len(namespaces.Items) != 1 {
-		return "", fmt.Errorf("expected to get 1 namespace got %d", len(namespaces.Items))
 	}
-
-	return namespaces.Items[0].Name, nil
+	return configMap.Items[0].Namespace, nil
```
blumamir, line 39: here we take an arbitrary namespace from the first configmap we found. previous code would check if the results indicates more than one namespace with the label and failed in this case.
consider keeping this check to make the cli more robust

blumamir (approved):
LGTM - added a nit about the removal of multiple matching namespace check which can be nice to keep (we saw this once in a user environment)

cli/cmd/resources/namespace.go
RonFed: +1

cli/cmd/resources/namespace.go
```diff
@@ -23,23 +24,23 @@ func NewNamespace(name string) *v1.Namespace {
...
 		return "", err
 	}
 
-	if len(namespaces.Items) == 0 {
+	if len(configMap.Items) == 0 {
 		return "", errNoOdigosNamespaceFound
-	} else if len(namespaces.Items) != 1 {
-		return "", fmt.Errorf("expected to get 1 namespace got %d", len(namespaces.Items))
+	} else if len(configMap.Items) != 1 {
+		return "", fmt.Errorf("expected to get 1 namespace got %d", len(configMap.Items))
```
RonFed, line 40: need to make sure here that they are from different namespaces?

AvihuHenya: Since we query with a fieldSelector, doesn't that ensure they're from different namespaces already (because we cannot have two ConfigMaps with the same name in the same namespace)

PR #2870, opened 2025-05-15
feat: support multiple manifest from one profile and migrate kratos specific ones

blumamir:
## Description

This PR makes the following changes:

- it allows any profile to list the name of the manifest files it wants to apply
- moves kratos specific profiles, from being in the profiles list, to simply be a manifest file for kratos.
- add k8sattributes action for kratos to collect replicaset name.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

Added profile, removed, checked all resources are created as needed

## Kubernetes Checklist

- [x] Changes how Odigos interacts with Kubernetes - schedualer now interacts with k8sattributesresolver resource
- [x] Changes RBAC permissions - add permissions for managing the new manifest for the profile

## User Facing Changes

No. only kratos users will see replicaset being collected automatically


profiles/allprofiles.go
```diff
@@ -15,15 +15,12 @@ var AllProfiles = []profile.Profile{
 	aggregators.KratosProfile,
 	aggregators.GreatwallProfile,
 
-	attributes.CategoryAttributesProfile,
 	attributes.CodeAttributesProfile,
 	attributes.CopyScopeProfile,
```
RonFed, line 19: we can do the same for the copy scope profile?

cli/cmd/resources/scheduler.go
```diff
@@ -97,6 +97,11 @@ func NewSchedulerRole(ns string) *rbacv1.Role {
 				Resources: []string{"processors", "instrumentationrules"},
 				Verbs:     []string{"get", "list", "watch", "patch", "delete", "create"},
 			},
+			{ // apply profiles as actions
+				APIGroups: []string{"actions.odigos.io"},
+				Resources: []string{"k8sattributesresolvers"},
+				Verbs:     []string{"get", "list", "watch", "patch", "delete", "create"},
```
RonFed, line 103: why do we need all of these? the scheduler only does `Apply`, right?

RonFed (approved):
Since we're deprecating the individuall actions CRD, I think it os a good chance to make the k8sattributes action into the generic one - as this PR adds permissions and code that handles the deprecated one - it might be better to do in the new way and this would also help us for future profiles regarding other actions (all under the generic one).

Don'r want to block on this, I just think it will be nice to have - to move the migration forward.

PR #2871, opened 2025-05-15
ui: Add Workload Conditions to Source view

damemi:
## Description

Our UI currently only tracks InstrumentationConfig conditions, stopping at `rollout triggered successfully` and showing `Everything Successful` if we reach that point

But in reality, the rollout could have failed on the Workload's end (due to resource quotas, scheduling issues, etc). This is confusing to the user because Odigos says everything should be instrumented and working.

This tries to grab the Workload's conditions and shows them alongside the instrumentation. This will be much more convenient than trying to manually debug the workload with kubectl.

The alternative would be adding Workload conditions to Sources/InstrumentationConfigs in the backend. This would require a controller to watch for updates to every workload in the cluster, which would be very expensive and unnecessary, since the user only cares about these conditions when they're looking at the Source in particular.

For comparison, this is what a currently (failing) instrumentation looks like:

Screenshot 2025-05-15 at 10 11 12 AM

That same app with these changes shows why instrumentation failed (resource quota):

Screenshot 2025-05-15 at 10 01 44 AM

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/frontend/graph/conversions.go
+++ b/frontend/graph/conversions.go
@@ -1,10 +1,14 @@
 package graph
 
 import (
+	"context"
+	"fmt"
 	"time"
 
 	"github.com/odigos-io/odigos/api/odigos/v1alpha1"
 	"github.com/odigos-io/odigos/frontend/graph/model"
+	"github.com/odigos-io/odigos/frontend/kube"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -29,7 +33,7 @@ func k8sLastTransitionTimeToGql(t v1.Time) *string {
 	return &str
 }
 
-func instrumentationConfigToActualSource(instruConfig v1alpha1.InstrumentationConfig) *model.K8sActualSource {
+func instrumentationConfigToActualSource(ctx context.Context, instruConfig v1alpha1.InstrumentationConfig) (*model.K8sActualSource, error) {
 	var containers []*model.SourceContainer
 
 	// Map the containers runtime details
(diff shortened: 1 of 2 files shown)


BenElferink:
Looks like "Replica Failure" shows as status success instead of error.
Other than that, good stuff! 👑

BenElferink (approved):

RonFed:
Very cool. I wonder if it might be too verbose in case everything is fine. 
Also wonder if we should somehow separate the conditions we generate and those k8s does.

damemi:
> Looks like "Replica Failure" shows as status success instead of error. Other than that, good stuff! 👑

Good catch, that's because the Deployment reports it as `type: ReplicaFailure` and `status: True`. We assume any `True` status to be success. Added a check in here for that. That seems to be the only relevant weird condition currently in k8s (see [here](https://github.com/kubernetes/api/blob/82c0029eaa2f5bdbfbbbce9305349a893e6c86dc/apps/v1/types.go#L547) and no other standard conditions for [statefulsets](https://github.com/kubernetes/api/blob/82c0029eaa2f5bdbfbbbce9305349a893e6c86dc/apps/v1/types.go#L318) or [daemonsets](https://github.com/kubernetes/api/blob/82c0029eaa2f5bdbfbbbce9305349a893e6c86dc/apps/v1/types.go#L739))

> Very cool. I wonder if it might be too verbose in case everything is fine. Also wonder if we should somehow separate the conditions we generate and those k8s does.

I actually was going to split them then @BenElferink suggested just doing it in one drop down, and I agree that's probably less confusing to just have the overall status reported in one place. We can probably iterate on the UX to be clearer or filter out irrelevant conditions too

RonFed:
@damemi Can you share an image of how it looks when everything is fine? Will it show all the deployment conditions?

damemi:
@RonFed yeah, sorry. It does show the deployment conditions. I think that's actually nice though because now instead of stopping at "rollout triggered" you see the full rollout success. New rollouts replace the old conditions so it's not like the list constantly grows over time, it's always just relevant to the latest changes

Screenshot 2025-05-15 at 12 02 00 PM

PR #2874, opened 2025-05-16
Allow user workload instrumentation in system namespace + exclude Odigos components

damemi:
## Description

This change is meant to allow users to instrument their own workloads in the same namespace that Odigos is installed. 

Some users, especially in small test environments, may not be able to create a separate namespace for Odigos. Currently, it is possible (but not easy) to do so because we exclude the system namespace by default in the `effective-config`. 

Users can manually edit this configmap to remove the system namespace, and from manual testing everything seems to work from that point. But this is not ideal for those users.

Importantly, we do not want to allow users to instrument Odigos components (to prevent event hotloops). This change updates the UI and Source instrumentation helpers to exclude any workloads with the `odigos-system` label instead of the entire namespace.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/autoscaler/controllers/clustercollector/sync.go
+++ b/autoscaler/controllers/clustercollector/sync.go
@@ -18,6 +18,7 @@ import (
 var (
 	ClusterCollectorGateway = map[string]string{
 		k8sconsts.OdigosCollectorRoleLabel: string(k8sconsts.CollectorsRoleClusterGateway),
+		k8sconsts.OdigosSystemLabelKey:     k8sconsts.OdigosSystemLabelValue,
 	}
 )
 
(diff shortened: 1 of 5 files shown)


RonFed (approved):
lgtm, better to hear @blumamir opinion on this as well.

damemi:
> lgtm, better to hear @blumamir opinion on this as well.

sounds good, no rush on this. happy to wait for more opinions

RonFed:
As discussed, this should be blocked by resolving the uninstall bug when installing in a user provided ns. 
cc @AvihuHenya

PR #2845, opened 2025-05-08
Update otel dependencies to 1.0.125

damemi:
## Description

This updates the OTel dependencies to the latest

In VScode, this is showing an error for the prometheus receiver that I can't quite figure out:

Screenshot 2025-05-08 at 5 06 06 PM

And similar in components.go:

Screenshot 2025-05-08 at 5 07 09 PM

I think this might be a go.work problem or an issue with upstream, looking into it because the collector seems to still build in a docker container

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

- [ ] Changes how Odigos interacts with Kubernetes
- [ ] Introduces additional calls to the API Server (potential performance impact)
- [ ] New Query/feature supported in all the k8s versions supported by Odigos
- [ ] Modifies Odigos manifests (addressed in both CLI and Helm)
- [ ] Changes RBAC permissions

## User Facing Changes

- [ ] Users need to take action before upgrading
- [ ] Automatic migration will modify existing objects (backward compatible)
- [ ] Changes UI, CLI, or K8s Manifests aspects in a way that users need to be aware of
- [ ] Documentation updated accordingly

PR #2875, opened 2025-05-16
Update Otel dependencies to 1.0.126

damemi:
## Description

Replaces https://github.com/odigos-io/odigos/pull/2845

Breaking changes from upstream:

* Remove deprecated use of `service.telemetry.metrics.address` (https://opentelemetry.io/docs/collector/internal-telemetry/#configure-internal-metrics)
* Remove use of deprecated ExporterBatcher (https://github.com/open-telemetry/opentelemetry-collector/pull/12780, example usage https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/39073). Note that the replacement BatcherConfig is also deprecated 🙃 so this will break in the future again
* Update collector metrics names in UI to include `_total` suffixes. I couldn't find the upstream change that did this, but it seems to be some prometheus normalization/compatibility change. I've seen other issues where they are changing how metric names are normalized to match prometheus standards, so I think our collectors are now appending `_total` and `bytes_total` to metrics by default.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


BenElferink:
> I think the metrics failures are due to `service.telemetry.metrics.address` being ignored since v123: https://opentelemetry.io/docs/collector/internal-telemetry/#configure-internal-metrics

Good job, let's gooo 🥳

autoscaler/controllers/clustercollector/configmap.go
```diff
@@ -97,7 +97,18 @@ func addSelfTelemetryPipeline(c *config.Config, ownTelemetryPort int32) error {
...
+			{
+				"pull": config.GenericMap{
+					"exporter": config.GenericMap{
+						"prometheus": config.GenericMap{
+							"host": "0.0.0.0",
+							"port": ownTelemetryPort,
+						},
+					},
+				},
+			},
```
RonFed, line 110: I don't fully understand how that works.
we have the `prometheus/self-metrics` receiver that we configure to scrape the collector metrics server,
so I'm not really sure what this configuration does.

RonFed (approved):

autoscaler/controllers/clustercollector/configmap.go
damemi: This exposes the collector metrics on a prometheus endpoint so it can scrape itself. The current `metrics.address` config field is ignored now. In the node collectors, using that config caused a `failed to scrape prometheus endpoint` error from the self-receiver. I didn't see that error in the cluster collector, but I did notice that it's using a different port, so maybe the metrics are exposed on that port by default.

PR #2876, opened 2025-05-16
feat: use cert-controller for certificates rotation and creation

RonFed:
### tl;dr
Changing the way we manage the certificates for our webhooks - making use of [cert-controller](https://github.com/open-policy-agent/cert-controller) to manage the certificates in the controller instead of doing it in CLI/Helm.

### Issues with our current approach
Today our certificates are managed by CLI/helm, this has the following issues:
* The certificates are refreshed every time an `install`, `upgrade`, `profile`, `config` or `helm upgrade` commands are executed. This is both not necessary and causes the webhooks to have a period of 30-100s of down time - this is a result of the kubelet sync period (https://ahmet.im/blog/kubernetes-secret-volumes-delay/). The implications are that after one of the mentioned commands are executed we'll fail to create sources and instrument newly created pods.
* The certificate rotation mechanism is not deterministic and depends on when the commands are running, in theory a cluster which has odigos installed for the duration of the certificate validity will become un-usable once the certificate is expired. 
* Having this managed by the CLI and helm can cause mismatches when Odigos is installed with Helm and CLI commands are ran - we saw this issue because the CA name in the CLI was `odigos-instrumentor` and in helm `sserving-cert` this caused bad certificate errors.
* We have the `crypto` package in the CLI that handles the certificates creation, this is not ideal since we don't have a mechanism to detect security patches required for it.

### Using `cert-controller`
* The `cert-controller` packages allows to register a controller that watches the certificate secret and the webhooks configurations and aligns the certificate and the `caBundle` on the webhooks. It has a configurable period of 12 hours to check if the certificates need rotation (about to expire) - it can be configured with `LookaheadInterval`.
* Using the new package also for the initial certificate creation allows us to remove this from the CLI and Helm and remove any possible mismatch.
* The package provides a `IsReady` channel which we use to delay the registration of the webhooks. The modified init flow is:
1. Create the manager.
2. Register `cert-controller` with the manager.
3. Start the manager (in a goroutine)
4. Wait for the IsReady signal (in a goroutine) - after that has arrived -> register the webhooks -> mark the readiness probe to return true.
This flow will guarantees that the pod will be marked ready only after the webhooks are usable.
It is used in the instrumentor and the autoscaler. Added readiness probe for the autoscaler.

### Cons with the new approach
* New permissions are required for the instrumentor/autoscaler pods - watching and listing webhook configurations and watching and updating the certificate secret. These permissions are as minimal as possible - limited to our namespace and specific resources to update - to minimize the permission additions.
* Once the certificate is rotated, the kubelet sync period will cause the same affect. With this change this will happen `LookaheadInterval` before the cert is expired. Assuming 1 year expiration and 100 sec kubelet sync period that is 99.99968% availability - for more details: https://github.com/open-policy-agent/cert-controller/issues/50
* The `certRotator` should probably only be launched in the leader instance, but this is not possible to do in a reliable way due to https://github.com/open-policy-agent/cert-controller/issues/53. So currently we have the cert controller running on both leader and non leader instances. This can lead to conflict error if the 2 instances are trying to update the same resource at the same time - but the reconcile retry logic will handle this.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [x] Updated e2e Tests
The `env-injection` test is updated to remove un-required sleeps that were there because of the webhook down times after `odigos config` command.
Since the readiness probes have been updated for the instrumentor and the autoscaler, the "wait for odigos to be installed" part of the tests was refactored to a script that waits for the pods to be ready.
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

Adding another controller from the package.
The controller comes with more queries but very low QPS and should only be active at init phase or when the cert is rotated.
Helm and CLI are updated to no longer generate the certificates and populdate them in the webhook configs.
Yes, see above.

## User Facing Changes

- [x] Automatic migration will modify existing objects (backward compatible)
Yes, the controler wil reconcile the secret and webhook configs if required.


autoscaler/controllers/manager.go
```diff
@@ -173,3 +172,7 @@ func SetupWithManager(mgr manager.Manager, imagePullSecrets []string, odigosVers
 
 	return nil
 }
+
+func RegisterWebhooks(mgr manager.Manager) error {
+	return actions.RegisterWebhooks(mgr)
```
damemi, line 177: how is this used?

scripts/rbac-docgen/main.go
```diff
@@ -37,12 +38,15 @@ This page lists the Kubernetes Roles and ClusterRoles used by Odigos and the Odi
 `
 
 	cmd := exec.Command("helm", "template", "odigos", "../../helm/odigos", "--set", "openshift.enabled=true")
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
 	out, err := cmd.Output()
+	result := string(out)
 	if err != nil {
 		fmt.Println("Error:", err)
+		fmt.Println("Stderr:", stderr.String())
```
damemi, line 47: 👍

damemi (approved):
overall lgtm, nice improvement from our current approach! I'm curious how it interacts with a current install, I guess the new controllers will just update the secrets automatically? Pretty cool library, good find

RonFed:
> overall lgtm, nice improvement from our current approach! I'm curious how it interacts with a current install, I guess the new controllers will just update the secrets automatically? Pretty cool library, good find

exactly, they reconcile the secret and align it with the webhook config - since the CA definitions I pass to the rotator are not identical to what was done in the CLI/Helm - on upgrade from current installation the secret will be updated.

tamirdavid1 (approved):

cli/cmd/resources/instrumentor.go
```diff
@@ -312,18 +344,10 @@ func NewSourceValidatingWebhookConfiguration(ns string, caBundle []byte) *admiss
 		},
 	}
 
-	if caBundle == nil {
```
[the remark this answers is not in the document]
RonFed: Even before this change we have no dependency or integration with cert manager.
This solution aims to be self contained to Odiogs.

instrumentor/instrumentor.go
```diff
@@ -37,6 +49,41 @@ func New(opts controllers.KubeManagerOptions, dp *distros.Provider) (*Instrument
...
+
+		// we could set RequireLeaderElection to true here but that will make the readiness probe fail for non-leader
+		// instances (since the IsReady channel will not be closed in non-leader instances).
+
+		// these are the defaults, but we set them explicitly for clarity
+		CaCertDuration:         10 * 365 * 24 * time.Hour, // 10 years
+		ServerCertDuration:     1 * 365 * 24 * time.Hour,  // 1 year
+		RotationCheckFrequency: 12 * time.Hour,            // 12 hours
+		LookaheadInterval:      90 * 24 * time.Hour,       // 90 days
+	})
```
blumamir, line 82: can we share this part which is duplicated in instroumentor and autoscaler via k8sutils?

helm/odigos/templates/autoscaler/webhooks.yaml
```diff
@@ -48,6 +43,4 @@ metadata:
   annotations:
     "helm.sh/hook": "pre-install,pre-upgrade"
     "helm.sh/hook-delete-policy": "before-hook-creation"
-data:
-  tls.crt: {{ $cert.Cert | b64enc }}
-  tls.key: {{ $cert.Key | b64enc }}
+# the actual certificate will be populated by the controller
```
blumamir, line 46: If someone is using helm-controller, then this controller might "fight" over this resource and will attempt to reconcile it back to secret with empty data, no?

helm/odigos/templates/instrumentor/webhooks.yaml
```diff
@@ -114,6 +107,4 @@ metadata:
   annotations:
     "helm.sh/hook": "pre-install,pre-upgrade"
     "helm.sh/hook-delete-policy": "before-hook-creation"
```
blumamir, line 109: are these relevant after the change? 
as it is now managed by the controller, I am not sure if these instructions to helm about creation and deletion are actually needed.

blumamir (approved):
Looks much better than our current certificate managing 🏆

Added few non-blocking questions / suggestions

cli/cmd/resources/autoscaler.go
```diff
@@ -451,14 +478,10 @@ func NewAutoscalerTLSSecret(ns string, cert *crypto.Certificate) *corev1.Secret
 				"helm.sh/hook-delete-policy": "before-hook-creation",
 			},
 		},
-		Data: map[string][]byte{
```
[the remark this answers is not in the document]
RonFed: as we talked and I double verified this - both Helm and the CLI have "least changes required" policy - which means if we don't specify the `Data` there they will not overwrite it with empty value.
In addition the controllers are marked as the `managers` for the `caBundle` field.

PR #2901, opened 2025-05-21
feat: add support for HTTP Headers instrumentation rule in UI backend (+ add docs)

BenElferink:
Following up after: https://github.com/odigos-io/odigos/pull/2826

---

This PR adds API support for the new HTTP Headers rule.
This PR also adds the rule to Odigos Docs.

The UI itself still needs an update via the UI-Kit: https://github.com/odigos-io/ui-kit/pull/134


docs/pipeline/rules/headerscollection.mdx
```diff
@@ -0,0 +1,60 @@
...
+<Warning>
+  Before enabling **headers collection**, please note the following:
+  - The support for headers collection varies between instrumentation libraries and languages.
+</Warning>
+
+## Configuration Options
+
+<AccordionGroup>
+  <Accordion title="headerKeys">
+    **headerKeys** `string[]` : Limit payload collection to specific header keys.
```
blumamir, line 24: I think it can be valuable to also describe how this key is collected.

e.g. that an header key "foo" with value "bar" will be collected as span attribute named `http.request.header.foo` with attribute value `["bar"]` (or `http.response.header.foo` if this is for response)

@edeNFed just making sure java is implemented this way ([semconv](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/registry/attributes/http.md#http-request-header))

docs/pipeline/rules/headerscollection.mdx
```diff
@@ -0,0 +1,60 @@
...
+description: "The \"Headers Collection\" rule can be used to add span attributes by collecting HTTP Headers data."
+sidebarTitle: "Headers Collection"
+icon: "webhook"
+---
+
+<Info>
+  This Instrumentation Rule is currently only available with the Odigos **Enterprise** plan.<br />
+  [Contact us](https://odigos.io/) for more information.
+</Info>
+
```
blumamir, line 12: for http we have both client/server, and request/response.

I guess this configuration option will apply to all of them? might be worth mentioning here

docs/pipeline/rules/headerscollection.mdx
```diff
@@ -0,0 +1,60 @@
...
+<AccordionGroup>
+  <Accordion title="headerKeys">
+    **headerKeys** `string[]` : Limit payload collection to specific header keys.
+    - This field is *optional*
+  </Accordion>
+</AccordionGroup>
+
+<Note>
+  Any unspecified options will not be recorded.
```
blumamir, line 30: What does this mean? should it be unspecified "header" instead of "option"?

docs/pipeline/rules/headerscollection.mdx
```diff
@@ -0,0 +1,60 @@
...
+---
+
+<Info>
+  This Instrumentation Rule is currently only available with the Odigos **Enterprise** plan.<br />
+  [Contact us](https://odigos.io/) for more information.
+</Info>
+
+## Considerations
+
+<Warning>
```
blumamir, line 15: consider adding a warning that http headers might contain PII data (like cookies or api keys) thus it is recommended to review each requested header before configuring it here to avoid sensitive information.

blumamir (approved):
LGTM.

Added few suggestion to the docs which can be nice to add (but optional)

alonkeyval (approved):

PR #2904, opened 2025-05-22
Add GCP/OTLP destination + Options to mount Secrets and set Env Vars in destinations

damemi:
## Description

This adds support for the [GCP OTLP Telemetry API](https://cloud.google.com/stackdriver/docs/reference/telemetry/overview), which uses the OTLPHTTP exporter along with the [googleclientauth extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/627de0d6ca98a0e6fbe4f966aafc2c27fc63d2af/extension/googleclientauthextension). And example for using this endpoint is available at https://github.com/GoogleCloudPlatform/otlp-k8s-ingest

For this destination to work when running Odigos outside of GCP (for example, a local k8s cluster that wants to send to GCP's OTLP backend), users need to authenticate the collector with a GCP service account. This is done by mounting a service account JSON key and defining the location of that file via the `GOOGLE_APPLICATION_CREDENTIALS` environment variable. The googleclientauth extension reads this environment variable to use the credentials. See https://cloud.google.com/docs/authentication/application-default-credentials

To support this, I made a few changes to the destination UI model and CRD:

* Add a `secretFile` boolean to the model, indicating that a field should be mounted as a file from the generated secret
* Add an `envVars` map to the model, which defines environment variables that should be set if the field is used
* Add matching fields to the Destination CRD so they can be propagated in the collector gateway deployment

I'm open to any feedback to this approach, these are just the base requirements for this destination. The current GCP exporter can also use application credentials for off-gcp users, and we can add the option there too.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


common/config/gcp.go
```diff
@@ -4,6 +4,12 @@ import (
 	"github.com/odigos-io/odigos/common"
 )
 
+const (
+	gcpProjectIdKey              = "GCP_PROJECT_ID"
+	gcpBillingProjectIdKey       = "GCP_BILLING_PROJECT"
+	gcpApplicationCredentialsKey = "GCP_APPLICATION_CREDENTIALS"
```
RonFed, line 10: not used?

frontend/graph/schema.resolvers.go
```diff
@@ -602,14 +611,24 @@ func (r *mutationResolver) UpdateDestination(ctx context.Context, id string, des
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to get destination: %v", err)
 	}
 
+	dest.Spec.EnvVars = map[string]string{}
+	for key := range dest.Spec.EnvVars {
+		if _, exists := envVars[key]; !exists {
+			delete(dest.Spec.EnvVars, key)
+		}
+	}
```
RonFed, line 627: we're creating an empty map and iterating over it, I guess a typo?

autoscaler/controllers/clustercollector/deployment.go
```diff
@@ -252,6 +267,35 @@ func getSecretsFromDests(destList *odigosv1.DestinationList) []corev1.EnvFromSou
 	return result
 }
 
+func getSecretVolumesFromDests(destList *odigosv1.DestinationList) ([]corev1.Volume, []corev1.VolumeMount) {
```
RonFed, line 270: consider the case where a user configured credentials and want to update them from the UI - we need to make sure the collector will get the updated value.
It looks like the env var and the config will remain the same - we have a secret reconciler - but I'm not sure it will trigger a restart in this case.

autoscaler/controllers/clustercollector/deployment.go
```diff
@@ -252,6 +267,35 @@ func getSecretsFromDests(destList *odigosv1.DestinationList) []corev1.EnvFromSou
...
+			volumes = append(volumes, corev1.Volume{
+				Name: volumeName,
+				VolumeSource: corev1.VolumeSource{
+					Secret: &corev1.SecretVolumeSource{
+						SecretName: dst.Spec.SecretRef.Name,
+						Items:      items,
+					},
+				},
+			})
+		}
```
RonFed, line 294: how will this handle multiple destinations that have file references?
I think that if we add the same mount path twice, the collector will fail to rollout

api/odigos/v1alpha1/destination_types.go
```diff
@@ -35,6 +35,20 @@ type DestinationSpec struct {
...
 	SourceSelector *SourceSelector `json:"sourceSelector,omitempty"`
+
+	// SecretFiles defines keys from the SecretEnvRef to mount as files alongside
+	// any environment variable secrets.
+	// +optional
+	SecretFiles []SecretFileConfig `json:"secretFiles,omitempty"`
+
+	// EnvVars defines environment variables to add to the Collector deployment.
+	// +optional
+	EnvVars map[string]string `json:"envVars,omitempty"`
```
RonFed, line 46: Not sure why this should be part of the API - it looks like we can completely control this internally per destination and don't expose this to the user, or maybe I'm missing something

RonFed (changes requested):

docs/backends/googlecloudotlp.mdx
```diff
@@ -0,0 +1,97 @@
...
+    kind: Destination
+    metadata:
+      name: googlecloudotlp-example
+      namespace: odigos-system
+    spec:
+      data:
+        GCP_PROJECT_ID: <GCP Project ID>
+        # Note: The commented fields below are optional.
+        # GCP_BILLING_PROJECT: <GCP Billing Project ID>
+        # GCP_APPLICATION_CREDENTIALS: <Google Application Credentials>
```
BenElferink, line 84: Please update the docs-generating script (`docs/sync-dest-doc.py`) to include the new YAML values: `secretFile`, `envVars`

(assuming these values are not hard-coded and need user configuration)

destinations/data/googlecloudotlp.yaml
```diff
@@ -0,0 +1,44 @@
...
+        required: false
+        placeholder: 'myproject'
+        tooltip: The GCP project to count toward billing and quota.
+
+    - name: GCP_APPLICATION_CREDENTIALS
+      displayName: Google Application Credentials
+      componentType: textarea
+      secretFile: true
+      envVars:
+        GOOGLE_APPLICATION_CREDENTIALS: "/secrets/GCP_APPLICATION_CREDENTIALS"
```
BenElferink, line 40: Are these always going to be hardcoded values?

If yes, then please handle directly in the destination `.go` file, the `.yaml` is for user configuration (and doc generation).

If no, then please add a `spec.note` to this YAML telling the user that we handle the path internally.
See usage of `spec.note` here: https://docs.odigos.io/adding-new-dest#allowed-properties
See example of "internal handling" note here: https://docs.odigos.io/backends/hyperdx

api/config/crd/bases/odigos.io_destinations.yaml
```diff
@@ -47,6 +47,27 @@ spec:
                 type: object
               destinationName:
                 type: string
+              envVars:
+                additionalProperties:
+                  type: string
+                description: EnvVars defines environment variables to add to the Collector
+                  deployment.
+                type: object
+              secretFiles:
```
BenElferink, line 56: Please add to docs under `Allowed properties for spec.fields[]`: https://docs.odigos.io/adding-new-dest#allowed-properties-for-spec-fields

api/config/crd/bases/odigos.io_destinations.yaml
```diff
@@ -47,6 +47,27 @@ spec:
                 type: object
               destinationName:
                 type: string
+              envVars:
```
BenElferink, line 50: Please add to docs under `Allowed properties for spec.fields[]`: https://docs.odigos.io/adding-new-dest#allowed-properties-for-spec-fields

BenElferink (changes requested):

api/odigos/v1alpha1/destination_types.go
damemi: I think @BenElferink's comment and my reply below is relevant here: https://github.com/odigos-io/odigos/pull/2904#discussion_r2106168735

destinations/data/googlecloudotlp.yaml
RonFed: That sounds like a valid option to me.
It can be an optional interface for a destination to implement something like ("CollectorSpecModifier").
A tricky part might be, that all the `Configer` implementation are currently under `common` - which intentionally does not have any dependency on k8s, as it used in VMs as well.

damemi:
@RonFed @BenElferink I made some updates based on your feedback, ptal!

* Added `CollectorSpecConfigurer` interface to `common/config.go`. 
  * While this is only meant for k8s, the dependence on the current `common.ExporterConfigurer` interface would have created an import cycle putting it anywhere else.
  *  This interface provides the `GetCollectorSpec(dest ExporterConfigurer)` function, which accepts the `ExporterConfigurer` destination object, and allows a destination to optionally configure the collector spec based on the config (in this case, we only want to mount the secret if that field was used)
* Persist redacted secret fields in the Destination CRD
  * Please check if this is breaking, but as part of the above the backend needs a good way to know if the secret field was set from the CRD. The presence of the SecretRef alone isn't enough, because those values could just be environment variables

Let me know your thoughts and if there's anything we should tweak. Thanks!

common/go.mod
```diff
@@ -7,12 +7,26 @@ require (
 	github.com/stretchr/testify v1.10.0
 	go.opentelemetry.io/otel/trace v1.35.0
 	gopkg.in/yaml.v2 v2.4.0
+	k8s.io/api v0.33.1
```
RonFed, line 10: we can't add this k8s dependency here,
if an import cycle happens we can create a new package for the new interface under `k8sutils` - the consumers of that new package would be the `api`, `autoscaler` and the ui?

instrumentation/go.sum
```diff
@@ -26,8 +26,8 @@ github.com/odigos-io/runtime-detector v0.0.7 h1:5jguo5bfvQ/pRz18inCB29sf+pOR+fhH
 github.com/odigos-io/runtime-detector v0.0.7/go.mod h1:59ei9q7pYZVs7ClPbB35vPj5AGpDYtENajLhQaikcGU=
 github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 h1:Jamvg5psRIccs7FGNTlIRMkT8wgtp5eCXdBlqhYGL6U=
 github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
-github.com/rogpeppe/go-internal v1.13.1 h1:KvO1DLK/DRN07sQ1LQKScxyZJuNnedQ5/wKSR38lUII=
-github.com/rogpeppe/go-internal v1.13.1/go.mod h1:uMEvuHeurkdAXX61udpOXGD/AzZDWNMNyH2VO9fmH0o=
+github.com/rogpeppe/go-internal v1.11.0 h1:cWPaGQEPrBb5/AsnsZesgZZ9yb1OQ+GOISoDNXVBh4M=
+github.com/rogpeppe/go-internal v1.11.0/go.mod h1:ddIwULY96R17DhadqLgMfk9H9tvdUzkipdSkR5nkCZA=
```
RonFed, line 30: don't think this should have any diff

common/config/config.go
```diff
@@ -51,3 +54,22 @@ type Pipeline struct {
 	Processors []string `json:"processors"`
 	Exporters  []string `json:"exporters"`
 }
+
+// CollectorSpecConfigurer is an interface that allows destinations to configure
+// their collector deployment specifications, such as environment variables and volume mounts.
+type CollectorSpecConfigurer interface {
+	// GetCollectorSpec returns the collector deployment specifications for this destination
+	GetCollectorSpec(dest ExporterConfigurer) *CollectorSpec
```
RonFed, line 62: not sure about returning a pointer here.
If each `CollectorSpecConfigurer` can have some modifications to the spec - it makes sense to pass the pointer as an argument and let each one add its relevant changes (that's how our current implementation works).

Error handling is probably better in the current approach in the PR - since we can identify conflicts in one place and not rely on each implementation to return an error.

seeing everything is in slices we could take the first approach - let all the implementations add their values and check for conflicts once in the end. WDYT?

common/config/config.go
```diff
@@ -14,6 +16,7 @@ type ExporterConfigurer interface {
 	// expected to be unique across all instances of exporters used in collector config, [a-zA-Z0-9-_]+
 	GetID() string
 	GetConfig() map[string]string
+	GetSecretRef() *corev1.LocalObjectReference
```
RonFed, line 19: we can make this optional interface and declare it under `k8sutils` - see other comment in go.mod

common/go.mod
damemi: We could do without the k8s dependency, this is just for the v1 EnvVar and volume fields, that could just be a map or other simple struct

The real import cycle is because if we define `CollectorSpecConfigurer` somewhere outside of `common`, then it needs to import `common` to accept the `ExporterConfigurer` to `GetCollectorSpec()`. Then, implementing the new optional interface in any of the destinations (which are currently under `common/config/*.go`), creates a loop between `common->new package->common`

I think it's important that 1) the `GetCollectorSpec()` function accepts an `ExporterConfigurer`, because the collector spec is configured based on the exporter configuration. And 2) that we should implement the optional interface in the same spot that we implement the current interface (it will be confusing to split up the destinations across 2 packages just for this)

It might make sense here to move the destination `.go` files outside of `common/config`, maybe to `destinations`. That would fix the import cycle and keep the destination code in a unified spot. That also puts the destination Go files under the same parent as their YAML data, so overall I think that's a better move. What do you think?

BenElferink:
> @RonFed I pushed a new commit [eed09ab](https://github.com/odigos-io/odigos/commit/eed09ab79e7cda00c0f87cdb396c281e4277052c) that just moves the destinations from `common/config` -> `destinations/config` and updates all references. I think this is actually a better spot for the destination configs since it aligns with the `data` files and centralizes all the destination code, but I'm curious what you think

Please update the /common references in: https://docs.odigos.io/adding-new-dest

PR #2888, opened 2025-05-19
Migrate api to hub and spoke

TheErikNaz:
## Description
Defined storage versions for all CRDs
Added a hub definition for conversions generation

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [x] Manual Load Test

## Kubernetes Checklist

- [x] Changes how Odigos interacts with Kubernetes
- [ ] Introduces additional calls to the API Server (potential performance impact)
- [ ] New Query/feature supported in all the k8s versions supported by Odigos
- [x] Modifies Odigos manifests (addressed in both CLI and Helm)
- [ ] Changes RBAC permissions

## User Facing Changes

- [ ] Users need to take action before upgrading
- [x] Automatic migration will modify existing objects (backward compatible)
- [ ] Changes UI, CLI, or K8s Manifests aspects in a way that users need to be aware of
- [ ] Documentation updated accordingly

PR #2919, opened 2025-05-26
Adopt otel connectors

tamirdavid1:
## Description

This PR introduces a new structure for organizing our OpenTelemetry pipelines. The main goals of this change are:

1. Reduce Data Duplication: Minimize the amount of data sent to each pipeline in the **"Data-Stream world".**

2. Enable Per Data-Stream Actions: Establish a foundation for applying specific processors (e.g., samplers) to individual data streams.

Previously, when a new Data Stream was created, **all telemetry data was forwarded to all destination pipelines, regardless of its relevance.** Each pipeline used a filtergroupsprocessor to extract only the relevant data. However, this approach caused significant data duplication, as every destination pipeline received the entire dataset.

Previous pipeline structure (per destination):
```
Receiver: OTLP (from data collection)
Processors: [..., filtergroupsprocessor]
Exporter: Jaeger (or other destination)
```

### New Architecture
With this PR, the pipeline architecture is redesigned into three distinct layers:

1. **Root Pipelines (per signal)**
Receive raw data from OTLP receivers for each signal type.

Example:
`traces/in, metrics/in, logs/in`

```
receiver: otlp
processors: [...]
exporter: odigosroutingconnector/<signal>
```

2. **Datastream Pipelines**
Receive only relevant data routed from the Root Pipelines via the odigosroutingconnector.
```
receiver: odigosroutingconnector/traces
processors: [...]
exporter: forward/traces/coralogix
```
3. **Destination Pipelines**
Receive grouped data and apply destination-specific processing.
```
receiver: forward/traces/coralogix
processors: [destination-specific processors]
exporter: Coralogix exporter
```

### Datastreams Routing Logic
Datastreams are still computed using the same label and destination Datastream selectors logic. They are supplied to the routing connector in the following format:
```
odigosrouterconnector/traces:
  dataStreams:
    - name: B
      sources:
        - namespace: default
          kind: Deployment
          name: frontend
        - namespace: default
          kind: Deployment
          name: membership
        - namespace: default
          kind: Deployment
          name: postgres
        - namespace: default
          kind: Deployment
          name: coupon
        - namespace: default
          kind: Deployment
          name: currency
      destinations:
        - destinationname: odigos.io.dest.coralogix-wzpll
          configuredsignals:
            - METRICS
            - TRACES
            - LOGS
```

## How Has This Been Tested?

- [x] Added Unit Tests
- [x] Updated e2e Tests
- [X] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


.github/workflows/build.yaml
```diff
@@ -93,7 +93,6 @@ jobs:
           push: false
           tags: odigos-collector:pr-${{ github.event.number }}
           provenance: false
-          context: ./collector
```
RonFed, line 96: looking at our release flow, this change seems to be not consistent with how we build the collector.
see https://github.com/odigos-io/odigos/blob/9de40f787da19405ac1b179eb987b5c16fe6a832/.github/workflows/publish-modules.yml#L301

If this change is required, we should update this everywhere including in the release

common/config/root.go
```diff
@@ -305,15 +107,15 @@ func isSignalExists(dest SignalSpecific, signal common.ObservabilitySignal) bool
...
 
-func isMetricsEnabled(dest SignalSpecific) bool {
+func IsMetricsEnabled(dest SignalSpecific) bool {
 	return isSignalExists(dest, common.MetricsObservabilitySignal)
 }
 
-func isLoggingEnabled(dest SignalSpecific) bool {
+func IsLoggingEnabled(dest SignalSpecific) bool {
 	return isSignalExists(dest, common.LogsObservabilitySignal)
 }
```
RonFed: nit: we don't have to export these for the autoscaler
since the autoscaler works with destinations from the api, and those have `GetSignals` function - the autoscaler can use that.
as a rule of thumb, if the code is in a k8s specific location (like the autoscaler) - it can use k8s specific functions like those from the api dir - but all the the different config files under common are not specific to k8s and hence it makes sense for them to use these common functions (like they do today).

operator/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/odigos-io/odigos/operator
 go 1.24.0
 
 require (
-	github.com/odigos-io/odigos/api v0.0.0
+	github.com/odigos-io/odigos/api v1.0.187
```
RonFed: is this intentional?

api/odigos/v1alpha1/source_types.go
```diff
@@ -94,12 +94,12 @@ type SourceSelector struct {
...
 	Namespaces []string `json:"namespaces,omitempty"`
-	// Workloads (sources) are assigned to groups via labels (odigos.io/group-backend: true), allowing a more flexible selection mechanism.
+	// Workloads (sources) are assigned to Datastreams via labels (odigos.io/data-stream: true), allowing a more flexible selection mechanism.
 	// Example:
-	// groups: ["backend", "monitoring"]
+	// dataStreams: ["backend", "monitoring"]
 	// This means the destination will receive data only from sources labeled with "backend" or "monitoring".
 	// +optional
-	Groups []string `json:"groups,omitempty"`
+	DataStreams []string `json:"dataStreams,omitempty"`
```
RonFed: it is a bit confusing that this is used only by destinations by defined in the sources file.

common/pipelinegen/datastreams.go
```diff
@@ -0,0 +1,69 @@
+// Package pipelinegen contains helper structures for dynamically generating OpenTelemetry
+// pipelines based on source groups and their associated destinations.
+//
+// This file (`groups.go`) defines the core data structures used to:
```
RonFed:
```suggestion
// This file (`datastreams.go`) defines the core data structures used to:
```

common/pipelinegen/datastreams.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/odigos-io/odigos/common"
+)
+
+// DataStreams defines a logical group of source workloads and the destination exporters
+// that they are allowed to send observability data to.
+type DataStreams struct {
+	Name         string            `mapstructure:"name"`         // Unique identifier for the group (used as pipeline name suffix)
+	Namespaces   []NamespaceFilter `mapstructure:"namespaces"`   // List of namespaces belonging to this group [marked as future select]
+	Sources      []SourceFilter    `mapstructure:"sources"`      // List of workloads belonging to this group
```
RonFed: please add a comment about the relation between namespaces and sources (sources take precedence over ns)

common/pipelinegen/datastreams.go
RonFed, line 1: Not sure about this being under `common`.
1. It is not consistent with all of our other collector components that declare their config inside the collector.
2. The `autoscaler` eventually deals with generic maps.
I don't have a strong opinion about this.

common/pipelinegen/pipeline_builder.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/odigos-io/odigos/common/config"
+	"github.com/odigos-io/odigos/common/consts"
+)
+
+type OtelPipelines map[string]config.Pipeline
```
RonFed: not used?

common/pipelinegen/pipeline_builder.go
```diff
@@ -0,0 +1,53 @@
...
+) map[string]config.Pipeline {
+	pipelines := make(map[string]config.Pipeline)
+
+	for _, dataStream := range dataStreams {
+		for _, signal := range []string{"logs", "metrics", "traces"} {
+			pipelineName := fmt.Sprintf("%s/%s", signal, dataStream.Name)
+
+			pipeline := config.Pipeline{
+				Receivers:  []string{fmt.Sprintf("odigosrouterconnector/%s", signal)},
+				Processors: []string{consts.GenericBatchProcessor}, // every group pipeline should have a generic batch processor
```
RonFed: why do we need another batcher for each pipeline? we have batching in earlier steps as well? or should we have batching only in the `in` steps?

common/pipelinegen/pipeline_builder.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/odigos-io/odigos/common/config"
+	"github.com/odigos-io/odigos/common/consts"
+)
+
+type OtelPipelines map[string]config.Pipeline
+
+// BuildDataStreamPipelines constructs data stream pipelines for logs, metrics, traces.
+// Each pipeline receives from its routing connector and exports to all destinations relevant to the data stream.
+func BuildDataStreamPipelines(
```
RonFed: if this function is only used in this package it should not be exported

instrumentor/controllers/sources_webhooks.go
```diff
@@ -41,6 +42,7 @@ type SourcesDefaulter struct {
 }
 
 var _ webhook.CustomDefaulter = &SourcesDefaulter{}
+var defaultDataStreamLabel = k8sconsts.SourceDataStreamLabelPrefix + consts.DefaultDataStream
 
 // TODO: uncomment when Data Streams are ready to use
 // var defaultDataStreamLabel = k8sconsts.SourceGroupLabelPrefix + consts.DefaultDataStream
```
RonFed: should be removed?

collector/connectors/odigosrouterconnector/routingmap.go
```diff
@@ -0,0 +1,112 @@
...
+//	        "metrics": ["dataStream-B"],
+//	    },
+//	    {ns2/statefulset/db}: {
+//	        "traces": ["dataStream-B"],
+//	    },
+//	    {ns3/*/*}: {
+//	        "traces": ["dataStream-Default"],
+//	    },
+//	}
+type SignalRoutingMap map[string]RoutingIndex
```
RonFed, line 28: nit/suggestions: we can have a struct as the key here which will have the name, namespace and kind fields - this will save all the string operations.
The frontend backend has a similar struct for identifying sources. I think it can make the code cleaner.

collector/connectors/odigosrouterconnector/routingmap.go
```diff
@@ -0,0 +1,112 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/odigos-io/odigos/common"
+	"github.com/odigos-io/odigos/common/pipelinegen"
+)
+
+// RoutingIndex maps signals (logs/metrics/traces) to dataStream pipelines
+type RoutingIndex map[string][]string // signal -> []pipeline names
```
RonFed: instead of string for signal we can use an "enum" (which I think we have declared in common)

collector/connectors/odigosrouterconnector/connector.go
```diff
@@ -0,0 +1,379 @@
...
+	if !ok {
+		return nil, ""
+	}
+
+	return pipelines, key
+}
+
+func (r *routerConnector) ConsumeTraces(ctx context.Context, td ptrace.Traces) error {
+	cfg := r.tracesConfig
+	groups := make(map[consumer.Traces]ptrace.Traces)
```
RonFed:
```suggestion
	tracesByConsumer := make(map[consumer.Traces]ptrace.Traces)
```

collector/connectors/odigosrouterconnector/connector.go
```diff
@@ -0,0 +1,379 @@
...
+
+	config := cfg.(*Config)
+
+	defaultTracesConsumer, err := tr.Consumer(
+		collectorpipeline.NewIDWithName(collectorpipeline.SignalTraces, consts.DefaultDataStream),
+	)
+	if err != nil {
+		set.Logger.Warn("failed to get default traces consumer")
+		// Do not return the error — just continue with nil fallback
+		defaultTracesConsumer = nil
```
RonFed, line 75: can this cause panics if we try to call consume for the default consumer?
what can cause an error here? maybe we should return an error if we can't set up a default consumer

autoscaler/controllers/clustercollector/configmap.go
```diff
@@ -242,3 +249,218 @@ func patchConfigMap(existing *v1.ConfigMap, desired *v1.ConfigMap, ctx context.C
...
+		return false
+	}
+	// we should add the odigostrafficmetrics processor to all destination pipelines
+	if slices.Contains(destinationPipelines, name) {
+		return true
+	}
+	return false
+}
+
+func findOrCreateDataStream(dataStreams *[]pipelinegen.DataStreams, name string) *pipelinegen.DataStreams {
```
RonFed: I think we should consider building a map once and not iterate over the slice each time - it will make the code clearer IMO.

This whole "avoiding duplication" logic is classic for a map.

This also relates to my point in other comment regarding sharing the same data structure between the collector and the autoscaler.

The use of slices is required for the config encoding/decoding but for our code clarity and efficiency it makes more sense to use a different data structure.

autoscaler/controllers/clustercollector/configmap.go
```diff
@@ -242,3 +249,218 @@ func patchConfigMap(existing *v1.ConfigMap, desired *v1.ConfigMap, ctx context.C
...
+	}
+	return configuredSignals
+}
+
+// For the default data stream, include all sources that don't have any data stream labels assigned
+// and sources that have the default data stream label assigned
+func getSourcesForDefaultDataStream(ctx context.Context, kubeClient client.Client, group string) (*odigosv1.SourceList, *odigosv1.SourceList, error) {
+
+	defaultStreamSources := &odigosv1.SourceList{}
+	err := kubeClient.List(ctx, defaultStreamSources, &client.ListOptions{})
```
RonFed, line 409: List Options can have a query for "give me everything that don't have this label" maybe we can use this here instead of filtering after the query

RonFed (commented):
did a first pass, didn't look at `common` and tests yet

operator/go.mod
tamirdavid1: No, thanks!

common/pipelinegen/pipeline_builder.go
tamirdavid1: Discussed with @blumamir on this, basically the `in` pipeline (AKA Root Pipeline) will forward the spans in batch.
but in each pipeline the the spans will be processed and forwarded not in batches, adding it can potentially make the processing more efficient.
at the worst case the batch will reachout timeout or maximum amount of spans and will forward the batch to the step

instrumentor/controllers/sources_webhooks.go
tamirdavid1: I know @BenElferink want to keep it as a reminder for un-comment in some case later.
I'm not sure why. @BenElferink can u please help here and elaborate? thanks

tamirdavid1:
- After review face 2 face, we will temporary need to move the small batches processor to the destination pipeline.

common/pipelinegen/pipeline_builder.go
tamirdavid1: After discuss face to face on this, better to keep the batches in each pipeline.

api/odigos/v1alpha1/source_types.go
tamirdavid1: Agree, moved to `destination_types.go`

collector/connectors/odigosrouterconnector/connector.go
tamirdavid1: It cant cause a panic because in the connector logic itself, before forwarding to default consumer i'm checking if it is not set to nil

```
// Fallback, if any spans unmatched
if defaultTraces.ResourceSpans().Len() > 0 {
if cfg.defaultCons != nil {
	if err := cfg.defaultCons.ConsumeTraces(ctx, defaultTraces); err != nil {
		cfg.logger.Debug("failed to send traces to the default pipeline", zap.Error(err))
	}
}
```

autoscaler/controllers/clustercollector/configmap.go
tamirdavid1: The problem is that the datastreams are in the label name itself and not the value (maybe worth change it)

odigos.io/data-stream-EXAMPLE: "true" instead of

odigos.io/data-stream: "EXAMPLE"

tamirdavid1: the cons is that we will need to maintain a list of strings slice like and parse it in case we will want to have multiple datastreams per source

tamirdavid1:
Enhance e2e tests:
1. Add to the tests one service that will be sent to all of them
2. Source that does not exists in any of them

api/odigos/v1alpha1/destination_types.go
```diff
@@ -85,3 +85,25 @@ func (dest Destination) GetConfig() map[string]string {
...
+	// +optional
+	DataStreams []string `json:"dataStreams,omitempty"`
+
+	// Selection Semantics:
+	// If both `Namespaces` and `Groups` are specified, the selection follows an **OR** logic:
+	// - A source is included **if** it belongs to **at least one** of the specified namespaces OR groups.
+	// - If `Namespaces` is empty but `Groups` is specified, only sources in those groups are included.
+	// - If `Groups` is empty but `Namespaces` is specified, all sources in those namespaces are included.
+	// - If SourceSelector is nil, the destination receives data from all sources.
+}
```
damemi:
```suggestion
	// If both `Namespaces` and `DataStreams` are specified, the selection follows an **OR** logic:
	// - A source is included **if** it belongs to **at least one** of the specified namespaces OR data streams.
	// - If `Namespaces` is empty but `DataStreams` is specified, only sources in those data streams are included.
	// - If `DataStreams` is empty but `Namespaces` is specified, all sources in those namespaces are included.
	// - If SourceSelector is nil, the destination receives data from all sources.
}
```

api/odigos/v1alpha1/destination_types.go
```diff
@@ -85,3 +85,25 @@ func (dest Destination) GetConfig() map[string]string {
...
+	// namespaces: ["default", "production"]
+	// This means the destination will receive data from all sources in "default" and "production" namespaces.
+	// +optional
+	Namespaces []string `json:"namespaces,omitempty"`
+	// Workloads (sources) are assigned to Datastreams via labels (odigos.io/data-stream: true), allowing a more flexible selection mechanism.
+	// Example:
+	// dataStreams: ["backend", "monitoring"]
+	// This means the destination will receive data only from sources labeled with "backend" or "monitoring".
+	// +optional
+	DataStreams []string `json:"dataStreams,omitempty"`
```
damemi, line 101: not sure but I think changing this name in the crd from `groups`->`dataStreams` might cause API issues for existing sources stored in etcd

damemi (commented):
love the approach, still reading the implementation code but overall this sounds like what we should be doing

api/odigos/v1alpha1/destination_types.go
RonFed: +1 that's a good point to check

tamirdavid1: As far as i know, we were not set the groups anytime so the whole `SourceSelector` object is nil.
@BenElferink in main now we block setting it right?

tamirdavid1: Also, if the cli-upgrade passed I guess it is not a problem because now we have groups in the crd and post upgrade it is changing to dataStreams

BenElferink: @tamirdavid1 I don't think this setting is blocked - if I'm not mistaken, @alonkeyval had Imran test Data Streams a few weeks ago (manifests only, no UI).

damemi: @tamirdavid1 I don't think cli-upgrade actually sets any groups on a Source does it? That would be the problem. The definition itself will upgrade no problem, but the issue is if a Source exists in the cluster that sets the `groups` field. When the definition is upgraded, and the API goes to fetch the existing source, it will either:

* Just drop the (now unknown) field and return a blank `dataStreams` field
* Return a schema error for an unknown field

I think it will probably just do the first one. The only way around this would be adding a new api version and a conversion webhook similar to https://github.com/odigos-io/odigos/pull/2888.

common/pipelinegen/datastreams.go
damemi: I think since the autoscaler uses this package, it's better to not have a dependency on the collector there. I see this as more of a common type that happens to be used by the collector config, but the structures are shared across packages outside of the collector if I understand correct

autoscaler/controllers/clustercollector/configmap.go
```diff
@@ -242,3 +250,190 @@ func patchConfigMap(existing *v1.ConfigMap, desired *v1.ConfigMap, ctx context.C
...
+func GetDataStreamsWithDestinations(
+	ctx context.Context,
+	kubeClient client.Client,
+	dests *odigosv1.DestinationList,
+	logger logr.Logger,
+) ([]pipelinegen.DataStreams, error) {
+
+	dataStreamsMap := make(map[string]*pipelinegen.DataStreams)
+
+	for _, dest := range dests.Items {
```
damemi, line 288: it's interesting that the pipeline is built in reverse, working backward from destinations->sources. The comment for this function with the example output is great, could you please also explain the logic/flow there since that's counterintuitive (at least to me)

collector/connectors/odigosrouterconnector/connector.go
```diff
@@ -0,0 +1,373 @@
...
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/connector"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+	collectorpipeline "go.opentelemetry.io/collector/pipeline"
+	semconv1_21 "go.opentelemetry.io/otel/semconv/v1.21.0"
```
damemi: is there a reason for v1.21 of semconv? it's up to v1.36 now, and binding to this might be a pain. we'll have to keep it in sync with the semconv versions used by the instrumentation

collector/connectors/odigosrouterconnector/routingmap.go
```diff
@@ -0,0 +1,96 @@
...
+// This is used to forward all signals for signal data stream pipelines e.g. logs/groupA, traces/groupC, metrics/groupB.
+func GetSignalsForDataStream(dataStream pipelinegen.DataStreams) []common.ObservabilitySignal {
+	signals := []common.ObservabilitySignal{}
+	seen := make(map[common.ObservabilitySignal]struct{})
+
+	for _, destination := range dataStream.Destinations {
+		for _, sig := range destination.ConfiguredSignals {
+			if _, exists := seen[sig]; !exists {
+				seen[sig] = struct{}{}
+				signals = append(signals, sig)
```
damemi, line 93: could you `break` once `len(signals) == 3`?

common/pipelinegen/config_builder.go
```diff
@@ -0,0 +1,269 @@
...
+	processors []config.ProcessorConfigurer,
+	memoryLimiterConfig config.GenericMap,
+	applySelfTelemetry func(c *config.Config, destinationPipelineNames []string, signalsRootPipelines []string) error,
+	groupDetails []DataStreams,
+) (string, error, *config.ResourceStatuses, []common.ObservabilitySignal) {
+	currentConfig := GetBasicConfig(memoryLimiterConfig)
+	return CalculateGatewayConfig(currentConfig, dests, processors, applySelfTelemetry, groupDetails)
+}
+
+func CalculateGatewayConfig(
```
damemi, line 26: similar to the datastreams map, I am a little confused at first about how the pipelines are built in reverse (destination->forward->root), a comment on this function explaining the logic would be really helpful.

It seems like the basic reasoning is that we don't know what signals to build forwarding pipelines for until we look at the destinations, then we don't know what forwarding pipelines to build until we look at the sources. is that right?

collector/connectors/odigosrouterconnector/connector.go
tamirdavid1: Great comment, I was searching for semconv in our repo and saw that 1.26 is the most common used (at least in go). 100% agree that we should manage it one place across instrumentations + core code.

collector/connectors/odigosrouterconnector/routingmap.go
tamirdavid1: Done, Thanks

instrumentor/controllers/instrumentationconfig/source_controller.go
```diff
@@ -28,15 +52,78 @@ func (r *SourceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	sourceDataStreamsLabels := sourceutils.GetSourceDataStreamsLabels(source)
+
+	// Namespace Source Reconciliation
 	if source.Spec.Workload.Kind == k8sconsts.WorkloadKindNamespace {
-		// Namespace Source does not contain configuration for a specific workload
+		// Step 1: Load all workload sources in the namespace
+		var workloadSources odigosv1alpha1.SourceList
```
blumamir: isn't all of this taken care of by the [source instrumentation controllers](https://github.com/odigos-io/odigos/blob/039dcc14cff8d63af8b53279caef0f44111b737a/instrumentor/controllers/sourceinstrumentation/namespace_controller.go#L44). Do we have to calculate this here again?

common/pipelinegen/config_builder.go
```diff
@@ -0,0 +1,269 @@
...
+			"resource/odigos-version": config.GenericMap{
+				"attributes": []config.GenericMap{
+					{
+						"key":    "odigos.version",
+						"value":  "${ODIGOS_VERSION}",
+						"action": "upsert",
+					},
+				},
+			},
+			"batch/generic-batch-processor": config.GenericMap{}, // Currently configured with default values
```
blumamir: nit: you've added a const for that called `GenericBatchProcessor`.

Can we use it here?

common/consts/consts.go
```diff
@@ -109,6 +109,13 @@ const (
 	KarpenterStartupTaintKey = "odigos.io/needs-init"
 )
 
+// Batch processor related consts
+const (
+	GenericBatchProcessor = "batch/generic-batch-processor"
```
blumamir: nit: it's the processor name in the collector config map. might be a bit more readable this way

```suggestion
	GenericBatchProcessorConfigKey = "batch/generic-batch-processor"
```

common/pipelinegen/config_builder.go
blumamir, line 26: I can see this function is used in testing, but ideally it should be private the the module (start with lowercase letter), as it's an internal one. the `GetGatewayConfig` is the public function a caller is expected to use

common/pipelinegen/config_builder.go
```diff
@@ -0,0 +1,269 @@
...
+		Service: config.Service{
+			Pipelines:  map[string]config.Pipeline{},
+			Extensions: []string{"health_check", "pprof"},
+		},
+	}
+}
+
+func filterSmallBatchesProcessor(tracesProcessors []string) ([]string, bool) {
+	smallBatchesEnabled := false
+	var filtered []string
```
blumamir: nit (golang best practice): since you know the expected size, you can initiate it with exact capacity so it does not need to scale during the iteration.

```suggestion
	filtered := make([]string, 0, len(tracesProcessors))
```

common/pipelinegen/config_builder.go
```diff
@@ -0,0 +1,269 @@
...
+		return "", err, status, nil
+	}
+
+	return string(data), nil, status, signals
+}
+
+func prepareRootPipelines(currentConfig *config.Config, groupDetails []DataStreams, tracesProcessors,
+	metricsProcessors, logsProcessors []string, signals []common.ObservabilitySignal) {
+	// for each signal, create a root pipeline and a connector
+	if slices.Contains(signals, common.TracesObservabilitySignal) {
```
blumamir, line 154: nit (style): the 3 functions for the 3 signals are almost the same, aside from the "traces" / "metrics" / "logs" being used, and the `ObservabilitySignal` being used. 

consider adding a function `applyRootPipelineForSignal(signal common.ObservabilitySignal, ...)` and calling it 3 times to make the code more manageable

common/pipelinegen/datastreams.go
```diff
@@ -0,0 +1,61 @@
...
+type Destination struct {
+	DestinationName   string                       `mapstructure:"destinationname"`
+	ConfiguredSignals []common.ObservabilitySignal `mapstructure:"configuredsignals"`
+}
+
+// telemetryRootPipelinesBySignal maps observability signal types to their corresponding
+// root pipelines. This mapping helps identify the initial pipeline for a given signal
+// when building the telemetry configuration.
+// and also to be single source of truth for the root pipelines
+var telemetryRootPipelinesBySignal = map[string]string{
```
blumamir: suggestion: instead of having the map key as `string` and calling `strings.ToLower`, we can change this map key to `common.ObservabilitySignal` which the caller already know. it will make the code a bit more type safe and reads better.

common/pipelinegen/config_builder.go
```diff
@@ -0,0 +1,269 @@
...
+	// Final marshal to YAML
+	data, err := yaml.Marshal(currentConfig)
+	if err != nil {
+		return "", err, status, nil
+	}
+
+	return string(data), nil, status, signals
+}
+
+func prepareRootPipelines(currentConfig *config.Config, groupDetails []DataStreams, tracesProcessors,
```
blumamir: nit: consider calling this function with a name that indicates it has side effects (e.g. going to mutate the config)

```suggestion
func insertRootPipelinesToConfig(currentConfig *config.Config, groupDetails []DataStreams, tracesProcessors,
```

or "applyRootPipelines"

blumamir: nit: consider moving this function to `pipeline_builder.go` since it's actually pipeline related

common/pipelinegen/config_builder.go
```diff
@@ -0,0 +1,269 @@
...
+			case strings.HasPrefix(pipelineName, "logs/"):
+				logsEnabled = true
+			}
+
+			// save the updated pipeline with the new receiver
+			currentConfig.Service.Pipelines[pipelineName] = pipeline
+		}
+	}
+	// track which signals are enabled
+	signals := []common.ObservabilitySignal{}
```
blumamir: nit: name better represents it's content

```suggestion
	enabledSignals := []common.ObservabilitySignal{}
```

common/pipelinegen/config_builder.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+
+	//  Add pipelines that receive from routing connectors and forward to destinations
+	groupPipelines := buildDataStreamPipelines(groupDetails, destForwardConnectors)
+	for name, pipe := range groupPipelines {
+		currentConfig.Service.Pipelines[name] = pipe
+	}
+
+	// Create root pipelines for each signal and connectors
+	prepareRootPipelines(currentConfig, groupDetails, tracesProcessors, metricsProcessors, logsProcessors, signals)
```
blumamir: nit: the above function `buildDataStreamPipelines` returns the pipelines to be applied, while this one modifies the `currentConfig` it receive as parameter.

I suggest to be consistent and consider also returning the root pipelines here and apply them on `currentConfig` explicitly in here.

common/pipelinegen/pipeline_builder.go
```diff
@@ -0,0 +1,51 @@
...
+	dataStreams []DataStreams,
+	forwardConnectorByDest map[string][]string,
+) map[string]config.Pipeline {
+	pipelines := make(map[string]config.Pipeline)
+
+	for _, dataStream := range dataStreams {
+		for _, signal := range []string{"logs", "metrics", "traces"} {
+			pipelineName := fmt.Sprintf("%s/%s", signal, dataStream.Name)
+
+			pipeline := config.Pipeline{
```
blumamir, line 23: nit: this pipeline will only be "used" if we found an exporter for it (which I guess is not the case most of the time since we always iterate all signals).

consider only creating this object once we actually going to use it (e.g. if any exporters for this datastream and signals is configured)

autoscaler/controllers/clustercollector/configmap.go
```diff
@@ -242,3 +250,190 @@ func patchConfigMap(existing *v1.ConfigMap, desired *v1.ConfigMap, ctx context.C
...
+func destinationExists(list []pipelinegen.Destination, item string) bool {
+	for _, v := range list {
+		if v.DestinationName == item {
+			return true
+		}
+	}
+	return false
+}
+
+func isOdigosTrafficMetricsProcessorRelevant(name string, rootPipelines []string, destinationPipelines []string) bool {
```
blumamir: do we use the `rootPipelines` for anything?
looks like it was an effort to propgate it all the way here

common/pipelinegen/pipeline_builder.go
```diff
@@ -0,0 +1,51 @@
...
+				if !exists {
+					continue
+				}
+
+				for _, connectorName := range connectors {
+					if strings.HasPrefix(connectorName, fmt.Sprintf("forward/%s/", signal)) {
+						pipeline.Exporters = append(pipeline.Exporters, connectorName)
+					}
+				}
+			}
```
blumamir, line 42: consider extracting this to function that returns list of exporters for a giving destinations, connectors, and signal. 

as this is already heavily nested, it can aid in readability and figuring out what is going on, IMO.

common/pipelinegen/pipeline_builder.go
```diff
@@ -0,0 +1,51 @@
...
+			// Add forward connectors for each destination in the group to route telemetry data
+			// Forward connectors follow the naming pattern: forward/<signal>/<destination-id>
+			for _, dest := range dataStream.Destinations {
+				connectors, exists := forwardConnectorByDest[dest.DestinationName]
+				if !exists {
+					continue
+				}
+
+				for _, connectorName := range connectors {
+					if strings.HasPrefix(connectorName, fmt.Sprintf("forward/%s/", signal)) {
```
blumamir, line 38: nit: here we need to calculate the `fmt.Sprintf` on every iteration. we can pre-calculate a `forwardConnectorPrefix` once before the loop, and use it here which is also a bit more readable :)

autoscaler/controllers/clustercollector/configmap.go
```diff
@@ -242,3 +250,190 @@ func patchConfigMap(existing *v1.ConfigMap, desired *v1.ConfigMap, ctx context.C
...
+//	        Sources: []SourceFilter{
+//	            {Namespace: "ns2", Kind: "StatefulSet", Name: "db"},
+//	        },
+//	        Destinations: []Destination{
+//	            {DestinationName: "jaeger",
+//				 ConfiguredSignals: []ObservabilitySignal{TracesObservabilitySignal}},
+//	        },
+//	    },
+//	}
+func GetDataStreamsWithDestinations(
```
blumamir: nit: internal to this go package:
```suggestion
func getDataStreamsWithDestinations(
```

blumamir: it does include destinations in the response, but also sources. 
how about: "calculateDataStreams" which plays well with what this function returns - an array of `pipelinegen.DataStreams`.

autoscaler/controllers/clustercollector/configmap.go
```diff
@@ -242,3 +250,190 @@ func patchConfigMap(existing *v1.ConfigMap, desired *v1.ConfigMap, ctx context.C
...
+	ctx context.Context,
+	kubeClient client.Client,
+	dests *odigosv1.DestinationList,
+	logger logr.Logger,
```
blumamir: with controller-runtime, you can always extract a logger from the `ctx` with 

```go
logger := ctrl.LoggerFrom(ctx)
```

so no need to pass this `logger` around as function parameters.

autoscaler/controllers/clustercollector/configmap.go
```diff
@@ -242,3 +250,190 @@ func patchConfigMap(existing *v1.ConfigMap, desired *v1.ConfigMap, ctx context.C
...
+			Namespace: instrumentationConfig.Namespace,
+			Kind:      string(workloadKind),
+			Name:      workloadName,
+		})
+	}
+
+	return sourcesFilters, nil
+}
+
+func destinationExists(list []pipelinegen.Destination, item string) bool {
```
blumamir:
```suggestion
func destinationExists(existingDestinations []pipelinegen.Destination, destinationName string) bool {
```

autoscaler/controllers/clustercollector/configmap.go
```diff
@@ -242,3 +250,190 @@ func patchConfigMap(existing *v1.ConfigMap, desired *v1.ConfigMap, ctx context.C
...
+				dataStreamDetails.Destinations = append(dataStreamDetails.Destinations, pipelinegen.Destination{
+					DestinationName:   dest.Name,
+					ConfiguredSignals: dest.GetSignals(),
+				})
+			}
+		}
+	}
+
+	// Convert map to slice, this is the final result as will be used in the configmap
+	var dataStreamDetailsList []pipelinegen.DataStreams
```
blumamir: nit: create it with it's known capacity so it doesn't need runtime scaling:

```suggestion
	dataStreamDetailsList := make([]pipelinegen.DataStreams, 0, len(dataStreamsMap))
```

autoscaler/controllers/clustercollector/configmap.go
```diff
@@ -242,3 +250,190 @@ func patchConfigMap(existing *v1.ConfigMap, desired *v1.ConfigMap, ctx context.C
...
+				dataStreamDetails = &pipelinegen.DataStreams{
+					Name:         dataStream,
+					Sources:      []pipelinegen.SourceFilter{},
+					Destinations: []pipelinegen.Destination{},
+				}
+
+				// SourcesFilters and NamespacesFilters are attached to the DataStream itself.
+				// They are independent of the Destinations that point to the DataStream.
+				// Therefore, we only load them once per unique data stream.
+				sourcesFilters, err := getSourcesForDataStream(ctx, kubeClient, dataStream, logger)
```
blumamir: nit: this is kind of an important and costly call, which can fail, being invoked within 4 nesting levels.

since we want to do it once per datastream, how about moving it below so that we:
1. collect all datastreams with destinations
2. for each datastream we found - calculate and add the `sourceFilters` for this datastream.

instrumentor/controllers/instrumentationconfig/source_controller.go
tamirdavid1: yes, leftovers from old approach before our sync, deleted.

damemi (approved):

PR #2920, opened 2025-05-27
Fix namespace deletion bug

AvihuHenya:
## Description
This PR introduces changes in how Odigos behaves during the install and uninstall process, depending on the installation method (CLI or Helm).

---

###  Helm Installation

- All resources deployed via Helm now include the Odigos system label:  
  `odigos.io/system-object=true`
- We preserve the existing behavior: if Helm creates the namespace during installation, it will **not** delete it during uninstallation.

---

### CLI Installation

- The Odigos system label is now added to `Role` and `RoleBinding` resources - and they will be deleted too in uninstall.
- During installation:
- The label is added to the `Namespace` only if the namespace was newly created by Odigos.
- During uninstallation: The namespace will only be deleted if it contains the Odigos system label — ensuring we delete only those namespaces created by Odigos (that will resolbe the main bug).

---

###  Documentation
- Added Note about namespace deletions in the condition i mention.
- Added instructions explaining how to pass the `onPremToken` when installing Odigos via Helm (added photo).

Screenshot 2025-05-27 at 8 32 08

---
## Updates
- The `cleanup-job` pass `in-cluster` flag and uses this label to identify and remove only code-created resources, and the rest deletion handle by helm (Avoiding duplicate deletion process).
- Add patch to deletion process: since `odigos-config` resource has a Helm hook annotation, it is not managed by Helm and running `helm uninstall` doesnt delete this object (and the corresponding effective-config). so as a workaround, we explicitly delete it.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [X] Manual Testing
- [ ] Manual Load Test


damemi (approved):

cli/cmd/uninstall.go
```diff
@@ -161,6 +177,16 @@ func UninstallClusterResources(ctx context.Context, client *kube.Client, ns stri
 
 }
 
+func namespaceHasOdigosLabel(ctx context.Context, client *kube.Client, ns string, labelKey, expectedValue string) (bool, error) {
```
RonFed, line 180: since the `labelkey` and `expectedValue` are consts - I think we can remove these 2 arguments to the function and just use the constants inside the function.

cli/cmd/uninstall.go
```diff
@@ -72,17 +73,32 @@ and rollback any metadata changes made to your objects.`,
...
-			// Deleting the namespace during uninstallation will cause Helm to fail due to the loss of the release state.
-			if !k8sutils.IsRunningInKubernetes() {
-				createKubeResourceWithLogging(ctx, fmt.Sprintf("Uninstalling Namespace %s", ns),
-					client, ns, k8sconsts.OdigosSystemLabelKey, uninstallNamespace)
+				hasSystemLabel, err := namespaceHasOdigosLabel(ctx, client, ns, k8sconsts.OdigosSystemLabelKey, k8sconsts.OdigosSystemLabelValue)
+				if err != nil {
+					fmt.Printf("\033[31mERROR\033[0m Failed to check if namespace %s has Odigos label: %s\n", ns, err)
+					os.Exit(1)
+				}
+				if hasSystemLabel {
```
RonFed, line 86: maybe worth to add a comment here saying we're only going to delete the ns if we created (labeled) it

cli/cmd/uninstall.go
```diff
@@ -713,8 +771,29 @@ func uninstallNamespace(ctx context.Context, client *kube.Client, ns, _ string)
...
+		return err
+	}
+
+	for _, cm := range list.Items {
+		if cm.Name == consts.OdigosConfigurationName {
+			return client.CoreV1().ConfigMaps(ns).Delete(ctx, cm.Name, metav1.DeleteOptions{})
+		}
+	}
+
+	return nil
```
RonFed, line 790: The `odigos-config` is the one we should explicitly delete.
`effective-config` is managed by our controller in the `scheduler` - and has an owner reference to the `odigos-config` - so deleting the first one would also delete the second.

RonFed (commented):
overall looks great.
I think we should add to `tests/common/assert_odigos_uninstalled.sh` script, a part that asserts that the RBAC related objects (roles, cluster roles, binding etc') are deleted as well - since we are not checking this currently and this PR fix that.

cli/cmd/uninstall.go
```diff
@@ -161,6 +190,15 @@ func UninstallClusterResources(ctx context.Context, client *kube.Client, ns stri
 
 }
 
+func namespaceHasOdigosLabel(ctx context.Context, client *kube.Client, ns string) (bool, error) {
+	nsObj, err := client.CoreV1().Namespaces().Get(ctx, ns, metav1.GetOptions{})
+	if err != nil {
+		return false, err // namespace may not exist or API error
+	}
+	val, exists := nsObj.Labels[k8sconsts.OdigosSystemLabelKey]
```
RonFed, line 198: maybe worth adding a nil check for `nsObj.Labels`

cli/cmd/uninstall.go
```diff
@@ -72,11 +74,38 @@ and rollback any metadata changes made to your objects.`,
...
+					// PATCH: Currently, the odigos-config resource has a Helm hook annotation,
+					// so it is not managed by Helm (i.e., it lacks Helm metadata labels).
+					// As a result, running `helm uninstall` does not delete this object
+					// (or the corresponding effective-config).
+					// As a workaround, we explicitly delete it here, and let helm delete all other resources.
+					uninstallOdigosConfiguration(ctx, client, ns)
+					fmt.Printf("Uninstalling OdigosConfiguration from namespace %s\n", ns)
+					os.Exit(0)
+				}
+				os.Exit(0)
```
RonFed, line 96: better to use return here

RonFed (approved):
🥇

PR #2924, opened 2025-05-28
Added a rollback for failed instrumentations

TheErikNaz:
## Description
Added an rollback feature for failed instrumentations

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/common/consts/consts.go
+++ b/common/consts/consts.go
@@ -79,8 +79,8 @@ const (
 	UserInstrumentationEnvsProperty   = "user-instrumentation-envs"
 	NodeSelectorProperty              = "node-selector"
 	KarpenterEnabledProperty          = "karpenter-enabled"
-	RollbackDisabledProperty          = "rollback-disabled"
-	RollbackGraceTimeProperty         = "rollback-grace-time"
+	RollbackDisabledProperty          = "instrumentation-auto-rollback-disabled"
+	RollbackGraceTimeProperty         = "instrumentation-auto-rollback-grace-time"
 )
 
 var (


damemi (commented):
this is interesting, but I'm not sure if we want to always automatically roll back because of a crash loop. There isn't necessarily enough to know that we are the reason for the crash loop, and if we are then we probably want to be able to inspect the crashing pod to see why, right?

TheErikNaz:
> this is interesting, but I'm not sure if we want to always automatically roll back because of a crash loop. There isn't necessarily enough to know that we are the reason for the crash loop, and if we are then we probably want to be able to inspect the crashing pod to see why, right?

Yes and no, Most of the clusters that we are running on (Or at least I choose to think so) have some kind of monitoring enabled, so the logs do exist somewhere in some storage.

damemi:
Also, what happens if the pod is still crashlooping after the roll back? Or say a user tries to instrument a pod that is currently crashlooping, would this keep trying to roll that pod back?

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -51,6 +64,51 @@ func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.Instrumentation
...
+					Message: "Rolled back application instrumentation",
+				})
+
+				meta.SetStatusCondition(&ic.Status.Conditions, metav1.Condition{
+					Type:    odigosv1alpha1.AgentEnabledStatusConditionType,
+					Status:  metav1.ConditionFalse,
+					Reason:  string(odigosv1alpha1.AgentEnabledReasonCrashLoopBackOff),
+					Message: "Pods entered CrashLoopBackOff; instrumentation disabled",
+				})
+				return true, ctrl.Result{RequeueAfter: requeueWaitingForWorkloadRollout}, nil
```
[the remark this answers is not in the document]
TheErikNaz: We need to requeue because getting to the CrashLoopBackOff might take some time (I think it requires a min of 2 restarts)

instrumentor/controllers/agentenabled/sync.go
```diff
@@ -341,25 +350,45 @@ func containerInstrumentationConfig(containerName string,
...
+			DistroParams:   distroParameters,
+		}
+	} else {
+		return odigosv1.ContainerAgentConfig{
+			ContainerName:       containerName,
+			AgentEnabled:        false,
+			AgentEnabledReason:  odigosv1.AgentEnabledReasonCrashLoopBackOff,
+			AgentEnabledMessage: "Pods entered CrashLoopBackOff; instrumentation disabled",
+			OtelDistroName:      distroName,
+			DistroParams:        distroParameters,
```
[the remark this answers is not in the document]
TheErikNaz: Yeah, But i still want the IC to be constructed correctly

damemi (approved):

cli/cmd/config.go
```diff
@@ -41,6 +41,8 @@ var configCmd = &cobra.Command{
 	- "user-instrumentation-envs": JSON string defining per-language env vars to customize instrumentation, e.g., {"languages":{"java":{"enabled":true,"env":{"OTEL_INSTRUMENTATION_COMMON_EXPERIMENTAL_VIEW_TELEMETRY_ENABLED":"true"}}}}
 	- "agent-env-vars-injection-method": Method for injecting agent environment variables into the instrumented processes. Options include loader, pod-manifest and loader-fallback-to-pod-manifest.
 	- "node-selector": Apply a space-separated list of Kubernetes NodeSelectors to all Odigos components (ex: "kubernetes.io/os=linux mylabel=foo").
+	- "rollback-disabled": Disable auto rollback feature for failing instrumentations.
+	- "rollback-grace-time": Grace time before uninstrumenting an application [default: 3m].
```
RonFed: needs to be done in helm as well, 
BTW, default in the code is 5m

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -193,3 +276,169 @@ func rolloutCondition(rolloutErr error) metav1.Condition {
...
+}
+
+// timeSinceCrashLoopStart checks all Pods belonging to the given workload (Deployment, StatefulSet, or DaemonSet),
+// finds those that are in CrashLoopBackOff, and returns how much time has elapsed since the earliest‐started
+// crashing Pod was created. If no Pod is in CrashLoopBackOff, it returns (0, nil).
+//
+// Example return values:
+//   - If a Pod entered CrashLoopBackOff 2m30s ago, this will return ≈2m30s.
+//   - If multiple Pods are crashing, it returns the duration since the oldest Pod.StartTime among them.
+func timeSinceCrashLoopStart(ctx context.Context, c client.Client, obj client.Object) (time.Duration, error) {
```
RonFed: no need to duplicate this from `isInCrashLoopBackOff` - we can have one function called once in the above flow to return the time since the crashloop or 0 if no crashloop was dtected

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -49,11 +65,77 @@ func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.Instrumentation
...
+				return false, ctrl.Result{}, err
+			}
+
+			if crashLoopBackOff && ic.Spec.AgentInjectionEnabled {
+				// Allow grace time for worklaod to stabelize before uninstrumenting it
+				conf, err := k8sutils.GetCurrentOdigosConfig(ctx, c)
+				if err != nil {
+					return false, ctrl.Result{}, err
+				}
+				rollbackGraceTime := "5m"
```
RonFed: use as const

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -49,11 +65,77 @@ func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.Instrumentation
...
+					return false, ctrl.Result{}, fmt.Errorf("invalid duration %q: %w", rollbackGraceTime, err)
+				}
+
+				timeSinceCrashLoop, err := timeSinceCrashLoopStart(ctx, c, workloadObj)
+				if err != nil {
+					return false, ctrl.Result{}, err
+				}
+
+				if timeSinceCrashLoop < graceTime {
+					return false, ctrl.Result{RequeueAfter: requeueWaitingForWorkloadRollout}, nil
```
RonFed: I think the requeu here should be
`graceTime - timeSinceCrashLoop + <some small amount>`
since we know exactly when this condition will become true

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -49,11 +65,77 @@ func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.Instrumentation
 
 	savedRolloutHash := ic.Status.WorkloadRolloutHash
 	newRolloutHash := ic.Spec.AgentsMetaHash
-
 	if savedRolloutHash == newRolloutHash {
+		if !isWorkloadRolloutDone(workloadObj) && !rollbackDisabled {
+			crashLoopBackOff, err := isInCrashLoopBackOff(ctx, c, workloadObj)
```
RonFed: we can also get here the time since the crashloop and avoid extra API calls

cli/cmd/config.go
RonFed:
```suggestion
	- "instrumentation-auto-rollback-disabled": Disable auto rollback feature for failing instrumentations.
	- "instrumentation-auto-rollback-grace-time": Grace time before uninstrumenting an application [default: 3m].
```

cli/cmd/config.go
```diff
@@ -239,6 +243,12 @@ func setConfigProperty(config *common.OdigosConfiguration, property string, valu
 		}
 		config.NodeSelector = nodeSelectorMap
 
+	case consts.RollbackGraceTimeProperty:
+		if len(value) != 1 {
+			return fmt.Errorf("%s expects exactly one value", property)
+		}
+		config.RollbackGraceTime = value[0]
```
RonFed, line 250: I think we should parse the string here for a valid time using the `time` package and return an error here if the user provided invalid time.

instrumentor/controllers/agentenabled/rollout/rollout.go
TheErikNaz: That's only one of the cases.
If an app just waits for some dependency to start after instrumenting it might be much faster (Where it would balance out instead of uninstrumenting it)

cli/cmd/config.go
TheErikNaz: It would work in the CLI, but will break the helm definition (It'll require it to be manually calculated) I prefer to leave them symmetric

common/consts/consts.go
```diff
@@ -79,6 +79,8 @@ const (
 	UserInstrumentationEnvsProperty   = "user-instrumentation-envs"
 	NodeSelectorProperty              = "node-selector"
 	KarpenterEnabledProperty          = "karpenter-enabled"
+	RollbackDisabledProperty          = "rollback-disabled"
+	RollbackGraceTimeProperty         = "rollback-grace-time"
```
RonFed: should align to config flags description?
```
	- "instrumentation-auto-rollback-disabled": Disable auto rollback feature for failing instrumentations.
	- "instrumentation-auto-rollback-grace-time": Grace time before uninstrumenting an application [default: 5m]
```

api/odigos/v1alpha1/instrumentationconfig_types.go
```diff
@@ -153,6 +158,7 @@ func IsReasonStatusDisabled(reason string) bool {
 		string(AgentEnabledReasonIgnoredContainer),
 		string(AgentEnabledReasonNoAvailableAgent),
 		string(AgentEnabledReasonOtherAgentDetected),
+		string(AgentEnabledReasonCrashLoopBackOff),
```
blumamir, line 161: I think this will make it so that we don't see error in the UI

docs/instrumentations/overview.mdx
```diff
@@ -62,3 +62,5 @@ Select a language to learn how to enrich your data with OpenTelemetry APIs:
 5. Kubernetes detects the changes in the manifest, and rollout-restarts the pods with the new `Instrumentation Devices`.
 6. New pods are scheduled and started. Odiglet resolves the resource request by mounting the auto instrumentation code and relevant environment variables into the container.
 7. The auto instrumentation code starts the OpenTelemetry SDK and sends telemetry data to the odigos pipeline.
+
+In case of a failure to instrument a workload, Odigos will disable the instrumentation and rollback the workload, This behaviour can be disabled by running ```bash odigos config set rollback-disabled true```
```
blumamir, line 66: this assumes user is using cli. please also add doc for helm

cli/cmd/config.go
```diff
@@ -41,6 +41,8 @@ var configCmd = &cobra.Command{
 	- "user-instrumentation-envs": JSON string defining per-language env vars to customize instrumentation, e.g., {"languages":{"java":{"enabled":true,"env":{"OTEL_INSTRUMENTATION_COMMON_EXPERIMENTAL_VIEW_TELEMETRY_ENABLED":"true"}}}}
 	- "agent-env-vars-injection-method": Method for injecting agent environment variables into the instrumented processes. Options include loader, pod-manifest and loader-fallback-to-pod-manifest.
 	- "node-selector": Apply a space-separated list of Kubernetes NodeSelectors to all Odigos components (ex: "kubernetes.io/os=linux mylabel=foo").
+	- "instrumentation-auto-rollback-disabled": Disable auto rollback feature for failing instrumentations.
+	- "instrumentation-auto-rollback-grace-time": Grace time before uninstrumenting an application [default: 5m].
```
blumamir, line 45: add these flags also to install command

helm/odigos/values.yaml
```diff
@@ -251,3 +251,7 @@ nodeSelector: {}
 # https://docs.odigos.io/setup/odigos-with-karpenter
 karpenter:
   enabled: false
+
+autoRollback:
+  disabled: false
+  graceTime: 5m
```
blumamir, line 257: add documentation comments about these values like other values above.

A user that reads `autoRollback` without any context might not understand what this does

common/consts/consts.go
```diff
@@ -105,3 +107,7 @@ const (
 const (
 	KarpenterStartupTaintKey = "odigos.io/needs-init"
 )
+
+const (
+	AutoRollbackGraceTime = "5m"
```
blumamir, line 112:
```suggestion
	DefaultAutoRollbackGraceTime = "5m"
```

this can be overwritten by user

instrumentor/controllers/agentenabled/sync.go
```diff
@@ -46,8 +47,17 @@ func reconcileAll(ctx context.Context, c client.Client, dp *distros.Provider) (c
 		return ctrl.Result{}, listErr
 	}
 
+	conf, err := k8sutils.GetCurrentOdigosConfig(ctx, c)
+	rollbackDisabled := false
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
blumamir, line 54: more idiomatic: 

```suggestion
	if err != nil {
		return ctrl.Result{}, err
	}
	rollbackDisabled := false	
```

instrumentor/controllers/agentenabled/sync.go
```diff
@@ -145,9 +154,18 @@ func updateInstrumentationConfigSpec(ctx context.Context, c client.Client, pw k8
...
 
+	// If the source was already marked for instrumentation, but has caused a CrashLoopBackOff we'd like to stop
+	// instrumentating it and to disable future instrumentation of this service
+	cond := meta.FindStatusCondition(ic.Status.Conditions, odigosv1.AgentEnabledStatusConditionType)
+	crashDetected := false
+	if cond != nil {
+		if cond.Reason == string(odigosv1.AgentEnabledReasonCrashLoopBackOff) {
+			crashDetected = true
+		}
+	}
```
blumamir, line 165: nit: avoid deriving logic based on status conditions

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -193,3 +272,128 @@ func rolloutCondition(rolloutErr error) metav1.Condition {
...
+	if err := c.List(
+		ctx, &pods,
+		client.InNamespace(ns),
+		client.MatchingLabelsSelector{Selector: sel},
+	); err != nil {
+		return false, fmt.Errorf("workloadHasOdigosAgents: listing pods failed: %w", err)
+	}
+
+	// Because the cache already filters on k8sconsts.OdigosAgentsMetaHashLabel,
+	// any non-empty list means the workload still runs instrumented pods.
```
blumamir, line 397: controllers should not rely on cache logic for correctness. consider adding this selector explicitly which is more readable and robust.

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -40,6 +44,19 @@ func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.Instrumentation
 	}
 
 	if ic == nil {
+		// If ic is nil and the PodWorkload is missing the odigos.io/agents-meta-hash label,
+		// it means it is a rolled back application that shouldn't be rolled out again.
+		hasAgents, err := workloadHasOdigosAgents(ctx, c, workloadObj)
```
blumamir, line 49: it the pod starts (or the cache is updated) after we check this, there is not controller to handle the new pod.

ideally we should reconcile the pods also to avoid this (not sure if we want to do it now but we should be aware)

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -49,11 +66,72 @@ func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.Instrumentation
 
 	savedRolloutHash := ic.Status.WorkloadRolloutHash
 	newRolloutHash := ic.Spec.AgentsMetaHash
-
 	if savedRolloutHash == newRolloutHash {
+		if !isWorkloadRolloutDone(workloadObj) && !rollbackDisabled {
+			TimeSinceCrashLoopBackOff, err := crashLoopBackOffDuration(ctx, c, workloadObj)
```
blumamir, line 71:
```suggestion
			timeSinceCrashLoopBackOff, err := crashLoopBackOffDuration(ctx, c, workloadObj)
```

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -49,11 +66,72 @@ func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.Instrumentation
...
+			if err != nil {
+				logger.Error(err, "Failed to check crashLoopBackOff")
+				return false, ctrl.Result{}, err
+			}
+
+			if TimeSinceCrashLoopBackOff > 0 && ic.Spec.AgentInjectionEnabled {
+				// Allow grace time for worklaod to stabelize before uninstrumenting it
+				conf, err := k8sutils.GetCurrentOdigosConfig(ctx, c)
```
blumamir, line 79: this is an expensive operation and we already do it prior. consider reusing the object which we already have to get this value

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -49,11 +66,72 @@ func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.Instrumentation
...
+				rollbackGraceTime := consts.AutoRollbackGraceTime
+				if conf.RollbackGraceTime != "" {
+					rollbackGraceTime = conf.RollbackGraceTime
+				}
+				graceTime, err := time.ParseDuration(rollbackGraceTime)
+				if err != nil {
+					return false, ctrl.Result{}, fmt.Errorf("invalid duration %q: %w", rollbackGraceTime, err)
+				}
+
+				if TimeSinceCrashLoopBackOff < graceTime {
```
blumamir, line 92: The check MUST be from instrumentation time.
this feature is only meant to roll-back workloads after initial instrumentation. If we do it this way, any application crash will make us uninstrument a workload which is problematic.

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -49,11 +66,72 @@ func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.Instrumentation
...
+				}
+
+				for i := range ic.Spec.Containers {
+					ic.Spec.Containers[i].AgentEnabled = false
+					ic.Spec.Containers[i].AgentEnabledReason = odigosv1alpha1.AgentEnabledReasonCrashLoopBackOff
+				}
+				ic.Spec.AgentInjectionEnabled = false
+
+				if err := c.Update(ctx, ic); err != nil {
+					logger.Error(err, "failed to persist spec rollback")
```
blumamir, line 103: this should be removed.
if the error is "conflict" we don't want it in the log, and `K8SUpdateErrorHandler` will take care to retry.

If this is a real error, controller runtime will take care of logging

api/odigos/v1alpha1/instrumentationconfig_types.go
TheErikNaz: It sets the source as disabled, That is the wanted behavior

PR #2929, opened 2025-05-30
ui: Only show workload conditions that occur after instrumentation

damemi:
## Description

Follow up to https://github.com/odigos-io/odigos/pull/2871 and https://github.com/odigos-io/odigos/pull/2925

We currently combine all Workload conditions and InstrumentationInstance conditions, but this could show irrelevant Workload conditions in the UI and lead to incorrect errors. This change only returns Workload conditions that have occurred *after* the first Instrumentation instance. For example,

1. Workload has a failed rollout condition
2. Odigos instruments the app
3. Odigos rolls out a new replica
4. Workload rolls out a new replica

Currently, there is a period where the UI will show an error after step 2, even though we weren't even touching the workload at that point. After this change, we should not consider the previous error condition

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/frontend/services/sources.go
+++ b/frontend/services/sources.go
@@ -686,10 +686,26 @@ func GetOtherConditionsForSources(ctx context.Context, namespace string, name st
 	}
 	for _, workloadItem := range workloadsConditions {
 		key := fmt.Sprintf("%s/%s/%s", workloadItem.Namespace, workloadItem.Name, workloadItem.Kind)
-		if _, exists := conditionsMap[key]; !exists {
-			conditionsMap[key] = workloadItem
-		} else {
-			conditionsMap[key].Conditions = append(conditionsMap[key].Conditions, workloadItem.Conditions...)
+		if _, exists := conditionsMap[key]; exists {
+			for _, condition := range workloadItem.Conditions {
+				if condition.LastTransitionTime != nil {
+					// Only add conditions that are newer than the first instance condition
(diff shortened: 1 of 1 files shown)


BenElferink (changes requested):
As discussed in Slack:

This PR might cause problems because it's heavily relying on the instrumentation instances by:
- Assuming there is always an instance-condition (you removed the handling for `!exists`)
- Comparing with the time of an instance-condition (without handling a case where there is no instance-condition)

It is important to handle all workload conditions, without relying on the instrumentation instance conditions, because instances don't always exists - as far as I'm aware it's only for Python and maybe 1 more programming language.

The PR in current state will basically remove the workload conditions from majority of the sources in UI.

---

Also, TBH I'm not sure this needs much attention because:

- When workloads are already instrumented, then this case of "condition relevancy" wouldn't even happen, because the workloads aren't in rollout state.
- This "condition relevancy" is only valid for new instrumentations - and usually the customer would select a group/chunk of apps, instead of single apps; which in the UI we wait for 100% of instrumentation configs to be created before we start fetching them, and only after we fetched all then we get the "other conditions". This flow means that by the time the workload condition was fetched, the cluster is mostly stable, and if not then we would refetch some of those workloads as soon as a k8s-modified event was triggered.

damemi:
Closing for https://github.com/odigos-io/odigos/pull/2957

PR #2951, opened 2025-06-02
Implement k8sattributes action as generic Action CRD

damemi:
## Description

Following https://github.com/odigos-io/odigos/pull/2783, implementing the more complex k8sattributes action (which is a unified output of all k8sattributes actions)

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


BenElferink:
The UI still uses actions where each action is it's own CRD.
Should we sync later to get the "generic action" changes applied in the UI too?

damemi:
> The UI still uses actions where each action is it's own CRD. Should we sync later to get the "generic action" changes applied in the UI too?

Yup! This is just shifting each one, then we'll have a translation layer to deprecate the old CRDs (kind of like we did with Sources)

damemi:
Addressing Cursor comments for reference:

> The ActionReconciler and K8sAttributesResolverReconciler are both configured to create a Processor resource named "odigos-k8sattributes" in the same namespace. This results in a resource conflict and race condition, as both controllers attempt to manage the same object.

This is intentional, as we are migrating from the old individual K8sAttributes Action CRD to the new generic Action CRD, we want both to be supported. So, both controllers are reconciling different input Action objects to the same output processor.

However, I did have a bug where the legacy K8sAttributes action objects weren't actually being reconciled anymore to the unified processor. Because I update the `convertToUnifiedProcessor` function to call a shared sub-function `k8sAttributesConfig` (which is used by both the Generic Action and Legacy Action), I was not passing the `K8sAttributesResolverList` to this function anymore. I updated this by adding a small conversion function to convert the legacy Action list to the new Generic action so the sub function can be shared.

> The ActionReconciler now creates this processor for individual K8sAttributes actions, while the K8sAttributesResolverReconciler is designed to create a single, unified "odigos-k8sattributes" processor by aggregating all K8sAttributes configurations from Action resources.

This is essentially referring to the bug I mentioned above

PR #2966, opened 2025-06-04
Add size profiles to all Odigos components

damemi:
## Description

Currently the `size_*` profiles can be used to [configure the pipeline](https://github.com/odigos-io/odigos/blob/a990aea3eb7abb3ed75c4b108fe43b3177750e3f/docs/pipeline/configuration.mdx#1-using-profiles). This updates them to also configure the instrumentor, autoscaler, scheduler, and UI and adds a new `xsmall` size.

If multiple size profiles are configured, the largest one will be chosen.

The main changes are:

* checks in the resource managers for each component to determine the profile and set resource requests appropriate. Doing this through the resource manager allows us to automatically get this behavior through install, upgrade, and config CLI commands as well as the operator
* the scheduler reconciles size profiles by patching the component deployments. this allows editing the odigos-config configmap directly

In the future, we could add more fine-grained size profiles (such as `pipeline_size_s` and `component_size_s`, etc)

This also adds a docs page for all profiles

## How Has This Been Tested?

- [ ] Added Unit Tests
- [x] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


RonFed (changes requested):
* This requires helm charts changes as well.
* I'm not sure if we want to have the same sizing for instrumentor, scheduler and sutoscaler - we might want to provide different values for each and support configuring them.
* I think we should aim for a [guaranteed QoS](https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/#guaranteed) - which means having request=limit for memory and CPU.
* In general, I think we should allow users to have full control to configure these values and not enforce a closed set on them, as I think it can have a pretty big variance depending on the environment. We could have the "open" option and also the T-shirt options - but that might be too much options and might confuse. So my opinion is to only have "open" configuration for users both for resources and HPA with docs about it with our recommendations - but other would probably won't agree with that.

damemi:
Thanks @RonFed, I'll add the Helm chart changes

On your other points though, the intent here is to be simple. We currently set all of these components to the same values, this just gives a basic slider to scale that and doesn't take much thought to adjust.

Giving wide open config gives more freedom, but it also puts more thought on the user. We can definitely have that option, like we do already with the collector requests, but I think the outcome is actually the opposite -- most users won't want to fine tune their requests per component. Many people don't even fine tune their own apps on k8s.

So this PR just takes what we already have and gives a broad setting to adjust it. I think giving more wide open or per component config would be an extension of that which is going to be a pretty niche use case if none of these settings are what the user wants

tamirdavid1:
> Thanks @RonFed, I'll add the Helm chart changes
> 
> On your other points though, the intent here is to be simple. We currently set all of these components to the same values, this just gives a basic slider to scale that and doesn't take much thought to adjust.
> 
> Giving wide open config gives more freedom, but it also puts more thought on the user. We can definitely have that option, like we do already with the collector requests, but I think the outcome is actually the opposite -- most users won't want to fine tune their requests per component. Many people don't even fine tune their own apps on k8s.
> 
> So this PR just takes what we already have and gives a broad setting to adjust it. I think giving more wide open or per component config would be an extension of that which is going to be a pretty niche use case if none of these settings are what the user wants

I agree with the approach of having both tshirt sizing for most users but we have to give the ability to configure them manually and take priority over tshirt sizing profile

docs/profiles/available-profiles.mdx
```diff
@@ -0,0 +1,111 @@
...
+
+### Collector Resource Requirements
+
+The profiles also configure resource requirements for the OpenTelemetry collectors:
+
+**Cluster Gateway Collector**:
+
+| Profile | Min Replicas | Max Replicas | Memory Request | Memory Limit | CPU Request | CPU Limit |
+|---------|--------------|--------------|----------------|--------------|-------------|------------|
+| `size_xs` | 1 | 4 | 225Mi | 225Mi | 113m | 225m |
```
tamirdavid1, line 42: I feel like it is duplicate with 
https://docs.odigos.io/pipeline/configuration
is'nt ?

damemi: Yeah, I did duplicate it because I thought it would be nice to have the pipeline resources also spelled out in the pipeline section, but I see now that we don't really need that and it's just more docs to keep in sync. Will remove

damemi:
Thanks @tamirdavid1, Ron and I talked about it and I think what we'll do is add the t-shirt sizes + open config per component that takes priority just like the current collector settings

damemi:
@RonFed @tamirdavid1 I updated this to make the instrumentor, scheduler, autoscaler, and ui deployments follow the same pattern as the current collector options: odigos-config values take priority, otherwise use default values provided by the profile. Also added options to helm chart

For this, I also updated the current `modifySizingConfig` function to independently check for values from each component. Right now, it won't try to update the nodeCollector if there are any gatewayCollector settings (@tamirdavid1 not sure if this was intentional? please let me know)

I didn't add any config for odiglet yet until we are done benchmarking it, that can be a follow up I think

AvihuHenya (commented):
is there a reason why resource requests/limits helm templating isn't applied for the central-proxy as well?

blumamir:
@damemi now that the sizing config has been removed from the profile is expected to be set only in helm, we can come back to this PR and add this option.

You can see what @tamirdavid1 did for odiglet data-collection for reference. let's chat about it and deiced if we want to finalize it or close for now

PR #2972, opened 2025-06-05
feat: add support for optional OIDC/OAuth2 in UI

BenElferink:
## Description

This PR adds support for OIDC/OAuth2 in the UI, it will behave as following:
- If the user did not define OIDC variables, we will load the UI as normal.
- If the user did define OIDC variables, we will handle the OIDC flow to authenticate the user, and only then will we load the UI as normal.

## How Has This Been Tested?

- [x] Manual Testing

https://github.com/user-attachments/assets/67de8570-1d7b-4e67-b7d5-123ab3a4afa9


frontend/services/oidc.go
```diff
@@ -0,0 +1,264 @@
...
+	if idToken.Nonce != nonce {
+		c.JSON(http.StatusInternalServerError, gin.H{"nonce": "mismatch"})
+		return
+	}
+
+	setCallbackCookie(c.Writer, c.Request, "id_token", rawIDToken, "/", int(idToken.Expiry.Unix()-idToken.IssuedAt.Unix()))
+
+	c.Redirect(http.StatusFound, "/")
+
+	// Redact the access token for security reasons
```
aviadhahami: cleardis :D

BenElferink: Done :)

aviadhahami (approved):

PR #2971, opened 2025-06-05
Add application rollback for failed instrumentation 

TheErikNaz:
## Description
Added a feature to automatically rollback crashing instrumented applications.
The feature defines a stability windows where a decision to uninstrument can happen.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [x] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

- [ ] Changes how Odigos interacts with Kubernetes
- [x] Introduces additional calls to the API Server (potential performance impact)
- [ ] New Query/feature supported in all the k8s versions supported by Odigos
- [x] Modifies Odigos manifests (addressed in both CLI and Helm)
- [ ] Changes RBAC permissions

## User Facing Changes

- [ ] Users need to take action before upgrading
- [ ] Automatic migration will modify existing objects (backward compatible)
- [x] Changes UI, CLI, or K8s Manifests aspects in a way that users need to be aware of
- [x] Documentation updated accordingly

PR #2975, opened 2025-06-06
fix(instrumentor): aggregate result of reconcile all instead of breaking

blumamir:
## Description

#2971 introduced a change that after initial instrumentation, the agentenabled controller in instrumentor will requeue for 10 seconds to check for crashloop backoff.

This works well for 1 workload, but when called from `reconcileAll`, this change makes it so that we now break early from the loop before handling all workload, leading to agents being started one by one with 10 second intervals.

This PR makes it so that we always process all workloads, and aggregate the results from all calls to fix this issue

### DIFF
--- a/instrumentor/controllers/agentenabled/sync.go
+++ b/instrumentor/controllers/agentenabled/sync.go
@@ -2,6 +2,7 @@ package agentenabled
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


instrumentor/controllers/agentenabled/sync.go
```diff
@@ -52,14 +53,23 @@ func reconcileAll(ctx context.Context, c client.Client, dp *distros.Provider) (c
...
 	for _, ic := range allInstrumentationConfigs.Items {
-		res, err := reconcileWorkload(ctx, c, ic.Name, ic.Namespace, dp, &conf)
-		if err != nil || !res.IsZero() {
-			return res, err
+		res, workloadErr := reconcileWorkload(ctx, c, ic.Name, ic.Namespace, dp, &conf)
+		if workloadErr != nil {
+			allErrs = errors.Join(allErrs, workloadErr)
+		}
+		if !res.IsZero() {
+			if aggregatedResult.RequeueAfter == 0 {
```
RonFed, line 64: nit
```suggestion
			if aggregatedResult.IsZero() {
```

RonFed (approved):

PR #2983, opened 2025-06-08
feat: UI DataStreams toggle-on, add Namespace support

BenElferink:
## Description

Cannot merge yet, depends on: https://github.com/odigos-io/ui-kit/pull/164

- Turn on `Data Streams` feature in the UI.
- Bring back `Namespace Instrumentation` toggle.
- Add support for `Namespace Instrumentation` with `Data Streams` in the UI.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


frontend/services/namespaces.go
```diff
@@ -15,40 +15,44 @@ import (
...
 
-type GetNamespacesResponse struct {
-	Namespaces []model.K8sActualNamespace `json:"namespaces"`
-}
-
-func GetK8SNamespaces(ctx context.Context) (GetNamespacesResponse, error) {
-	ns := env.GetCurrentNamespace()
+func GetK8SNamespaces(ctx context.Context, namespaceName *string) ([]*model.K8sActualNamespace, error) {
+	var namespaces []corev1.Namespace
+	var response []*model.K8sActualNamespace
```
alonkeyval, line 26: consider splitting the GetK8SNamespaces function into two separate functions — one for retrieving a single namespace and one for retrieving all relevant namespaces

frontend/services/data_stream.go
```diff
@@ -54,6 +50,54 @@ func ExtractDataStreamsFromEntities(sources []v1alpha1.Source, destinations []v1
...
+// - The 'secondarySource' is expected to be a Namespace source.
+func ExtractDataStreamsFromSource(primarySource *v1alpha1.Source, secondarySource *v1alpha1.Source) []*string {
+	seen := make(map[string]bool)
+	forbidden := make(map[string]bool)
+	result := make([]*string, 0)
+
+	// Get all data stream names from the workload source
+	if primarySource != nil {
+		for labelKey, labelValue := range primarySource.Labels {
+			if strings.Contains(labelKey, k8sconsts.SourceGroupLabelPrefix) {
```
alonkeyval:
```suggestion
			if strings.HasPrefix(labelKey, k8sconsts.SourceGroupLabelPrefix) {
```

Using Contains could allow keys like abc/odigos.io/source-group/foo to pass

frontend/services/data_stream.go
```diff
@@ -54,6 +50,54 @@ func ExtractDataStreamsFromEntities(sources []v1alpha1.Source, destinations []v1
...
+				dsName := strings.TrimPrefix(labelKey, k8sconsts.SourceGroupLabelPrefix)
+
+				if labelValue == "false" {
+					forbidden[dsName] = true
+				}
+
+				if !seen[dsName] && !forbidden[dsName] {
+					seen[dsName] = true
+					result = append(result, &dsName)
```
alonkeyval: not sure, but I think passing the pointer like this will always points the last value of the for loop
```suggestion
ds := dsName
result = append(result, &ds)
```

frontend/services/data_stream.go
```diff
@@ -54,6 +50,54 @@ func ExtractDataStreamsFromEntities(sources []v1alpha1.Source, destinations []v1
 	return dataStreams
 }
 
+// ExtractDataStreamsFromSource extracts data stream names from the given primary and secondary sources.
+// It ensures that the data stream names are unique and that the 'false' labels are respected.
+// - The 'primarySource' is expected to be a Workload source (or a Namespace source when alone).
+// - The 'secondarySource' is expected to be a Namespace source.
```
alonkeyval, line 56: I think the comment here is unclear, I think we need to add here the priority system between the source and namespace object

frontend/services/data_stream.go
```diff
@@ -144,58 +188,54 @@ func UpdateDestinationsCurrentStreamName(ctx context.Context, destinations *v1al
...
+				if strings.TrimPrefix(labelKey, k8sconsts.SourceGroupLabelPrefix) == currentStreamName {
 					// remove the old label
 					_, err := UpdateSourceCRDLabel(ctx, source.Namespace, source.Name, k8sconsts.SourceGroupLabelPrefix+currentStreamName, "")
 					if err != nil {
 						return fmt.Errorf("failed to update source %s: %v", source.Name, err)
 					}
 
 					// add the new label
-					_, err = UpdateSourceCRDLabel(ctx, source.Namespace, source.Name, k8sconsts.SourceGroupLabelPrefix+newStreamName, "true")
+					_, err = UpdateSourceCRDLabel(ctx, source.Namespace, source.Name, k8sconsts.SourceGroupLabelPrefix+newStreamName, labelValue)
```
alonkeyval: nit:
```suggestion
dataStreamLabelKey := k8sconsts.SourceGroupLabelPrefix+newStreamName
_, err = UpdateSourceCRDLabel(ctx, source.Namespace, source.Name, dataStreamLabelKey, labelValue)
```

* same in remove the old label on top

frontend/services/sources.go
```diff
@@ -310,65 +308,54 @@ func EnsureSourceCRD(ctx context.Context, nsName string, workloadName string, wo
...
-
-	if nsSource != nil {
-		// namespace source exists.
-		// we need to create a workload source and add "DisableInstrumentation" label,
-		// or remove the relevant data-stream label (if source is in multiple streams)
-
-		// note: create will also return an existing crd (if exists) without throwing an error
-		source, err := EnsureSourceCRD(ctx, nsName, workloadName, workloadKind, currentStreamName)
-		if err != nil {
+		nsSource = source
```
alonkeyval: in this case we pass ExtractDataStreamsFromSource the same source?

frontend/services/sources.go
```diff
@@ -310,65 +308,54 @@ func EnsureSourceCRD(ctx context.Context, nsName string, workloadName string, wo
...
 
-		dataStreamNames := GetSourceDataStreamNames(source)
+	dataStreamNames := ExtractDataStreamsFromSource(source, nsSource)
+	isWorkloadWithNamespace := workloadKind != WorkloadKindNamespace && nsSource != nil
 
-		if len(dataStreamNames) > 1 && currentStreamName != "" {
+	// we remove the current data-stream
+	if currentStreamName != "" {
+		if isWorkloadWithNamespace {
+			_, err = UpdateSourceCRDLabel(ctx, nsName, source.Name, k8sconsts.SourceGroupLabelPrefix+currentStreamName, "false")
```
alonkeyval: use as var(k8sconsts.SourceGroupLabelPrefix+currentStreamName) and use in both UpdateSourceCRDLabel

frontend/graph/schema.resolvers.go
```diff
@@ -152,11 +135,27 @@ func (r *computePlatformResolver) Sources(ctx context.Context, obj *model.Comput
...
 			if err != nil && !apierrors.IsNotFound(err) {
 				return err
 			}
-			srcList[idxCopy] = src
+
+			// Get namespace source CRD
+			val, exists := nsSources.Load(icCopy.Namespace)
+			var nsSource *v1alpha1.Source
+			if exists {
+				nsSource = val.(*v1alpha1.Source)
```
alonkeyval: val.(*v1alpha1.Source) can panics if val is nil or the wrong type

```suggestion
				source, ok := val.(*v1alpha1.Source)
    if !ok {
        return fmt.Errorf("invalid value in nsSources map")
    }
    nsSource = source
```

alonkeyval:
frontend side looks good

frontend/services/namespaces.go
BenElferink: It was already kind of "split" in schema resolvers.
I purposely removed the code from there and made this function work for both scenarios to prevent duplicate/copied code.

frontend/services/data_stream.go
BenElferink: It worked perfectly fine during all of my tests, but OK will change anyway

frontend/services/data_stream.go
BenElferink: wdym?

frontend/services/sources.go
BenElferink: Yes but it doesn't change anything, it would simple get the labels once.
The reason it's written like this is to keep the code clean and short as possible, previously the whole thing was copied and pasted in if/else.

alonkeyval (approved):

PR #2988, opened 2025-06-09
chore: generalise GetSource func

tamirdavid1:
## Description
This function can be used for any PodWorkload and it is not really need to get a  generic `client.Object`.
We use all over Odigos the PodWorkload object.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/api/odigos/v1alpha1/source_types.go
+++ b/api/odigos/v1alpha1/source_types.go
@@ -112,21 +112,21 @@ type SourceSelector struct {
 // GetSources returns a WorkloadSources listing the Workload and Namespace Source
 // that currently apply to the given object. In theory, this should only ever return at most
 // 1 Namespace and/or 1 Workload Source for an object. If more are found, an error is returned.
-func GetSources(ctx context.Context, kubeClient client.Client, obj client.Object) (*WorkloadSources, error) {
+func GetSources(ctx context.Context, kubeClient client.Client, workloadObj k8sconsts.PodWorkload) (*WorkloadSources, error) {
 	var err error
 	workloadSources := &WorkloadSources{}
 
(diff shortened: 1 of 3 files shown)


api/odigos/v1alpha1/source_types.go
```diff
@@ -112,21 +112,21 @@ type SourceSelector struct {
 // GetSources returns a WorkloadSources listing the Workload and Namespace Source
 // that currently apply to the given object. In theory, this should only ever return at most
 // 1 Namespace and/or 1 Workload Source for an object. If more are found, an error is returned.
-func GetSources(ctx context.Context, kubeClient client.Client, obj client.Object) (*WorkloadSources, error) {
+func GetSources(ctx context.Context, kubeClient client.Client, workloadObj k8sconsts.PodWorkload) (*WorkloadSources, error) {
```
blumamir: nit: we are not passing the object here so the name `workloadObj` can be refined.

In other places we call it `pw` (shortcut to podWorkload)

tamirdavid1:
> I'm a little confused why the change? The places you've update where this function is called seem to all be using a `client.Object` anyway to fill out the `PodWorkload` struct. Are there other places where we are already using PodWorkload that you want to use GetSources?

Hi Mike, you right i had to clarify this better on a comment.
I wanted to use getSource return object to calculate datastreams config in a place where the original function is called.
i'll tag you on the next PR so it will make more sense for you

damemi:
@tamirdavid1 sounds good thanks for explaining! I have no problem with the change just wanted to know the context

PR #2989, opened 2025-06-09
helm chart: add component resource configuration

damemi:
## Description

Subset of changes from https://github.com/odigos-io/odigos/pull/2966, this allows configuring resources through the helm chart for autoscaler, instrumentor, odiglet, scheduler, and ui. It follows the pattern that already exists for central proxy, and only sets odiglet resources if values are set (to match the current default of no resources set)

Follow ups to this will add config support through other methods (cli/operator), but those need to be done in a way that doesn't conflict with helm.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


AvihuHenya (approved):
🥇

PR #2991, opened 2025-06-10
Reconcile sources to instrumentationconfig

tamirdavid1:
## Description

In the Datastreams world we agreed to use the instrumenationConfig as the single point of truth for building the configuration and also for the UI to consume.
This PR ensure that this data will reconcile into the  instrumenationConfig for each workload with the workload source and also the namespace source.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/k8sutils/pkg/source/source.go
+++ b/k8sutils/pkg/source/source.go
@@ -132,19 +132,8 @@ func HandleInstrumentationConfigDataStreamsLabels(ctx context.Context,
 	}
 
 	// Extract labels from both sources (may be nil)
-	var workloadLabels map[string]string
-	if workloadSource != nil {
-		workloadLabels = getSourceDataStreamsLabels(workloadSource)
-	} else {
-		workloadLabels = map[string]string{}
-	}
-
-	var namespaceLabels map[string]string
-	if namespaceSource != nil {
-		namespaceLabels = getSourceDataStreamsLabels(namespaceSource)
-	} else {
-		namespaceLabels = map[string]string{}
-	}
+	workloadLabels := getSourceDataStreamsLabels(workloadSource)
+	namespaceLabels := getSourceDataStreamsLabels(namespaceSource)
 
 	// Start merging logic:
 	mergedLabels := make(map[string]string)
@@ -169,10 +158,10 @@ func HandleInstrumentationConfigDataStreamsLabels(ctx context.Context,
 	}
 
 	// Apply merged labels into InstrumentationConfig, and return true if changed
-	return mergeInstrumentationConfigLabels(ic, mergedLabels)
+	return setInstrumentationConfigDataStreamLabels(ic, mergedLabels)
 }
 
-func mergeInstrumentationConfigLabels(instConfig *odigosv1.InstrumentationConfig, desiredLabels map[string]string) (updated bool) {
+func setInstrumentationConfigDataStreamLabels(instConfig *odigosv1.InstrumentationConfig, desiredLabels map[string]string) (updated bool) {
 	if instConfig.Labels == nil {
 		instConfig.Labels = make(map[string]string)
 	}
@@ -205,8 +194,12 @@ func isDataStreamLabel(labelKey string) bool {
 func getSourceDataStreamsLabels(source *odigosv1.Source) map[string]string {
 	result := make(map[string]string)
 
+	if source == nil {
+		return result
+	}
+
 	for k, v := range source.Labels {
-		if strings.HasPrefix(k, k8sconsts.SourceDataStreamLabelPrefix) {
+		if isDataStreamLabel(k) {
 			result[k] = v
 		}
 	}
--- a/k8sutils/pkg/source/source.go
+++ b/k8sutils/pkg/source/source.go
@@ -3,6 +3,7 @@ package source
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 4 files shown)


k8sutils/pkg/source/source.go
```diff
@@ -123,3 +117,99 @@ func GetClientObjectFromSource(ctx context.Context, kubeClient client.Client, so
...
+		}
+	}
+
+	// Remove datastream labels not present in desiredLabels
+	for key := range instConfig.Labels {
+		if _, exists := desiredLabels[key]; !exists && isDataStreamLabel(key) {
+			delete(instConfig.Labels, key)
+			updated = true
+		}
+	}
```
BenElferink, line 162: Right now IC labels are only for DS, so it makes sense for now.
But if we add non-DS labels in the future, this will remove all of them.

I think it's best not to remove any labels.
The `desiredLabels` contains DS labels with `false` values, I think that's the only kind of "removal" we would need for now.

```suggestion
        // do not remove
```

k8sutils/pkg/source/source.go
```diff
@@ -123,3 +117,99 @@ func GetClientObjectFromSource(ctx context.Context, kubeClient client.Client, so
...
+func isDataStreamLabel(labelKey string) bool {
+	return strings.HasPrefix(labelKey, k8sconsts.SourceDataStreamLabelPrefix)
+}
+
+// GetSourceDataStreamsLabels extracts only datastream labels from the Source object.
+func getSourceDataStreamsLabels(source *odigosv1.Source) map[string]string {
+	result := make(map[string]string)
+
+	for k, v := range source.Labels {
+		if strings.HasPrefix(k, k8sconsts.SourceDataStreamLabelPrefix) {
```
blumamir:
```suggestion
		if isDataStreamLabel(k) {
```

k8sutils/pkg/source/source.go
```diff
@@ -123,3 +117,99 @@ func GetClientObjectFromSource(ctx context.Context, kubeClient client.Client, so
...
+			mergedLabels[k] = v
+		}
+	}
+
+	// Apply overrides from workload source (force false where set)
+	for k, v := range workloadLabels {
+		if v == "false" {
+			mergedLabels[k] = "false"
+		}
+	}
```
blumamir: nit: I think this can be simplified by starting with namespace labels and then overriding them with workload labels. let's talk in person

k8sutils/pkg/source/source.go
```diff
@@ -123,3 +117,99 @@ func GetClientObjectFromSource(ctx context.Context, kubeClient client.Client, so
...
+
+	// Case 1: no workload source and no namespace source — nothing to do
+	if workloadSource == nil && namespaceSource == nil {
+		logger.Info("No workload source and no namespace source found, skipping")
+		return false
+	}
+
+	// Extract labels from both sources (may be nil)
+	var workloadLabels map[string]string
+	if workloadSource != nil {
```
blumamir: nit: consider handling `nil` in the `getSourceDataStreamsLabels` so this function is simpler

k8sutils/pkg/source/source.go
```diff
@@ -123,3 +117,99 @@ func GetClientObjectFromSource(ctx context.Context, kubeClient client.Client, so
...
+func HandleInstrumentationConfigDataStreamsLabels(ctx context.Context,
+	workloadSources *odigosv1.WorkloadSources, ic *odigosv1.InstrumentationConfig) bool {
+	logger := log.FromContext(ctx)
+
+	workloadSource := workloadSources.Workload
+	namespaceSource := workloadSources.Namespace
+
+	// Case 1: no workload source and no namespace source — nothing to do
+	if workloadSource == nil && namespaceSource == nil {
+		logger.Info("No workload source and no namespace source found, skipping")
```
blumamir: do we expect this to happen? if not should this be an error?

k8sutils/pkg/source/source.go
```diff
@@ -123,3 +117,99 @@ func GetClientObjectFromSource(ctx context.Context, kubeClient client.Client, so
...
+		if v == "false" {
+			mergedLabels[k] = "false"
+		}
+	}
+
+	// Apply merged labels into InstrumentationConfig, and return true if changed
+	return mergeInstrumentationConfigLabels(ic, mergedLabels)
+}
+
+func mergeInstrumentationConfigLabels(instConfig *odigosv1.InstrumentationConfig, desiredLabels map[string]string) (updated bool) {
```
blumamir:
```suggestion
func setInstrumentationConfigDataStreamLabels(instConfig *odigosv1.InstrumentationConfig, desiredLabels map[string]string) (updated bool) {
```

blumamir (approved):

k8sutils/pkg/source/source.go
tamirdavid1: Thank for the comment!
i'm checking in this function `isDataStreamLabel(key)`
so i wont delete key if it is no Datastream label

k8sutils/pkg/source/source.go
tamirdavid1: It should not happen due to this code:
```
	sources, err := odigosv1.GetSources(ctx, k8sClient, pw)
	enabled, markedForInstrumentationCondition, err := sourceutils.IsObjectInstrumentedBySource(ctx, sources, err)
	if err != nil {
		return ctrl.Result{}, err
	}
```

PR #2994, opened 2025-06-10
Helm topologySpreadConstraints configuration

AvihuHenya:
## Description
This PR enables `topologySpreadConstraints` configuration for all relevant Odigos `Deployment` components, allowing the user to control how Odigos pods are distributed across different topology domains.

This configuration is enabled via the `.values` file, and shares the same constraints configuration across all components.
In the Helm templating process, we add the specific app label and apply it into `spec.template.spec`, resulting in this configuration being applied at the component level.

## How Has This Been Tested?
I've tested it with two scenarios using my kind cluster:

1) Basic setup: control-plane (with taint) + worker  
    Applied:  
    `kubectl label node kind-control-plane topology.kubernetes.io/zone=zone-a`  
    `kubectl label node kind-worker topology.kubernetes.io/zone=zone-b`

    Installed Odigos via Helm with the following config:
    ```yaml
    topologySpread:
      enabled: true
      constraints:
        - maxSkew: 1
          topologyKey: "topology.kubernetes.io/zone"
          whenUnsatisfiable: "DoNotSchedule"
    ```

    Screenshot 2025-06-10 at 13 41 27

2) 3 nodes (control-plane + 2 workers)  
    Applied:  
    `kubectl label node kind-worker topology.kubernetes.io/zone=zone-a`  
    `kubectl label node kind-worker2 topology.kubernetes.io/zone=zone-b`
     Installed Odigos via Helm with thesame configuration as above, and updated instrumentor replicas to 6 (for testing), resulting in instrumentor pods being scheduled evenly across the topology zones (as shown in image)

    Screenshot 2025-06-10 at 14 38 32


helm/odigos/values.yaml
```diff
@@ -252,6 +252,17 @@ nodeSelector: {}
...
+
+topologySpread:
+  enabled: false
+  constraints:
+    - maxSkew: 1
+      topologyKey: "topology.kubernetes.io/zone"
+      whenUnsatisfiable: "ScheduleAnyway"
+      labelSelector:
+        matchLabels:
+          app.kubernetes.io/instance: odigos
```
damemi: is the labelSelector value used anywhere? Looks like each deployment template just overwrites with its own app selector

AvihuHenya: Good catch, thanks!
That was my old way of injecting the labels.
I added docs too.

autoscaler/controllers/clustercollector/deployment.go
```diff
@@ -43,14 +44,20 @@ func syncDeployment(dests *odigosv1.DestinationList, gateway *odigosv1.Collector
 	}
 	odigletNodeSelector := odigletDaemonset.Spec.Template.Spec.NodeSelector
 
+	InstrumentorDeployment := &appsv1.Deployment{}
+	if err := c.Get(ctx, client.ObjectKey{Namespace: gateway.Namespace, Name: k8sconsts.AutoScalerDeploymentName}, InstrumentorDeployment); err != nil {
+		return nil, err
+	}
+	autoScalerTopologySpreadConstraints := InstrumentorDeployment.Spec.Template.Spec.TopologySpreadConstraints
```
damemi:
```suggestion
	autoscalerDeployment := &appsv1.Deployment{}
	if err := c.Get(ctx, client.ObjectKey{Namespace: gateway.Namespace, Name: k8sconsts.AutoScalerDeploymentName}, autoscalerDeployment); err != nil {
		return nil, err
	}
	autoScalerTopologySpreadConstraints := autoscalerDeployment.Spec.Template.Spec.TopologySpreadConstraints
```

AvihuHenya:
close this PR to make the configuration aligns with community conventions. updates soon.

PR #2931, opened 2025-06-01
feat: enable metrics server in Odiglet when not using hostNetwork. Add MeterProvider for custom metrics. Add ebpf manager metrics.

RonFed:
## Description

Following #2928, we can safely expose metrics server in the odiglet. For k8s versions >= 1.26 we are running without `hostNetwork` - and we can bind to a const and known port without handling conflicts. Using the default metrics port used by controller-runtime.
controller-runtime has a builtin metrics server using prometheus - it includes metrics for all the controllers and the go runtime.
This PR **does not aim** to scrape the metrics, and this is left for follow ups. The goal in this PR is to setup the metrics server and add `MeterProvider` for custom metrics.
We are setting up an OTel `MetreProvider`  that is configured to export in prometheus format and integrates with the prometheus metric registry exposed by controller-runtime.
The `MeterProvider` can be passed to different components in the Odiglet, each can declare its own specific metrics.
Adding the `MeterProvider` to the eBPF instrumentation manager as an initial step - including 2 metrics: first for the number of currently instrumented processes, and the second for the number of processes we failed to instrument.
In the future the `MeterProvider` can be passed and used by the `OpAmp` server, the `runtime-detector` and the different instrumentations libraries - each can add specific metrics which will end up in the metrics server.

```bash
kubectl port-forward -n odigos-system odiglet-swwwh 8080:8080
...

# odigos specific metrics
curl http://localhost:8080/metrics | grep odigos

# HELP odigos_ebpf_instrumentation_manager_failed_instrumentations_total Number of processes that failed to be instrumented
# TYPE odigos_ebpf_instrumentation_manager_failed_instrumentations_total counter
odigos_ebpf_instrumentation_manager_failed_instrumentations_total{otel_scope_name="github.com/odigos.io/odigos/instrumentation",otel_scope_version="v0.1.0"} 1
# HELP odigos_ebpf_instrumentation_manager_instrumented_processes Number of processes currently instrumented
# TYPE odigos_ebpf_instrumentation_manager_instrumented_processes gauge
odigos_ebpf_instrumentation_manager_instrumented_processes{otel_scope_name="github.com/odigos.io/odigos/instrumentation",otel_scope_version="v0.1.0"} 3
otel_scope_info{otel_scope_name="github.com/odigos.io/odigos/instrumentation",otel_scope_version="v0.1.0"} 1


# controller metrics - only showing a few here
curl http://localhost:8080/metrics | grep controller
[...]

PR #2996, opened 2025-06-10
operator: bump to v1.0.196

damemi:
## Description

This makes a few updates to fix changes:

* https://github.com/odigos-io/odigos/pull/2931 caused the odiglet to crashloop due to port in use. Debugging this showed that the k8s(openshift) cluster version was not being propagated through the context used by the operator to the cli install commands. So, hostnetwork was being set to `true`. The cli parses this cluster info by default through the local kubecontext, when in cluster we want to use the k8s server version. Updated operator to pass k8s version to install.
* Python/Java/JS apps were failing with the USDT 10 retries error. This turned out to be because the agent environment variables weren't injected by the webhook. Looking at the Instrumentor (pod webhook) logs, the agent injection method variable was not set in the Odigos config. Update the operator CRD to allow configuring this value and add a default
* Digging further, the Scheduler was not able to set a default value in the odigos configmap because it couldn't make any changes to it due to https://github.com/odigos-io/odigos/pull/2861 adding an ownerreference, which requires extra permissions in OpenShift (similar to the [instrumentor](https://github.com/odigos-io/odigos/blob/6837ba3cfcfa4f3c0415afde8b33e5c9ee08c65f/helm/odigos/templates/instrumentor/clusterrole.yaml#L62-L71))

```
2025-06-12T15:20:34Z	ERROR	Reconciler error	{"controller": "odigosconfig-odigosconfig", "controllerGroup": "", "controllerKind": "ConfigMap", "ConfigMap": {"name":"odigos-config","namespace":"odigos-operator-system"}, "namespace": "odigos-operator-system", "name": "odigos-config", "reconcileID": "1440478b-1d4f-48d3-a8ce-63dbc4c933b1", "error": "configmaps \"effective-config\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:347
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:294
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:255
```

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


BenElferink:
Yesterday released `v1.0.195`, and today we need to release `v1.0.196`.
Should this PR make the jump to .196 instead of .193?

BenElferink (approved):

damemi:
@BenElferink updated to 196, and also had to fix a couple bugs in the operator on openshift.

Note that we should merge these operator fixes now, but they won't be able to be published in the operatorhub until 197 is built/released with those fixes in place

PR #2999, opened 2025-06-11
feat: add workload rollout restarts to UI

BenElferink:
## Description

This PR adds the option to rollout restart workloads from the UI.
Depends on: https://github.com/odigos-io/ui-kit/pull/168

## How Has This Been Tested?

- [x] Manual Testing

### DIFF
--- a/frontend/graph/schema.resolvers.go
+++ b/frontend/graph/schema.resolvers.go
@@ -1225,15 +1225,3 @@ type computePlatformResolver struct{ *Resolver }
 type k8sActualNamespaceResolver struct{ *Resolver }
 type mutationResolver struct{ *Resolver }
 type queryResolver struct{ *Resolver }
-
-// !!! WARNING !!!
-// The code below was going to be deleted when updating resolvers. It has been copied here so you have
-// one last chance to move it out of harms way if you want. There are two reasons this happens:
-//  - When renaming or deleting a resolver the old code will be put in here. You can safely delete
-//    it when you're done.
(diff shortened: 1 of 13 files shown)


frontend/graph/schema.resolvers.go
```diff
@@ -1205,3 +1225,15 @@ type computePlatformResolver struct{ *Resolver }
...
+// The code below was going to be deleted when updating resolvers. It has been copied here so you have
+// one last chance to move it out of harms way if you want. There are two reasons this happens:
+//  - When renaming or deleting a resolver the old code will be put in here. You can safely delete
+//    it when you're done.
+//  - You have helper methods in this file. Move them out to keep these resolver files clean.
+/*
+	func (r *mutationResolver) RestartWorkload(ctx context.Context, sourceID model.K8sSourceID) (bool, error) {
+
+}
+*/
```
damemi: should this be removed?

BenElferink: Yes thank you!

BenElferink:
> what's the default mode? Today it's to restart the workloads right, will that still be the case?

When we instrument and uninstrument, the CLI already does a rollout.

Sometimes the user needs a manual rollout (maybe a pod is hung on some old error etc.), so this PR adds the option to restart workloads from the UI - on a button click.

---

Maybe the auto-rollout triggered by CLI will change in the near future, but that is unrelated to this change :)

damemi:
@BenElferink ahh thank you, I misunderstood

PR #3004, opened 2025-06-12
Initial commit to allow instrumenting cronjobs

TheErikNaz:
## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/instrumentor/controllers/agentenabled/rollout/rollout.go
+++ b/instrumentor/controllers/agentenabled/rollout/rollout.go
@@ -48,7 +48,7 @@ func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.Instrumentation
 			Type:    odigosv1alpha1.WorkloadRolloutStatusConditionType,
 			Status:  metav1.ConditionTrue,
 			Reason:  string(odigosv1alpha1.WorkloadRolloutReasonWaitingForRestart),
-			Message: "Waiting to job to trigger on itself",
+			Message: "Waiting for job to trigger by itself",
 		})
 		return statusChanged, ctrl.Result{}, nil
 	}
--- a/instrumentor/controllers/sourceinstrumentation/conditions.go
+++ b/instrumentor/controllers/sourceinstrumentation/conditions.go
@@ -45,14 +45,30 @@ func initiateRuntimeDetailsConditionIfMissing(ic *v1alpha1.InstrumentationConfig
 		return true
 	}
 
-	ic.Status.Conditions = append(ic.Status.Conditions, metav1.Condition{
-		Type:               v1alpha1.RuntimeDetectionStatusConditionType,
-		Status:             metav1.ConditionUnknown,
-		Reason:             string(v1alpha1.RuntimeDetectionReasonWaitingForDetection),
-		Message:            "Waiting for odiglet to initiate runtime detection in a node with running pod",
-		LastTransitionTime: metav1.NewTime(time.Now()),
-	})
-
+	isCronJob := false
+	for _, ref := range ic.OwnerReferences {
+		if ref.Controller != nil && *ref.Controller && ref.Kind == "CronJob" {
+			isCronJob = true
+			break
+		}
+	}
+	if isCronJob {
+		ic.Status.Conditions = append(ic.Status.Conditions, metav1.Condition{
+			Type:               v1alpha1.RuntimeDetectionStatusConditionType,
+			Status:             metav1.ConditionUnknown,
+			Reason:             string(v1alpha1.RuntimeDetectionReasonWaitingForDetection),
+			Message:            "Runtime detection pending Job to start",
+			LastTransitionTime: metav1.NewTime(time.Now()),
+		})
+	} else {
+		ic.Status.Conditions = append(ic.Status.Conditions, metav1.Condition{
+			Type:               v1alpha1.RuntimeDetectionStatusConditionType,
+			Status:             metav1.ConditionUnknown,
+			Reason:             string(v1alpha1.RuntimeDetectionReasonWaitingForDetection),
+			Message:            "Waiting for odiglet to initiate runtime detection in a node with running pod",
+			LastTransitionTime: metav1.NewTime(time.Now()),
+		})
+	}
 	return true
 }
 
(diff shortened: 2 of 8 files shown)


cli/cmd/resources/instrumentor.go
```diff
@@ -178,6 +178,11 @@ func NewInstrumentorClusterRole(ownerPermissionEnforcement bool) *rbacv1.Cluster
 				Resources: []string{"pods"},
 				Verbs:     []string{"get", "list", "watch"},
 			},
+			{ // Read instrumentation labels from statefulsets and apply pod spec changes
+				APIGroups: []string{"batch"},
+				Resources: []string{"cronjobs"},
+				Verbs:     []string{"get", "list", "watch"},
+			},
```
damemi, line 185: does it need `job` permissions too?

damemi:
Does a CronJob create Pods directly, or does it create a Job that owns the Pods? I don't recall

damemi:
Can you also add support to `odigos sources enable/disable`? https://github.com/odigos-io/odigos/blob/8e58bdeafd5f6410d89b35bf8561db872fec7977/cli/cmd/sources.go#L565-L582

cli/cmd/resources/instrumentor.go
TheErikNaz: No we don't.
Since were listing only CronJob (currently we won't and can't support regular jobs).

TheErikNaz:
> Does a CronJob create Pods directly, or does it create a Job that owns the Pods? I don't recall

A cron jobs schedules a job which in turn schedules a pod

instrumentor/controllers/sourceinstrumentation/conditions.go
```diff
@@ -32,7 +33,8 @@ func initiateRuntimeDetailsConditionIfMissing(ic *v1alpha1.InstrumentationConfig
 	}
 
 	// if the workload has no available replicas, we can't detect the runtime
-	if workloadObj.AvailableReplicas() == 0 {
+	if workloadObj.AvailableReplicas() == 0 &&
```
blumamir, line 36: what will be the initial message on these sources until we are able to detect the runtime for this job?

should we have some other initial message to indicate what is going on?

instrumentor/controllers/sourceinstrumentation/instrumentationconfig_controller.go
```diff
@@ -55,6 +56,12 @@ func getObjectByOwnerReference(ctx context.Context, k8sClient client.Client, own
 		return ss, err
 	}
 
+	if ownerRef.Kind == "CronJob" {
```
blumamir, line 59: Do we need to also check kind "Job" here?

k8sutils/pkg/container/container.go
```diff
@@ -112,13 +124,23 @@ func AllContainersReady(pod *v1.Pod) bool {
...
 		containerStatus := &pod.Status.ContainerStatuses[i]
-		if !containerStatus.Ready || containerStatus.Started == nil || !*containerStatus.Started {
+
+		if !containerStatus.Ready {
+			return false
+		}
+
+		// For long-running pods (RestartPolicy=Always) ensure the container
+		// has actually entered the running state (`Started == true`).
+		if !skipStarted && containerStatus.Started != nil && !*containerStatus.Started {
```
blumamir: if `containerStatus.Started == nil`, it's considered as the container did not started, but this function will not `return false` (was like that before the change).

Is it intentional?

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -43,6 +43,16 @@ func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.Instrumentation
 		return false, ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	if pw.Kind == k8sconsts.WorkloadKindCronJob || pw.Kind == k8sconsts.WorkloadKindJob {
+		statusChanged := meta.SetStatusCondition(&ic.Status.Conditions, metav1.Condition{
+			Type:    odigosv1alpha1.WorkloadRolloutStatusConditionType,
+			Status:  metav1.ConditionTrue,
+			Reason:  string(odigosv1alpha1.WorkloadRolloutReasonWaitingForRestart),
+			Message: "Waiting to job to trigger on itself",
```
BenElferink:
```suggestion
			Message: "Waiting for job to trigger by itself",
```

BenElferink (approved):

instrumentor/controllers/sourceinstrumentation/instrumentationconfig_controller.go
TheErikNaz: No, Since we do not support instrumenting regular jobs

k8sutils/pkg/container/container.go
```diff
@@ -103,6 +103,18 @@ func podContainerDeviceName(container *v1.Container) *string {
 	return nil
 }
 
+// isCronJobPod returns true if the Pod is ultimately controlled by a CronJob.
+// In practice it’s enough to check for a Job controller: only Jobs/CronJobs
+// produce Pods with Started == nil.
+func isCronJobPod(pod *v1.Pod) bool {
+	for _, ref := range pod.OwnerReferences {
+		if ref.Controller != nil && *ref.Controller && (ref.Kind == "Job" || ref.Kind == "CronJob") {
```
blumamir, line 111: will the pod ever have "CronJob" as the owner controller?

k8sutils/pkg/workload/ownerreference.go
```diff
@@ -59,9 +59,23 @@ func GetWorkloadNameAndKind(ownerName, ownerKind string) (string, k8sconsts.Work
 	if ownerKind == "ReplicaSet" {
 		return extractDeploymentInfo(ownerName)
 	}
+	// A CronJobs owner is a Job kind, but since we're not supporting instrumenting (at a source level) regular Jobs, this will suffice
+	if ownerKind == "CronJob" || ownerKind == "Job" {
```
blumamir, line 63: Here also - the owner name is expected to be "Job" but can it be "CronJob"?

PR #3008, opened 2025-06-15
feat: add CronJob sources to UI

BenElferink:
## Description

Following up after: https://github.com/odigos-io/odigos/pull/3004
This PR adds support for CronJob workloads in the UI.

## How Has This Been Tested?

- [x] Manual Testing

## Kubernetes Checklist

## User Facing Changes

- [x] Changes UI, CLI, or K8s Manifests aspects in a way that users need to be aware of


frontend/services/sources.go
```diff
@@ -275,7 +298,7 @@ func toggleSourceWithAPI(c *gin.Context, enabled bool) {
 	})
 }
 
-func stringToWorkloadKind(workloadKind string) (WorkloadKind, bool) {
+func stringToWorkloadKind(workloadKind string) (model.K8sResourceKind, bool) {
```
alonkeyval, line 325: dont we have similar function in the conversion file?

alonkeyval (approved):

frontend/services/sources.go
BenElferink: True :)
But one is in the `graph` module and the other is in `services` module.
We still need to figure out dependency loop between the 2 modules.

PR #3027, opened 2025-06-17
fix: check-links

BenElferink:
## Description

The following PR broke the Clickhouse default DDL link: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/40536
This PR locks the link at a specific tree to prevent 404 not found.

Also, the following broken links were fixed:
- MariaDB MySqlConnector for ADO.NET


blumamir (approved):

SpencerTorres:
Good catch, I need to update this to say `internal/sqltemplates`. The DDL is now in one central location instead of being pasted in the code and an examples directory

PR #3028, opened 2025-06-17
feat: add container overrides to source CRD

blumamir:
WIP: add feature to override the runtime info for specific container in a workload source.

### DIFF
--- a/docs/api-reference/operator.odigos.io.v1alpha1.mdx
+++ b/docs/api-reference/operator.odigos.io.v1alpha1.mdx
@@ -1,0 +1,256 @@
+
+
+## Resource Types 
+
+
+  
+    
+    
+
+## `Odigos` <a id="Odigos"></a>
+    
+<p>Odigos is the Schema for the odigos API</p>
+
+<table class="table">
+<thead><tr><th width="30%">Field</th><th>Type</th><th>Description</th></tr></thead>
+<tbody>
+    
+  
+<tr>
+<td>
+<code>metadata</code> <B>[Required]</B>
+</td>
+<td>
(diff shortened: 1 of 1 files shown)


api/odigos/v1alpha1/source_types.go
```diff
@@ -46,6 +46,16 @@ type Source struct {
 	Status SourceStatus `json:"status,omitempty"`
 }
 
+type ContainerOverride struct {
+	// The name of the container to override.
+	ContainerName string `json:"containerName"`
```
BenElferink, line 51: `RuntimeDetailsByContainer` already has `ContainerName`

api/odigos/v1alpha1/source_types.go
```diff
@@ -60,6 +70,14 @@ type SourceSpec struct {
...
 	// +optional
 	OtelServiceName string `json:"otelServiceName,omitempty"`
+
+	// Specify specific override values for containers.
+	// For example, allow to set the runtime info in case the automatic detection fails.
+	// Containers are listed by their names.
+	// Any container not in this list will retain it's default behavior.
+	// +kubebuilder:validation:Optional
+	// +optional
+	ContainerOverrides []ContainerOverride `json:"containerOverrides,omitempty"`
```
BenElferink, line 81:
```suggestion
	ContainerOverrides []RuntimeDetailsByContainer `json:"containerOverrides,omitempty"`
```

BenElferink (changes requested):

api/odigos/v1alpha1/source_types.go
blumamir: That is intentional, since in the future we may want to override other container properties like "service.name" or other resource attributes.

Issue #3026, opened 2025-06-17
Helm Installation of Odigos Failed due to Conflicts With Rancher's `mutating-webhook-configuration`


jundi77:
**Describe the bug**
When trying to install Odigos on our rancher cluster with helm, there's a failure that is related to `resource that already exists`.

```
helm install --namespace=odigos-dev --timeout=16m40s --values=/home/shell/helm/values-odigos-1.0.198.yaml --version=1.0.198 --wait=true odigos /home/shell/helm/odigos-1.0.198.tgz
Error: INSTALLATION FAILED: rendered manifests contain a resource that already exists. Unable to continue with install: MutatingWebhookConfiguration "mutating-webhook-configuration" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "odigos"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "odigos-dev"
```

It seems that Odigos use the same `MutatingWebhookConfiguration` name as the one used in Rancher `capi-webhooks`.

Odigos `mutating-webhook-configuration`:
https://github.com/odigos-io/odigos/blob/04bc6efe0a729532c3d777e3f57e6aa97e7a2feb/helm/odigos/templates/instrumentor/webhooks.yaml#L4

Rancher `mutating-webhook-configuration`:
https://github.com/rancher/rancher/blob/d3335d87ac86ab99c5aaced559925a5401d01c8d/pkg/crds/provisioningv2/capi-webhooks.yaml#L6

**To Reproduce**
1. Add Odigos Helm chart repository to Rancher.
2. Install Odigos via app installation menu in Rancher with default values.

**Expected behavior**
Installation via Helm charts should install Odigos successfully.

**Additional context**
Rancher v2.8.2
Helm v2.16.8-rancher2
rancher-machine v0.15.0-rancher106

damemi:
Thanks for reporting, if you haven't already it might be good to open an issue in the Rancher repo to let them know as well. It seems like a poor practice on both of our parts to use such a generic name, I'll update ours

PR #3032, opened 2025-06-17
CORE-365 Update webhook names to have odigos prefix

damemi:
## Description

Fixes https://github.com/odigos-io/odigos/issues/3026

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

Should be covered by existing e2es and cli-upgrade

## Kubernetes Checklist

## User Facing Changes


BenElferink (approved):

dan505512 (approved):

dan505512 (changes requested):
CA not working

RonFed (changes requested):
Blocking because:
1. CLI install is deprecated and should not be changed. 
2. Better to name "odigos-pod-mutating..." instead of "odigos-mutating..."
3. Need to verify the upgrade manually - or maybe it did got verified? Just want to make sure

PR #3034, opened 2025-06-17
CORE-97: Add automatic Go offsets support via cron job

damemi:
## Description

This adds a new Odigos configuration + cli commands to set automatic Go offsets updates. It uses the CLI image we already use for helm chart cleanup to create a cronjob that runs the update-offsets command on the schedule set by the user.

Enable the cronjob with:

```
odigos config set go-auto-offsets-cron "0 0 * * *"
```

(to run nightly at midnight)

Disable with:

```
odigos config set go-auto-offsets-cron ""
```

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/docs/permissions.mdx
+++ b/docs/permissions.mdx
@@ -80,7 +80,7 @@ Below are the ClusterRoles used by Odigos components.
 |---|---|---|---|
 | odigos.io | instrumentationconfigs | \* | get<br />list<br />watch |
 | \* | configmaps/finalizers | \* | update |
-| batch | cronjobs | odigos-go-offsets-updater | list<br />watch |
+| batch | cronjobs | \* | list<br />watch |
 | \* | configmaps | \* | list |
 
 ### odigos-ui
@@ -175,7 +175,7 @@ Below are the Roles used by Odigos components. These Roles are only scoped to th
 | odigos.io | collectorsgroups/status | \* | get |
 | odigos.io | instrumentationrules<br />processors | \* | get<br />list<br />watch<br />patch<br />delete<br />create |
 | \* | secrets | \* | get<br />list<br />watch |
-| batch | cronjobs | \* | get<br />list<br />watch<br />create<br />update<br />patch<br />delete |
+| batch | cronjobs | odigos-go-offsets-updater | get<br />list<br />watch<br />create<br />update<br />patch<br />delete |
 | apps | daemonsets | odiglet | patch |
 
 ### odigos-ui
--- a/helm/odigos/templates/scheduler/clusterrole.yaml
+++ b/helm/odigos/templates/scheduler/clusterrole.yaml
@@ -25,8 +25,6 @@ rules:
       - 'batch'
     resources:
       - cronjobs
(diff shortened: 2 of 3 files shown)


edeNFed (approved):
Looks good to me but @RonFed should probably take a look as well

cli/cmd/resources/scheduler.go
```diff
@@ -97,6 +97,16 @@ func NewSchedulerRole(ns string) *rbacv1.Role {
 				Resources: []string{"secrets"},
 				Verbs:     []string{"get", "list", "watch"},
 			},
+			{ // manage go offsets CronJob
+				APIGroups: []string{"batch"},
+				Resources: []string{"cronjobs"},
```
RonFed: Can er narrow it to just the required CronJob name?

cli/cmd/resources/scheduler.go
```diff
@@ -97,6 +97,16 @@ func NewSchedulerRole(ns string) *rbacv1.Role {
...
+				Verbs:     []string{"get", "list", "watch", "create", "update", "patch", "delete"},
+			},
+			{ // restart odiglet daemonset after updating go offsets
+				APIGroups: []string{"apps"},
+				Resources: []string{"daemonsets"},
```
RonFed: Can we narrow it to just the odiglet?
BTW, where is it used? couldn't find where we do the restart

scheduler/controllers/odigosconfiguration/odigosconfiguration_controller.go
```diff
@@ -168,6 +179,120 @@ func (r *odigosConfigurationController) persistEffectiveConfig(ctx context.Conte
...
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to delete go offsets CronJob: %v", err)
+		}
+		return nil
+	}
+	return nil
+}
+
+func reconcileCronJob(ctx context.Context, kubeClient client.Client, ns string, cronJob client.Object, config *common.OdigosConfiguration) error {
```
RonFed: nit
```suggestion
func applyCronJob(ctx context.Context, kubeClient client.Client, ns string, cronJob client.Object, config *common.OdigosConfiguration) error {
```

scheduler/go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/go-logr/zapr v1.3.0
 	github.com/golang-jwt/jwt/v4 v4.5.2
 	github.com/odigos-io/odigos/api v0.0.0
-	github.com/odigos-io/odigos/common v0.0.0
+	github.com/odigos-io/odigos/common v1.0.63
```
RonFed: I think this is not intentional?

RonFed (approved):
lgtm, if we can narrow the RBAC a bit it would be great

PR #3035, opened 2025-06-17
feat: use container overrides for distro selection if it exists

blumamir:
## Description

So any runtime info provided as an override in `Source` CR will be honored and take precedence over automatic detection value


BenElferink:
was approved verbally

PR #3040, opened 2025-06-18
Dynatrace agent detection all lang - CloudNative mode

AvihuHenya:
## Description
This PR introduces detection logic for existing Dynatrace agent installations in the cluster. The goal is to prevent conflicts when Odigos instruments user applications.
 
Dynatrace supports 2 deployment modes, and this changes covers all languages except php when working with `CloudNativeFullStack`, and for python apps in `ClassicFullStack` mode. 
This apparoach relies on `LD_PRELOAD` environment variable the agent inject into the app with specific values it uses. 

The detection mechanism relies on inspecting the LD_PRELOAD environment variable, which Dynatrace injects into application processes with known path patterns (e.g., /opt/dynatrace/oneagent...). By identifying these values, we can infer the presence of the Dynatrace agent and skip instrumentation as needed.
## How Has This Been Tested?

kv-mall apps on eks cluster

Screenshot 2025-06-18 at 12 26 39
Screenshot 2025-06-18 at 13 34 08

## User Facing Changes


procdiscovery/pkg/process/process.go
```diff
@@ -36,14 +36,16 @@ var LangsVersionEnvs = map[string]struct{}{
...
+	DynatraceDynamizerExeSubString   = "oneagentdynamizer"
+	DynatraceFullStackEnvValuePrefix = "/opt/dynatrace/"
 )
 
 var OtherAgentEnvs = map[string]string{
-	NewRelicAgentEnv:      "New Relic Agent",
-	DynatraceDynamizerEnv: "Dynatrace Agent",
+	NewRelicAgentEnv:                 "New Relic Agent",
+	DynatraceDynamizerEnv:            "Dynatrace Agent",
+	DynatraceFullStackEnvValuePrefix: "Dynatrace Agent",
```
tamirdavid1: This approach is for using envs that point us to understand specific agent is install such as `DT_DYNAMIZER_TARGET_EXE` for dynatrace go and `NEW_RELIC_CONFIG_FILE` for newrelic python.

adding `DynatraceFullStackEnvValuePrefix: "Dynatrace Agent"` do nothing in this case. the reason it worked for you is that we're already collecting the `LD_PRELOAD` env few lines above: in 
```
	// Add LD_PRELOAD to the list of relevant environment variables
	relevantOverwriteEnvVars[consts.LdPreloadEnvVarName] = nil
```

procdiscovery/pkg/process/process.go
```diff
@@ -36,14 +36,16 @@ var LangsVersionEnvs = map[string]struct{}{
 }
 
 const (
-	NewRelicAgentEnv               = "NEW_RELIC_CONFIG_FILE"
-	DynatraceDynamizerEnv          = "DT_DYNAMIZER_TARGET_EXE"
-	DynatraceDynamizerExeSubString = "oneagentdynamizer"
+	NewRelicAgentEnv                 = "NEW_RELIC_CONFIG_FILE"
+	DynatraceDynamizerEnv            = "DT_DYNAMIZER_TARGET_EXE"
+	DynatraceDynamizerExeSubString   = "oneagentdynamizer"
+	DynatraceFullStackEnvValuePrefix = "/opt/dynatrace/"
```
tamirdavid1: nit:
I think better to check if `dynatrace` without /opt/ because maybe at some point they will change the path of it but most likely the `dynatrace` substring will still be there.
if u choose this approach maybe better used `contains` and not prefix

AvihuHenya: good point. thanks

tamirdavid1 (approved):

PR #3047, opened 2025-06-19
installed Datadog agent detection 

AvihuHenya:
## Description
Following the same approach as [this PR](https://github.com/odigos-io/odigos/pull/3040), we want the ability to check if Datadog is already installed in the cluster in order to avoid conflicts with Odigos instrumentation.
 

We look for a specific environment variable that the Datadog agent injects into the pod. If it exists, we propagate it to the otherAgent configuration.

This detection logic been checked and works well for all supported languages except PHP.
## How Has This Been Tested?
Manual testing

Screenshot 2025-06-19 at 16 45 34


tamirdavid1:
@AvihuHenya is it the same env injected for all languages?
Thanks

AvihuHenya:
@tamirdavid1 Yes, there are five environment variables with the `DD_` prefix that get injected in all languages. I just picked the one with the most significant name.

tamirdavid1 (approved):

tamirdavid1:
> @tamirdavid1 Yes, there are five environment variables with the `DD_` prefix that get injected in all languages. I just picked the one with the most significant name.

Approved

PR #3048, opened 2025-06-20
Add Gin middleware handler to allow instrumentation

damemi:
## Description

Gin middleware is difficult to instrument, because each middleware function has a unique name and the Gin framework uses a small internal function to call each one that is inlined due to its size. It also does not provide any information about each middleware.

This adds a middleware wrapper that can be inserted to allow Odigos to instrument Gin middleware spans.

The output spans look like the following:

Screenshot 2025-06-20 at 1 08 43 PM

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


blumamir (approved):
🥇

PR #3050, opened 2025-06-22
feat: runtime override indication in UI

BenElferink:
## Description

This PR adds a pre-override warning, and a post-override indicator to the UI.

## Pre-override warning:

Screenshot 2025-06-22 at 10 54 17

## Post-override indicator:

Screenshot 2025-06-22 at 10 54 39


frontend/graph/schema.graphqls
```diff
@@ -63,6 +63,7 @@ type SourceContainer {
   containerName: String!
   language: String!
   runtimeVersion: String!
+  overriden: Boolean!
```
alonkeyval, line 66: for me "overriden" is out of context here, maybe runtimeInfoOverriden

alonkeyval (approved):
approved with small comment

frontend/graph/schema.graphqls
BenElferink: It's not only for runtime though, it's for the whole container, Amir plans on allowing other values to be overriden in the near future. Prefixing the boolean will require a refactor to remove the prefix again when other values are added.

alonkeyval: If we plan to support overrides for other fields like serviceName in the future, each override should ideally have its own flag (e.g., serviceNameOverriden, runtimeInfoOverriden, etc.). This allows us to be more explicit about which fields were customized, without overloading a single boolean.

PR #3059, opened 2025-06-23
style(instrumentor): cleanups and improvments in source inst controller

blumamir:
Just some cleanups of few things in instrumentor `sourceinstrumentation` controllers:

- remove redundant duplicate creation or `PodWorkload`
- the `WorkloadKindFromClientObject` function was used only for this one case, so not needed anymore
- check for `disabled` early and remove big part in indentation.

### DIFF
--- a/instrumentor/controllers/sourceinstrumentation/common.go
+++ b/instrumentor/controllers/sourceinstrumentation/common.go
@@ -68,7 +68,7 @@ func syncNamespaceWorkloads(
 			if client.IgnoreNotFound(err) != nil {
 				return collectiveRes, err
 			}
-			res, err := syncWorkload(ctx, k8sClient, runtimeScheme, obj, kind)
+			res, err := syncWorkload(ctx, k8sClient, runtimeScheme, obj)
 			if err != nil {
 				errs = errors.Join(errs, err)
 			}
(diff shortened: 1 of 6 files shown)


instrumentor/controllers/sourceinstrumentation/common.go
```diff
@@ -98,97 +98,91 @@ func syncWorkload(ctx context.Context, k8sClient client.Client, scheme *runtime.
 		return ctrl.Result{}, err
 	}
 
-	podWorkload := k8sconsts.PodWorkload{
```
blumamir, line 101: reviewer note: we already have `pw` above, so no need to create it again.

instrumentor/controllers/sourceinstrumentation/common.go
```diff
@@ -98,97 +98,91 @@ func syncWorkload(ctx context.Context, k8sClient client.Client, scheme *runtime.
...
-		Name:      obj.GetName(),
-		Namespace: obj.GetNamespace(),
-		Kind:      workload.WorkloadKindFromClientObject(obj),
```
blumamir, line 104: reviewer note: this is the only placed where we used this function, so now it can be removed from odigos

instrumentor/controllers/sourceinstrumentation/common.go
```diff
@@ -98,97 +98,91 @@ func syncWorkload(ctx context.Context, k8sClient client.Client, scheme *runtime.
...
+	if !enabled {
```
blumamir, line 101: reviewer note: this was the `else` part of the original test, far below.

Since it's only returning, I moved it up here and the entire section beflow does not need to indent

PR #3063, opened 2025-06-23
Add disable-gin profile

damemi:
## Description

This adds a profile to disable `gin-gonic/gin` instrumentation for its apparent redundance with some net/http spans. It adds this profile to the `kratos` aggregator

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


api/odigos/v1alpha1/instrumentationrule_type.go
```diff
@@ -72,6 +72,9 @@ type InstrumentationRuleSpec struct {
 
 	// Allows to configure the collection of http headers for different types of payloads.
 	HeadersCollection *instrumentationrules.HttpHeadersCollection `json:"headersCollection,omitempty"`
+
+	// Configure the tracing configuration for the library.
```
blumamir, line 76: This can be set also when `InstrumentationLibraries` is nil, and then it should act on all libraries. might be worth thinking if this is supported, and document any limitation here in the comment

api/odigos/v1alpha1/instrumentationrules/trace_config.go
```diff
@@ -0,0 +1,7 @@
+package instrumentationrules
+
+// +kubebuilder:object:generate=true
+// +kubebuilder:deepcopy-gen=true
+type TraceConfig struct {
+	Enabled *bool `json:"enabled,omitempty"`
```
blumamir: The default value (`nil`) should be "traces on", right? so how about we name it `Disabled` so when it's missing (default) it's the same as false, and user need to opt-in to `disable`,

blumamir: Consider adding a comment explaining what exactly is being enabled/disabled (might be trivial from the context, bug still nice to have written down)

instrumentor/controllers/instrumentationconfig/common.go
```diff
@@ -87,6 +89,23 @@ func updateInstrumentationConfigForWorkload(ic *odigosv1alpha1.InstrumentationCo
...
+				Enabled: rule2.Enabled,
+			}
+		}
+		return nil
+	} else if rule2 == nil {
+		return rule1
+	}
+
+	mergedRules := odigosv1alpha1.InstrumentationLibraryConfigTraces{}
+	mergedRules.Enabled = merge2RuleBooleans(rule1.Enabled, rule2.Enabled)
```
blumamir: calling it `disabled` instead of `enabled` also has the benefit that when there are multiple rules to apply, if any if them set `disabled` to true, then the library will be disabled, regardless of other rules that might set it to "disabled: false" (e.g. enabled).

This guarantee that if someone disables a library in one rule, it will never show up

api/odigos/v1alpha1/instrumentationrule_type.go
damemi: Added, depends on https://github.com/odigos-io/odigos-enterprise/pull/1046

api/odigos/v1alpha1/instrumentationrules/trace_config.go
damemi: I totally agree, I just did it this way so it's a little less confusing having to invert the enabled/disabled between IR and IC. But I pushed an update to do what you're suggesting here, ptal

blumamir (approved):

damemi:
Note: the new instrumentation rule is currently shown as "unknown" in the UI which will need a ui-kit update (cc @BenElferink 🙏 ). This doesn't seem to cause any other issues and the rule can't be edited or deleted since it's system-owned

PR #3064, opened 2025-06-24
fix(instrumentor): fetch the object once in sync function instead of each caller

blumamir:
## Description

This PR addresses 2 things:
- fix bug, where we first list a group of objects and then reconcile them with possibly non-up-to-date state, thus leading to in-consistent state for an object. we already fixed a bunch of those in the past, but this one still looks like it needs addressing.
- the `syncWorkload` function in the `source instrumentation` controllers is called 4 times, and every caller needs to run the exact same lines: 

```
	obj := workload.ClientObjectFromWorkloadKind(objKind)
	err := k8sClient.Get(ctx, key, obj)
	if err != nil {
	}
```

We already have the ctx and k8s client in the function, so better to do it there just once.

This one also guarantee that we are always working with a fresh copy from cache, and not something we `List`ed maybe some time ago and during the iteration became invalid.

## How Has This Been Tested?

- [x] Manual Testing

## Kubernetes Checklist

No K8s changes. We are still making the same calls but now in one central place. All calls are to the cache and not api server

## User Facing Changes

Not noticeable, beside for the bug fix which should be quite rare

### DIFF
--- a/instrumentor/controllers/sourceinstrumentation/common.go
+++ b/instrumentor/controllers/sourceinstrumentation/common.go
@@ -32,6 +32,7 @@ func syncNamespaceWorkloads(
 	runtimeScheme *runtime.Scheme,
 	namespace string) (ctrl.Result, error) {
 
+	workloadsToSync := make([]k8sconsts.PodWorkload, 0)
 	collectiveRes := ctrl.Result{}
 	var errs error
 	for _, kind := range []k8sconsts.WorkloadKind{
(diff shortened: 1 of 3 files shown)


instrumentor/controllers/sourceinstrumentation/common.go
```diff
@@ -46,52 +47,63 @@ func syncNamespaceWorkloads(
...
 // or update the existing InstrumentationConfig if necessary.
-func syncWorkload(ctx context.Context, k8sClient client.Client, scheme *runtime.Scheme, obj client.Object) (ctrl.Result, error) {
+func syncWorkload(ctx context.Context, k8sClient client.Client, scheme *runtime.Scheme, pw k8sconsts.PodWorkload) (ctrl.Result, error) {
 	logger := log.FromContext(ctx)
 
-	pw := k8sconsts.PodWorkload{
-		Name:      obj.GetName(),
-		Namespace: obj.GetNamespace(),
-		Kind:      k8sconsts.WorkloadKind(obj.GetObjectKind().GroupVersionKind().Kind),
+	obj := workload.ClientObjectFromWorkloadKind(pw.Kind)
```
blumamir, line 98: @tamirdavid1  this call now happens once here, it used to be invoked be each caller to `syncWorkload` (3 times, one for each caller)

blumamir, line 98: one more nice side-effect - the user cannot forget to pull in a fresh copy of the workload object. the function guarentee here that an up-to-date object is being used everytime,

PR #3045, opened 2025-06-19
feat: DataStreams

BenElferink:
## Description

- Turn on `Data Streams` feature in the UI with `Namespace Source` support.
- Migrate from `Processor` to `Connector`

## How Has This Been Tested?

- [x] Added Unit Tests
- [x] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

PR #3066, opened 2025-06-24
refactor: update instrumentation config only once in source controller

blumamir:
Just a small refactor, in instrumentor `sourceinstrumentation` controllers.

Instead of calling `Update` twice, the data streams labels is:
1. added to new instrumentation configs upon creation of the resource (instead of create + update)
2. When reconciled on existing object - checks if any updates are needed and calling `Update` once, possibly with other changes to the resource (instead of separate updates per property)


blumamir:
Waiting #3045  to merge before bringing in this one

edeNFed (approved):

PR #3068, opened 2025-06-24
Odigos config hook fix

AvihuHenya:
## Description
This PR renames odigos-config to odigos-configuration globally.

The main reason for this change is that the old `odigos-config` had a legacy `helm.sh/hook` annotation in its Helm template, and that caused a few annoying issues:
- Helm didn’t manage that ConfigMap like the rest of the resources, so we had to manually delete it during uninstall.
- There’s also been a bug between Argo CD and Helm that seems related to this annotation. I wasn’t able to reproduce it consistently, but this rename should help avoid it altogether.

And beside of that, a small nit:  the `odigos.io/system-object=true` label was added to the `odigos-gateway` ConfigMap cli manifest to keep things consistent with the rest of the resources.

#### CLI (odigos update command)
During the update process, the CLI reads the existing odigos-config ConfigMap and preserves its content.
It then creates a new odigos-configuration ConfigMap with the same data.
The old ConfigMap is then deleted since it becomes resource with old value in `odigos.io/config` label (see `DeleteOldOdigosSystemObjects` func for more detailes).

#### Helm
On upgrade:
A new odigos-configuration ConfigMap is created automatically.
A post-upgrade Helm hook Job deletes the legacy odigos-config if it still exists.

On uninstall:
A pre-delete Job ensures the legacy odigos-config is deleted before uninstall completes.

### Short Summary of the Changes
1. Global renaming of odigos-config to odigos-configuration in both CLI and Helm code.
2. nit: added the odigos.io/managed: "true" label to the odigos-gateway ConfigMap cli resource.

### How Has This Been Tested?
Upgraded between Odigos versions and verified:
- Existing config is preserved.
- The instrumentation process keeps running smoothly.
- Ran odigos config and odigos diagnose locally to make sure things still work as expected.

Tested Helm-based upgrade and uninstall in a cluster:
Confirmed that the legacy ConfigMap is properly deleted by the new Jobs.


damemi (commented):
just curious what was the fix? Was something expecting `odigos-configuration` for the configmap name?

AvihuHenya:
I opened this because I want the CI tests to run against my code changes.

This is related to issues we’re encountering due to the `odigos-config` template in the Helm chart including an unnecessary `helm.sh/hook` annotation. This causes several problems—for example, Helm doesn’t manage this resource, so we have to delete it explicitly during the uninstall process. It also results in inconsistent installation behavior with both Helm and ArgoCD.

Since removing the hook would break existing installations and require an awkward migration, the final solution will be to **rename** `odigos-config` and create a Helm-managed resource without the hook annotation.

For now, I’m working on identifying all the places where it is used and testing the renaming process, Expect more changes to follow as this evolves.

AvihuHenya:
Open question – should I rename these as well for consistency?
effective-config → effective-configuration
(CLI command) odigos config set → odigos configuration set

helm/odigos/templates/odigos-config-cm.yaml
RonFed, line 1: filename to odigos-configuration-cm

autoscaler/go.mod
```diff
@@ -8,6 +8,7 @@ require (
 	github.com/go-logr/logr v1.4.3
 	github.com/go-logr/zapr v1.3.0
 	github.com/odigos-io/odigos/api v0.0.0
+	github.com/odigos-io/odigos/cli v0.0.0-20250702105647-24f4dfd7631d
```
RonFed: I don't think we should have a dependency on the CLI. here.
If we're using a common label it should be declared in the `api` probably?

common/consts/consts.go
```diff
@@ -9,7 +9,8 @@ const (
 	OdigosVersionEnvVarName     = "ODIGOS_VERSION"
 	OdigosTierEnvVarName        = "ODIGOS_TIER"
 	DefaultOdigosNamespace      = "odigos-system"
-	OdigosConfigurationName     = "odigos-config"
+	OdigosConfigurationName     = "odigos-configuration"
+	OdigosLegacyConfigName      = "odigos-config"
```
RonFed, line 14: nit:
```suggestion
         // Deprecated: only used for migrations
	DeprecatedOdigosConfigName      = "odigos-config"
```

helm/odigos/templates/cleanup/cleanup-role.yaml
```diff
@@ -7,14 +7,19 @@ metadata:
...
 # This is necessary because it has a Helm hook annotation and is not managed by Helm directly.
 # For more context, see the implementation in /cli/cmd/uninstall.go.
 - apiGroups: [""]
   resources: ["configmaps"]
   resourceNames: ["odigos-deployment"]
   verbs: ["get"]
+- apiGroups: [""]
+  resources: ["configmaps"]
+  resourceNames: ["odigos-configuration"]
+  verbs: ["get","delete"]
```
RonFed: why do we need the cleanup job to clean the new configuration? it should be cleaned by Helm automatically I think.

helm/odigos/templates/legacy-cm-migration/migration-rolebinding.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: migration-rolebinding
```
RonFed:
```suggestion
  name: config-migration-rolebinding
```

helm/odigos/templates/legacy-cm-migration/migration-rolebinding.yaml
```diff
@@ -0,0 +1,15 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    odigos.io/system-object: "true"
+subjects:
+- kind: ServiceAccount
+  name: migration-sa
+  namespace: {{ .Release.Namespace }}
+roleRef:
+  kind: Role
+  name: migration-role
```
RonFed:
```suggestion
  name: config-migration-role
```

helm/odigos/templates/legacy-cm-migration/serviceaccount.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: migration-sa
```
RonFed:
```suggestion
  name: config-migration-sa
```

helm/odigos/templates/legacy-cm-migration/legacy-cm-migration-job.yaml
```diff
@@ -0,0 +1,28 @@
...
+    "helm.sh/hook-delete-policy": hook-succeeded
+spec:
+  ttlSecondsAfterFinished: 60
+  template:
+    spec:
+      serviceAccountName: config-migration-sa
+      restartPolicy: Never
+      containers:
+        - name: cleanup
+          image: bitnami/kubectl:1.28.2
```
RonFed: TODO: think how to handle this in closed environments that can't pull from Dockerhub.
consider implementing this in the CLI image with a dedicated flag?

AvihuHenya: fixed by moving the logic into instrumentor controller

autoscaler/controllers/clustercollector/configmap.go
```diff
@@ -188,6 +188,8 @@ func syncConfigMap(dests *odigosv1.DestinationList, allProcessors *odigosv1.Proc
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      k8sconsts.OdigosClusterCollectorConfigMapName,
 			Namespace: gateway.Namespace,
+			Labels:    map[string]string{ k8sconsts.OdigosSystemLabelKey:
+				k8sconsts.OdigosSystemLabelValue},
```
blumamir: why is it needed?
The system label is used only for objects that are generated by the CLI (this one is generated by a controller)

cli/cmd/resources/applyresources.go
```diff
@@ -46,12 +47,18 @@ func DeleteOldOdigosSystemObjects(ctx context.Context, client *kube.Client, ns s
 
 func GetCurrentConfig(ctx context.Context, client *kube.Client, ns string) (*common.OdigosConfiguration, error) {
 	configMap, err := client.CoreV1().ConfigMaps(ns).Get(ctx, consts.OdigosConfigurationName, metav1.GetOptions{})
-	if err != nil {
-		return nil, err
+	if err != nil && apierrors.IsNotFound(err) {
```
blumamir, line 50: nit: you can omit the nil checking in this case

cli/cmd/uninstall.go
```diff
@@ -105,26 +104,27 @@ Note: Namespaces created during Odigos CLI installation will be deleted during u
...
-					// As a result, running `helm uninstall` does not delete this object
-					// (or the corresponding effective-config).
-					// As a workaround, we explicitly delete it here, and let helm delete all other resources.
-					uninstallOdigosConfiguration(ctx, client, ns)
-					fmt.Printf("\n\u001B[32mSUCCESS:\u001B[0m Odigos uninstalled instrumentation resources and odigos configuration ConfigMap successfuly\n")
+				// MIGRATION: In older versions of Odigos, a legacy ConfigMap named "odigos-config" was used.
+				// It has since been replaced by "odigos-configuration", which is Helm-managed and does not include hook annotations.
+				// As part of the migration, we explicitly delete the legacy ConfigMap if it still exists.
+				config, err := client.CoreV1().ConfigMaps(ns).Get(ctx, consts.OdigosLegacyConfigName, metav1.GetOptions{})
+				if err != nil {
```
blumamir: I think we should return here only for `NotFound` errors, and not any error

docs/mint.json
```diff
@@ -89,7 +89,7 @@
             "cli/odigos_upgrade",
             "cli/odigos_version",
             "cli/odigos_sources",
-            "cli/odigos_config"
+            "cli/odigos_configuration"
```
blumamir, line 92: this value should have a file with this name under `docs/cli/odigos_configuration`. we should either keep the current name, or rename the docs file also.
You can run the docs website locally to validate it is not broken

instrumentor/instrumentor.go
```diff
@@ -50,6 +52,11 @@ func New(opts controllers.KubeManagerOptions, dp *distros.Provider) (*Instrument
 		Namespace: env.GetCurrentNamespace(),
 		Name:      k8sconsts.DeprecatedInstrumentorWebhookSecretName,
 	}})
+	// remove the legacy configmap if it exists
+	mgr.Add(&configmaps.ConfigMapDeleteMigration{Client: mgr.GetClient(), Logger: opts.Logger, ConfigMap: types.NamespacedName{
+		Namespace: env.GetCurrentNamespace(),
+		Name:      consts.OdigosLegacyConfigName,
+	}})
```
blumamir: since the scheduler is the one that handles config, does it make sense to move this functionality there? it will also limit the permissions for instrumentor which is a very important component

instrumentor/instrumentor.go
```diff
@@ -50,6 +52,11 @@ func New(opts controllers.KubeManagerOptions, dp *distros.Provider) (*Instrument
 		Namespace: env.GetCurrentNamespace(),
 		Name:      k8sconsts.DeprecatedInstrumentorWebhookSecretName,
 	}})
+	// remove the legacy configmap if it exists
+	mgr.Add(&configmaps.ConfigMapDeleteMigration{Client: mgr.GetClient(), Logger: opts.Logger, ConfigMap: types.NamespacedName{
```
blumamir: why host these temporary runnable in k8sutils?

If it's used by just one component, it should be part of this component only

k8sutils/pkg/configmaps/configmap_migration.go
```diff
@@ -0,0 +1,56 @@
...
+		}
+		err = m.Client.Delete(ctx, &cm)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, nil
+		}
+		m.Logger.Info("Successfully deleted deprecated configMap", "configMap", m.ConfigMap)
+		return true, nil
```
blumamir, line 52: do we have permissions for that? I see the instrumentor only has 

```
			{
				APIGroups:     []string{""},
				Resources:     []string{"configmaps"},
				ResourceNames: []string{consts.OdigosEffectiveConfigName},
				Verbs:         []string{"get", "list", "watch"},
			},

```

docs/mint.json
AvihuHenya: I updated all reference in docs to use odigos-configuration

blumamir (approved):

PR #3070, opened 2025-06-25
refactor(instrumentor): use distro instead of otelsdk to inject signal envs

blumamir:
## Description

There is an ongoing effort to migrate from the "OtelSDK" notation to the distro notation, managing properties in distro yaml files instead of doing it with "if"s in code in dozens of places.

This PR is another step in the way, migrating the "static signals otel env vars" to derive and be calculated as part of distro flow.

- a new property is added to distro called "signalsAsStaticOtelEnvVars"
- set this property to `true` for community `java`, `php`, `ruby` which were set to use it in code.
- When this property is set to `true`, the pods webhook will inject `OTEL_LOGS_EXPORTER`, `OTEL_METRICS_EXPORTER`, `OTEL_TRACES_EXPORTER` with values either `otlp`, or `none` based on the enabled signals in the node collectors group.
- notice that these env vars are static for the time of pod creation. this was the case previously, but it's encouraged to dynamically propagate these values into the agent so to have all instrumented signals collected correctly at all times.

I am still fetching the collectors group. now instead of doing this in internal function, it is done in the main webhook call. In the future, we should write the enabled signals per container and avoid using the collectors group object in the webhook.

## How Has This Been Tested?

- [x] Manual Testing
- [x] e2e testing - added a missing assert for these env vars in e2e for frontend. php and ruby already have it @BenElferink great job 💪

## Kubernetes Checklist

We are now always `Get`ing the collectors group object from Cache, even for languages that do not need it. this is temporary and will be refactored in a followup PR.

## User Facing Changes

No

### DIFF
--- a/common/envOverwrite/originalenv.go
+++ b/common/envOverwrite/originalenv.go
@@ -1,5 +1,0 @@
-package envOverwrite
-
-// Deprecated - left for migration purposes.
-// TODO: delete in odigos v1.1
-type OriginalEnv map[string]*string
--- a/common/envOverwrite/overwriter.go
+++ b/common/envOverwrite/overwriter.go
@@ -91,12 +91,6 @@ func CleanupEnvValueFromOdigosAdditions(envVarName string, envVarValue string) s
 		envVarValue = strings.ReplaceAll(envVarValue, value, "")
 	}
 
-	// remove any odigos special values if they exist
-	if envVarName == "JAVA_OPTS" || envVarName == "JAVA_TOOL_OPTIONS" {
-		envVarValue = strings.ReplaceAll(envVarValue, " -javaagent:/opt/sre-agent/sre-agent.jar", "")
-		envVarValue = strings.ReplaceAll(envVarValue, " newrelic/bootstrap", "")
-	}
-
 	return envVarValue
 }
 
--- a/k8sutils/pkg/envoverwrite/origenv.go
+++ b/k8sutils/pkg/envoverwrite/origenv.go
@@ -5,17 +5,16 @@ import (
 	"fmt"
 
 	"github.com/odigos-io/odigos/common/consts"
-	"github.com/odigos-io/odigos/common/envOverwrite"
 )
 
 // Deprecated: Used for migration purposes only.
 // remove in odigos v1.1
 type OrigWorkloadEnvValues struct {
-	OrigManifestValues map[string]envOverwrite.OriginalEnv
+	OrigManifestValues map[string]map[string]*string
 }
 
 func NewOrigWorkloadEnvValues(workloadAnnotations map[string]string) (*OrigWorkloadEnvValues, error) {
-	manifestValues := make(map[string]envOverwrite.OriginalEnv)
+	manifestValues := make(map[string]map[string]*string)
 	if workloadAnnotations != nil {
 		if currentEnvAnnotation, ok := workloadAnnotations[consts.ManifestEnvOriginalValAnnotation]; ok {
 			err := json.Unmarshal([]byte(currentEnvAnnotation), &manifestValues)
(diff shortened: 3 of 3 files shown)


blumamir:
> ### Bug: Webhook Fails to Handle Missing Collector Group
> The webhook's signal injection logic is flawed due to two issues:
> 
> 1. The constant used to fetch the `CollectorsGroup` changed to `k8sconsts.OdigosNodeCollectorCollectorGroupName`. If this constant refers to a non-existent or incorrect resource, the fetch will fail.
> 2. If the `CollectorsGroup` is not found, `client.IgnoreNotFound(err)` causes the webhook to proceed with an empty list of enabled signals. This results in all OTEL signal exporter environment variables (e.g., `OTEL_TRACES_EXPORTER`) being explicitly set to "none", overriding OTEL's default behavior. Previously, if the collector group was not found, these environment variables would not be set, allowing OTEL defaults to apply.
> 
> `instrumentor/controllers/agentenabled/pods_webhook.go#L78-L83`
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OmI4YWE2ZDUwLWJkMmItNDBmNC1iMWQ3LWI5ZWViNDgxNWU2ZSIsImVuY3J5cHRpb25LZXkiOiJOQ2ZuYndzWWNOeDBUX0JOajBxTkJDQkNmUHNWMllVR3JCRDhLbEVQbFRZIiwiYnJhbmNoIjoic2lnbmFscy1lbnYtZGlzdHJvIn0sImlhdCI6MTc1MDgxMzI0OSwiZXhwIjoxNzUxNDE4MDQ5fQ.a-x9_VPrVmnOc2iOGBnwxGM_HPg8eH5_iF4Vq7rQ7SK_9cvCwt4SiyN02N0OGUlEiSZLj7x7-nr1YOkmXA8okwIZYvVjxNbNTM3ZpElMmf3BJXIGfDlwyHvOJERR27wSYE-WaGhOoerMP2QzAzv4pw0clHxnz4708dwo2KxAMj9TeaAWFhQ7eij6PJvZYeb0YBng0mbkS3xKM5n3oAjZ_O5RNjAJ65CMSFgJMc_oeWaScA0q60cRK_bmNxQ1Fk3frlsInLG1ZW-DrpG7uCXz9FypnsXM1aaFEKhckDZyme7LESiyUpTuuySxeNf3HgT0vNheM1oQLhpZHGhYnTLDWg)
> 
> _Was this report helpful? Give feedback by reacting with 👍 or 👎_

This is right but practically not relevant, and is temporary until I will pass the signals per container in instrumentation config directly as a followup PR.

instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -274,47 +258,6 @@ func getContainerEnvVarPointer(containerEnv *[]corev1.EnvVar, envVarName string)
 	return nil
 }
 
-func setOtelSignalsExporterEnvVars(ctx context.Context, logger logr.Logger, container *corev1.Container, client client.Client) {
-	odigosNamespace := env.GetCurrentNamespace()
-
-	var nodeCollectorGroup odigosv1.CollectorsGroup
-	err := client.Get(ctx, types.NamespacedName{
-		Namespace: odigosNamespace,
-		Name:      k8sconsts.OdigosNodeCollectorDaemonSetName,
```
blumamir, line 283: note: there was a (minor) bug here, we were using the daemonset name instead of the collector group name. they happen to have the same value.

instrumentor/controllers/agentenabled/podswebhook/env.go
```diff
@@ -86,6 +87,27 @@ func InjectOtlpHttpEndpointEnvVar(existingEnvNames EnvVarNamesMap, container *co
...
 }
 
+func signalOtlpExporterEnvValue(enabledSignals []common.ObservabilitySignal, signal common.ObservabilitySignal) string {
+	if slices.Contains(enabledSignals, signal) {
+		return "otlp"
+	}
+	return "none"
+}
+
+func InjectSignalsAsStaticOtelEnvVars(existingEnvNames EnvVarNamesMap, container *corev1.Container, enabledSignals []common.ObservabilitySignal) EnvVarNamesMap {
```
tamirdavid1, line 97: Maybe consider change the func name to something with exporter signals ?

tamirdavid1 (approved):

PR #3071, opened 2025-06-25
refactor(odiglet): use distros to read "append" env var names for detection

blumamir:
## Description

There is an ongoing effort to consolidate all "distro" related properties into a single place which is the distro manifest.
At the moment, we maintain an `overwriter.go` file with specific env variables per language we use for appending odigos values, and for specifying the override per language.

This PR is another step towards retiring these old code "maps" and "ifs" that controls distro-based logic. once we migrate all uses of this overwrite file, we can delete it and maintain everything env related in the distro manifest.

- migrate odiglet runtime detection to use distro for append environment
- move this configuration to reside under the `environmentVaribales` under each distro - which makes more sense.
- change the configuration option for this property, to allow more flexible replace pattern

Notice: these setting should also be fixed as followup in:
- odiglet ebpf manager
- instrometor webhook
- enterprise loader

## How Has This Been Tested?

Run manually + existing tests that should cover many of these cases.

## Kubernetes Checklist

Not related to k8s in any way

## User Facing Changes

Internal refactor, no changes to user.

### DIFF
--- a/procdiscovery/pkg/process/process.go
+++ b/procdiscovery/pkg/process/process.go
@@ -8,9 +8,6 @@ import (
 	"io"
 	"os"
 	"strings"
-
-	"github.com/odigos-io/odigos/common/consts"
-	"github.com/odigos-io/odigos/common/envOverwrite"
 )
 
 const (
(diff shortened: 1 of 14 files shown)


procdiscovery/pkg/process/process.go
```diff
@@ -8,9 +8,6 @@ import (
 	"io"
 	"os"
 	"strings"
-
-	"github.com/odigos-io/odigos/common/consts"
-	"github.com/odigos-io/odigos/common/envOverwrite"
```
blumamir, line 13: note: this was the main intent. to retire the `envOverwrite` package, we need to migrate all it's users to consume it via the distros.

here we are stop using it in one place, and consume further changes via distro getter which pulls it from the manifests

tamirdavid1 (approved):

PR #3073, opened 2025-06-25
feat: add enabled signals to instrumentation config containers spec

blumamir:
## Description

This PR adds 3 new properties in instrumentation config container agent spec.
These are now either `nil` or empty config, and are used to set if a specific signal is enabled or disabled for collection.

Later on, the webhook can use this info to inject otel signals environment varaibles instead of querying the collectors group.
It will also be extended in the future ro:
- avoid injecting agent if the signals are not supported.
- instrumentations - to dynamically set the relevant signals (replacement for the sdk config which are per-language and are not so easy to work with).

## How Has This Been Tested?

- [x] Manual Testing

## Kubernetes Checklist

## User Facing Changes

Only changes in internal objects


BenElferink (approved):
👑 nice feat!

PR #3074, opened 2025-06-25
Service graph

tamirdavid1:
## Description
Service-to-service interactions is not only a powerful observability feature, but also a valuable debugging tool.
This service graph view helps verify that traces are flowing correctly, since the graph is built directly from trace-based metrics.
It also provides real-time visibility into service connectivity patterns and traffic volume.

This PR adds support for exposing Service Graph metrics in the Odigos Gateway to enable UI visualization of service-to-service interactions.

1. Parses servicegraph_traces_service_graph_request_total metrics received from cluster collectors.
2. Maintains an in-memory structure (ServiceGraph) to store cumulative request counts between services.
3. Exposes the service graph via a new GetServiceGraphEdges() method.
4.Extends the gateway to forward these metrics to the UI layer.
5. The graph is represented as a `map[client][server]ServiceGraphEdge`, capturing both request count and last update time.

This functionality lays the groundwork for visualizing live service dependency maps.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/common/pipelinegen/config_builder.go
+++ b/common/pipelinegen/config_builder.go
@@ -10,6 +10,7 @@ import (
 	"github.com/odigos-io/odigos/common"
 	"github.com/odigos-io/odigos/common/config"
 	"github.com/odigos-io/odigos/common/consts"
+	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
 )
 
 func GetGatewayConfig(
@@ -213,7 +214,7 @@ func insertServiceGraphPipeline(currentConfig *config.Config) {
 		"store": config.GenericMap{
 			"ttl": "5m",
 		},
-		"dimensions": []string{"service.name"},
+		"dimensions": []string{string(semconv.ServiceNameKey)},
 	}
 
 	// Add the service graph pipeline to receive the service graph metrics from the root traces pipeline
--- a/common/pipelinegen/config_builder.go
+++ b/common/pipelinegen/config_builder.go
@@ -139,7 +139,9 @@ func CalculateGatewayConfig(
 			return "", err, status, nil
 		}
 	}
-
+	if tracesEnabled {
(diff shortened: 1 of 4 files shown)


common/pipelinegen/config_builder.go
```diff
@@ -196,6 +199,37 @@ func applyRootPipelineForSignal(currentConfig *config.Config, signal common.Obse
 	}
 }
 
+func insertServiceGraphPipeline(currentConfig *config.Config) {
+	// Add the service graph exporter to expose the service graph metrics to prometheus
+	currentConfig.Exporters["prometheus/servicegraph"] = config.GenericMap{
+		"endpoint":  "localhost:9090",
```
daemon1024, line 205: Should we keep the port configurable or atleast declare it as a const (for future configurability) in case end user want it to be configurable?

common/pipelinegen/config_builder.go
```diff
@@ -266,3 +300,49 @@ func filterSmallBatchesProcessor(tracesProcessors []string) ([]string, bool) {
...
 	return filtered, smallBatchesEnabled
 }
+
+func AddServiceGraphScrapeConfig(c *config.Config) {
+	servicegraphScrape := config.GenericMap{
+		"job_name":        "servicegraph",
+		"scrape_interval": "10s",
+		"static_configs": []config.GenericMap{
+			{
+				"targets": []string{"127.0.0.1:9090"},
```
daemon1024, line 310: We are using `127.0.0.1` , `0.0.0.0` and `localhost` interchangably across the code? Is there a need to standardize this? Use it from a const?

daemon1024 (commented):
Small nit, not a blocker for PR merge.

PR #3081, opened 2025-06-26
Service graph

tamirdavid1:
## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/cli/cmd/config.go
+++ b/cli/cmd/config.go
@@ -58,7 +58,7 @@ var configCmd = &cobra.Command{
 	- "%s": Sets the client ID of the OIDC application.
 	- "%s": Sets the client secret of the OIDC application.
 	- "%s": Sets the port for the Odiglet health probes (readiness/liveness).
-  - "%s": Enable or disable the service graph feature [default: false].
+  	- "%s": Enable or disable the service graph feature [default: false].
 	`,
 		consts.TelemetryEnabledProperty,
 		consts.OpenshiftEnabledProperty,
--- a/docs/api-reference/odigos.io.v1alpha1.mdx
+++ b/docs/api-reference/odigos.io.v1alpha1.mdx
@@ -938,6 +938,18 @@ these settings are used to protect the collectors instances from:</p>
 </ul>
 </td>
 </tr>
+<tr>
+<td>
+<code>serviceGraphDisabled</code> <B>[Required]</B>
+</td>
+<td>
+<code>bool</code>
+</td>
+<td>
+   <p>ServiceGraphEnabled is a feature that allows you to visualize the service graph of your application.
+It is enabled by default and can be disabled by setting the enabled flag to false.</p>
+</td>
+</tr>
 </tbody>
 </table>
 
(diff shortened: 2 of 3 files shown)


common/pipelinegen/config_builder.go
```diff
@@ -198,6 +201,43 @@ func applyRootPipelineForSignal(currentConfig *config.Config, signal common.Obse
...
+	// adding the service graph scrape config to prometheus/self-metrics receiver
+	err := AddServiceGraphScrapeConfig(currentConfig)
+	if err != nil {
+		return
+	}
+
+	// Add the service graph connector to receive the service graph metrics from the root traces pipeline
+	currentConfig.Connectors[consts.ServiceGraphConnectorName] = config.GenericMap{
+		"store": config.GenericMap{
+			"ttl": "5m",
```
blumamir: 5m is a long time to keep span in memory, no? we can blow it up quite fast if there is a lot of traffic. otel default is `2s`

common/pipelinegen/config_builder.go
```diff
@@ -198,6 +201,43 @@ func applyRootPipelineForSignal(currentConfig *config.Config, signal common.Obse
...
+	// Add the service graph pipeline to receive the service graph metrics from the root traces pipeline
+	currentConfig.Service.Pipelines["metrics/servicegraph"] = config.Pipeline{
+		Receivers: []string{consts.ServiceGraphConnectorName},
+		Exporters: []string{"prometheus/servicegraph"},
+	}
+
+	// Add the service graph exporter to the root traces pipeline
+	rootPipelineName := GetTelemetryRootPipelineName(common.TracesObservabilitySignal)
+	pipeline, exists := currentConfig.Service.Pipelines[rootPipelineName]
+	if !exists {
```
blumamir, line 241: nit: we will only enter this function if `tracesEnabled` is `true`, so this check should always hold. since this function does not return any error, this looks like a valid path in code.

consider on of the following:
1. adding a comment that the pipeline should always exist.
2. change the condition to enter this function to the existence of the pipeline
3. return a proper error so it's logged.

blumamir (approved):

common/odigos_config.go
```diff
@@ -100,6 +100,10 @@ type CollectorGatewayConfiguration struct {
 	// this is when go runtime will start garbage collection.
 	// if not specified, it will be set to 80% of the hard limit of the memory limiter.
 	GoMemLimitMib int `json:"goMemLimitMiB,omitempty"`
+
+	// ServiceGraphEnabled is a feature that allows you to visualize the service graph of your application.
+	// It is enabled by default and can be disabled by setting the enabled flag to false.
+	ServiceGraphEnabled *bool `json:"serviceGraphEnabled,omitempty"`
```
blumamir: default is true -> Disabled instead of Enabled

PR #3090, opened 2025-06-26
chore: remove unused enum for runtime detection migration

blumamir:
Remove unused `RuntimeUpdateState` property from runtime detection which was used for the migration which is already removed.


blumamir:
will be re-implemented after the webhook migration to distro will be completed and webhook no longer use this property

PR #3091, opened 2025-06-26
Should fix the owner resolve

TheErikNaz:
## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


damemi (approved):
~~can ignore operator rbac failure -- operator already has [batch/jobs `get,list,watch`](https://github.com/odigos-io/odigos/blob/469007d48ca24504733c85242b607920d20121ee/operator/internal/controller/odigos_controller.go#L116)~~

damemi (changes requested):
sorry, misread, operator only has `cronjobs`. You'll need to add `jobs` to the resources on that line and run `USE_IMAGE_DIGESTS=true make generate manifests bundle`

PR #3096, opened 2025-06-27
chore: move deprecated envoverwrite utils from k8sutils to cli

blumamir:
This is one more step towards deprecating our old envoverwriter logic which is no longer relevant.

- move cleanup code from common to cli, add deprecation info, and reduce the complexity
- Move some `OrigWorkloadEnvValues` deprecated code used only in "uninstall" from `k8sutils` to `cli` module.

After a lot of cleanup over the time, cli uninstall is the only one still using this `OrigWorkloadEnvValues` and cleanup function, so I want it to move away from common modules to clean them up, and since they should contain only shared functionallity.

I think it's not actually relevant anymore to anyone, but decided to keep it around for some more time just in case someone is using a very old version.

### DIFF
--- a/cli/cmd/uninstall.go
+++ b/cli/cmd/uninstall.go
@@ -10,7 +10,6 @@ import (
 	"time"
 
 	"github.com/odigos-io/odigos/api/k8sconsts"
-	"github.com/odigos-io/odigos/common/envOverwrite"
 	"github.com/odigos-io/odigos/k8sutils/pkg/installationmethod"
 
 	"github.com/odigos-io/odigos/cli/cmd/resources"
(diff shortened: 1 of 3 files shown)


cli/cmd/uninstall.go
```diff
@@ -10,13 +10,12 @@ import (
 	"time"
 
 	"github.com/odigos-io/odigos/api/k8sconsts"
-	"github.com/odigos-io/odigos/common/envOverwrite"
```
blumamir, line 13: this is the motivation, to stop using "github.com/odigos-io/odigos/common/envOverwrite" everywhere and git rid of it

PR #3103, opened 2025-06-29
chore: configurable liveness/readiness port on odiglet to avoid port conflicts for hostnetwork mood

tamirdavid1:
…conflicting

## Description

For Kubernetes versions below 1.26, Odigos installs the Odiglet with hostNetwork enabled. This can lead to port conflicts with other services running on the same node.

Previously, the Odiglet used port 8081 for its liveness and readiness probes. However, in some environments, this port was already in use, causing issues during deployment.

To address this, we need to make the probe port configurable—either through the Helm chart or via a dedicated configuration option—so users can avoid these conflicts by selecting a custom port.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/odiglet/cmd/main.go
+++ b/odiglet/cmd/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"flag"
 	"os"
 
 	"github.com/odigos-io/odigos/distros"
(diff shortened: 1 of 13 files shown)


odiglet/cmd/main.go
```diff
@@ -22,6 +23,10 @@ import (
 )
 
 func main() {
+	var healthProbeBindPort int
+	flag.IntVar(&healthProbeBindPort, "health-probe-bind-port", 8081, "The port the probe endpoint binds to.")
```
blumamir: change default port to new default and use the const

```suggestion
	flag.IntVar(&healthProbeBindPort, "health-probe-bind-port", consts.OdigletDefaultHealthProbeBindPort, "The port the probe endpoint binds to.")
```

PR #3112, opened 2025-06-30
Pdb for odigos components

tamirdavid1:
## Description

This PR introduces optional support for Kubernetes Pod Disruption Budgets (PDBs) across core Odigos components. The goal is to improve workload availability and resilience during voluntary disruptions (e.g., node drains, upgrades) while maintaining operational flexibility.

1. PDBs are available only via Helm and can be enabled using a global pdb.enabled flag in values.yaml (default: false).

2. Each component is assigned a disruption policy based on its criticality, with minAvailable: 1 for critical services (e.g., instrumentor, gateway) and maxUnavailable: 1 for non-critical, single-replica components.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


tamirdavid1:
> lgtm, but why are they only available for helm?

Basically it is a request coming from a user using helm, wanted to push it quickly for him.
Eventually i'll expand it also for cli per request

PR #3115, opened 2025-06-30
otlp destinations: Set tls insecure:true if TLS is not enabled

damemi:
## Description

In https://github.com/odigos-io/odigos/pull/3069, the tls config was moved to only be set if configured by a user explicitly or through oauth

However, this means that disabling TLS in these destinations does not set `tls.insecure: true`. This updates both destinations to fix the regression

This can cause errors like:

```
rpc error: code = Unavailable desc = connection error: desc = \"transport: authentication handshake failed: tls: first record does not look like a TLS handshake
```

or

```
connection error: desc = "transport: Error while dialing: dial tcp 10.96.133.107:4317: connect: connection refused"
```

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/common/config/genericotlp.go
+++ b/common/config/genericotlp.go
@@ -60,10 +60,10 @@ func (g *GenericOTLP) ModifyConfig(dest ExporterConfigurer, currentConfig *Confi
 	}
 
 	// Only add TLS config if TLS is needed (user-enabled or OAuth2-required)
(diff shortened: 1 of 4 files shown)


BenElferink (approved):
LGTM

PR #3119, opened 2025-07-01
chore: adding the operation name fo the span name if missing

tamirdavid1:
## Description

We have a processor that heuristically determines the SQL operation name by extracting the first word from the db.query.text. Once identified, we append this operation name to the span names to avoid scenarios where users see spans labeled only as "DB", which can be ambiguous.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/collector/processors/odigossqldboperationprocessor/processor.go
+++ b/collector/processors/odigossqldboperationprocessor/processor.go
@@ -2,6 +2,7 @@ package odigossqldboperationprocessor
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	"go.opentelemetry.io/collector/pdata/ptrace"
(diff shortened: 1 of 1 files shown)


collector/processors/odigossqldboperationprocessor/processor.go
```diff
@@ -56,6 +57,10 @@ func (sp *DBOperationProcessor) processTraces(ctx context.Context, td ptrace.Tra
 				// Only set the `db.operation.name` if the detected operation name is not "UNKNOWN"
 				if operationName != OperationUnknown {
 					span.Attributes().PutStr(string(semconv.DBOperationNameKey), operationName)
+
+					// Append operation name to the span name
+					originalName := span.Name()
+					span.SetName(fmt.Sprintf("%s %s", originalName, operationName))
```
RonFed, line 63: should we always do this? if the span already has a full correct span name according to semconv - we probably don't want to touch it. Maybe worth to add a check only if the span name is `DB` ?

blumamir (approved):

collector/processors/odigossqldboperationprocessor/processor.go
tamirdavid1: thats a good point @RonFed but i also saw cases of
`posgresql - DB`
image

PR #3122, opened 2025-07-01
Add initial odiglet plugin proto defintion

damemi:
## Description

This adds the initial proto design for odiglet plugins from https://github.com/RonFed/odigos_fork/tree/plugins_poc. It does not actually use it anywhere yet, but merging this will make developing the plugins easier.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


instrumentation/plugin/proto/v1/instrumentation.proto
```diff
@@ -0,0 +1,32 @@
...
+  optional int32 process_id = 1;
+}
+
+service Instrumentation {
+  rpc Start(InstrumentationRequest) returns (google.protobuf.Empty) {}
+
+  rpc ApplyConfig(InstrumentationConfigRequest) returns (google.protobuf.Empty) {}
+
+  rpc Close(InstrumentationCloseRequest) returns (google.protobuf.Empty) {}
+}
```
edeNFed: I think we should also add an api to let the plugin identify itself to the odiglet. Which runtimes this plugin support, what kinds of telemetry is it able to produce, is it requiring process restart. This all will probably be used in a later stage but I do see us writing plugins not just for tracing for example David is working on a way to extract Java Runtime metrics (GC, threads, etc) via eBPF. 

Also, in the future I think one of our goals is to eliminate the need for application restart entirely and using the plugin mechanism would be a good way to gradually roll that capability (will probably require big changes to our current instrumentations and we will want to run side by side for some time).

We have clients that would like to extend Odiglet by writing their own plugins even for non observability use cases so maybe a better naming would be attach and detach instead of instrumentation. 

I don't remember if versioning is something we get out of the box with go-plugin package but if not may be we should also add some sort of versioning to the proto.

PR #3131, opened 2025-07-03
fix: allow action only in odigos ns

blumamir:
## Description

Autoscaler will attempt to reconcile an action into a processor CR, which is limited to the odigos namespace.

This PR fix this issue, by only allowing actions to be set in the namespace where they are supported:

- cluster role -> role in manifests
- limit the controller runtime manager cache to pull only from odigos ns
- validate in webhook that a new action has the correct namespace.

## How Has This Been Tested?

Created action in the "default" namespace: 

```
The Action "add-cluster-info-example" is invalid: namespace: Invalid value: "default": actions are only allowed in the odigos namespace 'odigos-system'
```

Create action in "odigos-system" namespace and verified it is successfull.

- [x] Manual Testing

## Kubernetes Checklist

## User Facing Changes

The new unified actions are not yes used by users, so no user is expected to be effected


damemi (approved):
Operator RBAC check can be ignored, RBAC changes are just shifting a clusterrole to a role, which the operator already has 👍

PR #3133, opened 2025-07-03
Update operator README with more details

damemi:
## Description

Add a lot more info about how the operator works. Also adds support for setting `operator-rbac-approved` to PRs that fail the operator RBAC check

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


.github/workflows/check-operator-rbac.yml
```diff
@@ -17,6 +17,14 @@ jobs:
 
       - name: Check for RBAC changes
         run: |
+          # Check if PR has operator-rbac-approved label
+          PR_LABELS="${{ join(github.event.pull_request.labels.*.name, ' ') }}"
+          
+          if echo "$PR_LABELS" | grep -q "operator-rbac-approved"; then
+            echo "✅ PR has 'operator-rbac-approved' label - any RBAC changes have been approved"
+            exit 0
+          fi
```
blumamir, line 26: 🥇

operator/Makefile
```diff
@@ -277,6 +277,7 @@ endif
 
 .PHONY: bundle
 bundle: manifests kustomize operator-sdk ## Generate bundle manifests and metadata, then validate generated files.
+	rm -rf bundle/manifests/*
```
blumamir, line 280: it looks like we run `bundle: manifests` above to create the manifests and then `rm` them. perhaps we should not create them in the first place?

operator/README.md
```diff
@@ -3,116 +3,79 @@ The Odigos Operator is another installation, management, and upgrade method for
...
+This tool is essentially a wrapper around Kubebuilder with some added functionality (mainly around building and packaging), so
+it shares much of the support for Kubebuilder.
 
-**Deploy the Manager to the cluster with the image specified by `IMG`:**
+### Manifests, CSV, and Bundle
 
-```sh
-make deploy IMG=<some-registry>/odigos-operator:tag
-```
+Operator manifests are defined under the [`config/`](config) directory as Kustomize files. Many of these were scaffolded by default
```
blumamir, line 29: now that they are `scaffolded`, do we manage them manually? or are they derived-autogenerated from some other files?

operator/README.md
```diff
@@ -3,116 +3,79 @@ The Odigos Operator is another installation, management, and upgrade method for
...
-**Create instances of your solution**
-You can apply the samples (examples) from the config/sample:
+The CSV is a CoreOS/Operator Framework concept that defines metadata about the operator, its dependent images, CRDs, manifests, etc.
+It is essentially a single YAML file that allows the Operator to be installed and managed by the [Operator Lifecycle Manager (OLM)](https://olm.operatorframework.io/).
 
-```sh
-kubectl apply -k config/samples/
-```
+The base CSV file is at [`config/manifests/bases/odigos-operator.clusterserviceversion.yaml`](config/manifests/bases/odigos-operator.clusterserviceversion.yaml).
+This file can be edited to update fixed metadata like description, keywords, maintainer info, etc.
```
blumamir, line 38: I see it also contains the config options. do we need to consider add each new config option to this file as well? what is the process to update these?

operator/README.md
```diff
@@ -3,116 +3,79 @@ The Odigos Operator is another installation, management, and upgrade method for
...
+Running `make bundle` uses the base CSV and all other manifest Kustomize files to generate the final CSV in [`bundle/manifests/odigos-operator.clusterserviceversion.yaml`](bundle/manifests/odigos-operator.clusterserviceversion.yaml).
+This final CSV is what is distributed to OperatorHub on OpenShift (see [DEVELOPMENT.md](DEVELOPMENT.md#publishing-the-release-to-openshift)).
 
-Label the installed namespace with:
+### Versioning
 
-```
-kubectl label ns/<namespace> odigos.io/system-object=true
-```
+Currently, the operator is pinned to the same version of OpenShift. Meaning `v1.0.205` of the Operator will only install `v1.0.205` of Odigos.
```
blumamir:
```suggestion
Currently, the operator is pinned to the same version of Odigos. Meaning `v1.0.205` of the Operator will only install `v1.0.205` of Odigos.
```

operator/README.md
```diff
@@ -3,116 +3,79 @@ The Odigos Operator is another installation, management, and upgrade method for
...
+from this directory. This will update the operator manifests and generate a new operator bundle (which may include other irrelevant generated changes such
+as image references and timestamps -- this is normal).
 
-## Project Distribution
+#### Why ClusterRoles?
 
-Following are the steps to build the installer and distribute this project to users.
+In the interest of generating as much code as possible, we use `kubebuilder` markers to define operator permissions. These markers support
+generating namespaced Roles, but the namespace must be coded into the marker to do so. Since the operator can be installed in any namespace,
+this makes it currently infeasible to generate namespaced Roles. This is a TODO to be improved.
```
blumamir, line 77: Does users really need to set the namespace in openshift? if we assume it's always going to be `odigos-system` we can relax some of the permissions we use which is a big win

operator/Makefile
damemi: this command is creating the manifest files that get packaged into the bundle (under `bundle/manifests`. the `manifests` command generates the original files under `config` that then get packaged into the bundle

So basically:

kubebuilder markers -> config/ -> bundle -> bundle image

We might be able to improve this, or possibly just not even commit the bundle folder itself since it is in some ways redundant (it gets copied exactly to the Red Hat release repo)

operator/README.md
damemi: this is a little messy, and I don't have a good answer. Some things are generated (such as `config/rbac/role.yaml`) while others are manually managed. I'll do another follow up with more details on which is which but I actually need to dig more into that

operator/README.md
damemi: Similar to above, this is a weird one. The config options are automatically generated (try removing one and then re-running the `make generate manifests bundle` command). But the description and metadata aren't. I'll add more documentation on this when I have more info

operator/README.md
damemi: Yeah, we can propose a recommended namespace (https://github.com/odigos-io/odigos/pull/2543), but ultimately it's up to the user. That's a design best practice for the operator framework in general to not assume a namespace so a lot of the tooling is built around that https://sdk.operatorframework.io/docs/best-practices/best-practices/

PR #3135, opened 2025-07-05
feat: calculate env injection method in reconcile 

blumamir:
## Description

This is another step in the journey for deprecating otel-sdk and env-overwite.

Today, pods webhook applies some logic deciding which env injection to use, based on:
- odigos configuration (`agentEnvVarsInjectionMethod` which can have 3 values)
- runtime detection (`securedEcecution` and `LD_PRELOAD` in EnvVars)
- `otelSdk` as this logic will only run for languages found in this static map in code.

The initiative is to move this calculation to the reconcile phase, and only storing single enum on each container agent configuration to instruct it what to use in injection:
- less complexity in webhook which is sensitive component that might get run frequently
- better visibility in ui and instrumentation config to what odigos is doing
- less dependencies in the webhook - consolidate all it's inputs into a single place which gets updated atomically in reconcile which is less room for race conditions and conflicts.
- one more place where we don't rely on otelSdk and migrate to use distro manifests. this will help us deprecate it and cleanup our code a bit.

## How Has This Been Tested?

- [x] Manual Testing

## Kubernetes Checklist

writes one more enum into each agent config in "InstrumentationConfig" object

~~- [ ] Introduces additional calls to the API Server (potential performance impact)~~
~~- [ ] New Query/feature supported in all the k8s versions supported by Odigos~~
added to CRD

~~- [ ] Changes RBAC permissions~~

## User Facing Changes

Internal refactor.

### DIFF
--- a/api/odigos/v1alpha1/instrumentationconfig_types.go
+++ b/api/odigos/v1alpha1/instrumentationconfig_types.go
@@ -290,6 +290,15 @@ type ContainerAgentConfig struct {
 	// Keys are parameter names (like "libc") and values are the value to use for that parameter (glibc / musl)
 	DistroParams map[string]string `json:"distroParams,omitempty"`
 
+	// the reconciler resolves which injection method to use for this container.
+	// it is affected (at the moment) by config, runtime inspection and the user defined overrides.
+	// This field carries the following semantics:
+	// - nil: do not inject any "append variables" (PYTHONPATH, NODE_OPTIONS, etc.) at all.
+	// - "loader": inject the LD_PRELOAD env var to the pod manifest which will trigger the odigos loader.
+	// - "pod-manifest": inject the runtime specific agent loading env vars (e.g PYTHONPATH, NODE_OPTIONS) as specified in the distro manifest.
+	// - "loader-fallback-to-pod-manifest": use "pod-manifest". it means we tried LD_PRELOAD and it failed, so we falled-back to the pod manifest.
+	EnvInjectionMethod *common.EnvInjectionMethod `json:"agentInjectionMethod,omitempty"`
+
 	// Each enabled signal must be set with a non-nil value (even if the config content is empty).
 	// nil means that the signal is disabled and should not be instrumented/collected by the agent.
 	Traces  *AgentTracesConfig  `json:"traces,omitempty"`
--- a/instrumentor/controllers/agentenabled/sync.go
+++ b/instrumentor/controllers/agentenabled/sync.go
@@ -14,6 +14,7 @@ import (
 	"github.com/odigos-io/odigos/common"
 	commonconsts "github.com/odigos-io/odigos/common/consts"
 	"github.com/odigos-io/odigos/distros"
+	"github.com/odigos-io/odigos/distros/distro"
 	distroTypes "github.com/odigos-io/odigos/distros/distro"
 	"github.com/odigos-io/odigos/instrumentor/controllers/agentenabled/rollout"
 	"github.com/odigos-io/odigos/k8sutils/pkg/utils"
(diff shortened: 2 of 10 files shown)


blumamir:
> * It attempts loader injection for distros that support `LD_PRELOAD` but lack `AppendOdigosVariables`, a scenario previously prevented.

This is OK, it aligns with the configuration in the mainfest without applying any additional assumptions. at the moment they are aligned, but we might have different cases going forward which we would want to have in the manifest and let reconciler follow as is

api/odigos/v1alpha1/instrumentationconfig_types.go
```diff
@@ -290,6 +290,15 @@ type ContainerAgentConfig struct {
 	// Keys are parameter names (like "libc") and values are the value to use for that parameter (glibc / musl)
 	DistroParams map[string]string `json:"distroParams,omitempty"`
 
+	// the reconciler resolves which injection method to use for this container.
+	// it is affected (at the moment) by config, runtime inspection and the user defined overrides.
+	// This field carries the following semantics:
+	// - nil: do not inject any "append variables" (PYTHONPATH, NODE_OPTIONS, etc.) at all.
+	// - "loader": inject the LD_PRELOAD env var to the pod manifest which will trigger the odigos loader.
+	// - "pod-manifest": inject the runtime specific agent loading env vars (e.g PYTHONPATH, NODE_OPTIONS) as specified in the distro manifest.
```
RonFed:
```suggestion
	// - "pod-manifest": inject the runtime specific agent loading env vars (e.g PYTHONPATH, NODE_OPTIONS) to the pod manifest as specified in the distro manifest.
```

api/odigos/v1alpha1/instrumentationconfig_types.go
```diff
@@ -290,6 +290,15 @@ type ContainerAgentConfig struct {
 	// Keys are parameter names (like "libc") and values are the value to use for that parameter (glibc / musl)
 	DistroParams map[string]string `json:"distroParams,omitempty"`
 
+	// the reconciler resolves which injection method to use for this container.
```
RonFed: I think this line is not relevant for explaining this field (it is an implementation detail)

instrumentor/controllers/agentenabled/sync.go
```diff
@@ -295,6 +296,117 @@ func getEnvVarFromRuntimeDetails(runtimeDetails *odigosv1.RuntimeDetailsByContai
...
+			AgentEnabled:        false,
+			AgentEnabledReason:  odigosv1.AgentEnabledReasonInjectionConflict,
+			AgentEnabledMessage: "container is running in secure execution mode and injection method is set to 'loader'",
+		}
+	}
+
+	// check if the LD_PRELOAD env var is not already present in the manifest and the runtime details env var is not set or set to the odigos loader path.
+	odigosLoaderPath := filepath.Join(k8sconsts.OdigosAgentsDirectory, commonconsts.OdigosLoaderDirName, commonconsts.OdigosLoaderName)
+	ldPreloadValue, _ := getEnvVarFromRuntimeDetails(runtimeDetails, "LD_PRELOAD")
+	ldPreloadUnsetOrExpected := ldPreloadValue == "" || strings.Contains(ldPreloadValue, odigosLoaderPath) // treat empty value as unset (TODO: revisit)
```
RonFed: I think we should be conservative here and treat empty LD_PRELOAD as a user defined value, but I'm not sure.

instrumentor/controllers/agentenabled/sync.go
```diff
@@ -295,6 +296,117 @@ func getEnvVarFromRuntimeDetails(runtimeDetails *odigosv1.RuntimeDetailsByContai
...
+			ContainerName:       containerName,
+			AgentEnabled:        false,
+			AgentEnabledReason:  odigosv1.AgentEnabledReasonInjectionConflict,
+			AgentEnabledMessage: "container is running in secure execution mode and injection method is set to 'loader'",
+		}
+	}
+
+	// check if the LD_PRELOAD env var is not already present in the manifest and the runtime details env var is not set or set to the odigos loader path.
+	odigosLoaderPath := filepath.Join(k8sconsts.OdigosAgentsDirectory, commonconsts.OdigosLoaderDirName, commonconsts.OdigosLoaderName)
+	ldPreloadValue, _ := getEnvVarFromRuntimeDetails(runtimeDetails, "LD_PRELOAD")
```
RonFed:
```suggestion
	ldPreloadValue, _ := getEnvVarFromRuntimeDetails(runtimeDetails, commonconsts.LdPreloadEnvVarName)
```

instrumentor/controllers/agentenabled/sync.go
```diff
@@ -295,6 +296,117 @@ func getEnvVarFromRuntimeDetails(runtimeDetails *odigosv1.RuntimeDetailsByContai
...
+//
+// returns "nil" on success, and a container agent config with reason and message if not supported.
+func isLoaderInjectionSupportedByRuntimeDetails(containerName string, runtimeDetails *odigosv1.RuntimeDetailsByContainer) *odigosv1.ContainerAgentConfig {
+	// check for conditions to inject ldpreload when it is the only method configured.
+	secureExecution := runtimeDetails.SecureExecutionMode == nil || *runtimeDetails.SecureExecutionMode
+	if secureExecution {
+		return &odigosv1.ContainerAgentConfig{
+			ContainerName:       containerName,
+			AgentEnabled:        false,
+			AgentEnabledReason:  odigosv1.AgentEnabledReasonInjectionConflict,
```
RonFed, line 312: I wouldn't classify this as `Injection Conflict`

instrumentor/controllers/agentenabled/sync.go
```diff
@@ -295,6 +296,117 @@ func getEnvVarFromRuntimeDetails(runtimeDetails *odigosv1.RuntimeDetailsByContai
...
+		}
+	}
+
+	return nil
+}
+
+// Will calculate the env injection method for the container based on the relevant parameters.
+// returned paramters are:
+// - the env injection method to use for this container (may be nil if no injection should take place)
+// - a container agent config to signal any failures in using the loader in these conditions.
```
RonFed, line 335: might be worth as a follow up but the `ContainerAgentConfig` is more of a condition/status I guess

instrumentor/controllers/agentenabled/sync.go
```diff
@@ -295,6 +296,117 @@ func getEnvVarFromRuntimeDetails(runtimeDetails *odigosv1.RuntimeDetailsByContai
...
+			AgentEnabled:        false,
+			AgentEnabledReason:  odigosv1.AgentEnabledReasonInjectionConflict,
+			AgentEnabledMessage: "no injection method configured for odigos agent",
+		}
+	}
+
+	// If we should try loader, check for this first
+	distroSupportsLoader := distro.RuntimeAgent != nil && distro.RuntimeAgent.LdPreloadInjectionSupported
+	configRequiresLoader := *effectiveConfig.AgentEnvVarsInjectionMethod == common.LoaderEnvInjectionMethod
+	configRequestLoader := configRequiresLoader || *effectiveConfig.AgentEnvVarsInjectionMethod == common.LoaderFallbackToPodManifestInjectionMethod
```
RonFed: This `requires` vs `request` is confusing IMO - What do you think about keeping the original enum terminoligy here:
```Go
    loaderRequested := *effectiveConfig.AgentEnvVarsInjectionMethod == common.LoaderEnvInjectionMethod || *effectiveConfig.AgentEnvVarsInjectionMethod == common.LoaderFallbackToPodManifestInjectionMethod
    fallbackRequested := *effectiveConfig.AgentEnvVarsInjectionMethod == common.LoaderFallbackToPodManifestInjectionMethod

	if distroSupportsLoader && loaderRequested {
		err := isLoaderInjectionSupportedByRuntimeDetails(containerName, runtimeDetails)
		if err == nil {
			// loader is requested by config and distro, and supported by the runtime details.
			// thus, we can use the loader injection method in webhook.
			loaderInjectionMethod := common.LoaderEnvInjectionMethod
			return &loaderInjectionMethod, nil
		} else {
			// loader is requested by config and distro, but not supported by the runtime details.
			if !fallbackRequested {
				// config requires us to use loader when it is supported by distro,
				// thus we can't use it and need fail the injection.
				return nil, err
			} // else - we will not return and continue to the next injection method.
		}
	}
```

RonFed (approved):
Wrote few nits, overall looks great

api/odigos/v1alpha1/instrumentationconfig_types.go
blumamir: deleted 👍  and changed the text

instrumentor/controllers/agentenabled/sync.go
blumamir: It was there to begin with, but we actually have test just for that [here](https://github.com/odigos-io/odigos/blob/3b5c53c96ec4c3134936dfa06ff14fdd32941a0a/tests/e2e/workload-lifecycle/services/java-http-server/java-supported-docker-env.Dockerfile#L8). looks like you added it, so current implementation reserves current behavior. do you remember why you've added it?

instrumentor/controllers/agentenabled/sync.go
blumamir: User instructed us to use loader, but we can't follow this. It makes sense to me but I am open for other suggestions

instrumentor/controllers/agentenabled/sync.go
blumamir: yes, it's actually the error here. 

so if it's nil - no error.
if not nil - there was an error (cannot inject env into container) and the returned value is the error why we can't do it. it's just convenient to use the existing struct, but I agree it can be improved.

instrumentor/controllers/agentenabled/sync.go
blumamir: These names are consistent with what this function is checking (it uses "requested" instead of "loaderRequested || fallbackRequested", and it communicates in code that something is "required" and not just equals some enum value.

the function abstract these enums and uses a terminology which is more accurate to the calculation being performed IMO, let's discuss it face to face

instrumentor/controllers/agentenabled/sync.go
RonFed: its a nit:
In the other case where we have a user defined some other LD_PRELOAD value it make sense to set `AgentEnabledReason: Conflict`
This secure execution case seems to be something that is not a "conflict"

So what these 2 cases (user defined value and secure execution) have in common that we couldn't put the loader even though we were requested to, but this common thing is not expressed in the generic "conflict" enum value, maybe we can find better semantics for this.

instrumentor/controllers/agentenabled/sync.go
RonFed:
```Go
		runtimeDetailsVal, foundInInspection := getEnvVarFromRuntimeDetails(runtimeDetails, commonconsts.LdPreloadEnvVarName)
		ldPreloadUnsetOrExpected := !foundInInspection || strings.Contains(runtimeDetailsVal, odigosLoaderPath)
```

PR #3136, opened 2025-07-05
ci: create release PR

blumamir:
Experimenting with creating automatic releases from CI instead of doing it from local machines.
Currently mostly cursor code


blumamir:
this PR is adding a new workflow which currently does not touch odigos source or existing processes.
I need to merge it to test if it works in CI, so merging and will approve this once it consolidate to a stable state

PR #3137, opened 2025-07-05
ci: run create release PR with the relevant token

blumamir:
Add the token to avoid these errors in CI:

```
 * [new branch]        release/v1.0.206 -> release/v1.0.206
gh: To use GitHub CLI in a GitHub Actions workflow, set the GH_TOKEN environment variable. Example:
  env:
    GH_TOKEN: ${{ github.token }}
```


blumamir:
does not affect any existing processes. merging to test until it stabilizes

PR #3141, opened 2025-07-05
refactor: migrate ebpf instrumentation mangaer to distro

blumamir:
## Description

Part of the effort to get rid of the deprecated "OtelSdk" and migrate all usages to the new distro module with all distro related in one single file. 

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

No effect on k8s

## User Facing Changes

Internal refactor, no user facing changes


k8sutils/pkg/container/container.go
```diff
@@ -1,111 +1,11 @@
...
 	odigosv1 "github.com/odigos-io/odigos/api/odigos/v1alpha1"
-	"github.com/odigos-io/odigos/common"
-)
-
-var (
-	ErrDeviceNotDetected     = errors.New("device not detected")
-	ErrContainerNotInPodSpec = errors.New("container not found in pod spec")
 )
 
-func LanguageAndSdk(pod *v1.Pod, containerName string, distroName string) (common.ProgrammingLanguage, common.OtelSdk, error) {
```
blumamir, line 18: removing this function was the main motivation. many references to `common.OtelSdk` which we unlink 🎉

instrumentation/types.go
```diff
@@ -4,19 +4,10 @@ import (
...
-	"github.com/odigos-io/odigos/common"
+	"github.com/odigos-io/odigos/distros/distro"
 	"github.com/odigos-io/odigos/instrumentation/detector"
 )
 
-// OtelDistribution is a customized version of an OpenTelemetry component.
-// see https://opentelemetry.io/docs/concepts/distributions  and https://github.com/odigos-io/odigos/pull/1776#discussion_r1853367917 for more information.
-// TODO: This should be moved to a common root package, since it will require a bigger refactor across multiple components,
-// we use this local definition for now.
-type OtelDistribution struct {
```
blumamir, line 15: now the instrumentation module uses the exact same distro as odigos, so no need for duplication

odiglet/cmd/main.go
```diff
@@ -108,11 +108,8 @@ func deviceInjectionCallbacks() instrumentation.OtelSdksLsf {
...
 
-func ebpfInstrumentationFactories() map[commonInstrumentation.OtelDistribution]commonInstrumentation.Factory {
-	return map[commonInstrumentation.OtelDistribution]commonInstrumentation.Factory{
-		commonInstrumentation.OtelDistribution{
-			Language: common.GoProgrammingLanguage,
-			OtelSdk:  common.OtelSdkEbpfCommunity,
-		}: sdks.NewGoInstrumentationFactory(),
+func ebpfInstrumentationFactories() map[string]commonInstrumentation.Factory {
+	return map[string]commonInstrumentation.Factory{
+		"golang-community": sdks.NewGoInstrumentationFactory(),
```
RonFed, line 113: Can we do this without having the distro name here hard coded?
I guess we can have a field for each distro like `requiresEbpf` and only have those here but not sure if it is worth it.

blumamir: regarding the string, we should probably have consts for this in the distro package, but I don't want to change in this PR.
Regarding `requiresEbpf`, it's not only the boolean value, but also the implementation `NewGoInstrumentationFactory` which cannot be decoded into the yaml.

PR #3144, opened 2025-07-06
helm: set default GOMEMLIMIT for odiglet

AvihuHenya:
## Description
This PR introduces a default value for the GOMEMLIMIT environment variable (odiglet.goMemLimit) in the Helm chart to improve Go runtime memory management and provide safer defaults for users.

This PR introduces logic to set both the `resources` block and `GOMEMLIMIT` for the odiglet helm template:

`resources` logic:
If .resources.limits is set → use it as-is.
If only .resources.requests is set → use it and also copy it to .limits.
If neither is set → default both limits and requests to:

This ensures resource constraints are always present, even if the user doesn't explicitly configure them.

`GOMEMLIMIT` logic:
Derived from the `limits.memory` value (falling back to requests.memory, or 512Mi if unset).
Computes 80% of the memory using float support (e.g., 1024.5Mi is valid).
Ensures the output ends in a valid Go runtime format like MiB or GiB (adds B suffix if missing).


AvihuHenya:
> ### Bug: Memory Parsing Bug in `odigos.gomemlimitFromLimits`
> The `odigos.gomemlimitFromLimits` template incorrectly calculates `GOMEMLIMIT` for floating-point memory values. The `regexFind` pattern `^[0-9]+` and the `int` function only process the integer part of the memory string (e.g., "1024" from "1024.5Mi"), silently truncating any decimal portion. This leads to `GOMEMLIMIT` being calculated based on an incorrect, lower memory value, contradicting the claimed float support.
> 
> `helm/odigos/templates/_helpers.tpl#L76-L82`
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OmQ3ZmNkZTMzLTNmMTQtNDQ2NS1hMzc2LTVlY2I1M2U1NDBmZSIsImVuY3J5cHRpb25LZXkiOiJXOHBfSEQ0RHNpODVTWnpqeFo4dHpjaHJ4cVVxUVVmaGtLS19BNU9henNBIiwiYnJhbmNoIjoib2RpZ2xldC1nby1tZW0tbGltaXQifSwiaWF0IjoxNzUxOTY4MjA5LCJleHAiOjE3NTI1NzMwMDl9.V7X69q62MObF7CP-eB_zVxIktJTFO9Pc6c_7muU3xWh2OlTdf_RTHYadCv2yswtYAnzkCgtVbwbPAbNv3uA_lLshkcSF6seGvpjE4fp69lags8GE2oehdgJtnlkWnBqS0pEhVYFtkyUijDnaNcRsAAnAA1rt11e9bxtLPto-1hqA-SXBpan-zshMAm6Qd-vxUJM38H1g8r3lNqthdf2jMEyx6mg57GZ1_zF9fdKuIiWavElw7KS6fhjSyjOhxzNv3af7RXXcjCJ8t0GYwI07gT0cl6SV4Has_GMW1Z0_Cm3V8GL8rnmBJ_ExofERJMfUCpyg_2gxbjwolKEtJfrb1Q) • [Fix in Web](https://cursor.com/agents?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9XRUIiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OmQ3ZmNkZTMzLTNmMTQtNDQ2NS1hMzc2LTVlY2I1M2U1NDBmZSIsImVuY3J5cHRpb25LZXkiOiJXOHBfSEQ0RHNpODVTWnpqeFo4dHpjaHJ4cVVxUVVmaGtLS19BNU9henNBIiwiYnJhbmNoIjoib2RpZ2xldC1nby1tZW0tbGltaXQiLCJyZXBvT3duZXIiOiJvZGlnb3MtaW8iLCJyZXBvTmFtZSI6Im9kaWdvcyIsInByTnVtYmVyIjozMTQ0LCJjb21taXRTaGEiOiIzMzc2NTFjMjgwODVhNzMwNGNlZTNjOGZiMTQ2OTRhODg4YTk2YjNjIn0sImlhdCI6MTc1MTk2ODIwOSwiZXhwIjoxNzUyNTczMDA5fQ.f3MpEPqaAvw3oIiMu6pkpNekCbzCWsFeJy3ykOxUWvJrlCwU4zfABpsgG0-DemJeaCMkPpeO71vYVc7O6KS3ppAB4beuPpcZ8P7N8QuFoinvEMlWEeQIdtUEsZimdIMF0-xf7Kf84B6tz-Bhepp_z7eDOsbrJxntH31o37IlHXjWuUOYchfH2VxfgmjeIiM4pp47ZRle-YMUdNVK1DXWwIxLt0TJgAQiaLMCUyzGpyiFrPOt34X7RqGejh3JT9FuOf0ZdSprg0C9oAFttSJA1hg1SD02oHOu8IDUNVufyQ5agdasnKYSZe-7vFpSSbE7qACFv4o-l1WALp-Tu8iycA)
> 
> _Was this report helpful? Give feedback by reacting with 👍 or 👎_

Since I want `GOMEMLIMIT` to be calculated based on `resources.limits`, and resource values must be integers, I think we can ignore the float case.

helm/odigos/templates/_helpers.tpl
```diff
@@ -44,3 +44,46 @@ true
...
+{{- end }}
+{{- if and (empty $limits) (empty $requests) -}}
+  {{- $_ := set $resources "limits" $defaults -}}
+  {{- $_ := set $resources "requests" $defaults -}}
+{{- end }}
+{{- toYaml $resources | indent 12 }}
+{{- end }}
+
+
+{{- define "odigos.gomemlimitFromLimits" -}}
```
blumamir: this function is specifically for odiglet, right? 
please add `odiglet` to it's name or make it generic (accept resources as input)

helm/odigos/templates/_helpers.tpl
```diff
@@ -44,3 +44,46 @@ true
...
+
+{{- $resources := .Values.odiglet.resources | default dict -}}
+{{- $limits := get $resources "limits" | default dict -}}
+{{- $requests := get $resources "requests" | default dict -}}
+
+{{/* Use limits.memory if set, otherwise fallback to requests.memory, or default to 512Mi */}}
+{{- $raw := get $limits "memory" | default (get $requests "memory" | default "512Mi") | trim -}}
```
blumamir: this one can get out of sync with the value set above. please use the same definition so they both are calculated from the same source

helm/odigos/templates/_helpers.tpl
```diff
@@ -44,3 +44,46 @@ true
...
+
+{{- $number := regexFind "^[0-9]+" $raw -}}
+{{- $unit := regexFind "[a-zA-Z]+$" $raw -}}
+
+{{- if and $number $unit }}
```
blumamir, line 76: else? should we report an error?

helm/odigos/templates/_helpers.tpl
```diff
@@ -44,3 +44,46 @@ true
...
+  {{- $num := int $number -}}
+  {{- $val := div (mul $num 80) 100 -}}
+  {{- if hasSuffix $unit "B" -}}
```
blumamir: why is this here?
please add a comment explaining the exception and why it's needed

docs/setup/docker-registry.mdx
```diff
@@ -171,7 +171,7 @@ odigos install --image-prefix=$CUSTOM_DOCKER_REGISTRY
 ```bash
 helm repo update
 helm upgrade --install odigos odigos/odigos --namespace odigos-system --create-namespace --set imagePrefix=$CUSTOM_DOCKER_REGISTRY
-kubectl label namespace odigos-system odigos.io/system-object="true"
+kubectl label namespace odigos-test odigos.io/system-object="true"
```
blumamir: remove this change :)

helm/odigos/templates/_helpers.tpl
```diff
@@ -44,3 +44,43 @@ true
...
+  {{- $_ := set $resources "limits" $defaults -}}
+  {{- $_ := set $resources "requests" $defaults -}}
+{{- end }}
+
+{{- toYaml $resources -}}
+{{- end }}
+
+{{- define "odigos.odiglet.resources" -}}
+{{- include "odigos.odiglet.resolvedResources" . | indent 12 }} 
+{{- end }}
```
blumamir: do we really need a function for that? can we just call `odigos.odiglet.resolvedResources` directly from the template?

helm/odigos/templates/_helpers.tpl
```diff
@@ -44,3 +44,43 @@ true
...
+{{- $requests := get $resources "requests" -}}
+
+{{- $raw := (get $limits "memory") | default (get $requests "memory") -}}
+{{- $number := regexFind "^[0-9]+" $raw -}}
+{{- $unit := regexFind "[a-zA-Z]+$" $raw -}}
+
+{{- if and $number $unit }}
+  {{- $num := int $number -}}
+  {{- $val := div (mul $num 80) 100 -}}
+  {{- printf "%d%sB" $val $unit -}}
```
blumamir, line 83: Do we always have to add the `B` here? what if the unit already contains `B`?

blumamir: Also, please add a comment about your findings with GOMEMLIMIT supported units, so next person will also know it :)

helm/odigos/templates/odiglet/daemonset.yaml
```diff
@@ -84,6 +83,26 @@ spec:
...
+              port: {{ .Values.odiglet.readinessAndLivenessProbePort }}
+            initialDelaySeconds: 15
+            periodSeconds: 20
+          readinessProbe:
+            httpGet:
+              path: /readyz
+              port: {{ .Values.odiglet.readinessAndLivenessProbePort }}
+            periodSeconds: 10
+          resources:
+          resources:
```
blumamir: this line is here twice

PR #3149, opened 2025-07-07
feat: under memory pressure, reject data in gateway collector at receiver time

blumamir:
## Description

See detailed description in collector/config/configgrpc/README.md

This is a fix to avoid decoding new items in otlp receiver if the collector is under memory pressure and must refuse data.
Doing it as a processor is too late, as at the request has already spent memory which can exhaust the safety buffer that protect collector from OOM

## How Has This Been Tested?

- [x] Added Unit Tests
- [x] Manual Testing
- [x] Manual Load Test

## Kubernetes Checklist

No changes in k8s api or manifest

## User Facing Changes

The collectors should be more stable and not crash due to OOM in extreme traffic and high load, and when downstream is slow or faulty


collector/config/configgrpc/Makefile
```diff
@@ -0,0 +1 @@
+include ../../Makefile.Common
```
blumamir, line 1: note: this belongs to the upstream repo, and has no effect in odigos repo. It is here only to make future upgrades easier, so there are minimum changes between us and upstream

collector/builder-config.yaml
```diff
@@ -129,3 +130,4 @@ replaces:
   - go.opentelemetry.io/collector/confmap/provider/odigosk8scmprovider => ../providers/odigosk8scmprovider
   - github.com/odigos-io/odigos/collector/connectors/odigosrouterconnector => ../connectors/odigosrouterconnector
   - github.com/odigos-io/odigos/common => ../../common
+  - go.opentelemetry.io/collector/config/configgrpc => ../config/configgrpc
```
RonFed, line 133: nit: maybe worth to add a comment here as well

RonFed (approved):

collector/config/configgrpc/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+Once the collector adopts the new "middlewares" approach, this module can be retired and removed.
+
+## Maintenance
+
+### How to Update
+
+The code in this module is copied from the collector repository using the current tag (`v0.126.0` as of this writing). When upgrading the collector version, this module must also be updated accordingly.
+
+- Remove mono-repo replaces in `go.mod` as they confuse the `make go-mod-tidy`
```
damemi: can you explain more what this means?

collector/config/configgrpc/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+The OpenTelemetry Collector SIG is working on refactoring how receiver middlewares are applied. As of 06/07/2025, this is still under development with several draft PRs, including [PR #13265](https://github.com/open-telemetry/opentelemetry-collector/pull/13265).
+
+Once the collector adopts the new "middlewares" approach, this module can be retired and removed.
+
+## Maintenance
+
+### How to Update
+
+The code in this module is copied from the collector repository using the current tag (`v0.126.0` as of this writing). When upgrading the collector version, this module must also be updated accordingly.
```
damemi, line 50: would it be possible to wrap the changes around the upstream module and import it to apply the extra functionality on top?

collector/config/configgrpc/README.md
blumamir: updated the docs. thanks

collector/config/configgrpc/README.md
blumamir: maybe, but it's all temporary and I want to first validate this even fixes the issue :) let's delay it to future work once we have confidence in this fix

PR #3193, opened 2025-07-16
refactor: migrate LD_PRELOAD injeciton from otelsdk

blumamir:
## Description

#3135  added the env injection method into the instrumentation config agent containers section in the spec.
now it's calculated in reconcile, and webhook do not need to choose injection method anymore.

This PR migrate the LD_PRELOAD injection to use the new field. 
pod manifest will be handled in a followup PRs.

After this PR, we removed some code from `webhook_env_injector.go` which uses the deprecated otelsdk.

## How Has This Been Tested?

This is an internal refactor which is covered by env-injection e2e tests.

## Kubernetes Checklist

No changes in k8s

## User Facing Changes

Internal refactor

### DIFF
--- a/.github/workflows/create-patch.yml
+++ b/.github/workflows/create-patch.yml
@@ -111,16 +111,11 @@ jobs:
                 --body "This is an automated PR for the patch release ${{ steps.calculate_patch_version.outputs.patch_version }}"
             fi
 
-      - name: Notify Slack Success
-        if: success()
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.ODIGOS_RELEASE_STATUS_WEBHOOK_URL }}
-        run: |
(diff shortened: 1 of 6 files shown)


instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -202,6 +204,24 @@ func (p *PodsWebhook) injectOdigosInstrumentation(ctx context.Context, pod *core
...
+			// container config exists, but no agent should be injected by webhook to this container
+			continue
+		}
+		if containerConfig.EnvInjectionMethod != nil &&
+			(*containerConfig.EnvInjectionMethod == common.LoaderFallbackToPodManifestInjectionMethod ||
+				*containerConfig.EnvInjectionMethod == common.LoaderEnvInjectionMethod) {
+			// lodaer has been migrated to use distros.
+			// we only process "pod-manifest" containers here.
+			continue
+		}
```
blumamir: this is ugly but temporary, we will get rid of it in future PR.
I don't want this one to change tons of things in one huge PR

tamirdavid1 (approved):

PR #3199, opened 2025-07-19
chore(odiglet): simplify device plugin container in odiglet

blumamir:
## Description

Odigos has 2 modes of mounting the `/var/odigos` directory into instrumented containers: "k8s-host-path" and "virtual-device".

The "virtual device" is the default one, and used by most of odigos users. Instrumentor webhook will add a "resource" request on instrumented containers with the name of the device, and then k8s scheduler will consider only nodes with this device as relevant for instrumentation.

This mechanism is important for the stability of the cluster as it affects k8s scheduler.

This PR simplifies the "device-plugin" container, removing potential failure points from it that are no longer used or can be refactored,, and reducing our failure surface for this critical component.

- Remove unused devices. enterprise still uses the dotnet legacy device. after it is remove we can clean this up even further.
- remove the query to get `enabled signals` from the `collectors group`. it was used by java virtual device which has migrated to generic many months ago and is being removed in this PR.
- remove `collectors group` RBAC permissions for odiglet as they are no longer needed.
- fallback to device amount of `1024` instead of setting it to allocatable pods. the device is used per container, not per pod, and the query is another fail point. we don't have an actual limit so using a static number is good enough.

## How Has This Been Tested?

Run it manually and made sure for the changes in the manifest. saw apps starting and traces.
this is a refactor and is tested in e2e.

## Kubernetes Checklist

removed many devices which will not be registered to the node anymore.

~~ [ ] Introduces additional calls to the API Server (potential performance impact)~~
~~ [ ] New Query/feature supported in all the k8s versions supported by Odigos~~
~~ [] Modifies Odigos manifests (addressed in both CLI and Helm)
removed `CollectrosGroup` from odiglet as it is not used anymore

## User Facing Changes

Internal refactor. removing unused things. no impact on user


deviceplugin/pkg/instrumentation/lister.go
```diff
@@ -48,22 +46,15 @@ func (l *lister) NewPlugin(s string) dpm.PluginInterface {
 // Otel SDKs frequently requires to set some environment variables and mount some fs dirs for it to work.
 type OtelSdksLsf map[common.ProgrammingLanguage]map[common.OtelSdk]LangSpecificFunc
```
blumamir, line 40: note: this is still being used in enterprise thus not removed completely.
will be removed once we retire dotnet-legacy hopefully soon

common/instrumentationdevice.go
```diff
@@ -31,29 +29,3 @@ func InstrumentationPluginName(language ProgrammingLanguage, otelSdk OtelSdk, li
...
-	return OdigosInstrumentationDevice(OdigosResourceNamespace + "/" + pluginName)
-}
-
-func InstrumentationDeviceNameToComponents(deviceName string) (ProgrammingLanguage, OtelSdk) {
-	pluginName := strings.Split(deviceName, "/")[1]
-	return InstrumentationPluginNameToComponents(pluginName)
-}
-
-func IsResourceNameOdigosInstrumentation(resourceName string) bool {
-	return strings.HasPrefix(resourceName, OdigosResourceNamespace)
```
blumamir, line 58: these are not in use.
we should have cleaned them up before but I just discovered it and added it here on the fly

deviceplugin/pkg/instrumentation/lister.go
```diff
@@ -48,22 +46,15 @@ func (l *lister) NewPlugin(s string) dpm.PluginInterface {
...
 
-func NewLister(ctx context.Context, clientset *kubernetes.Clientset, otelSdksLsf OtelSdksLsf) (dpm.ListerInterface, error) {
-	maxPods, err := getInitialDeviceAmount(clientset)
```
RonFed, line 52: we can remove `getInitialDeviceAmount` if it is not used anymore

RonFed (approved):
🥇 
left one small comment

edeNFed (approved):
Lgtm but I would wait for Tamir to review as well

PR #3203, opened 2025-07-20
Publish image with agents

tamirdavid1:
## Description

This PR introduces a new component called init-container, which supports a new mount method. When this mount method is selected (in a separate PR), the webhook will inject the init-container as an initContainer in the instrumented pod. The init container copies the required agent files into a shared volume for the application containers to use.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/.github/workflows/publish-modules.yml
+++ b/.github/workflows/publish-modules.yml
@@ -254,7 +254,7 @@ jobs:
     strategy:
       matrix:
         dockerfile: ['Dockerfile', 'Dockerfile.rhel']
-        service: ['autoscaler', 'scheduler', 'instrumentor', 'collector', 'odiglet', 'ui', 'operator']
+        service: ['autoscaler', 'scheduler', 'instrumentor', 'collector', 'odiglet', 'ui', 'operator', 'init-container']
         include:
           - service: autoscaler
             runner: ubuntu-latest
@@ -284,6 +284,10 @@ jobs:
             runner: ubuntu-latest
             summary: 'Odigos Operator'
             description: 'The Odigos Operator installs and manages Odigos in a cluster'
+          - service: init-container
+            runner: ubuntu-latest
+            summary: 'Odigos Init Container'
+            description: 'The Odigos Init Container is used to inject the Odigos agent into the user workloads.'
     runs-on: ${{ matrix.runner }}
     steps:
       - name: Checkout repository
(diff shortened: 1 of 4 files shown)


.github/workflows/publish-modules.yml
```diff
@@ -284,6 +284,10 @@ jobs:
             runner: ubuntu-latest
             summary: 'Odigos Operator'
             description: 'The Odigos Operator installs and manages Odigos in a cluster'
+          - service: init-container
```
blumamir: it's a very generic name which can potentially confuse people.

How about `odigos-agents` or `otel-distributions` or something along those lines that explains what is inside?

init-container/Dockerfile
```diff
@@ -0,0 +1,170 @@
...
+WORKDIR /src
+ARG NODEJS_AGENT_VERSION=main
+RUN git clone https://github.com/odigos-io/opentelemetry-node.git && cd opentelemetry-node && git checkout $NODEJS_AGENT_VERSION
+
+# The second build stage 'nodejs-agent-src' prepares the actual code we are going to compile and embed in odiglet.
+# By default, it uses the previous 'nodejs-agent-clone' stage, but one can override it by setting the
+# --build-context nodejs-agent-src=../opentelemetry-node flag in the docker build command.
+# This allows us to use the agent sources and test changes during development.
+# The output of this stage is the resolved source code to be used in the next stage.
+FROM scratch AS nodejs-agent-src
```
blumamir: this was built for the odiglet. we can reuse it here also but it requires changes to the makefile :)

.github/workflows/publish-modules.yml
```diff
@@ -254,7 +254,7 @@ jobs:
     strategy:
       matrix:
         dockerfile: ['Dockerfile', 'Dockerfile.rhel']
-        service: ['autoscaler', 'scheduler', 'instrumentor', 'collector', 'odiglet', 'ui', 'operator']
+        service: ['autoscaler', 'scheduler', 'instrumentor', 'collector', 'odiglet', 'ui', 'operator', 'init-container']
```
damemi: note to me: this will need a new component in the red hat catalog

damemi (commented):
@tamirdavid1 is this going to be an option that someone can configure? Want to know if this will need any operator updates when it's added

tamirdavid1:
> @tamirdavid1 is this going to be an option that someone can configure? Want to know if this will need any operator updates when it's added

@damemi yea it will be another mount-method, i'll share the OSS PR soon

PR #3207, opened 2025-07-21
feat: enbale k8s_cluster metrics via config in the helm

tamirdavid1:
## Description

To expand Odigos' support for metrics, I’ve added integration for the k8s_cluster receiver, which collects cluster-level metrics and entity events directly from the Kubernetes API server.
Enabling this feature will expose the following additional metrics:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/k8sclusterreceiver/metadata.yaml#L226

Currently, users can enable this feature only via the Helm chart, as we agreed to release it quickly for an initial user request.
## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


helm/odigos/templates/autoscaler/clusterrole.yaml
```diff
@@ -46,4 +46,4 @@ rules:
     resourceNames:
       - action-validating-webhook-configuration
     verbs:
-      - update
```
damemi, line 49: only role change looks like an added line here, can skip operator rbac check 👍

damemi (approved):

common/odigos_config.go
```diff
@@ -105,6 +105,10 @@ type CollectorGatewayConfiguration struct {
 	// ServiceGraphDisabled is a feature that allows you to visualize the service graph of your application.
 	// It is enabled by default and can be disabled by setting the disabled flag to true.
 	ServiceGraphDisabled *bool `json:"serviceGraphDisabled,omitempty"`
+
+	// ClusterMetricsEnabled is a feature that allows you to enable the cluster metrics.
+	// It is disabled by default and can be enabled by setting the enabled flag to true.
```
blumamir: Consider adding a link explaining what "ClusterMetrics" means, and that it is relevant only when metrics destination is present

PR #3212, opened 2025-07-24
CORE-98 Init container support

tamirdavid1:
## Description

This PR adds support for a new mount-method called `k8s-init-container`, providing an alternative to the existing hostPath and device plugin-based methods. It was designed to address user concerns around scheduling issues that may occur when relying on device plugins or node affinity, particularly in cases where Odiglet is unavailable or fails to run on certain nodes.

With this method, Odigos injects an InitContainer into the instrumented workload, which copies the required instrumentation files into a shared emptyDir volume. This eliminates the dependency on node-level file mounting and improves scheduling flexibility.

The webhook logic has been updated to skip node affinity and device settings when this method is selected. While this approach offers greater isolation and resilience, it introduces a slight delay in pod startup and additional disk usage per pod.

This method is opt-in and not enabled by default. It can be configured via the CLI, Helm chart, or directly in the odigos-configuration ConfigMap.

- [ ] Added Unit Tests
- [X] Updated e2e Tests
- [X] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


tests/e2e/mount-method/simple-demo-instrumented.yaml
```diff
@@ -0,0 +1,374 @@
...
+  containers:
+    - name: membership
+status:
+  containerStatuses:
+    - name: membership
+      ready: true
+      restartCount: 0
+      started: true
+  phase: Running
+---
```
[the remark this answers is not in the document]
tamirdavid1: This is okay because membership is go application and we're not copy any files to it.

RonFed (approved):
🚀

instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -358,3 +386,50 @@ func getRuntimeInfoForContainerName(ic *odigosv1.InstrumentationConfig, containe
 	// if both are not found, return we don't have runtime info for this container
 	return nil
 }
+
+func CreateInitContainer(pod *corev1.Pod, dirsToCopy map[string]struct{}, config common.OdigosConfiguration) {
+	const (
+		instrumentationsPath = "/instrumentations"
+	)
+	imageVersion := os.Getenv(consts.OdigosVersionEnvVarName)
+	imageName := config.ImagePrefix + "/" + k8sconsts.OdigosInitContainerName + ":" + imageVersion
```
blumamir: in other places we are checking more cases: https://github.com/odigos-io/odigos/blob/62539a6cc14dd4ea74fd6f24e510192a80ca320a/cli/pkg/containers/name.go#L8

should we consolidate them (maybe in k8sutils?)

blumamir: will it work for rhel? (ubi9 variant)

instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -239,9 +263,9 @@ func (p *PodsWebhook) injectOdigosToContainer(containerConfig *odigosv1.Containe
 	// inject various kinds of distro environment variables
 	existingEnvNames, err = podswebhook.InjectStaticEnvVarsToPodContainer(existingEnvNames, podContainerSpec, distroMetadata.EnvironmentVariables.StaticVariables, containerConfig.DistroParams)
 	if err != nil {
-		return false, err
+		return false, nil, err
 	}
-	existingEnvNames = podswebhook.InjectOdigosK8sEnvVars(existingEnvNames, podContainerSpec, distroName, pw.Namespace)
+	existingEnvNames = podswebhook.InjectOdigosK8sEnvVars(existingEnvNames, podContainerSpec, distroMetadata.Name, pw.Namespace)
```
blumamir: any reason for this change?

instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -256,10 +280,12 @@ func (p *PodsWebhook) injectOdigosToContainer(containerConfig *odigosv1.Containe
...
 
 	volumeMounted := false
+	containerDirsToCopy := make(map[string]struct{})
 	if distroMetadata.RuntimeAgent != nil {
-		if *config.MountMethod == common.K8sHostPathMountMethod {
-			// mount directory only if the mount type is host-path
+		if *config.MountMethod == common.K8sHostPathMountMethod || *config.MountMethod == common.K8sInitContainerMountMethod {
+			// mount directory only if the mount type is host-path or init container
 			for _, agentDirectoryName := range distroMetadata.RuntimeAgent.DirectoryNames {
+				containerDirsToCopy[agentDirectoryName] = struct{}{}
```
blumamir, line 291: the `containerDirsToCopy` is only for `init-container` but it will also be populated for `host-path`.

How about calling it `mountedAgentDirectories`, so it describe more precisely what it's role in the context of this function

instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -224,13 +247,14 @@ func (p *PodsWebhook) injectOdigosInstrumentation(ctx context.Context, pod *core
 	return nil
 }
 
-func (p *PodsWebhook) injectOdigosToContainer(containerConfig *odigosv1.ContainerAgentConfig, podContainerSpec *corev1.Container, pw k8sconsts.PodWorkload, serviceName string, config common.OdigosConfiguration) (bool, error) {
+func (p *PodsWebhook) injectOdigosToContainer(containerConfig *odigosv1.ContainerAgentConfig, podContainerSpec *corev1.Container,
+	pw k8sconsts.PodWorkload, serviceName string, config common.OdigosConfiguration) (bool, map[string]struct{}, error) {
```
blumamir, line 254: consider removing the first returned value `volumeMounted` as it now redundant (if the map is not empty it means a volume is mounted)

instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -268,12 +294,14 @@ func (p *PodsWebhook) injectOdigosToContainer(containerConfig *odigosv1.Containe
...
 					*config.AgentEnvVarsInjectionMethod == common.LoaderEnvInjectionMethod) {
+				containerDirsToCopy[filepath.Join(distro.AgentPlaceholderDirectory, consts.OdigosLoaderDirName)] = struct{}{}
 				podswebhook.MountDirectory(podContainerSpec, filepath.Join(k8sconsts.OdigosAgentsDirectory, consts.OdigosLoaderDirName))
 				volumeMounted = true
 			}
 		}
+
 		if distroMetadata.RuntimeAgent.K8sAttrsViaEnvVars {
-			podswebhook.InjectOtelResourceAndServiceNameEnvVars(existingEnvNames, podContainerSpec, distroName, pw, serviceName)
+			podswebhook.InjectOtelResourceAndServiceNameEnvVars(existingEnvNames, podContainerSpec, distroMetadata.Name, pw, serviceName)
```
blumamir: nit: also here, what is the motivation for the change?

`distroName` is a bit more readable IMO

instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -358,3 +386,50 @@ func getRuntimeInfoForContainerName(ic *odigosv1.InstrumentationConfig, containe
 	// if both are not found, return we don't have runtime info for this container
 	return nil
 }
+
+func CreateInitContainer(pod *corev1.Pod, dirsToCopy map[string]struct{}, config common.OdigosConfiguration) {
```
blumamir: This is a private function

```suggestion
func createInitContainer(pod *corev1.Pod, dirsToCopy map[string]struct{}, config common.OdigosConfiguration) {
```

instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -358,3 +386,50 @@ func getRuntimeInfoForContainerName(ic *odigosv1.InstrumentationConfig, containe
...
+	const (
+		instrumentationsPath = "/instrumentations"
+	)
+	imageVersion := os.Getenv(consts.OdigosVersionEnvVarName)
+	imageName := config.ImagePrefix + "/" + k8sconsts.OdigosInitContainerName + ":" + imageVersion
+
+	if len(dirsToCopy) == 0 {
+		return
+	}
```
blumamir: should we check this before calling the function? so we also know to skip other related mutations

blumamir (approved):

instrumentor/controllers/agentenabled/pods_webhook.go
tamirdavid1: @damemi can i have your comment on this?
meanwhile i'm merging this PR but let me know if any adjustments needs to be done to support ubi

damemi: good catch @blumamir, no I think you'll need a way to handle the -ubi9 suffix when it should be set. I can't remember what we do with the autoscaler collector images but it might be similar to what you need

tamirdavid1: Using it now also

tamirdavid1: @damemi and @blumamir i've added support for ubi-9 images

PR #3213, opened 2025-07-24
CLI Command: Odigos Describe Config

Jordan-G-CS:
## How Has This Been Tested?

- [ ] via command ./cli/odigos/ describe config

## User Facing Changes

- [ ] Users now has access to the odigos-config configmap with this  

for @damemi for review

### DIFF
--- a/cli/cmd/describe.go
+++ b/cli/cmd/describe.go
@@ -11,6 +11,13 @@ import (
 	"github.com/odigos-io/odigos/cli/pkg/kube"
 	"github.com/odigos-io/odigos/k8sutils/pkg/describe"
 	"github.com/spf13/cobra"
+
+	// new ones for the odigos describe config command
+	"github.com/odigos-io/odigos/cli/pkg/log"
+	//"github.com/odigos-io/odigos/common"
+	"github.com/odigos-io/odigos/common/consts"
+	// "github.com/odigos-io/odigos/cli/cmd/resources/odigospro"
+	// "github.com/odigos-io/odigos/cli/cmd/resources/resourcemanager"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


cli/cmd/describe.go
```diff
@@ -101,6 +104,204 @@ var describeSourceCmd = &cobra.Command{
...
+		}
+
+		log.Print(fmt.Sprintf("- %s: Time windows where the auto rollback can happen [default: 1h]. status: %d.\n", consts.OdigletHealthProbeBindPortProperty, config.OdigletHealthProbeBindPort))
+
+		if config.CollectorGateway == nil {
+			log.Print(fmt.Sprintf("- %s: Enable or disable the service graph feature [default: false]. status: not set\n", consts.ServiceGraphConnectorName))
+		} else if config.CollectorGateway.ServiceGraphDisabled == nil {
+			log.Print(fmt.Sprintf("- %s: Enable or disable the service graph feature [default: false]. status: not set\n", consts.ServiceGraphConnectorName))
+		} else {
+			log.Print(fmt.Sprintf("- %s: Enable or disable the service graph feature [default: false]. status: %t\n", consts.ServiceGraphConnectorName, *config.CollectorGateway.ServiceGraphDisabled))
```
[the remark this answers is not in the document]
Jordan-G-CS: fixed

common/consts/consts.go
```diff
@@ -99,6 +99,73 @@ const (
...
+	AutomaticRolloutDisabledProperty:   "Disable auto rollout feature for workloads when instrumenting or uninstrumenting.",
+	OidcTenantUrlProperty:              "Sets the URL of the OIDC tenant.",
+	OidcClientIdProperty:               "Sets the client ID of the OIDC application.",
+	OidcClientSecretProperty:           "Sets the client secret of the OIDC application.",
+	OdigletHealthProbeBindPortProperty: "Sets the port for the Odiglet health probes (readiness/liveness).",
+	ServiceGraphDisabledProperty:       "Enable or disable the service graph feature [default: false].",
+}
+
+// var ConfigValues map[string]interface{}
+var ConfigValues = map[string]interface{}{}
```
damemi: don't declare variables in this package that you're going to write at runtime. this package is for constants that should only be read, not written to. it's also not a good practice to be writing to variables that are in a different package, this should just be a local variable in your function

cli/cmd/describe.go
```diff
@@ -185,6 +219,206 @@ var describeSourceStatefulSetCmd = &cobra.Command{
...
+
+func printPointerFileStuff(featureSetting *common.MountMethod, featureName string) {
+	if featureSetting != nil {
+		log.Print(fmt.Sprintf("- %s: %s status: %s\n", featureName, consts.ConfigDisplay[featureName], *featureSetting))
+	} else {
+		log.Print(fmt.Sprintf("- %s: %s status: not set\n", featureName, consts.ConfigDisplay[featureName]))
+	}
+}
+
+func printPointerFileStuff2(featureSetting *common.EnvInjectionMethod, featureName string) {
```
damemi: try to use more descriptive function names and avoid `stuff`,`1`,`2`,etc

cli/cmd/config.go
```diff
@@ -27,69 +27,16 @@ import (
...
-		consts.OdigletHealthProbeBindPortProperty,
-		consts.ServiceGraphDisabledProperty,
-	),
+	Long:  "Displays information about features in odigos map, and options the user has to turn certain features on or off, or add things, etc.",
+	Run: func(cmd *cobra.Command, args []string) {
+
+		log.Print(`Manage Odigos configuration settings to customize system behavior.` + "\n")
+		log.Print(`Configurable properties:` + "\n")
+
+		for _, key := range consts.ConfigDisplayOrder {
```
damemi: instead of defining the displayorder as its own list, try just sorting the keys in the `consts.ConfigDisplay` map, then loop through the sorted list and call `consts.ConfigDisplay[key]`. that way we don't have to maintain this list in 2 spots

cli/cmd/describe.go
Jordan-G-CS: Fixed, please let me know if they are descriptive enough

cli/cmd/config.go
Jordan-G-CS: fixed and got rid of configdisplayorder

common/consts/consts.go
Jordan-G-CS: ok, made the new one in the describe.go file

common/odigos_config.go
```diff
@@ -1,5 +1,241 @@
...
+
+func (data *ConfigCollectorNode) ToString() {
+	if data == nil {
+		fmt.Println(": not set")
+		return
+	} else {
+		var placeholder ConfigString = ConfigString(data.K8sNodeLogsDirectory)
+		placeholder.ToString()
+	}
+}
```
damemi: we don't need to redefine our existing custom structs to add `ToString()` functions to them. You can just add `ToString` to the `CollectorNodeConfiguration` struct directly (same with the other types below)

common/odigos_config.go
```diff
@@ -1,5 +1,241 @@
...
+
+// go doesn't allow type to deal with pointers, so it is wrapping around MountMethod
```
damemi: can you explain what you mean by this comment?

common/odigos_config.go
```diff
@@ -1,5 +1,241 @@
...
+
+type ConfigString string
+
+func (data ConfigString) ToString() {
+	if data == "" {
+		fmt.Println(": not set")
+	} else {
+		fmt.Printf(": %s\n", data)
+	}
+}
```
damemi: the `else` in almost all of these is unnecessary if you just return from the first `if`. For example

```go
func (data ConfigString) ToString() {
  if data == "" {
    fmt.Println(": not set")
    return
  }
  fmt.Printf(": %s\n", data)
}

```

common/odigos_config.go
```diff
@@ -1,5 +1,241 @@
...
+		fmt.Printf(": not set\n")
+	} else {
+		fmt.Printf(": %v\n", *data.Value)
+	}
+}
+
+type ConfigInt int
+
+func (data ConfigInt) ToString() {
+	fmt.Printf(": %d\n", data)
```
damemi: shift the colon+space out of each of these print lines, so you don't have to repeat it everywhere and have this odd formatting pattern. it should just be on the main print below (see my other comment). Then each ToString only needs to print the value

common/odigos_config.go
```diff
@@ -1,5 +1,241 @@
...
+
+	var order []string
+	for k := range consts.ConfigDisplay {
+		order = append(order, k)
+	}
+
+	sort.Strings(order)
+
+	for _, key := range order {
+		fmt.Printf("	- %s", key)
```
damemi:
```suggestion
		fmt.Printf("	- %s: ", key)
```

add the colon+space here so you only need to write it once

common/odigos_config.go
```diff
@@ -1,5 +1,241 @@
...
+		fmt.Printf(": not set\n")
+	} else {
+		var placeholder ConfigString = ConfigString(data.ClientSecret)
+		placeholder.ToString()
+	}
+}
+
+func makeAMap(config *OdigosConfiguration) map[string]ConfigField {
+	displayData := map[string]ConfigField{
+		consts.TelemetryEnabledProperty:           ConfigBool(config.TelemetryEnabled),
```
damemi: instead of doing all these casts, update the actual types used in the OdigosConfig to the new redefined types. For our own custom structs, you can just add the ToString method directly to those types, so you don't even need to change them in the OdigosConfig (just add the function)

damemi (changes requested):

common/odigos_config.go
Jordan-G-CS: fixed

common/odigos_config.go
Jordan-G-CS: fixed

common/odigos_config.go
Jordan-G-CS: fixed

common/odigos_config.go
Jordan-G-CS: essentially, you get this kind of error
invalid receiver type ConfigMountMethod (pointer or interface type)compilerInvalidRecv
it won't be permitted
comments were made in code

Jordan-G-CS: comments were erased because it is redundant, i made your changes of adding ToString() to the actual structs instead of redefining them

common/odigos_config.go
Jordan-G-CS: fixed, however i kept it for the OIDC configuration struct because it is how i can differentiate them. it i make 3 ToString() functions all with the same input, i will get an error, so in order to differentiate them, the redefining works

common/odigos_config.go
damemi: Can you just define this function as

```go
func (data *MountMethod) ToString() {
  ...
}
```

?

common/odigos_config.go
```diff
@@ -1,5 +1,239 @@
...
+		consts.AgentEnvVarsInjectionMethod:        (*ConfigEnvInjection)(config.AgentEnvVarsInjectionMethod),
+		consts.NodeSelectorProperty:               ConfigNodeSelector(config.NodeSelector),
+		consts.KarpenterEnabledProperty:           ConfigBoolPointer{Value: karpenterValue},
+		consts.RollbackDisabledProperty:           ConfigBoolPointer{Value: config.RollbackDisabled},
+		consts.RollbackGraceTimeProperty:          ConfigString(config.RollbackGraceTime),
+		consts.RollbackStabilityWindow:            ConfigString(config.RollbackStabilityWindow),
+		consts.AutomaticRolloutDisabledProperty:   ConfigBoolPointer{Value: rolloutValue},
+		consts.OidcTenantUrlProperty:              (*ConfigOidcTenant)(config.Oidc),
+		consts.OidcClientIdProperty:               (*ConfigOidcClientId)(config.Oidc),
+		consts.OidcClientSecretProperty:           (*ConfigOidcClientSecret)(config.Oidc),
```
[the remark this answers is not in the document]
Jordan-G-CS: needed because i first have to do a nil check for the struct, otherwise an error

common/odigos_config.go
```diff
@@ -1,5 +1,241 @@
...
+		consts.RollbackGraceTimeProperty:          ConfigString(config.RollbackGraceTime),
+		consts.RollbackStabilityWindow:            ConfigString(config.RollbackStabilityWindow),
+		consts.AutomaticRolloutDisabledProperty:   (*ConfigRollout)(config.Rollout),
+		consts.OidcTenantUrlProperty:              (*ConfigOidcTenant)(config.Oidc),
+		consts.OidcClientIdProperty:               (*ConfigOidcClientId)(config.Oidc),
+		consts.OidcClientSecretProperty:           (*ConfigOidcClientSecret)(config.Oidc),
+		consts.OdigletHealthProbeBindPortProperty: ConfigInt(config.OdigletHealthProbeBindPort),
+		consts.ServiceGraphDisabledProperty:       (*ConfigCollectorGateway)(config.CollectorGateway),
+		consts.GoAutoOffsetsCronProperty:          ConfigString(config.GoAutoOffsetsCron),
+		consts.ClickhouseJsonTypeEnabledProperty:  ConfigBoolPointer{Value: config.ClickhouseJsonTypeEnabledProperty},
```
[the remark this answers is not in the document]
Jordan-G-CS: nope, that's the actual name

common/odigos_config.go
Jordan-G-CS: fixed

common/odigos_config.go
damemi: >  it i make 3 ToString() functions all with the same input, i will get an error

what error?

Jordan-G-CS: compiler error, because that one struct cannot have 3 ToString() functions
method OidcConfiguration.ToString already declared at /Users/jordangasaatura/Desktop/odigos-testing/common/odigos_config.go:139:32compilerDuplicateMethod

common/odigos_config.go
```diff
@@ -1,5 +1,242 @@
...
+// no need for a new type, since most of the ones comming up are already types that are structs
+// just give them this ToString() function to make it part of the interface
+func (data *CollectorNodeConfiguration) ToString() {
+	if data == nil {
+		fmt.Println("not set")
+		return
+	} else {
+		var placeholder ConfigString = ConfigString(data.K8sNodeLogsDirectory)
+		placeholder.ToString()
+	}
```
chen-keinan, line 68: `else`  block can be removed as `if`  block `return` and its the last action on method
replace `else` block with this:
```
                var placeholder ConfigString = ConfigString(data.K8sNodeLogsDirectory)
		placeholder.ToString()
```

common/odigos_config.go
```diff
@@ -1,5 +1,242 @@
...
+}
+
```
chen-keinan, line 70: same for all below `else` block is not needed as it the last action in method

common/odigos_config.go
damemi: you could have one ToString() function that prints all of the values in OidcConfiguration

common/odigos_config.go
Jordan-G-CS: fixed

common/odigos_config.go
Jordan-G-CS: fixed

PR #3219, opened 2025-07-26
ci: add git config for actor to fix pr creation

blumamir:
Trying the new workflow for release candidate PRs. i run into this:
https://github.com/odigos-io/odigos/actions/runs/16535004040/job/46767756028

image

This PR fixes the issue


blumamir:
merging to test if it works. does not impact any existing workflows, just this new one which is under test

PR #3241, opened 2025-07-29
feat: adding forked process application to python workload lifecycle …

tamirdavid1:
## Description

Adding gunicorn application to python workload lifecycle tests, this is to test also forked processes scenario.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


tests/e2e/workload-lifecycle/chainsaw-test.yaml
```diff
@@ -61,6 +61,19 @@ spec:
...
+              EXPECTED_COUNT=6
+              ACTUAL_COUNT=$(kubectl get instrumentationinstance | grep gunicorn | wc -l)
+
+              if [ "$ACTUAL_COUNT" -eq "$EXPECTED_COUNT" ]; then
+                echo "✅ Found $EXPECTED_COUNT gunicorn instrumentation instances."
+                exit 0
+              else
+                echo "❌ Expected $EXPECTED_COUNT gunicorn instances, but found $ACTUAL_COUNT."
+                exit 1
+              fi
```
RonFed, line 75: why not to do this in a yaml assert for the instances?

tamirdavid1: Because the the names include the pid which is changed between executions

PR #3245, opened 2025-07-29
CORE-106: Add comment to helm chart

damemi:
Make it clearer that `collectorNode` sets the `data-collection` pod (make it easier to search)

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


blumamir (approved):

helm/odigos/values.yaml
```diff
@@ -105,7 +105,7 @@ collectorGateway:
   # It is disabled by default and can be enabled by setting the enabled flag to true.
   clusterMetricsEnabled: false
 
-
+# Settings for Odigos data-collection daemonset pods
```
blumamir: LGTM. now it's part of the odiglet pod so the comment should be updated accordingly

PR #3262, opened 2025-07-31
CORE-111: add-allow-concurrent-agents-to-helm

AvihuHenya:
## Description
add concurrent agent configuration in `.values` file.

and small nit - typo fix for cli `userInstrumentationEnvs` configuration.  

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


helm/odigos/values.yaml
```diff
@@ -380,6 +380,9 @@ rollout:
 pdb:
   enabled: false
 
+# Setting for allowing Odigos to run concurrently with other agents
+allowConcurrentAgents:
+  enabled: true
```
RonFed: Do we want it to be enabled by default?

dan505512: I don't think we do

AvihuHenya: updated to false

PR #3263, opened 2025-07-31
chore: push tag if PR merged to releases/* branch

tamirdavid1:
## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


.github/workflows/push-realease-candidate-tag.yml
```diff
@@ -0,0 +1,51 @@
...
+
+jobs:
+  tag-rc-version:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0  # Full history required to create tags
```
blumamir, line 19: I think we also need the token set to 
```suggestion
          fetch-depth: 0  # Full history required to create tags
          token: ${{ secrets.RELEASE_BOT_TOKEN }}
```

otherwise the workflow will not run since it's a github bot commit

.github/workflows/push-realease-candidate-tag.yml
```diff
@@ -0,0 +1,51 @@
...
+          RC_TAG=$(echo "$COMMIT_MSG" | grep -oE 'v[0-9]+\.[0-9]+\.[0-9]+-rc[0-9]+')
+
+          if [[ -z "$RC_TAG" ]]; then
+            echo "No RC tag found in commit message. Skipping."
+            exit 0
+          fi
+
+          echo "Found RC tag: $RC_TAG"
+          echo "rc_tag=$RC_TAG" >> $GITHUB_OUTPUT
+
```
blumamir, line 38:
```suggestion
    - name: Configure Git
      run: |
        git config --global user.email "bot@keyval.dev"
        git config --global user.name "Odigos Release Bot"

```

.github/workflows/push-realease-candidate-tag.yml
```diff
@@ -0,0 +1,51 @@
...
+          fetch-depth: 0  # Full history required to create tags
+
+      - name: Extract RC tag from commit message
+        id: extract_tag
+        run: |
+          COMMIT_MSG="${{ github.event.head_commit.message }}"
+          echo "Commit message: $COMMIT_MSG"
+
+          # Extract the tag version from the message (e.g., v1.0.214-rc3)
+          RC_TAG=$(echo "$COMMIT_MSG" | grep -oE 'v[0-9]+\.[0-9]+\.[0-9]+-rc[0-9]+')
```
blumamir: This will process any commit to the release branch.
If we make an unrelated commit that is not release candidate, and happen to write a version in the PR description, it will also fire a release here, right?
Should we check for a specific text like: "This is an automated PR for the release candidate v1.0.214-rc3"?

.github/workflows/push-realease-candidate-tag.yml
```diff
@@ -0,0 +1,51 @@
...
+
+          # Check if tag already exists
+          if git rev-parse "$RC_TAG" >/dev/null 2>&1; then
+            echo "Tag $RC_TAG already exists. Skipping push."
+            exit 0
+          fi
+
+          # Create and push the tag
+          git tag "$RC_TAG"
+          git push origin "$RC_TAG"
```
blumamir, line 58: consider notify slack about it, also for failures

.github/workflows/push-realease-candidate-tag.yml
blumamir: or even better, use the actor like here: https://github.com/odigos-io/odigos/pull/3219

.github/workflows/push-realease-candidate-tag.yml
```diff
@@ -0,0 +1,75 @@
...
+          RC_TAG=$(echo "$COMMIT_MSG" | grep -oE '^This is an automated PR for the release candidate v[0-9]+\.[0-9]+\.[0-9]+-rc[0-9]+' | grep -oE 'v[0-9]+\.[0-9]+\.[0-9]+-rc[0-9]+')
+
+          if [[ -z "$RC_TAG" ]]; then
+            echo "No RC tag found in commit message. Skipping."
+            exit 0
+          fi
+
+          echo "Found RC tag: $RC_TAG"
+          echo "rc_tag=$RC_TAG" >> $GITHUB_OUTPUT
+          echo "TAG=$RC_TAG" >> $GITHUB_ENV
```
blumamir: nit: do we need to set the tag as both output and env?

.github/workflows/push-realease-candidate-tag.yml
```diff
@@ -0,0 +1,75 @@
...
+            echo "Tag $RC_TAG already exists. Skipping push."
+            exit 0
+          fi
+
+          # Create and push the tag
+          git tag "$RC_TAG"
+          git push origin "$RC_TAG"
+
+      - name: Notify Slack on Success
+        if: ${{ success() }}
```
blumamir: I think this will always send a slack message, even if the PR is not a rc, or if the tag already exists.

blumamir (approved):

.github/workflows/push-realease-candidate-tag.yml
```diff
@@ -0,0 +1,72 @@
...
+          COMMIT_MSG="${{ github.event.head_commit.message }}"
+          echo "Commit message: $COMMIT_MSG"
+
+          # Extract the tag version from the message (e.g., v1.0.214-rc3)
+          RC_TAG=$(echo "$COMMIT_MSG" | grep -oE '^This is an automated PR for the release candidate v[0-9]+\.[0-9]+\.[0-9]+-rc[0-9]+' | grep -oE 'v[0-9]+\.[0-9]+\.[0-9]+-rc[0-9]+')
+
+          if [[ -z "$RC_TAG" ]]; then
+            echo "No RC tag found in commit message. Skipping."
+            exit 0
+          fi
```
damemi, line 34: so the way this works is that the commit message has to have an `rc` version in it for the new tag to be added?

Looking forward, we may want to drop multiple commits into a release branch before releasing, for example if we start doing patch version releases and backports for bugfixes to older versions. We'll probably want to think about this pattern and how to prevent someone from accidentally publishing a new rc release based on their commit message

PR #3271, opened 2025-08-04
chore: add label-attribute profile and attach to greatwall

AvihuHenya:
## Description
Add all relevant files for generating the `label-attributes` processor, and append the corresponding manifest to a new `label-attributes` profile, which now aggregated into the `greatwall` profile.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/profiles/manifests/label-attributes.yaml
+++ b/profiles/manifests/label-attributes.yaml
@@ -1,0 +1,33 @@
+apiVersion: odigos.io/v1alpha1
+kind: Processor
+metadata:
+  name: label-attributes
+  namespace: odigos-system
+spec:
+  type: k8sattributes
+  processorName: Unified Kubernetes Attributes
+  notes: "Collect Kubernetes attributes for observability"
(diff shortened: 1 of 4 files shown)


profiles/manifests/label-attributes.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: odigos.io/v1alpha1
+kind: Action
+metadata:
+  name: label-attributes
+  namespace: odigos-system
```
blumamir, line 5: remove namespace

damemi (approved):
RBAC changes are already covered by operator cluster role 👍 https://github.com/odigos-io/odigos/blob/1b590a2f825bd7e2ecd7792c14499376d21c14ac/operator/config/rbac/role.yaml#L211-L216

PR #3273, opened 2025-08-04
feat: add security headers to HTTP server for enhanced protection

BenElferink:
This commit introduces security headers including Content-Security-Policy, Strict-Transport-Security, and X-Frame-Options to the HTTP server, improving the overall security posture of the application.

emergency-ticket id: EMR-7


frontend/main.go
```diff
@@ -145,6 +145,14 @@ func startHTTPServer(ctx context.Context, flags *Flags, logger logr.Logger, odig
 	// Enable CORS
 	r.Use(cors.Default())
 
+	// Enable security headers
+	r.Use(func(c *gin.Context) {
```
alonkeyval: nit: if it works correctly and does not break the ui as the cursor bot pointed out, I would move the headers to a func and use it:


```suggestion
// Add security headers middleware
r.Use(securityHeadersMiddleware())
```

alonkeyval (approved):

PR #3282, opened 2025-08-06
CORE-122 removed redundant memory and cpu limits from helm

dan505512:
## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

- [x] Automatic migration will modify existing objects (backward compatible)


dan505512:
https://linear.app/odigos/issue/CORE-122/memory-limits-should-be-defined-by-one-variable

PR #3296, opened 2025-08-10
feat: add option to configure HTTPS_PROXY for gateway exporter

blumamir:
Add the option to set `HTTPS_PROXY` for gateway collector to be used by exporters which uses http.

See more [https://opentelemetry.io/docs/collector/configuration/#proxy-support].

The value can be set with helm `--set collectorGateway.httpsProxyAddress=https://example.com` and not with cli at the moment.

The setting is global to the gateway collector and will be automatically used for any http destination configured in the cluster.


RonFed:
Very similar to https://github.com/odigos-io/odigos/pull/2904
cc @damemi

PR #3309, opened 2025-08-13
Creating ebpf reciever - CORE-137

tamirdavid1:
## Description

Add new `odigosebpfreceiver`

This change introduces a new `odigosebpfreceiver`.
Instead of odiglet writing to ring buffers and forwarding data to the data collection service, the data collection service now reads directly from a BPF map created by odiglet.

The goal is to reduce odiglet’s memory and CPU usage by shifting this work to the data collection process.

Initially, only Python applications will forward data using this new method. Support will be extended to all languages and signal types in future iterations.

In lab tests and Python test suites, we observed approximately:

90% reduction in odiglet CPU usage

30% reduction in odiglet memory usage
## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/frontend/gqlgen.yml
+++ b/frontend/gqlgen.yml
@@ -22,12 +22,12 @@ resolver:
   layout: follow-schema
   dir: graph
   package: graph
-  filename_template: '{name}.resolvers.go'
+  filename_template: "{name}.resolvers.go"
 
 # gqlgen will search for any type names in the schema in these go packages
 # if they match it will use them, otherwise it will generate them.
 autobind:
-  - 'github.com/odigos-io/odigos/frontend/graph/model'
+  - "github.com/odigos-io/odigos/frontend/graph/model"
 
 models:
   ID:
@@ -106,4 +106,9 @@ models:
   K8sWorkloadPodContainer:
     fields:
       processes:
+        resolver: true
+
+  K8sActualSource:
+    fields:
+      workloadOdigosHealthStatus:
         resolver: true
--- a/frontend/graph/generated.go
+++ b/frontend/graph/generated.go
@@ -41,6 +41,7 @@ type Config struct {
 type ResolverRoot interface {
 	ComputePlatform() ComputePlatformResolver
 	K8sActualNamespace() K8sActualNamespaceResolver
+	K8sActualSource() K8sActualSourceResolver
 	K8sWorkload() K8sWorkloadResolver
 	K8sWorkloadPodContainer() K8sWorkloadPodContainerResolver
 	K8sWorkloadTelemetryMetrics() K8sWorkloadTelemetryMetricsResolver
@@ -371,15 +372,16 @@ type ComplexityRoot struct {
 	}
 
 	K8sActualSource struct {
-		Conditions        func(childComplexity int) int
-		Containers        func(childComplexity int) int
-		DataStreamNames   func(childComplexity int) int
-		Kind              func(childComplexity int) int
-		Name              func(childComplexity int) int
-		Namespace         func(childComplexity int) int
-		NumberOfInstances func(childComplexity int) int
-		OtelServiceName   func(childComplexity int) int
-		Selected          func(childComplexity int) int
+		Conditions                 func(childComplexity int) int
+		Containers                 func(childComplexity int) int
+		DataStreamNames            func(childComplexity int) int
+		Kind                       func(childComplexity int) int
+		Name                       func(childComplexity int) int
+		Namespace                  func(childComplexity int) int
+		NumberOfInstances          func(childComplexity int) int
+		OtelServiceName            func(childComplexity int) int
+		Selected                   func(childComplexity int) int
+		WorkloadOdigosHealthStatus func(childComplexity int) int
 	}
 
 	K8sAnnotationAttribute struct {
@@ -816,6 +818,9 @@ type ComputePlatformResolver interface {
 }
 type K8sActualNamespaceResolver interface {
 	Sources(ctx context.Context, obj *model.K8sActualNamespace) ([]*model.K8sActualSource, error)
+}
+type K8sActualSourceResolver interface {
+	WorkloadOdigosHealthStatus(ctx context.Context, obj *model.K8sActualSource) (*model.DesiredConditionStatus, error)
 }
 type K8sWorkloadResolver interface {
 	ServiceName(ctx context.Context, obj *model.K8sWorkload) (*string, error)
@@ -2331,6 +2336,13 @@ func (e *executableSchema) Complexity(typeName, field string, childComplexity in
 		}
 
 		return e.complexity.K8sActualSource.Selected(childComplexity), true
+
+	case "K8sActualSource.workloadOdigosHealthStatus":
+		if e.complexity.K8sActualSource.WorkloadOdigosHealthStatus == nil {
+			break
+		}
+
+		return e.complexity.K8sActualSource.WorkloadOdigosHealthStatus(childComplexity), true
 
 	case "K8sAnnotationAttribute.annotationKey":
 		if e.complexity.K8sAnnotationAttribute.AnnotationKey == nil {
@@ -9050,6 +9062,8 @@ func (ec *executionContext) fieldContext_ComputePlatform_sources(_ context.Conte
 				return ec.fieldContext_K8sActualSource_containers(ctx, field)
 			case "conditions":
 				return ec.fieldContext_K8sActualSource_conditions(ctx, field)
+			case "workloadOdigosHealthStatus":
+				return ec.fieldContext_K8sActualSource_workloadOdigosHealthStatus(ctx, field)
 			}
 			return nil, fmt.Errorf("no field named %q was found under type K8sActualSource", field.Name)
 		},
@@ -9114,6 +9128,8 @@ func (ec *executionContext) fieldContext_ComputePlatform_source(ctx context.Cont
 				return ec.fieldContext_K8sActualSource_containers(ctx, field)
 			case "conditions":
 				return ec.fieldContext_K8sActualSource_conditions(ctx, field)
+			case "workloadOdigosHealthStatus":
+				return ec.fieldContext_K8sActualSource_workloadOdigosHealthStatus(ctx, field)
 			}
 			return nil, fmt.Errorf("no field named %q was found under type K8sActualSource", field.Name)
 		},
@@ -14485,6 +14501,8 @@ func (ec *executionContext) fieldContext_K8sActualNamespace_sources(_ context.Co
 				return ec.fieldContext_K8sActualSource_containers(ctx, field)
 			case "conditions":
 				return ec.fieldContext_K8sActualSource_conditions(ctx, field)
+			case "workloadOdigosHealthStatus":
+				return ec.fieldContext_K8sActualSource_workloadOdigosHealthStatus(ctx, field)
 			}
 			return nil, fmt.Errorf("no field named %q was found under type K8sActualSource", field.Name)
 		},
@@ -14896,6 +14914,57 @@ func (ec *executionContext) fieldContext_K8sActualSource_conditions(_ context.Co
 				return ec.fieldContext_Condition_lastTransitionTime(ctx, field)
 			}
 			return nil, fmt.Errorf("no field named %q was found under type Condition", field.Name)
+		},
+	}
+	return fc, nil
+}
+
+func (ec *executionContext) _K8sActualSource_workloadOdigosHealthStatus(ctx context.Context, field graphql.CollectedField, obj *model.K8sActualSource) (ret graphql.Marshaler) {
+	fc, err := ec.fieldContext_K8sActualSource_workloadOdigosHealthStatus(ctx, field)
+	if err != nil {
+		return graphql.Null
+	}
+	ctx = graphql.WithFieldContext(ctx, fc)
+	defer func() {
+		if r := recover(); r != nil {
+			ec.Error(ctx, ec.Recover(ctx, r))
+			ret = graphql.Null
+		}
+	}()
+	resTmp, err := ec.ResolverMiddleware(ctx, func(rctx context.Context) (any, error) {
+		ctx = rctx // use context from middleware stack in children
+		return ec.resolvers.K8sActualSource().WorkloadOdigosHealthStatus(rctx, obj)
+	})
+	if err != nil {
+		ec.Error(ctx, err)
+		return graphql.Null
+	}
+	if resTmp == nil {
+		return graphql.Null
+	}
+	res := resTmp.(*model.DesiredConditionStatus)
+	fc.Result = res
+	return ec.marshalODesiredConditionStatus2ᚖgithubᚗcomᚋodigosᚑioᚋodigosᚋfrontendᚋgraphᚋmodelᚐDesiredConditionStatus(ctx, field.Selections, res)
+}
+
+func (ec *executionContext) fieldContext_K8sActualSource_workloadOdigosHealthStatus(_ context.Context, field graphql.CollectedField) (fc *graphql.FieldContext, err error) {
+	fc = &graphql.FieldContext{
+		Object:     "K8sActualSource",
+		Field:      field,
+		IsMethod:   true,
+		IsResolver: true,
+		Child: func(ctx context.Context, field graphql.CollectedField) (*graphql.FieldContext, error) {
+			switch field.Name {
+			case "name":
+				return ec.fieldContext_DesiredConditionStatus_name(ctx, field)
+			case "status":
+				return ec.fieldContext_DesiredConditionStatus_status(ctx, field)
+			case "reasonEnum":
+				return ec.fieldContext_DesiredConditionStatus_reasonEnum(ctx, field)
+			case "message":
+				return ec.fieldContext_DesiredConditionStatus_message(ctx, field)
+			}
+			return nil, fmt.Errorf("no field named %q was found under type DesiredConditionStatus", field.Name)
 		},
 	}
 	return fc, nil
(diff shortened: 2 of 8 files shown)


collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
```diff
@@ -0,0 +1,260 @@
...
+	// We watch the root bpffs mount (/sys/fs/bpf) instead of the exact pinned map path
+	// because fsnotify cannot watch a file that does not yet exist. The pinned map file
+	// (/sys/fs/bpf/odiglet/traces) will be created dynamically by the odiglet.
+	// By watching the root directory, we can receive a fsnotify.Create event when the
+	// target file is eventually pinned anywhere under it, and then check if the path
+	// matches our target map before attempting to load it.
+	dir := "/sys/fs/bpf"
+
+	// Try to set up fsnotify.
+	watcher, wErr := fsnotify.NewWatcher()
```
edeNFed: I remember we had issues with fsnotify in the past, I don't remember what was the fix but I would just validate that the issue is no longer relevant

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
```diff
@@ -0,0 +1,260 @@
...
+				var span tracepb.ResourceSpans
+				err = proto.Unmarshal(record.RawSample[8:8+acceptedLength], &span)
+				if err != nil {
+					r.logger.Error("error unmarshalling span", zap.Error(err))
+					continue
+				}
+				td = convertResourceSpansToPdata(&span)
+			}
+
+			err = r.nextTraces.ConsumeTraces(ctx, td)
```
edeNFed: Do we need to put the resource attributes used to be put by Odiglet?

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
tamirdavid1: I also remember it, this is the reason I fallback to pooling if fail.
The problems was related to lack of `fs.inotify` objets on a machine

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
tamirdavid1: Nope, in python we're setting them via env var
e.g:
```
     - name: OTEL_RESOURCE_ATTRIBUTES                                                                                                                                 
       value: k8s.pod.name=$(ODIGOS_POD_NAME),k8s.container.name=simple-app,k8s.namespace.name=default,k8s.deployment.name=simple-app
```
But just in case i will compare them and share the results on this thread

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
```diff
@@ -0,0 +1,260 @@
...
+
+	go func() {
+		if err := r.readLoop(ctx); err != nil {
+			r.logger.Error("read loop failed", zap.String("component", TypeStr), zap.Error(err))
+		}
+	}()
+	return nil
+}
+
+func (r *ebpfReceiver) Shutdown(ctx context.Context) error {
```
RonFed, line 176: I think we need to add some waiting mechanism here, to make sure the goroutines exit gracefully

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
```diff
@@ -0,0 +1,260 @@
...
+	m, err := ebpf.LoadPinnedMap(r.mapPath, nil)
+	if err != nil {
+		return nil, err
+	}
+	return m, nil
+}
+
+// isNoSuchFileErr detects "no such file" errors even if wrapped.
+func isNoSuchFileErr(err error) bool {
+	return errors.Is(err, fs.ErrNotExist) ||
```
RonFed: Consider using
```Go
if _, err := os.Stat("/path/to/whatever"); os.IsNotExist(err) 
```

I also think this should not be its own function

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
```diff
@@ -0,0 +1,260 @@
...
+
+	r.logger.Info("pinned map not found; waiting for creation", zap.String("path", r.mapPath))
+
+	// We watch the root bpffs mount (/sys/fs/bpf) instead of the exact pinned map path
+	// because fsnotify cannot watch a file that does not yet exist. The pinned map file
+	// (/sys/fs/bpf/odiglet/traces) will be created dynamically by the odiglet.
+	// By watching the root directory, we can receive a fsnotify.Create event when the
+	// target file is eventually pinned anywhere under it, and then check if the path
+	// matches our target map before attempting to load it.
+	dir := "/sys/fs/bpf"
```
RonFed: nit: use the const

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
```diff
@@ -0,0 +1,260 @@
...
+				r.logger.Info("pinned map found and loaded", zap.String("path", r.mapPath))
+				return m, nil
+			}
+		}
+	}
+}
+
+func (r *ebpfReceiver) readLoop(ctx context.Context) error {
+	// Wait (via fsnotify or polling) until the map is available.
+	m, err := r.waitForPinnedMap(ctx)
```
RonFed: I think the code structure will not allow to scale this to multiple readers.
Now we have:
Start() calles readLoop in a goroutine.
But readLoop does also the map loading and waiting for it.

So when we'll add support for multiple concurrect readers this won't work since we'll need to load and wait for the map only once and not for all the readers.

I suggest split the readLoop function accordingly.

collector/receivers/odigosebpfreceiver/config.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+type Config struct {
+	receiver.Settings `mapstructure:",squash"`
+
+	// Maximum number of records to read per syscall [e.g., 1024 records per syscall].
+	MaxReadBatchSize int `mapstructure:"max_read_batch_size"`
+
+	// Polling interval in milliseconds [e.g., 5-10ms for latency/CPU balance]
+	PollInterval time.Duration `mapstructure:"poll_interval"`
```
RonFed: PollInterval for what?
Can you please rename and or update the comment?

collector/receivers/odigosebpfreceiver/config.go
```diff
@@ -0,0 +1,20 @@
...
+
+	// Max number of goroutines per ring buffer [e.g., 1-2 goroutines per buffer]
+	MaxGoroutinesPerBuffer int `mapstructure:"max_goroutines_per_buffer"`
```
RonFed: If we decide not to support this at first, I think we shouldn't have it in the config for now

collector/receivers/odigosebpfreceiver/factory.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+}
+
+func createTracesReceiver(
+	_ context.Context,
+	set receiver.Settings,
+	cfg component.Config,
+	next consumer.Traces,
+) (receiver.Traces, error) {
+	tracesEbpfMapPath := "/sys/fs/bpf/odiglet/traces"
```
RonFed: nit: use consts and `filepath.Join`

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
tamirdavid1: please look at the context it in use
```
if m, err := r.tryToLoadPinnedMap(); err == nil {
```
can return multiple error types, I dont want to return err for ErrNotExists

RonFed: I understand. What I mean is:
1. This one line check should not be in a function
2. Instead of using `strings.Contains` the way I suggested it the more portable and recommended way to check this.

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
```diff
@@ -0,0 +1,300 @@
...
+
+	// WaitGroup wait in a goroutine to support context cancellation
+	go func() {
+		r.wg.Wait()
+		close(done)
+	}()
+
+	select {
+	case <-ctx.Done():
+		r.logger.Warn("odigos-ebpf: receiver shutdown timed out", zap.String("mapPath", r.mapPath))
```
RonFed: what time out? where is the time out defined?

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
```diff
@@ -0,0 +1,300 @@
...
+					r.logger.Warn("fsnotify error; switching to 10s polling", zap.Error(werr))
+					// break to polling loop below
+					watcher.Close()
+					watcher = nil
+					break
+				}
+			}
+			if watcher == nil {
+				break
+			}
```
RonFed: This is inside 
```
if watcher != nil {
```

Can we remove this?

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
```diff
@@ -0,0 +1,300 @@
...
+		if err != nil {
+			r.logger.Warn("odigos-ebpf: failed to wait for pinned map", zap.Error(err))
+			return
+		}
+
+		// Step 2: start N concurrent readers using this map
+		const numReaders = 1 // 🔁 can bump this up later
+
+		var readersWg sync.WaitGroup
+		readersWg.Add(numReaders)
```
RonFed: Need to add each goroutine individually in the loop to avoid races

```Go
		for i := 0; i < numReaders; i++ {
			go func(id int) {
                 readersWg.Add(1)
				defer readersWg.Done()
```

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
```diff
@@ -0,0 +1,300 @@
...
+			if errors.Is(err, perf.ErrClosed) {
+				// Reader closed due to shutdown signal — exit gracefully
+				return nil
+			}
+			// Log unexpected read errors and continue
+			r.logger.Error("error reading from perf reader", zap.Error(err))
+			continue
+		}
+
+		// If kernel reports lost samples, log it
```
RonFed: All these comments are too much AI in my opinion - I would remove most of them

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
```diff
@@ -0,0 +1,300 @@
...
+		r.logger.Error("failed to open perf reader", zap.Error(err))
+		return err
+	}
+	defer reader.Close() // ensure proper cleanup on return
+
+	// record buffer to be reused
+	var record perf.Record
+
+	// 🆕 Goroutine: unblocks reader.ReadInto when context is cancelled
+	go func() {
```
RonFed, line 210: should we add waiting for this waitgroup as well?

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
tamirdavid1: `WaitGroup.Add()` should be called before starting the go routine, if not there’s a race where Wait() may return before the counter is added.
move it the to loop but outside the go.
also changed to range loop

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
tamirdavid1: Removed some of them, keep some that wasn't clear on my opinion

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
RonFed: sorry meant
```Go
		for i := 0; i < numReaders; i++ {
                         readersWg.Add(1)
			go func(id int) {
				defer readersWg.Done()
```

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
tamirdavid1: got you

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
tamirdavid1: No, we're using in case:
```
case werr := <-watcher.Errors
```
to break the watch loop and fallback to pooling

RonFed: How watcher will be nil here?

tamirdavid1:
```
			case werr := <-watcher.Errors:
				if werr != nil {
					r.logger.Warn("fsnotify error; switching to 10s polling", zap.Error(werr))
					// break to polling loop below
					watcher.Close()
					watcher = nil
					break
				}
			}
```

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
tamirdavid1: As far as i understood it, `waitGroup` is not needed inside readLoop because r.wg already tracks the lifecycle of the go routine, and the single shutdown helper goroutine always ends safely with the context.

api/k8sconsts/init-container.go
```diff
@@ -5,4 +5,7 @@ const (
 	OdigosInitContainerImageName  = "odigos-agents"
 	OdigosInitContainerName       = "odigos-agents"
 	OdigosInitContainerEnvVarName = "ODIGOS_INIT_CONTAINER_IMAGE"
+
+	ExchangeDir            = "/var/exchange"
+	ExchangeUnixSocketPath = "/var/exchange/exchange.sock"
```
RonFed: I don't think these need to be under init-container.go?
Maybe under some data collection file?

collector/receivers/odigosebpfreceiver/go.mod
```diff
@@ -0,0 +1,66 @@
+module github.com/odigos-io/odigos/collector/receivers/odigosebpfreceiver
+
+go 1.24.0
+
+require (
+	github.com/cilium/ebpf v0.19.0
+	github.com/odigos-io/odigos/common v0.0.0-00010101000000-000000000000
```
RonFed: the version looks weird?

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
```diff
@@ -0,0 +1,268 @@
...
+			if currentMap != nil {
+				currentMap.Close()
+			}
+		}()
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case newMap := <-updates:
```
RonFed: when the channel will be closed this will return a nil map
what do you think about doing
for newMap := range(updates) {...}

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
```diff
@@ -0,0 +1,268 @@
...
+				// Switch to the new map
+				currentMap = newMap
+				readerCtx, cancel := context.WithCancel(ctx)
+				readerCancel = cancel
+
+				r.logger.Info("switched to new eBPF map", zap.Int("fd", newMap.FD()))
+
+				// Start reading from the new map
+				go func() {
+					defer r.logger.Info("reader stopped")
```
RonFed: Do we also need to add waiting for this goroutine as well? do we want to make sure the previous reader loop is closed before we launch the new one?

common/unixfd/client.go
```diff
@@ -0,0 +1,141 @@
...
+			return ctx.Err()
+		default:
+		}
+
+		// Try to connect and get FD
+		fd, err := connectAndGetFD(socketPath)
+		if err != nil {
+			// Connection attempt failed—odiglet may be down or in the process of restarting.
+			// Retry the connection after a short delay.
+			time.Sleep(2 * time.Second)
```
RonFed: Maybe consider a log here? so we know if we are stuck in a loop of getting the fd?

common/unixfd/client.go
```diff
@@ -0,0 +1,141 @@
...
+// recvFD reads a file descriptor from the connection
+func recvFD(c *net.UnixConn) (int, error) {
+	buf := make([]byte, 16)
+	oob := make([]byte, unix.CmsgSpace(4))
+
+	n, oobn, flags, addr, err := c.ReadMsgUnix(buf, oob)
+	// We only need oobn from this call, ignore for linter
+	_ = n
+	_ = flags
+	_ = addr
```
RonFed:
```suggestion
	_, oobn, _, _, err := c.ReadMsgUnix(buf, oob)
```

common/unixfd/client.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	if len(msgs) != 1 {
+		return -1, fmt.Errorf("expected 1 control message, got %d", len(msgs))
+	}
+
+	fds, err := unix.ParseUnixRights(&msgs[0])
+	if err != nil {
+		return -1, fmt.Errorf("parse rights: %w", err)
+	}
+	if len(fds) == 0 {
```
RonFed:
```suggestion
	if len(fds) != 1 {
```

common/unixfd/protocol.go
```diff
@@ -0,0 +1,11 @@
+package unixfd
+
+// Protocol messages exchanged between odiglet (server) and data-collection (client).
+const (
+	// Client → Server
+	ReqGetFD = "GET_FD"
+)
+
+const (
+	DefaultSocketPath = "/var/exchange/exchange.sock"
```
RonFed, line 13: idealy we should have this defined in one place - so we can remove the one under api?

common/unixfd/server.go
```diff
@@ -0,0 +1,104 @@
...
+		return
+	}
+
+	s.Logger.Info("sent FD to client", "fd", fd)
+}
+
+// sendFD sends a file descriptor over the Unix socket
+func sendFD(conn *net.UnixConn, fd int) error {
+	rights := unix.UnixRights(fd)
+	_, _, err := conn.WriteMsgUnix([]byte("OK"), rights, nil)
```
RonFed: Do we check this OK? should it be part of the protocol?

api/k8sconsts/init-container.go
tamirdavid1: you right, changed.

collector/receivers/odigosebpfreceiver/go.mod
tamirdavid1: Changed to v0.0.0 like all custom collector componenets that uses common package

collector/receivers/odigosebpfreceiver/odigosebpfreceiver.go
tamirdavid1: Changed to
```
case newMap, ok := <-updates:
if !ok {
	return
}
```

RonFed (approved):

PR #3310, opened 2025-08-13
fix: only add hostPath mounts and privelged to the data-collection when required by a metrics or logs destination

RonFed:
In order to minimize permissions for the node collector and minimize the `hostPath` volumes:
* only add the host mounts when required - either by logs or metrics destination. When only sending traces host mounts are not required.
* the node collector is only set to being a privileged container if either logs or metrics destination is present. If either of those signals is required we set the container to privileged - which is the current behavior. When only sending traces the collector is run not privileged.  This can be further refined in the future when we'll add more alternatives for collection host metrics or logs.

Tested with logs and metrics destination, verified the daemonset is reconciled correctly once this destination is added. Once that destination is removed and we are only sending traces - the daemonset is reconciled back to remove the mounts and to un privileged.

### DIFF
--- a/autoscaler/controllers/nodecollector/daemonset.go
+++ b/autoscaler/controllers/nodecollector/daemonset.go
@@ -401,10 +401,6 @@ func getDesiredDaemonSet(datacollection *odigosv1.CollectorsGroup,
 	return desiredDs, nil
 }
 
-func boolPtr(b bool) *bool {
-	return &b
-}
-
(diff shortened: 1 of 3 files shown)


edeNFed (approved):

autoscaler/controllers/nodecollector/daemonset.go
```diff
@@ -345,7 +370,7 @@ func getDesiredDaemonSet(datacollection *odigosv1.CollectorsGroup,
 								},
 							},
 							SecurityContext: &corev1.SecurityContext{
-								Privileged: boolPtr(true),
```
blumamir, line 348: Did a quick check, I think this is the only place using the `boolPtr` function which can probably be cleaned up as well

PR #3331, opened 2025-08-19
[RUN-112] [WIP] Reduce hostPath mounts, permissions and hostPID for odiglet and data collection

RonFed:
WIP


RonFed:
Removing draft to see if CI runs

Makefile
```diff
@@ -482,3 +482,25 @@ build-cli-image:
 	DATE=$(shell date -u +'%Y-%m-%d_%H:%M:%S') \
 	ko build --bare --tags $(TAG) --local .
 
+# install getekeeper to prevent:
```
blumamir:
```suggestion
# install gatekeeper to prevent:
```

helm/odigos/values.yaml
```diff
@@ -328,6 +328,11 @@ odiglet:
   # into the Odiglet to ensure it can access the container runtime unix socket.
   customContainerRuntimeSocketPath: ''
 
+  # run the odiglet container as privileged.
+  privileged: true
+  allowHostPathMounts: true
+  hostPid: true
+
```
blumamir: add docs

helm/odigos/values.yaml
```diff
@@ -328,6 +328,11 @@ odiglet:
   # into the Odiglet to ensure it can access the container runtime unix socket.
   customContainerRuntimeSocketPath: ''
 
+  # run the odiglet container as privileged.
+  privileged: true
+  allowHostPathMounts: true
+  hostPid: true
```
blumamir: consider making these so the default "false".

e.g. when they are missing (default), it's like they are false

helm/odigos/templates/odiglet/daemonset.yaml
```diff
@@ -43,21 +43,22 @@ spec:
...
               readOnly: true
             - name: selinux
               mountPath: /host/etc/selinux
               mountPropagation: Bidirectional
-            {{- end }}
+          {{- end }}
+          {{- end }}
           securityContext:
+          {{- if .Values.odiglet.privileged}}
             privileged: true
```
blumamir, line 62: is this required

helm/odigos/templates/odiglet/daemonset.yaml
```diff
@@ -43,21 +43,22 @@ spec:
           args:
             - init
           volumeMounts:
+          {{- if .Values.odiglet.allowHostPathMounts}}
```
blumamir: check if we need the init container when we run with init container

helm/odigos/templates/odiglet/daemonset.yaml
```diff
@@ -216,33 +243,44 @@ spec:
       {{- if semverCompare "<1.26.0" $version }}
```
blumamir, line 246: the gatekeeper will work with k8s > 1.26

tests/gatekeeper/restrict-host-namespace.yaml
```diff
@@ -0,0 +1,84 @@
+apiVersion: templates.gatekeeper.sh/v1
```
blumamir: consider adding README that explains how to use this setup

blumamir (approved):

helm/odigos/templates/odiglet/daemonset.yaml
RonFed: removed the init container of the odiglet if host paths are not allowed

PR #3339, opened 2025-08-21
Move sizing from profile

tamirdavid1:
## Description

**Motivation**
This PR moves sizing configuration from the profile concept into explicit Helm/CLI configuration. The goal is to simplify installation and configuration of data collection by making sizing user-controlled instead of tied to internal profile logic.

**Changes**
1. Introduced sizingConfig values (size_s, size_m, size_l) with defaults in Helm charts.
2. Added CLI-side support for sizing configuration, with overlay behavior:
    a. Helm installation: sizing is driven by Helm values (with validation and mirroring rules).
    b. CLI installation: defaults are applied, but any user-specified fields directly override the defaults (no mirroring/validation).
3. Defined Go-level SizingConfig mapping (size_s, size_m, size_l) consistent with Helm defaults.
4. Removed dependency on the old profile-based sizing logic.

**Next Steps**
This is a step toward the larger refactor where data collection and Odiglet will run in a single pod as separate containers, enabling volume sharing and removing the need for hostPath mounts.
## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [X] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


k8sutils/pkg/sizing/sizing.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	// if odigos installed using cli
+	//  we want to set the sizing based on the sizing config [default: size_m].
+	if c.SizingConfig == "" {
+		c.SizingConfig = string(SizeMedium)
+	}
+
+	// Start from base sizing config
+	base := configs[Sizing(c.SizingConfig)]
```
blumamir: what if the value of user supplied string `c.SizingConfig` is not valid? the map lookup will return empty `base`.
need to add a check for that

k8sutils/pkg/sizing/sizing.go
```diff
@@ -0,0 +1,201 @@
...
+			GoMemLimitMib:              560, // 80% of MemoryLimiterLimitMiB
+		},
+	},
+}
+
+func ModifySizingConfig(c *common.OdigosConfiguration, helmInstallation bool) {
+	// If helm installation, we don't want to modify the sizing config.
+	if helmInstallation {
+		return
+	}
```
blumamir: what is the motivation for that? shouldn't the computation still result in same output?

k8sutils/pkg/sizing/sizing.go
```diff
@@ -0,0 +1,201 @@
...
+
+	if src.MemoryLimiterLimitMiB != 0 {
+		dst.MemoryLimiterLimitMiB = src.MemoryLimiterLimitMiB
+	}
+	if src.MemoryLimiterSpikeLimitMiB != 0 {
+		dst.MemoryLimiterSpikeLimitMiB = src.MemoryLimiterSpikeLimitMiB
+	}
+	if src.GoMemLimitMib != 0 {
+		dst.GoMemLimitMib = src.GoMemLimitMib
+	}
```
blumamir, line 158: This code can potentially mix user manual values with sizing value.

For example, if the user overrides `MaxReplicas` but `MinReplicas` is still used from size, then we can end up with max < min.

Also for all the other fields here. this file should apply similar logic to the helm checks and alignments.

helm/odigos/values.yaml
```diff
@@ -54,51 +54,73 @@ userInstrumentationEnvs:
       enabled: false
       env: {}
 
+# Sizing Configurations size_s, size_m, size_l are pre-defined configurations designed to simplify pipeline configurations.
+# The default value is 'size_m', which is the medium size configuration.
+# See https://docs.odigos.io/pipeline/configuration#1-using-sizing-configuration for more details.
+sizingConfig: size_m
```
blumamir: I wonder if the "config" in the name is meaningful or confusing.
This name will be hard to change if we ever want to in the future

helm/odigos/values.yaml
```diff
@@ -54,51 +54,73 @@ userInstrumentationEnvs:
...
+
+
 collectorGateway:
+  # HYBRID CONFIGURATION APPROACH:
+  # By default, all values below are automatically set based on the sizingConfig above.
+  # Uncomment and modify any value to override the automatic sizing for that specific parameter.
+  #
+  # IMPORTANT CONSTRAINTS:
+  # 1. minReplicas must be less than maxReplicas
```
blumamir: or equal (also worth making sure in the helm functions that we allow that)

helm/odigos/values.yaml
```diff
@@ -54,51 +54,73 @@ userInstrumentationEnvs:
...
+  # 2. If you set limitMemoryMiB without requestMemoryMiB, request will equal limit
+  # 3. If you set requestMemoryMiB without limitMemoryMiB, limit will equal request
+  # 4. Same logic applies to CPU settings
+  # 5. If you customize any memory/CPU setting, consider setting all related memory/CPU values
```
blumamir: I think we want the user to only set limit/request and avoid other settings (spike, gomemlimit etc) as they should be auto calculated.

We want only advanced users which know what they are doing to touch the "other" related settings preferably with guidance

helm/odigos/templates/_sizing-helpers.tpl
```diff
@@ -0,0 +1,248 @@
...
+  gatewayCPULimit: 300
+  nodeMemoryRequest: 150
+  nodeMemoryLimit: 300
+  nodeCPURequest: 150
+  nodeCPULimit: 300
+  {{- else if eq $s "size_l" }}
+  gatewayMinReplicas: 3
+  gatewayMaxReplicas: 12
+  gatewayMemoryRequest: 750
+  gatewayMemoryLimit: 850
```
blumamir: updated here as well to match request and limit for gateway

cli/cmd/config.go
```diff
@@ -63,7 +64,8 @@ var configCmd = &cobra.Command{
 	- "%s": Cron schedule for automatic Go offsets updates (e.g. "0 0 * * *" for daily at midnight). Set to empty string to disable.
 	- "%s": Enable or disable ClickHouse JSON column support. When enabled, telemetry data is written using a new schema with JSON-typed columns (requires ClickHouse v25.3+). [default: false]
 	- "%s": List of allowed domains for test connection endpoints (e.g., "https://api.honeycomb.io", "https://otel.example.com"). Use "*" to allow all domains. Empty list allows all domains for backward compatibility.
-	- "%s": Enable or disable data compression before sending data to the Gateway collector. [default: false]
+	- "%s": Enable or disable data compression before sending data to the Gateway collector. [default: false],
+	- "%s": Set the sizing configuration for the Odigos components. [default: size_m]
```
blumamir: suggestion: list all options here: 

```suggestion
	- "%s": Set the sizing configuration for the Odigos components (size_s, size_m [default], size_l).
```

helm/odigos/values.yaml
```diff
@@ -54,51 +54,73 @@ userInstrumentationEnvs:
...
+  # If you set only requestMemoryMiB, the limitMemoryMiB will be set to the same value.
   # requestMemoryMiB: 500
 
   # the memory limit for the cluster gateway collector deployment.
   # it will be embedded in the deployment as a resource limit
   # of the form "memory: <value>Mi".
   # default value is 625Mi
-  limitMemoryMiB: 625
+  # If you set only limitMemoryMiB, the requestMemoryMiB will be set to the same value.
+  # limitMemoryMiB: 625
```
blumamir, line 87: also here - need to match request and limit (to reflect what we are doing)

helm/odigos/values.yaml
tamirdavid1: Changed to `ResourceSizePreset`

blumamir (approved):

PR #3359, opened 2025-08-25
fix: check memory limit using runtime gc controller

blumamir:
Fixes: CORE-192

This PR targets the collector crashing due to OOM under high load or spikes.

Previous to this PR, the memory limit was enforced using the OpenTelemetry Collector memory limiter extension to periodically check if the limit is reached and respose by refusing data with RESOURCE_ EXHUSTED in this case.

OpenTelemetry uses the `runtime.ReadMemStats` function which gives consistent memory stats, but it also stops the world and is expensive to call, thus cannot work for high-frequency, per request checks.

It is replaced by similar code that:
- `go:linkname` directly into the gc controller to fetch the relevant values fast. The gc controller uses `atomic` to sync the access to these values.
- hooks into the same computation the go GC is making, to make sure we end up with the same result.

### DIFF
--- a/collector/odigosotelcol/go.mod
+++ b/collector/odigosotelcol/go.mod
@@ -470,7 +470,7 @@ require (
 	github.com/mtibben/percent v0.2.1 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/mwitkow/go-conntrack v0.0.0-20190716064945-2f068394615f // indirect
-	github.com/odigos-io/go-rtml v0.0.0-20250828115257-56abd1103712 // indirect
+	github.com/odigos-io/go-rtml v0.0.1 // indirect
 	github.com/odigos-io/odigos/common v0.0.0 // indirect
 	github.com/oklog/ulid v1.3.1 // indirect
 	github.com/oklog/ulid/v2 v2.1.1 // indirect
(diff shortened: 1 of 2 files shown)


collector/config/configgrpc/realtimememlimit.go
```diff
@@ -0,0 +1,162 @@
...
+var runtimeGCController gcControllerState
+
+//go:linkname runtimeHeapGoal runtime.(*gcControllerState).heapGoal
+func runtimeHeapGoal(*gcControllerState) uint64
+
+// following struct is a mirror of the exact struct used by the go runtime.
+// notice that it must match exactly (field order and types).
+// if go ever changes the internal struct, this need to be updated as well,
+// or we can get invalid values when accessing those fields.
+type gcControllerState struct {
```
RonFed: This increases the chance that something will break.
If some random field is added/removed - we can read wrong values.
How can we defend ourselves from this?
Someone who is not very familiar with this may upgrade the collector in the future (which will in turn upgrade the go version) - and this would break in runtime which makes it much more challenging to catch.

collector/config/configgrpc/go.mod
```diff
@@ -4,6 +4,7 @@ go 1.23.0
 
 require (
 	github.com/mostynb/go-grpc-compression v1.2.3
+	github.com/odigos-io/go-rtml v0.0.0-20250828115257-56abd1103712
```
RonFed: consider tag

PR #3366, opened 2025-08-25
CORE-173: Enable mirrored image automatic Go offsets updates

damemi:
## Description

This enables users to set automatic updates using our published cli-offsets docker image rather than pulling directly from the hosted file in GCS. The user still needs Internet access to pull the image, but that access is now shifted from the cluster to their own registry/image mirror, which may work better for some disconnected environments.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/api/k8sconsts/cli.go
+++ b/api/k8sconsts/cli.go
@@ -8,4 +8,7 @@ const (
 	MetricsDir = "Metrics"
 )
 
-const CliImageName = "odigos-cli"
+const (
+	CliImageName        = "odigos-cli"
+	CLIOffsetsImageName = "odigos-cli-offsets"
+)
--- a/api/k8sconsts/cli.go
+++ b/api/k8sconsts/cli.go
@@ -10,5 +10,5 @@ const (
 
 const (
 	CliImageName        = "odigos-cli"
-	CLIOffsetsImageName = "odigos-cli-offsets"
+	CliOffsetsImageName = "odigos-cli-offsets"
 )
--- a/api/k8sconsts/odiglet.go
+++ b/api/k8sconsts/odiglet.go
@@ -26,11 +26,13 @@ const (
 	OdigletDefaultHealthProbeBindPort = 55683
 
 	// ConfigMap used to store custom/updated Go instrumentation offsets
-	GoOffsetsConfigMap  = "odigos-go-offsets"
-	GoOffsetsFileName   = "go_offset_results.json"
-	GoOffsetsEnvVar     = "OTEL_GO_OFFSETS_FILE"
-	OffsetFileMountPath = "/offsets"
-	OffsetCronJobName   = "odigos-go-offsets-updater"
+	GoOffsetsConfigMap      = "odigos-go-offsets"
+	GoOffsetsFileName       = "go_offset_results.json"
+	GoOffsetsEnvVar         = "OTEL_GO_OFFSETS_FILE"
+	OffsetFileMountPath     = "/offsets"
+	OffsetCronJobName       = "odigos-go-offsets-updater"
+	OffsetCronJobModeDirect = "direct"
+	OffsetCronJobModeImage  = "image"
 
 	OdigletLocalTrafficServiceName = "odiglet-local"
 	OdigletMetricsServerPort       = 8080
--- a/api/k8sconsts/odiglet.go
+++ b/api/k8sconsts/odiglet.go
@@ -26,17 +26,39 @@ const (
 	OdigletDefaultHealthProbeBindPort = 55683
 
 	// ConfigMap used to store custom/updated Go instrumentation offsets
-	GoOffsetsConfigMap      = "odigos-go-offsets"
-	GoOffsetsFileName       = "go_offset_results.json"
-	GoOffsetsEnvVar         = "OTEL_GO_OFFSETS_FILE"
-	OffsetFileMountPath     = "/offsets"
-	OffsetCronJobName       = "odigos-go-offsets-updater"
-	OffsetCronJobModeDirect = "direct"
-	OffsetCronJobModeImage  = "image"
+	GoOffsetsConfigMap  = "odigos-go-offsets"
+	GoOffsetsFileName   = "go_offset_results.json"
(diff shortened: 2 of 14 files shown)


api/k8sconsts/cli.go
```diff
@@ -8,4 +8,7 @@ const (
 	MetricsDir = "Metrics"
 )
 
-const CliImageName = "odigos-cli"
+const (
+	CliImageName        = "odigos-cli"
+	CLIOffsetsImageName = "odigos-cli-offsets"
```
blumamir: nit: letter capitalization for consistency 
```suggestion
	CliOffsetsImageName = "odigos-cli-offsets"
```

api/k8sconsts/odiglet.go
```diff
@@ -26,11 +26,13 @@ const (
...
-	GoOffsetsEnvVar     = "OTEL_GO_OFFSETS_FILE"
-	OffsetFileMountPath = "/offsets"
-	OffsetCronJobName   = "odigos-go-offsets-updater"
+	GoOffsetsConfigMap      = "odigos-go-offsets"
+	GoOffsetsFileName       = "go_offset_results.json"
+	GoOffsetsEnvVar         = "OTEL_GO_OFFSETS_FILE"
+	OffsetFileMountPath     = "/offsets"
+	OffsetCronJobName       = "odigos-go-offsets-updater"
+	OffsetCronJobModeDirect = "direct"
+	OffsetCronJobModeImage  = "image"
```
blumamir: nit: consider creating an enum from those "OffsetCronJobMode" which I think is what we do in many other places.

cli/cmd/config.go
```diff
@@ -450,6 +454,16 @@ func setConfigProperty(ctx context.Context, client *kube.Client, config *common.
...
 		config.GoAutoOffsetsCron = cronValue
 
+	case consts.GoAutoOffsetsModeProperty:
+		if len(value) != 1 {
+			return fmt.Errorf("%s expects exactly one value", property)
+		}
+		modeValue := value[0]
+		if modeValue != "" && modeValue != k8sconsts.OffsetCronJobModeDirect && modeValue != k8sconsts.OffsetCronJobModeImage {
+			return fmt.Errorf("invalid mode: %s. Must be either '%s' or '%s'", modeValue, k8sconsts.OffsetCronJobModeDirect, k8sconsts.OffsetCronJobModeImage)
+		}
```
blumamir: nit: what do you think about adding an `"off"` mode to allow users to make it explicit?

cli/cmd/config.go
```diff
@@ -62,6 +63,7 @@ var configCmd = &cobra.Command{
 	- "%s": Sets the port for the Odiglet health probes (readiness/liveness).
   	- "%s": Enable or disable the service graph feature [default: false].
 	- "%s": Cron schedule for automatic Go offsets updates (e.g. "0 0 * * *" for daily at midnight). Set to empty string to disable.
+	- "%s": Mode for automatic Go offsets updates. Options include direct (default) and image.
```
blumamir, line 66: need to add these options also to helm

scheduler/controllers/odigosconfiguration/odigosconfiguration_controller.go
```diff
@@ -196,6 +196,27 @@ func (r *odigosConfigurationController) handleGoOffsetsCronJob(ctx context.Conte
...
+	var imageName string
+	var command []string
+
+	switch mode {
+	case k8sconsts.OffsetCronJobModeDirect:
+		imageName = k8sconsts.CliImageName
+		command = []string{"pro", "update-offsets"}
+	case k8sconsts.OffsetCronJobModeImage:
+		imageName = k8sconsts.CLIOffsetsImageName
+		command = []string{"pro", "update-offsets", "--from-file", "/odigos/offset_results_min.json"}
```
blumamir: I am probably missing something trivial, but where is the `odigos-cli-offsets` image coming from? I don't see it in the registry or the code for it

damemi: Ah looks like the push job is failing https://github.com/odigos-io/enterprise-go-instrumentation/pull/1697

damemi:
@blumamir thanks, sorry for the delay. Think I got to all you feedback ptal

scheduler/controllers/odigosconfiguration/odigosconfiguration_controller.go
```diff
@@ -214,9 +239,9 @@ func (r *odigosConfigurationController) handleGoOffsetsCronJob(ctx context.Conte
 			ServiceAccountName: k8sconsts.SchedulerServiceAccountName,
 			Containers: []corev1.Container{
 				{
-					Name:  k8sconsts.CliImageName,
-					Image: fmt.Sprintf("%s/%s:%s", config.ImagePrefix, k8sconsts.CliImageName, r.OdigosVersion),
-					Args:  []string{"pro", "update-offsets"},
+					Name:  imageName,
+					Image: fmt.Sprintf("%s/%s:%s", config.ImagePrefix, imageName, r.OdigosVersion),
+					Args:  command,
 				},
```
AvihuHenya, line 245: What do you think about adding an OwnerReference so that the CronJob is deleted along with the scheduler?

damemi:
Updated this after talking with Amir offline:

* Put blockers in cli config command and scheduler to make sure cron job is only created in Pro tier (otherwise it will just create a bunch of failed jobs)
* Moved the cron job setting to its own controller under odigospro. This has the following benefits:
    * Directly handles `odigos pro activate` flow -> upgrading to pro triggers cron job reconcile
    * That is mainly only useful if the odigos configmap already has cron job settings enabled. In this case, they will automatically take effect.
    * Otherwise, if an OSS user somehow manually set odigos configmap settings for cronjobs, that would flood the scheduler logs with a bunch of errors and prevent other configmap settings from being persisted.

Overall moving the Cronjob to its own controller under the odigospro manager is better organized and isolates failures. Thanks @blumamir for brainstorming that out with me

PR #3381, opened 2025-08-27
fix: remove unnecessary BUILD_DIR variable for Odigos Collector in Makefile

BenElferink:


damemi (approved):
I'm surprised this didn't cause any problems, since the dockerfile refers to things like `common` outside of that build-dir https://github.com/odigos-io/odigos/blob/b68f750e7024ab67c1c88fc79d6cdf32c30dd2e6/collector/Dockerfile#L4

PR #3387, opened 2025-08-28
CORE-169: Use NODE_IP env var for kubeletstats receiver

AvihuHenya:
## Description
This PR updates the `kubeletstats` receiver configuration to use the new `NODE_IP` environment variable, in order to support environments where the NODE_NAME value is not resolvable in cluster DNS.

* Added a new NODE_IP environment variable in `data-collection`.
* Updated the auto-scaler nodeCollector controller to create the `kubeletstats` receiver based on the NODE_IP value.

## How Has This Been Tested?
Configured a Prometheus destination in the cluster and verified that kubelet metrics were successfully fetched.


dan505512:
Could this break something that's already using `NODE_NAME`?

blumamir (approved):

PR #3419, opened 2025-09-02
Split workload-lifecycle

TheErikNaz:
Split the workload-lifecycle test suite into multiple test suites
* runtime-detection - Language & version detection
* instrumentation-lifecycle - Device injection & instances
* environment-variables - Env var handling
* trace-collection - End-to-end traces

emergency-ticket id: EMR-201


.github/workflows/e2e.yaml
```diff
@@ -57,33 +55,31 @@ jobs:
...
 
   kubernetes-test:
-    needs:
-      - build-odigos-images
-      - build-cli
-    # workload-lifecycle test scenario requires more resources, so we use a different instance type
-    # all other test scenarios use the default instance type which is free and fast
-    runs-on: warp-ubuntu-latest-x64-8x-spot
+    needs: [ build-odigos-images, build-cli ]
+    runs-on: depot-ubuntu-24.04-8
```
blumamir, line 67: All the other tests are running successfully on the a regular machine.

One of the motivations for breaking it up into smaller suites is that we don't need a big machines for the tests (both in terms of money and so we can run it on our locally dev machines).

I wonder if running on 8 cores is still required in this case

tests/e2e/runtime-detection/README.md
```diff
@@ -0,0 +1,125 @@
+# Runtime Detection Test Suite
+
+## Purpose
+This test suite verifies that Odigos correctly detects programming languages and their versions across different runtime environments and configurations.
```
blumamir: could be nice to describe here what fields of the runtime detection are tested, and which are not.

tests/e2e/runtime-detection/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+## What is Tested
+
+### Languages Covered
+- **Node.js** - JavaScript runtime detection
+- **Java** - JVM-based applications
+- **Python** - Python interpreter detection
+- **.NET** - .NET Core/Framework applications
+- **C++** - Unsupported language detection
```
blumamir: should we also add go here?

tests/e2e/runtime-detection/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+### Test Scenarios
+
+#### Node.js Runtime Detection
+- `nodejs-unsupported-version` - Detects v8.17.0 (below minimum support)
+- `nodejs-very-old-version` - Detects language but no version info
+- `nodejs-minimum-version` - Detects v14.0.0 (minimum supported)
+- `nodejs-latest-version` - Detects latest Node.js version
```
blumamir: list is missing some cases (env from manifest, env from dockerfile)

tests/e2e/runtime-detection/README.md
```diff
@@ -0,0 +1,125 @@
...
+- `python-gunicorn-server` - Detects Python with Gunicorn WSGI server
+
+#### .NET Runtime Detection
+- `dotnet8-musl` - Detects .NET 8 with musl libc (Alpine)
+- `dotnet6-musl` - Detects .NET 6 with musl libc
+- `dotnet8-glibc` - Detects .NET 8 with glibc (standard Linux)
+- `dotnet6-glibc` - Detects .NET 6 with glibc
+
+#### Unsupported Language Detection
+- `cpp-http-server` - Detects C++ as unsupported language
```
blumamir, line 43: It used to be unsupported, but it is detected successfully now.

tests/e2e/runtime-detection/README.md
```diff
@@ -0,0 +1,125 @@
...
+```
+Assert runtime detection → Verify Python instances → Summary report
+```
+
+## Key Validations
+
+### Runtime Detection Assertions
+- **Language identification** - Correct language detected for each workload
+- **Version detection** - Accurate version strings where available
+- **Support classification** - Proper supported/unsupported determination
```
blumamir: what do you mean by support?

tests/e2e/runtime-detection/README.md
```diff
@@ -0,0 +1,125 @@
...
+- **Environment detection** - Container vs manifest environment variables
+
+### Special Validations
+- **Python Instance Count** - Verifies exactly 5 Python instrumentation instances are created (excluding unsupported Python version)
```
blumamir: removed as unrelated to runtime detection

tests/e2e/runtime-detection/README.md
```diff
@@ -0,0 +1,125 @@
...
+└── 01-assert-runtime-detected.yaml    # Runtime detection assertions
+```
+
+## Dependencies
+- Requires Odigos installation with runtime detection capabilities
+- Uses common trace database for destination
+- Relies on public ECR images for test workloads
+
+## Duration
+Approximately 8-12 minutes depending on image pull times and cluster resources.
```
blumamir: Is it still taking 12 minutes??

tests/e2e/runtime-detection/chainsaw-test.yaml
```diff
@@ -0,0 +1,98 @@
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: runtime-detection
+spec:
+  description: |
+    This e2e test verifies runtime detection and version compatibility across different languages.
```
blumamir, line 7:
```suggestion
    This e2e test verifies some fields of runtime detection across different languages
```

tests/e2e/runtime-detection/chainsaw-test.yaml
```diff
@@ -0,0 +1,98 @@
...
+    It tests Node.js, Java, Python, and C++ workloads to ensure proper language and version detection.
```
blumamir, line 8:
```suggestion
    It tests Node.js, Java, Python, and C++ workloads to ensure proper language and version detection, and other properties of runtime detection like environment variables, other agent, etc.
```

tests/e2e/runtime-detection/chainsaw-test.yaml
```diff
@@ -0,0 +1,98 @@
...
+  skipDelete: true
+  steps:
+    - name: Prepare destination
+      try:
+        - apply:
+            file: ../../common/apply/simple-trace-db-deployment.yaml
```
blumamir, line 14: Since we don't execute any quires on this destination, we can consider replacing it with a `nop` one that will make the test faster, more stable, and require less resources

tests/e2e/runtime-detection/chainsaw-test.yaml
```diff
@@ -0,0 +1,98 @@
...
+
+    - name: Instrument Namespaces
+      try:
+        - apply:
+            file: ../../common/apply/instrument-default-ns.yaml
+
+    - name: Add Destination
+      try:
+        - apply:
+            file: ../../common/apply/add-simple-trace-db-destination.yaml
```
blumamir, line 63: nit: you don't need to wait for odigos to be ready to start these processes. might save a bit time :)

tests/e2e/configuration-reload/chainsaw-test.yaml
```diff
@@ -0,0 +1,75 @@
...
+            file: ../../common/apply/add-simple-trace-db-destination.yaml
+
+    - name: Wait for destination to be ready
+      try:
+        - script:
+            timeout: 60s
+            content: |
+              echo "Waiting for destination to be processed..."
+              sleep 10
+              echo "Destination ready"
```
blumamir: what do you mean by ready? what is it's role here?

tests/e2e/configuration-reload/chainsaw-test.yaml
```diff
@@ -0,0 +1,75 @@
...
+                if [ $? -eq 0 ]; then
+                  echo "✅ Collector configuration reloaded successfully"
+                  break
+                else
+                  echo "⏳ Waiting for collector config reload..."
+                  sleep 3
+                fi
+              done
+
+              echo "=== Configuration reload test completed successfully ==="
```
blumamir: The collector would reload itself also when the destination is added, when new sources are added, etc. This check I think is not asserting anything with actual value at the moment

tests/e2e/configuration-reload/chainsaw-test.yaml
```diff
@@ -0,0 +1,75 @@
...
+            timeout: 60s
+            content: |
+              echo "Waiting for destination to be processed..."
+              sleep 10
+              echo "Destination ready"
+
+    - name: Create cluster info action
+      try:
+        - apply:
+            file: 03-create-action.yaml
```
blumamir: the `03` is meant to follow `01`, `02` and help to identify the stage of the test.

tests/e2e/configuration-reload/README.md
```diff
@@ -0,0 +1,172 @@
...
+## Purpose
+This test suite verifies that the Odigos collector correctly reloads its configuration when actions are created or modified, ensuring that observability pipeline changes are applied dynamically without service restarts.
+
+## What is Tested
+
+### Configuration Management
+- **Action creation** - Adding new observability actions to the system
+- **Configuration generation** - Automatic collector configuration updates
+- **Configuration reload** - Hot-reloading collector configuration without restart
+- **Reload verification** - Confirming configuration changes are applied
```
blumamir: I think simply checking the logs does not mean that it's applied. we should check the generated telemetry for that

tests/e2e/environment-variables/chainsaw-test.yaml
```diff
@@ -0,0 +1,96 @@
...
+  steps:
+    - name: Prepare destination
+      try:
+        - apply:
+            file: ../../common/apply/simple-trace-db-deployment.yaml
+
+    - name: Install Test Apps
+      try:
+        - apply:
+            file: install-test-apps.yaml
```
blumamir, line 20: we need just a few apps here, all the "version" ones can add complexity and noise IMO

tests/e2e/environment-variables/chainsaw-test.yaml
```diff
@@ -0,0 +1,96 @@
...
+    - name: Install Odigos
+      try:
+        - script:
+            content: |
+              ../../../cli/odigos install --namespace odigos-test --version e2e-test
+              ../../common/verify_odigos_installation.sh odigos-test
+            timeout: 4m
+        - assert:
+            timeout: 3m
+            file: ../../common/assert/odigos-installed.yaml
```
blumamir, line 31: do we really need this assert if we already have `verify_odigos_installation.sh`?

tests/e2e/environment-variables/assert-env-vars.yaml
```diff
@@ -0,0 +1,105 @@
...
+      name: nodejs-dockerfile-env
+status:
+  runtimeDetailsByContainer:
+    - containerName: nodejs-dockerfile-env
+      envFromContainerRuntime:
+        - name: NODE_OPTIONS
+          value: "--require /app/execute_before.js --max-old-space-size=256"
+      language: javascript
+      runtimeVersion: 20.17.0
+      secureExecutionMode: false
```
blumamir: I think these does not belong here. It's adding noise and unrelated checks for a test that verifies something else

tests/e2e/runtime-detection/README.md
TheErikNaz: The detection is correct and works, the unsupported part is the instrumentation itself

tests/e2e/runtime-detection/chainsaw-test.yaml
TheErikNaz: The destination still needs to be up and valid to prevent random failures because of datacol errors and what not

tests/e2e/runtime-detection/chainsaw-test.yaml
TheErikNaz: They Don't really take time, it is just applying a crd.
But I do want to wait for everything to be stable before doing it

tests/e2e/runtime-detection/README.md
TheErikNaz: When it'll be supported :)

tests/e2e/runtime-detection/README.md
TheErikNaz: Unsupported for instrumented, detection should work

tests/e2e/environment-variables/chainsaw-test.yaml
TheErikNaz: Yeah, I prefer that in a future we'll migrate to only asserting .yamls over writing scripts and scrip blocks, leaving it in for now

tests/e2e/runtime-detection/chainsaw-test.yaml
TheErikNaz: I prefer to leave it in as is, and populate it with tests in the future

tests/e2e/runtime-detection/chainsaw-test.yaml
TheErikNaz: Like in the previous comment, the description is idealized and not fully represents the current state

blumamir (approved):

PR #3423, opened 2025-09-02
Support sampler actions as generic CRDs

damemi:
## Description

This PR supports the sampler actions as the new generic CRD. It includes the following changes:

* API: Create new `SamplersConfig` field in `Actions` CRD and implement `ActionConfig` interface for common samplers + probabilistic sampler
    * Add new `ClusterRoles()` function to `ActionConfig` interface and implement for existing generic actions
* Autoscaler: Add probabilistic+common sampler support
    * Based on new `samplersConfig` function which is now shared by the actions_controller and sampling_controller
* Autoscaler: Update `ActionHandler` interface:
    * Update all handler functions to support legacy action type and new generic action type (for example, `List()` now gets all legacy actions for that handler and merges with generic actions of that handler). Include deduplication code in handler and controller (to prevent duplicate rules and ownerreferences)
    * Add `ConvertLegacyToAction()` function to handler interface for convenience (this is an internal API -- remove this function after migration)
* Unit tests:
    * Enable unit testing for Autoscaler in CI
    * Needed: Fix up existing autoscaler unit tests that were broken and not running (node+cluster collector configmap generation)
    * Add unit test suite for Actions controller and Probabilistic sampler controller (existing sampling controller tests would be duplicate with cases covered by Actions controller)

After this all existing actions will be supported by the generic CRD, meaning we can migrate the UI to use the new object and implement the migration webhook. cc @BenElferink, sorry this took so long

## How Has This Been Tested?

- [x] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

emergency-ticket id: EMR-200


damemi:
Note will need to remove `build-autoscaler` from required CI tests and add `build-and-test-autoscaler`

autoscaler/controllers/actions/action_controller.go
```diff
@@ -200,11 +264,11 @@ func (r *ActionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	if err != nil {
 		statusErr := r.reportReconciledToProcessorFailed(ctx, action, odigosv1.ActionTransformedToProcessorReasonFailedToCreateProcessor, err)
 		if statusErr == nil {
-			return ctrl.Result{}, err // return original error on success
+			return utils.K8SUpdateErrorHandler(err)
 		} else {
 			logger := ctrl.LoggerFrom(ctx)
 			logger.Error(statusErr, "Failed to set status on action")
-			return ctrl.Result{}, err // return original error if the patch fails
+			return utils.K8SUpdateErrorHandler(err)
```
damemi, line 271: the sampling reconciler patches actions in the middle of reconciling to report success or failure (since it aggregates all sampling actions in a single reconcile). So, we could hit a conflict here doing a final update

autoscaler/controllers/actions/actions_webhook.go
```diff
@@ -119,6 +121,10 @@ func (a *ActionsValidator) validateAction(ctx context.Context, action *v1alpha1.
 		path := field.NewPath("spec").Child("k8sAttributes")
 		fields[path] = action.Spec.K8sAttributes
 	}
+	if action.Spec.Samplers != nil {
+		path := field.NewPath("spec").Child("samplers")
+		fields[path] = action.Spec.Samplers
```
RonFed, line 126: Is it valid to have multiple non nil fields inside the Samplers field?

autoscaler/controllers/actions/action_controller.go
```diff
@@ -109,6 +168,11 @@ func convertToDefaultProcessor(action *odigosv1.Action, actionConfig ActionConfi
 		return nil, err
 	}
 
+	collectorRoles := []odigosv1.CollectorsGroupRole{}
+	for _, role := range actionConfig.CollectorRoles() {
+		collectorRoles = append(collectorRoles, odigosv1.CollectorsGroupRole(role))
```
RonFed, line 173: Is it possible to have 2 collector roles in one action? what is the meaning of it?

autoscaler/controllers/actions/action_controller.go
```diff
@@ -92,6 +96,61 @@ func convertActionToProcessor(ctx context.Context, k8sclient client.Client, acti
...
+				return nil, err
+			}
+
+			// Create probabilistic sampler processor using its own config
+			return convertToDefaultProcessor(action, action.Spec.Samplers.ProbabilisticSampler, config)
+		} else {
+			// Handle other samplers using the unified sampling approach
+			// For non-probabilistic samplers, we need to create a composite config
+			// that can be used with the odigossampling processor
+			config, ownerReferences, err := samplersConfig(ctx, k8sclient, action.Namespace)
```
RonFed, line 114: nit: for readability I'd separate the owner reference part from the config part

autoscaler/controllers/actions/sampling/servicenamesampler.go
```diff
@@ -62,13 +120,24 @@ func (h *ServiceNameSamplerHandler) GetRuleConfig(action metav1.Object) []Rule {
 }
 
 func (h *ServiceNameSamplerHandler) GetActionReference(action metav1.Object) metav1.OwnerReference {
```
RonFed, line 122: All the `GetActionReference` seems to be identical to me, maybe we can have once function for this, instead of a per sampler approach.

autoscaler/controllers/actions/actions_webhook.go
damemi: good catch, updated to check for only one of each

fyi as part of the final migration to generic actions, I would like to support setting multiple actions in a single CRD. issue to track this: https://linear.app/odigos/issue/CORE-203/support-defining-multiple-types-of-action-in-one-generic-crd

autoscaler/controllers/actions/action_controller.go
damemi: I'm not sure, it's just right now the ProcessorSpec.CollectorRoles field takes a list, so I just made this helper work with a list. I think we could take a look at that as a different refactor

autoscaler/controllers/actions/action_controller.go
damemi: in the sampler actions and k8sattributes actions, ownerreferences are directly tied to the config, since the config is an aggregate of all the actions that go into it. So imo it makes sense to return "the config" + "objects that went into the config". We'd also have to do basically the same re-listing and processing to separate the logic

autoscaler/controllers/actions/sampling/servicenamesampler.go
damemi: Each type of legacy action needs to be cast to its own kind in order to get the info needed for the owner reference. So a single helper would basically be a type switch or a bunch of if/else

Once we remove the legacy actions and everything is a single Action CRD, yes we can definitely move to a single function. See also that I'd like to unify the `ActionConfig` and `ActionHandler` interface https://linear.app/odigos/issue/CORE-202/unify-actionconfig-and-actionhandler-interfaces

autoscaler/controllers/actions/actions_webhook.go
```diff
@@ -119,6 +121,46 @@ func (a *ActionsValidator) validateAction(ctx context.Context, action *v1alpha1.
...
+		if action.Spec.Samplers.ServiceNameSampler != nil {
+			path := field.NewPath("spec").Child("samplers").Child("serviceNameSampler")
+			samplerFields[path] = action.Spec.Samplers.ServiceNameSampler
+		}
+		if action.Spec.Samplers.ProbabilisticSampler != nil {
+			path := field.NewPath("spec").Child("samplers").Child("probabilisticSampler")
+			samplerFields[path] = action.Spec.Samplers.ProbabilisticSampler
+		}
+		if len(samplerFields) == 0 {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("samplers"), samplerFields, fmt.Sprintf("At least one of (%s) must be set", strings.Join(validSamplerFields, ", "))))
```
blumamir, line 158: since we already have a list of `validSamplerFields`, I wonder if we can iterate it somehow instead of repeating this code many times? that would also create consistency between those 2 related values, making sure in the future we don't add in one place and miss the other place

autoscaler/controllers/actions/action_controller.go
```diff
@@ -92,6 +96,61 @@ func convertActionToProcessor(ctx context.Context, k8sclient client.Client, acti
 		return processor, nil
 	}
 
+	if action.Spec.Samplers != nil {
```
blumamir, line 99: nit(style): this function is already becoming quite long. Consider if we can move some of the code that process single action into dedicated function

blumamir (approved):

autoscaler/controllers/actions/actions_webhook.go
damemi: I mentioned in https://github.com/odigos-io/odigos/pull/3423#discussion_r2327915820 (marked resolved so not sure if you saw it) that ideally we shouldn't even need all this validation, and you can configure multiple actions in one object. This is brittle and not a long term solution, but it keeps the code easier to read right now since each reconcile is only parsing one type of config

autoscaler/controllers/actions/action_controller.go
damemi: Yeah I think expanding the ActionHandler interface to all actions (not just sampler) will greatly simplify this section

PR #3436, opened 2025-09-06
chore: introduce an enum for env injection decision and use in reconcile

blumamir:
Fixes: CORE-205

Just renaming the existing property, and introducing an enum for the decisions.

Using one enum for both config and decision was confusing, as the `loader-fallback-to-pod-manifest` does not feet well into this step which is what we are actually going to use

### DIFF
--- a/common/env_injection_method.go
+++ b/common/env_injection_method.go
@@ -25,6 +25,6 @@ type EnvInjectionDecision string
 // this decision is based on the runtime inspection, user overrides, and the distro support,
 // and reflects what odigos is actually plan to use.
 const (
-	EnvInjectionDecisionLoader      EnvInjectionDecision = "loader"
-	EnvInjectionDecisionPodManifest EnvInjectionDecision = "pod-manifest"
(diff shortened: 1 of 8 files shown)


common/env_injection_method.go
```diff
@@ -16,3 +16,15 @@ const (
...
+// +kubebuilder:validation:Enum=loader;pod-manifest
+type EnvInjectionDecision string
+
+// The decision on the actual injection method to use for a specific pod.
+// While the configuration allows one to choose options with fallbaks,
+// this decision is based on the runtime inspection, user overrides, and the distro support,
+// and reflects what odigos is actually plan to use.
+const (
+	EnvInjectionDecisionLoader      EnvInjectionDecision = "loader"
+	EnvInjectionDecisionPodManifest EnvInjectionDecision = "pod-manifest"
```
RonFed: nit
```suggestion
	EnvInjectionDecisionLoader     =  EnvInjectionDecision(LoaderEnvInjectionMethod)
	EnvInjectionDecisionPodManifest = EnvInjectionDecision(PodManifestEnvInjectionMethod)
```

PR #3450, opened 2025-09-09
feat: UI migration to Generic Action CRD

BenElferink:
Following up after #2783 & #3423
This PR changes action CRDs used by the UI, from legacy CRDs to the Generic CRD.

Legacy CRDs do not need continued support in read-actions, because @damemi will follow up with a backend PR converting all legacy actions into the generic action.

emergency-ticket id: EMR-233


BenElferink:
Added Operator RBAC label because we already have this permission which should also apply to actions:
```
// +kubebuilder:rbac:groups=odigos.io,resources=*,verbs=*
```

damemi (dismissed):
PR looks good, just requesting changes to make sure https://github.com/odigos-io/odigos/pull/3456 merges first so we don't break all actions in the ui

alonkeyval (approved):

PR #3454, opened 2025-09-09
chore: comment out workload conditions (temporarily)

BenElferink:
Temporarily disable the retrieval and processing of workload conditions.
This change is made to revisit the displaying of errors in the UI.

emergency-ticket id: EMR-229


alonkeyval (approved):

damemi (approved):
sad face but understandable if this is causing trouble

BenElferink:
> sad face but understandable if this is causing trouble

It will be brought back, we just have to differentiate between the various kinds of errors - customers see red and panic 😅

PR #3456, opened 2025-09-09
CORE-76: Migrate legacy Actions to generic Actions

damemi:
## Description

This PR updates all legacy Action controllers to simply be in charge of migrating legacy Actions to the new Generic CRD.

If a legacy action is created or modified, a new Generic action will be made for it with the same name prefiexed with `migrated-legacy-`. The Generic action will have the legacy action as an ownerreference. No legacy actions will be directly used in the generation of any Processor configs, nor will they be listed as ownerreferences on the Processor (instead the generated Generic action will take their place).

The UI migration to Generic Actions CRD in https://github.com/odigos-io/odigos/pull/3450 means that no legacy Actions will be visible in the UI. Because of that, there is no need to continue supporting both code paths in each controller.

## How Has This Been Tested?

- [x] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


autoscaler/controllers/actions/addclusterinfo_controller.go
```diff
@@ -45,21 +44,28 @@ func (r *AddClusterInfoReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	processor, err := r.convertToProcessor(action)
+	// Migrate to odigosv1.Action
+	migratedActionName := v1.ActionMigratedLegacyPrefix + action.Name
+	odigosAction := &v1.Action{}
+	err = r.Get(ctx, client.ObjectKey{Name: migratedActionName, Namespace: action.Namespace}, odigosAction)
 	if err != nil {
-		r.ReportReconciledToProcessorFailed(ctx, action, FailedToTransformToProcessorReason, err.Error())
```
damemi, line 50: worth noting, in the new code I'm not updating the legacy action with its status anymore. This is because update conflicts caused a hotloop that I wasn't able to figure a good way around (if trying to update the action failed due to a conflict, then trying to update the status would still be a conflict, and requeues never seemed to sort it out)

BenElferink (approved):
LFG 🔥

autoscaler/controllers/actions/action_controller.go
```diff
@@ -46,6 +47,11 @@ func convertActionToProcessor(ctx context.Context, k8sclient client.Client, acti
 	}
 
 	if action.Spec.PiiMasking != nil {
+		/*for _, signal := range action.Spec.Signals {
+			if _, ok := piiMaskingSupportedSignals[signal]; !ok {
+				return nil, fmt.Errorf("unsupported signal in PiiMasking action: %s", signal)
+			}
+		}*/
```
RonFed: should be removed?

autoscaler/controllers/actions/addclusterinfo_controller.go
RonFed, line 1: Cam we add above `type AddClusterInfoReconciler struct` (and all other legacy controllers)
a comment saying something like
`Deprecated: now using generic Actions`
This will help when we will remove all the migration code in the future - and also helps in the IDE to mark types as deprecated

autoscaler/controllers/actions/k8sattributesresolver_controller.go
```diff
@@ -129,120 +125,81 @@ func (r *K8sAttributesResolverReconciler) Reconcile(ctx context.Context, req ctr
...
+		// Action doesn't exist, create new one
+		odigosAction = r.createMigratedAction(action, migratedActionName)
+		err = r.Create(ctx, odigosAction)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	} else {
+		// Action exists, update it
+		updatedAction := r.createMigratedAction(action, migratedActionName)
+		updatedAction.ResourceVersion = odigosAction.ResourceVersion
```
RonFed: Why is this needed?

blumamir: +1

autoscaler/controllers/actions/addclusterinfo_controller.go
```diff
@@ -45,21 +44,28 @@ func (r *AddClusterInfoReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		}
+		// Action doesn't exist, create new one
+		odigosAction = r.createMigratedAction(action, migratedActionName)
+		err = r.Create(ctx, odigosAction)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	} else {
+		// Action exists, update it
+		updatedAction := r.createMigratedAction(action, migratedActionName)
```
blumamir: If the user updates the "migrated" action from UI after it was created, this change will override his input back to the "non-migrated" action, right?

I think we should only migrate once and then never update the migrated generic object

autoscaler/controllers/actions/addclusterinfo_controller.go
```diff
@@ -45,21 +44,28 @@ func (r *AddClusterInfoReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	err = r.Patch(ctx, processor, client.Apply, client.FieldOwner(action.Name), client.ForceOwnership)
-	if err != nil {
-		r.ReportReconciledToProcessorFailed(ctx, action, FailedToCreateProcessorReason, err.Error())
-		return ctrl.Result{}, err
-	}
-
-	err = r.ReportReconciledToProcessor(ctx, action)
-	if err != nil {
-		return ctrl.Result{}, err
+		if client.IgnoreNotFound(err) != nil {
```
blumamir: nit: `IgnoreNotFound(err)` is useful when you plan to return this value from the function.

If you just want to check if it's not found or not, use `apierrors.IsNotFound(err)`.

`client.IgnoreNotFound(err) != nil` is quite confusing to read (for me at least) as it has multiple logical negative directives (`Ignore`, `NotFound`, `!=`)

autoscaler/controllers/actions/action_controller.go
damemi: should actually be uncommented, thanks

autoscaler/controllers/actions/addclusterinfo_controller.go
damemi: yup done

autoscaler/controllers/actions/k8sattributesresolver_controller.go
damemi: this pattern was confusing, I changed it

basically this approach was always initializing an empty Action object, regardless of if we were creating or updating. Updates need a resource version, so in the update case this was filling that in.

A better way is to get the object from the cluster and just update the fields that need to be changed, so resource version is already set from the get. That's what I'm doing now

autoscaler/controllers/actions/addclusterinfo_controller.go
damemi: This is a good question and goes to @RonFed's comment above https://github.com/odigos-io/odigos/pull/3456#pullrequestreview-3205617470

Updating allows users that are deploying with helm or manifests to take time migrating their manifests over, but still continue redeploying without issues.

This does create a problem if someone modifies the migrated action from another source, such as the UI.

However, that would be a problem anyway (even today with legacy actions) if a user is deploying from helm/manifest right?

damemi:
Chatted with @blumamir offline and we decided to make the following changes:

**Action migration will be one-way, one time only** - Similar to how we migrated Source labels->CRDs, the migration will happen once, and future updates to the legacy object will not take effect. This prevents us from having to be concerned with the multiple-source-of-truth problem between the manifest+UI. This will be logged in the Autoscaler to let the user know.

**Migrated Action will own Legacy Action** - At first, this PR was the other way around (if a user has a legacy Action, it would own the migrated Action). In practice we want it the opposite, because the UI and all of Odigos will now be interacting with the migrated Action. So, for example, if a user creates a legacy Action, it will show up in the UI (as a migrated Action). If the user deletes that migrated action from the UI, it needs to delete the legacy Action or there will be a hotloop where the legacy action recreates the migrated action.

**Not deleting legacy Actions** - The update+ownerreference change above should in spirit accomplish the same thing as @RonFed suggested by deleting the legacy action once it's migrated. The legacy Action will now have no effect on the cluster. It is a bit aggressive to delete the user's object at this point, though in the future we might do that as we do remove the CRD.

I also updated the actions docs to use the new CRD and will open a PR for a blog post announcing the change and the impact it has.

blumamir (approved):

PR #3462, opened 2025-09-10
CORE-251: clean up node labels in cleanup job 

AvihuHenya:
## Description
Add deletion logic for node lables.
Ensure node labels (odigos.io/odiglet-oss-installed, odigos.io/odiglet-enterprise-installed) are cleaned up during uninstall
Add test verification in assert_odigos_uninstalled.sh to check no Odigos labels remain on nodes


helm/odigos/templates/cleanup/cleanup-clusterrole.yaml
```diff
@@ -11,4 +11,10 @@ rules:
 - apiGroups: [""]
   resources: ["pods"]
   verbs: ["list"]
+- apiGroups: [""]
+  resources: ["nodes"]
+  verbs: ["list", "patch"]
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get", "delete"]
```
blumamir: Is this being used for the namespace detection?

Can we run the cleanup job `uninstall` with `--namespace` and avoid the cluster wide configmap permissions?

blumamir: anyhow, `delete` for configmaps in the entire cluster is not something we want to allow. Is it actually needed? better to only allow as a role and not in cluster scope

blumamir (commented):
Looks good, just one request for the new cluster permissions which we should try to avoid

blumamir (approved):

blumamir (approved):

PR #3476, opened 2025-09-12
chore: cleanup complexity from device-plugin

blumamir:
Fixes: CORE-271

The device-plugin implementation used to serve multiple languages and had quite some complexity to achieve this.

this component is very important to be run as stable as possible since the device health is required for pods to be scheduled and run on the node, thus we want to make it simple.

Now that we have just one device, some of the complexity is no longer needed, and we can clean it up and get a thiner implementation that is easier to maintain going forward.

Fixes:

- get rid of the Musl plugin
- use just one "Generic" plugin instead of having a plugin with response function as parameter
- remove the handling of the device id during allocate, as it is not needed in the single generic device response (used in very old implementation where the service name would be the device id and then the collector would substitute it)
- copy 2 consts into the device plugin code to remove tons of dependencies coming from odigos api just for a constant.
- remove code from webhook that was used for handling legaacy language specific, non generic devices which no longer exists.
- move the device constant from `common` to `api` since it's a k8s specific value (not relevant to vmagent)
- remove env from device plugin manifest (was used for setting up resource attributes and otlp endpoints in non-generic device). no longer used.
- reduce the resources for the device pluging container, since it's using wayyyy less than the amount we currently use (which is multiplied to be on every node on the cluster).
- remove the otel logs zap bridge. it was used for odigos cloud, but now just prints this into the logs: `2025/09/12 08:02:47 Post "https://localhost:4318/v1/logs": http: server gave HTTP response to HTTPS client`. Can be re-introduced sometime in the future if we want to support it again

### DIFF
--- a/cli/cmd/resources/odiglet.go
+++ b/cli/cmd/resources/odiglet.go
@@ -727,6 +727,10 @@ func NewOdigletDaemonSet(odigletOptions *OdigletDaemonSetOptions) *appsv1.Daemon
 										Port: intstr.FromInt(13133),
 									},
 								},
+								InitialDelaySeconds: 10,
+								FailureThreshold:    3,
+								PeriodSeconds:       10,
+								TimeoutSeconds:      5,
 							},
 							ReadinessProbe: &corev1.Probe{
 								ProbeHandler: corev1.ProbeHandler{
@@ -735,6 +739,10 @@ func NewOdigletDaemonSet(odigletOptions *OdigletDaemonSetOptions) *appsv1.Daemon
 										Port: intstr.FromInt(13133),
 									},
 								},
+								InitialDelaySeconds: 10,
+								FailureThreshold:    3,
+								PeriodSeconds:       10,
+								TimeoutSeconds:      5,
 							},
 							StartupProbe: &corev1.Probe{
 								ProbeHandler: corev1.ProbeHandler{
--- a/.github/actions/tag-modules/action.yml
+++ b/.github/actions/tag-modules/action.yml
@@ -1,0 +1,25 @@
+name: "Tag Odigos Modules"
+description: "It is needed so they can be consumed as libraries in other projects in the go.mod"
+
+inputs:
+  tag:
+    description: "The tag to use for all modules"
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Tag Modules
+      shell: bash
+      run: |
+        git tag api/${{ inputs.tag }}
+        git tag common/${{ inputs.tag }}
+        git tag instrumentation/${{ inputs.tag }}
+        git tag k8sutils/${{ inputs.tag }}
+        git tag odiglet/${{ inputs.tag }}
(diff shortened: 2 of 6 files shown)


api/k8sconsts/device.go
```diff
@@ -1,9 +1,11 @@
 package k8sconsts
 
+type OdigosInstrumentationDevice string
```
RonFed: Is this type used?

cli/cmd/resources/odiglet.go
```diff
@@ -759,40 +759,14 @@ func NewOdigletDaemonSet(odigletOptions *OdigletDaemonSetOptions) *appsv1.Daemon
...
-					"cpu":    resource.MustParse("100m"),
-					"memory": resource.MustParse("300Mi"),
+					"cpu":    resource.MustParse("10m"),
+					"memory": resource.MustParse("50Mi"),
 				},
 				Requests: corev1.ResourceList{
-					"cpu":    resource.MustParse("40m"),
-					"memory": resource.MustParse("200Mi"),
+					"cpu":    resource.MustParse("10m"),
+					"memory": resource.MustParse("50Mi"),
```
RonFed: Why is this changed? we used to have problems with liveness of this container

instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -321,7 +320,8 @@ func (p *PodsWebhook) injectOdigosToContainer(containerConfig *odigosv1.Containe
 	volumeMounted := false
 	containerDirsToCopy := make(map[string]struct{})
 	if distroMetadata.RuntimeAgent != nil {
-		if *config.MountMethod == common.K8sHostPathMountMethod || *config.MountMethod == common.K8sInitContainerMountMethod {
+		switch *config.MountMethod { // verfied as not nil in webhook start
+		case common.K8sHostPathMountMethod, common.K8sInitContainerMountMethod:
```
RonFed, line 337: why do we need to mount directories if the mount method is init container?

api/k8sconsts/device.go
blumamir: unused. good catch 👍

cli/cmd/resources/odiglet.go
blumamir: We have this container on every node, and it does not do anything beside replying a constant respond once every new container starts. I observed the actual values we use and it is requesting a lot more memory and cpu than what's actually needed. This change will reduce the about of resources we consume in the cluster

instrumentor/controllers/agentenabled/pods_webhook.go
blumamir: It's unrelated to current PR (was like this before my change).
What is being mounted is the volume to share files between init container and the instrumented pods

cli/cmd/resources/odiglet.go
blumamir: increased the CPU to `30m`

tamirdavid1 (approved):

PR #3478, opened 2025-09-12
chore: refactor distro params handling

blumamir:
Fixes: CORE-160

Distro params are a way to propagate parameters from the runtime detection into the webhook, to be used during the injection of the distro in the webhook.

It allows use to:
- verify the value from runtime detection is available when required
- the value is in the right format (for example - a version number)
- process the value to make it useful for the webhook (for example, calculate MAJOR.MINOR for ruby/php)
- simplify the code in webhook (get processed values without needing to validate or compute anything)
- transaction - the values are all coming from the reconcile, so if the runtime details are updated, there will be no race with the new result of the reconcile and the webhook that also access these values.

This PR extract the functions to handle the distro params to a dedicated file since the original function in the `sync` is quite long already.
It also introduces the "append env variables" found in CRI detection into the params. They will later be used to refactor the pod manifest env injection which currently uses otelsdk.

This is the error in UI when we have criError and env injection is set to pod manifest:

image

### DIFF
--- a/instrumentor/controllers/agentenabled/sync.go
+++ b/instrumentor/controllers/agentenabled/sync.go
@@ -328,19 +328,6 @@ func isLoaderInjectionSupportedByRuntimeDetails(containerName string, runtimeDet
 	return nil
 }
 
-func isPodManifestInjectionSupportedByRuntimeDetails(containerName string, runtimeDetails *odigosv1.RuntimeDetailsByContainer) *odigosv1.ContainerAgentConfig {
-	if runtimeDetails.CriErrorMessage != nil {
-		return &odigosv1.ContainerAgentConfig{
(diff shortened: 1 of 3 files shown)


instrumentor/controllers/agentenabled/sync.go
```diff
@@ -399,11 +385,6 @@ func getEnvInjectionDecision(
 		return nil, nil
 	}
 
-	err := isPodManifestInjectionSupportedByRuntimeDetails(containerName, runtimeDetails)
-	if err != nil {
-		return nil, err
-	}
```
blumamir, line 414: this check is now made when we calculate the "append env vars" to use as distro parameters later on.

edeNFed (approved):

PR #3484, opened 2025-09-14
Add Service Graph metrics when sending to Prometheus

edeNFed:
## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [X] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/common/config/prometheus.go
+++ b/common/config/prometheus.go
@@ -38,6 +38,12 @@ func (p *Prometheus) ModifyConfig(dest ExporterConfigurer, currentConfig *Config
 
 	currentConfig.Exporters[rwExporterName] = GenericMap{
 		"endpoint": fmt.Sprintf("%s/api/v1/write", url),
+		"resource_to_telemetry_conversion": GenericMap{
+			"enabled": true,
+		},
+		"external_labels": map[string]string{
+			"job": "odigos-remote-write",
+		},
 	}
 
 	resourceAttributesLabels, exists := config[prometheusResourceAttributesLabelsKey]
--- a/common/pipelinegen/config_builder.go
+++ b/common/pipelinegen/config_builder.go
@@ -257,6 +257,14 @@ func insertServiceGraphPipeline(currentConfig *config.Config) {
 	}
 	pipeline.Exporters = append(pipeline.Exporters, consts.ServiceGraphConnectorName)
 	currentConfig.Service.Pipelines[rootPipelineName] = pipeline
+
+	// Add ServiceGraphConnectorName receiver to every prometheus pipeline starts with metrics/prometheus-
+	for pipelineName, pipeline := range currentConfig.Service.Pipelines {
+		if strings.HasPrefix(pipelineName, "metrics/prometheus-") {
+			pipeline.Receivers = append(pipeline.Receivers, consts.ServiceGraphConnectorName)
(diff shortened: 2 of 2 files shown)


dan505512 (approved):
The one thing I want to note is that Prometheus has a hard time with any metrics coming out of order, I don't know if it's relevant here

common/config/prometheus.go
```diff
@@ -38,6 +38,12 @@ func (p *Prometheus) ModifyConfig(dest ExporterConfigurer, currentConfig *Config
 
 	currentConfig.Exporters[rwExporterName] = GenericMap{
 		"endpoint": fmt.Sprintf("%s/api/v1/write", url),
+		"resource_to_telemetry_conversion": GenericMap{
+			"enabled": true,
+		},
```
blumamir, line 43: the resource can contain a lot of entries. are we sure this is safe? (will not blow up number of series)

did you consider using the [`PROMETHEUS_RESOURCE_ATTRIBUTES_LABELS`](https://docs.odigos.io/backends/prometheus) or add more values to the default of this field?

common/pipelinegen/config_builder.go
```diff
@@ -257,6 +257,14 @@ func insertServiceGraphPipeline(currentConfig *config.Config) {
...
 	pipeline.Exporters = append(pipeline.Exporters, consts.ServiceGraphConnectorName)
 	currentConfig.Service.Pipelines[rootPipelineName] = pipeline
+
+	// Add ServiceGraphConnectorName receiver to every prometheus pipeline starts with metrics/prometheus-
+	for pipelineName, pipeline := range currentConfig.Service.Pipelines {
+		if strings.HasPrefix(pipelineName, "metrics/prometheus-") {
+			pipeline.Receivers = append(pipeline.Receivers, consts.ServiceGraphConnectorName)
+			currentConfig.Service.Pipelines[pipelineName] = pipeline
+		}
+	}
```
blumamir, line 267: This will make it always enabled for any Prometheus user, right? 
Just making sure that this is what we want, as there is no way to disable it (or opt in), and any existing users will now get it after upgrade.

Do you think we should include it in any metrics destination going forward? or make it an opt-in configurable setting?

common/pipelinegen/config_builder.go
```diff
@@ -257,6 +257,14 @@ func insertServiceGraphPipeline(currentConfig *config.Config) {
 	}
 	pipeline.Exporters = append(pipeline.Exporters, consts.ServiceGraphConnectorName)
 	currentConfig.Service.Pipelines[rootPipelineName] = pipeline
+
+	// Add ServiceGraphConnectorName receiver to every prometheus pipeline starts with metrics/prometheus-
+	for pipelineName, pipeline := range currentConfig.Service.Pipelines {
+		if strings.HasPrefix(pipelineName, "metrics/prometheus-") {
+			pipeline.Receivers = append(pipeline.Receivers, consts.ServiceGraphConnectorName)
```
blumamir, line 264: This assumes the connector is in the config, because metrics always enable traces, and traces enables service graph, right?
I think if one will disable the service graph in config, this part will make collector fail to load (probably not likely but supported).
wondering if we can assert that connector exist for safety for whatever scenario we might hit.

PR #3511, opened 2025-09-16
add --wasp-enabled argument to odiglet and instrumentor

RonFed:


blumamir:
did you consider reading it from effective config instead? seems that we don't use the cli flags for controllers, and reading it from the cluster will enable it to behave correctly when running locally in debugger

RonFed:
> did you consider reading it from effective config instead? seems that we don't use the cli flags for controllers, and reading it from the cluster will enable it to behave correctly when running locally in debugger

I chose this option because it allows us to choose if to launch a goroutine on start-up or not which is relevant to this config. Doing this using the odigos-configuration is subject to race conditions - (if the odiglet starts before the scheduler for example)

PR #3517, opened 2025-09-17
chore: fix re-upgrade without sources/destinations delete data collection cm

tamirdavid1:
## Description

This change ensures the odigos-data-collection ConfigMap is created once by Helm to let the collector start, but is skipped once the controller takes ownership.
If the ConfigMap still contains the default stub (no ownerReferences), Helm will continue managing it so it isn’t deleted on upgrades.
On uninstall, Helm cleans up the stub if it still owns it, while controller-managed ConfigMaps are left for the controller’s GC.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

emergency-ticket id: EMR-276

### DIFF
--- a/helm/odigos/templates/datacollection/configmap.yaml
+++ b/helm/odigos/templates/datacollection/configmap.yaml
@@ -1,17 +1,17 @@
 {{/*
-    This ConfigMap is special:
+  This ConfigMap is special:
 
-    - Normally Helm would re-apply (and overwrite) ConfigMaps on every upgrade.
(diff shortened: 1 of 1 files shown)


RonFed (approved):

RonFed (changes requested):
https://github.com/odigos-io/odigos/pull/3522
will handle this I think

PR #3519, opened 2025-09-17
CORE-227: Add GCP OTLP exporter

damemi:
## Description

GCP recently announced availability for their [Telemetry (OTLP) API](https://cloud.google.com/blog/products/management-tools/opentelemetry-now-in-google-cloud-observability) which supports the OTLP exporter instead of using the googlecloud API exporter.

However, using the OTLP exporter with GCP also requires configuring the googleclientauth extension and ensuring that the `gcp.project_id` attribute is set on traces. When running Odigos off of GCP and sending traces to the GCP OTLP endpoint, the same application credential auth is needed as was added in #3449.

This adds support for this leveraging the same k8sconfiger interface added in #3449 

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes


docs/backends/googlecloudotlp.mdx
```diff
@@ -0,0 +1,139 @@
...
+    !! Do not remove this comment, this acts as a key indicator in `docs/sync-dest-doc.py` !!
+    !! END CUSTOM EDIT !!
+*/}
+
+### Configuring Destination Fields
+
+<Accordion title="Supported Signals:">
+  ✅ Traces
+  ❌ Metrics
+  ❌ Logs
```
blumamir, line 70: To bad these aren't supported as well. Do you know if they plan to add them soon too?

destinations/data/googlecloudotlp.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+    - name: GCP_APPLICATION_CREDENTIALS
+      displayName: Application Credentials
+      componentType: textarea
+      secret: true
+      componentProps:
+        type: text
+        required: false
+        placeholder: 'Credentials config... (optional)'
+        tooltip: The JSON GCP Application Credentials file for the GCP project. Either a Workload Identity Federation config file or Service Account Key. Required if running Odigos outside of Google Cloud.
```
blumamir, line 43: if you add [testConnectionSupported: true](https://github.com/odigos-io/odigos/blob/af4c23317ff7d37bec3ad059d5e8cdce6b0f9927/destinations/data/otlp.yaml#L137) here, it will show up a button in the ui, that once pressed, will send an empty otlp message to the destination and show if it's successful or not.

replying success on empty message [is speced](https://opentelemetry.io/docs/specs/otlp/#full-success-1)

> If the server receives an empty request (a request that does not carry any telemetry data) the server SHOULD respond with success.

But it's safe to test once and make sure. This will help users of this destination to check if the destination is reachable and auth is setup properly

docs/backends/googlecloudotlp.mdx
```diff
@@ -0,0 +1,139 @@
...
+    **Obtaining Access Token**<br />
+    Go to **⚙️ > Access Tokens** and click **Create New**
+
+    !! Do not remove this comment, this acts as a key indicator in `docs/sync-dest-doc.py` !!
+    !! START CUSTOM EDIT !!
+*/}
+
+### Google Cloud Telemetry API
+
+This destination uses the [Google Cloud Telemetry (OTLP) API](https://cloud.google.com/stackdriver/docs/reference/telemetry/overview)
```
blumamir, line 29: nit: add similar comment in the other destination docs

docs/backends/googlecloudotlp.mdx
```diff
@@ -0,0 +1,139 @@
...
+to send data as OTLP to GCP Observability using an OTLP/HTTP exporter. It is different from the
+[Google Cloud Monitoring destination](https://docs.odigos.io/backends/googlecloudmonitoring)
+which uses the [googlecloud exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/d9769f7c1a0f3bea82c711af551d9b5b44d69e6f/exporter/googlecloudexporter).
```
blumamir, line 32: should we add some guidance on which one to use (recommended), or why use one over the other?

docs/mint.json
```diff
@@ -264,6 +264,7 @@
             "backends/elasticsearch",
             "backends/gigapipe",
             "backends/googlecloudmonitoring",
+            "backends/googlecloudotlp",
```
blumamir, line 267: wonder if we should also add it [here](https://github.com/odigos-io/odigos/blob/main/docs/backends-overview.mdx) for completeness. even though it is a special case (same product, different protocol)

blumamir (approved):

destinations/data/googlecloudotlp.yaml
damemi: cool, how does that work? like will this run the full collector config and call the k8sconfiger handler this needs?

docs/backends/googlecloudotlp.mdx
damemi: I don't know for sure what the plans are, imo it wouldn't surprise me if they eventually did but they have basically separate products for logs and metrics so 🤷

docs/mint.json
damemi: didn't know about that table, yeah I think so

damemi: oh looks like it was auto added

destinations/data/googlecloudotlp.yaml
blumamir: It's being sent from the UI pod which sets up a minimal collector exporter to initiate the empty otlp request.
https://github.com/odigos-io/odigos/blob/main/frontend/services/test_connection/otlphttp_test_connection.go

docs/backends/googlecloudotlp.mdx
```diff
@@ -0,0 +1,141 @@
...
+
+Odigos currently supports Standard (non-Autopilot) GKE clusters.
```
blumamir: It actually supports any cluster now, maybe we should rephrase this part to make it more obvious that this only relates to the automatic authentication?

blumamir: (also for the other destination docs I guess)

destinations/data/googlecloudotlp.yaml
damemi: It doesn't look like that will pull in the extra config needed with the gcp auth extension, but this is something we should probably TODO at some point

docs/backends/googlecloudotlp.mdx
damemi: I don't actually know what Google's guidance is at this point

docs/backends/googlecloudotlp.mdx
damemi: on GCP we still don't support Autopilot, but I made it clearer that this is only referring to GCP clusters

PR #3385, opened 2025-08-27
Combine odiglet datacollection - CORE-137

tamirdavid1:
## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

- [ ] Changes how Odigos interacts with Kubernetes
- [ ] Introduces additional calls to the API Server (potential performance impact)
- [ ] New Query/feature supported in all the k8s versions supported by Odigos
- [ ] Modifies Odigos manifests (addressed in both CLI and Helm)
- [ ] Changes RBAC permissions

## User Facing Changes

- [ ] Users need to take action before upgrading
- [ ] Automatic migration will modify existing objects (backward compatible)
- [ ] Changes UI, CLI, or K8s Manifests aspects in a way that users need to be aware of
- [ ] Documentation updated accordingly

PR #3522, opened 2025-09-18
handle data-collection config map creation in autoscaler reconciler, enhance config map provider in the collector

RonFed:


autoscaler/controllers/nodecollector/sync.go
```diff
@@ -28,17 +29,15 @@ func reconcileNodeCollector(ctx context.Context, c client.Client, scheme *runtim
...
-	if len(ics.Items) == 0 {
-		logger.V(3).Info("No odigos sources found, skipping data collection sync")
-		return ctrl.Result{}, nil
-	}
-
 	odigosNs := env.GetCurrentNamespace()
 
-	var dataCollectionCollectorGroup odigosv1.CollectorsGroup
-	err := c.Get(ctx, client.ObjectKey{Namespace: odigosNs, Name: k8sconsts.OdigosNodeCollectorCollectorGroupName}, &dataCollectionCollectorGroup)
+	dataCollectionCollectorGroup := new(odigosv1.CollectorsGroup)
```
damemi: curious why the new() instead of just declaring an empty one?

helm/odigos/templates/odiglet/daemonset.yaml
```diff
@@ -231,6 +231,12 @@ spec:
             httpGet:
               path: /
               port: 13133
+          startupProbe:
```
damemi, line 247: what endpoint does this startup probe actually hit that tells it whether it's ready or not?

autoscaler/controllers/nodecollector/configmap.go
```diff
@@ -38,14 +39,14 @@ func SyncConfigMap(sources *odigosv1.InstrumentationConfigList, signals []odigos
 
 	processors := commonconf.FilterAndSortProcessorsByOrderHint(allProcessors, odigosv1.CollectorsGroupRoleNodeCollector)
 
-	desired, err := getDesiredConfigMap(sources, signals, processors, datacollection, scheme)
+	desired, err := getDesiredConfigMap(ctx, sources, signals, processors, datacollection, scheme, c)
```
damemi: this function SyncConfigMap is only called in reconcileNodeCollector which is called by the collectorsgroup controller, instrumentationconfig controller, and processor controller

how does it work then that this gets triggered on an initial startup where there aren't any of these, which seems like the no op configmap is meant for? or is the nop cm just for when there aren't destinations?

helm/odigos/templates/odiglet/daemonset.yaml
blumamir: I guess it's the one from the [healthcheckextension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/healthcheckextension) which becomes ready once the collector starts. like the livness and readiness above.

being [setup here](https://github.com/odigos-io/odigos/blob/af4c23317ff7d37bec3ad059d5e8cdce6b0f9927/autoscaler/controllers/nodecollector/configmap.go#L334)

autoscaler/controllers/nodecollector/sync.go
blumamir: I usually also do 

```suggestion
	dataCollectionCollectorGroup := &odigosv1.CollectorsGroup{}
```

autoscaler/cmd/main.go
```diff
@@ -265,14 +265,9 @@ func main() {
...
-		setupLog.Error(err, "Instrumentor exited with error")
-	}
-
-	setupLog.Info("starting manager")
-	if err := mgr.Start(ctx); err != nil {
-		setupLog.Error(err, "problem running manager")
-		os.Exit(1)
+		setupLog.Error(err, "autoscaler exited with error")
 	}
+	setupLog.V(0).Info("autoscaler exiting")
```
blumamir: for error we will have 2 lines: exited -> exiting. should we write the 2nd log only on success?

autoscaler/controllers/nodecollector/configmap.go
```diff
@@ -95,24 +96,85 @@ func createConfigMap(desired *v1.ConfigMap, ctx context.Context, c client.Client
...
+		},
+		Service: config.Service{
+			Extensions: []string{"health_check"},
+			Pipelines: map[string]config.Pipeline{
+				"traces": {
+					Receivers:  []string{"otlp"},
+					Processors: []string{},
+					Exporters:  []string{"nop"},
+				},
+			},
```
blumamir, line 150: should we also include metrics here so if an agent is trying to send it to us for any reason, it will not fail?

autoscaler/controllers/nodecollector/configmap.go
```diff
@@ -95,24 +96,85 @@ func createConfigMap(desired *v1.ConfigMap, ctx context.Context, c client.Client
...
+			Namespace: env.GetCurrentNamespace(),
 		},
 		Data: map[string]string{
 			k8sconsts.OdigosNodeCollectorConfigMapKey: cmData,
 		},
 	}
 
-	if err := ctrl.SetControllerReference(datacollection, &desired, scheme); err != nil {
+	autoscalerDeployment := appsv1.Deployment{}
+	err = c.Get(ctx, client.ObjectKey{Namespace: env.GetCurrentNamespace(), Name: k8sconsts.AutoScalerDeploymentName}, &autoscalerDeployment)
```
blumamir: nit: not sure if supported, but do you think we can do it just once and store the deployment somewhere, and just use it here, without `err` and without needing to pass the client into this function?

since the deployment cannot change while this code runs :)

autoscaler/controllers/nodecollector/configmap.go
blumamir: I remember we chatted about it at the office, wonder too how this get's created the first time eventually.

RonFed: looking more into this the reconciler is defined with 
```Go
		Named("nodecollector-collectorsgroup").
		Owns(&appsv1.DaemonSet{}). // in case the node-collector ds is deleted or modified for any reason, this will reconcile and recreate it
		Owns(&corev1.ConfigMap{}). // in case the configmap of the node-collector is deleted or modified for any reason, this will reconcile and recreate it
		// we assume everything in the collectorsgroup spec is the configuration for the collectors to generate.
		// thus, we need to monitor any change to the spec which is what the generation field is for.
		WithEventFilter(
			predicate.Or(
				predicate.And(&odigospredicate.OdigosCollectorsGroupNodePredicate, &predicate.GenerationChangedPredicate{}),
				predicate.And(&odigospredicate.OdigosCollectorsGroupClusterPredicate, &odigospredicate.ReceiverSignalsChangedPredicate{}),
			)).
```

looking at the logs what triggers this is the creation of the gateway collectors group. The `ReceiverSignalsChangedPredicate` return true on Create events - so that is the trigger.

We can make this more explicit and avoid potential future bugs if something in the collectors group logic changes.
I also thing the `Owns(&appsv1.DaemonSet{})` is also not correct anymore after #3385 - but that is probably something better to do in a follow up PR.
WDYT? @blumamir @damemi

autoscaler/controllers/nodecollector/manager.go
```diff
@@ -25,10 +25,12 @@ func SetupWithManager(mgr ctrl.Manager, imagePullSecrets []string, odigosVersion
 				predicate.And(&odigospredicate.OdigosCollectorsGroupClusterPredicate, &odigospredicate.ReceiverSignalsChangedPredicate{}),
 			)).
 		Complete(&CollectorsGroupReconciler{
-			Client:           mgr.GetClient(),
-			Scheme:           mgr.GetScheme(),
-			ImagePullSecrets: imagePullSecrets,
-			OdigosVersion:    odigosVersion,
+			nodeCollectorBaseReconciler: nodeCollectorBaseReconciler{
```
blumamir, line 27: nit: now that it's all the same, we can create it just once and use the same instance for each controller in this file

autoscaler/controllers/nodecollector/configmap.go
```diff
@@ -32,24 +32,34 @@ const (
...
+	if b.autoscalerDeployment == nil {
+		// we only need to get the autoscaler deployment once since it can't change while this code is running
+		// (since we are running in the autoscaler pod)
+		autoscalerDeployment := &appsv1.Deployment{}
+		err := b.Client.Get(ctx, client.ObjectKey{Namespace: env.GetCurrentNamespace(), Name: k8sconsts.AutoScalerDeploymentName}, autoscalerDeployment)
+		if err != nil {
+			return err
+		}
+		b.autoscalerDeployment = autoscalerDeployment
+	}
```
blumamir, line 51: nit: can we do it once in `main` maybe and the use the deployment as a parameter while setting up manager?

autoscaler/controllers/nodecollector/configmap.go
RonFed: update: see https://github.com/odigos-io/odigos/pull/3522/commits/66d8aeff530b01d9f4b268d44f715750a5543ebc

I added a controller that reacts only to create event of the autoscaler deployment - and it "Owns" the config map - so any modifications to the config - map will be reconciled as well - since now the cg is no longer the owner of the config map.

autoscaler/controllers/nodecollector/configmap.go
damemi: I was wondering the same thing but I think if the autoscaler deployment does change at all (status could change for example, something small in k8s) then we might end up having to retry. which is also not bad but I think it's just not perfect either option

autoscaler/controllers/nodecollector/manager.go
```diff
@@ -25,10 +24,36 @@ func SetupWithManager(mgr ctrl.Manager, imagePullSecrets []string, odigosVersion
...
+		// this controller is supposed to react to:
+		// 1. 	creation of the autoscaler deployment -
+		// 	 	single event in the lifecycle of this pod which is part of the autoscaler deployment (one event on start up).
+		//      if no sources or destinations are defined this will create a no-op config map upon a fresh install.
+		// 2. 	changes to the node collector config map - which is owned by the autoscaler deployment -
+		// 		if for any reason the config map is modified or deleted, this will reconcile it.
+		WithEventFilter(
+			predicate.Or(
+				predicate.And(&odigospredicate.ObjectNamePredicate{AllowedObjectName: k8sconsts.AutoScalerDeploymentName}, &odigospredicate.CreationPredicate{}),
+				predicate.And(&odigospredicate.ObjectNamePredicate{AllowedObjectName: k8sconsts.OdigosNodeCollectorConfigMapName}, &odigospredicate.OnlyUpdatesPredicate{}),
```
blumamir, line 50: how does it work with `Own`? is the event filter called with the `NamespacedName` of the `Owned` object?

RonFed: yes from controller-runtime docs:
```
// WithEventFilter sets the event filters, to filter which create/update/delete/generic events eventually
// trigger reconciliations. For example, filtering on whether the resource version has changed.
// Given predicate is added for all watched objects and thus must be able to deal with the type
// of all watched objects.
```

helm/odigos/templates/datacollection/configmap.yaml
```diff
@@ -1,44 +0,0 @@
-{{/*
```
damemi, line 1: @RonFed I tried installing 1.5.0-rc1 from the helm chart locally (fresh install) and get this crashloop error in the odiglet data-collection pod:

```
$ kubectl logs pod/odiglet-nt5vh -c data-collection
Error: failed to get config: cannot resolve the configuration: cannot retrieve the configuration: failed to get configmap: configmaps "odigos-data-collection" not found
2025/09/23 15:20:58 collector server run finished with error: failed to get config: cannot resolve the configuration: cannot retrieve the configuration: failed to get configmap: configmaps "odigos-data-collection" not found
```

this should probably resolve itself, right? Still looks a little bad

RonFed: @damemi 1.5.0-rc1 does not include this change - so that is the behavior.
this should hopefully be part of rc2 or something - and these errors should go away - at max the collector will log that it is waiting for the config to get created we should not take a lot of  time - until the autoscaler starts.

damemi: hm actually I think I deployed from main, but used the rc1 image, sorry for the false alarm

PR #3524, opened 2025-09-18
ci: publish modules on promote

blumamir:
Fixes: CORE-267

This PR is a fix for us not publishing the tags for all the specific modules in promote rc to stable.

This PR extracts the tagging of the modules into a reusable action, and then using it from both "publish-modules" and "promote-rc-to-stable"

### DIFF
--- a/cli/cmd/resources/instrumentor.go
+++ b/cli/cmd/resources/instrumentor.go
@@ -408,7 +408,7 @@ func NewSourceMutatingWebhookConfiguration(ns string) *admissionregistrationv1.M
 					},
 				},
 				FailurePolicy:      ptrGeneric(admissionregistrationv1.Fail),
-				ReinvocationPolicy: ptrGeneric(admissionregistrationv1.IfNeededReinvocationPolicy),
+				ReinvocationPolicy: ptrGeneric(admissionregistrationv1.NeverReinvocationPolicy),
 				SideEffects:        ptrGeneric(admissionregistrationv1.SideEffectClassNone),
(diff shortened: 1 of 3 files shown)


.github/actions/tag-modules/action.yml
```diff
@@ -0,0 +1,25 @@
...
+        git tag api/${{ inputs.tag }}
+        git tag common/${{ inputs.tag }}
+        git tag instrumentation/${{ inputs.tag }}
+        git tag k8sutils/${{ inputs.tag }}
+        git tag odiglet/${{ inputs.tag }}
+        git tag opampserver/${{ inputs.tag }}
+        git tag procdiscovery/${{ inputs.tag }}
+        git tag profiles/${{ inputs.tag }}
+        git tag distros/${{ inputs.tag }}
+        git tag instrumentor/${{ inputs.tag }}
```
RonFed, line 24: nit: do you think we can "discover" the go.mod directories here and tag them?
it will help avoiding new modules being forgotten to add here - wdyt?

RonFed (approved):

.github/actions/tag-modules/action.yml
blumamir: we only need to tag what we end up using as a dependency in other repos, which are not all the modules in this repo. I think it is healthy to have this list explicit, and add something here after one understand what he does and what it is used for, as oppose to just creating new tags automatically where not required or noisy.

For example, `cli` is an internal module not consumed elsewhere and not part of this list. also `destinations` etc.

PR #3482, opened 2025-09-14
Adopt helm in cli install

tamirdavid1:
## Description

This PR introduces a new approach for installing, upgrading, and uninstalling Odigos.
We’re moving to a Helm-first strategy by wrapping installation logic with the Helm SDK. This simplifies our development workflow, ensures consistency with Helm’s ecosystem, and reduces the amount of custom Kubernetes logic we need to maintain.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

- [ ] Changes how Odigos interacts with Kubernetes
- [ ] Introduces additional calls to the API Server (potential performance impact)
- [ ] New Query/feature supported in all the k8s versions supported by Odigos
- [ ] Modifies Odigos manifests (addressed in both CLI and Helm)
- [ ] Changes RBAC permissions

## User Facing Changes

- [ ] Users need to take action before upgrading
- [ ] Automatic migration will modify existing objects (backward compatible)
- [ ] Changes UI, CLI, or K8s Manifests aspects in a way that users need to be aware of
- [ ] Documentation updated accordingly

PR #3533, opened 2025-09-21
Pods webhook improvments

blumamir:
Fixes: CORE-270

Minor improvements for the pods webhook:

- Run only on "CREATE" events (no need to run on "UPDATE" since we do our mutations only on created pods.
- Do not reinvoke the webhook - this was added to support more cases when there are other web-hooks, but it is not tested, and I think does not actually works at the moment. we can always re-introduce it after we test and design it properly.
- Allow to set the `TimeoutSeconds` property for the webhook timeout. default is 10 seconds and can be increased or decreased if any issues are observed


helm/odigos/templates/instrumentor/webhooks.yaml
```diff
@@ -20,15 +20,14 @@ webhooks:
...
         apiGroups: [""]
         apiVersions: ["v1"]
         resources: ["pods"]
         scope: Namespaced
     failurePolicy: Ignore
-    reinvocationPolicy: IfNeeded
+    reinvocationPolicy: Never
     sideEffects: None
-    timeoutSeconds: 10
+    timeoutSeconds: {{ .Values.instrumentor.podsWebhookTimeoutSeconds }}
```
edeNFed: Wouldn't this resolve to empty value if not set by the user?

damemi: +1

cli/cmd/resources/instrumentor.go
```diff
@@ -463,7 +462,7 @@ func NewPodMutatingWebhookConfiguration(ns string) *admissionregistrationv1.Muta
 					},
 				},
 				FailurePolicy:      ptrGeneric(admissionregistrationv1.Ignore),
-				ReinvocationPolicy: ptrGeneric(admissionregistrationv1.IfNeededReinvocationPolicy),
+				ReinvocationPolicy: ptrGeneric(admissionregistrationv1.NeverReinvocationPolicy),
 				SideEffects:        ptrGeneric(admissionregistrationv1.SideEffectClassNone),
 				TimeoutSeconds:     intPtr(10),
```
damemi, line 467: should this be configurable like in the helm chart?

helm/odigos/values.yaml
```diff
@@ -302,6 +302,11 @@ instrumentor:
   # loader-fallback-to-pod-manifest: try using the odigos loader first, and if it fails, fallback to adding the environment variables to the container manifest.
   agentEnvVarsInjectionMethod: ''
 
+  # the timeout in seconds for the pods webhook.
+  # default is 10 seconds and it is recommended to keep it as is unless you are experiencing issues with the webhook.
```
damemi, line 306: you could add something like: a shorter timeout will relieve API pressure, but may result in more pods failing to instrument

helm/odigos/templates/instrumentor/webhooks.yaml
blumamir: It will not get populated in this case, which defaults to 10seconds.

https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts

> The timeout for an admission webhook defaults to 10 seconds.

This is to make it clear that user is not interested in any specific value.

helm/odigos/values.yaml
blumamir: like

cli/cmd/resources/instrumentor.go
blumamir: for cli, we have way less configuration options right now and there are many examples.
If something is really niche we support it in helm only.

Perhaps once #3482 landed all those very advanced options can be supported in cli with a `values.yaml` file

helm/odigos/templates/instrumentor/webhooks.yaml
blumamir: made it explicit so it's less confusing 

```
timeoutSeconds: {{ default 10 .Values.instrumentor.podsWebhookTimeoutSeconds }}
```

PR #3547, opened 2025-09-21
refactor(autoscaler): extract own metrics part to seperate file

blumamir:
The config map calculation function for node collector has become so long and cluttered. 

I want to add the spanmetrics to this configmap, and thought to start by first doing some cleanups, removing the parts related to own-metrics calculation from the very long function and into it's own file.

Don't want to change too much at once, so doing this first step, which I might improve going forward

### DIFF
--- a/autoscaler/controllers/nodecollector/collectorconfig/ownmetrics.go
+++ b/autoscaler/controllers/nodecollector/collectorconfig/ownmetrics.go
@@ -1,0 +1,142 @@
+package collectorconfig
+
+import (
+	"fmt"
+
+	"github.com/odigos-io/odigos/common/config"
+	"github.com/odigos-io/odigos/common/consts"
+	semconv "go.opentelemetry.io/otel/semconv/v1.5.0"
+)
+
+// this processor should be added to any user telemetry pipeline to track the amount of data being exported by each source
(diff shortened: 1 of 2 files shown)


autoscaler/controllers/nodecollector/collectorconfig/ownmetrics.go
```diff
@@ -0,0 +1,142 @@
...
+// Notice: this config part requires that you add the OdigosTrafficMetricsProcessorName processor
+// to any pipeline for user telemetry to work correctly
+func CalculateOwnMetricsConfig(ownMetricsPort int32, odigosNamespace string) config.Config {
+	return config.Config{
+		Receivers:  calculateReceiversConfigForOwnMetrics(ownMetricsPort),
+		Exporters:  calculateExportersConfigForOwnMetrics(odigosNamespace),
+		Processors: calculateProcessorsConfigForOwnMetrics(),
+		Service: config.Service{
+			Pipelines: calculatePipelinesConfigForOwnMetrics(),
+			Telemetry: calculateServiceTelemetryConfigForOwnMetrics(ownMetricsPort),
```
RonFed, line 139: nit: consider removing the `calculate` prefix from all the names

PR #3550, opened 2025-09-21
refactor: extract common node collector config to file

blumamir:
Fixes: CORE-271

This is continue of #3547 , it extracts all the "common" things into a file, and refactor the code to use it.

To have logical units like "own-metrics", "common", "traces", "metrics", "logs", etc... I've introduced util function that can merge 2 collector configs into one. At the moment, just those 2 are combined into one, and rest will be followed in other PRs


edeNFed (approved):

common/config/config.go
```diff
@@ -51,3 +53,89 @@ type Pipeline struct {
...
+	if len(telemetry1.Metrics) > 0 && len(telemetry2.Metrics) == 0 {
+		return telemetry1, nil
+	} else if len(telemetry1.Metrics) == 0 && len(telemetry2.Metrics) > 0 {
+		return telemetry2, nil
+	}
+	// if both are empty return either one
+	if len(telemetry1.Metrics) == 0 && len(telemetry2.Metrics) == 0 {
+		return telemetry1, nil
+	}
+	return Telemetry{}, fmt.Errorf("service telemetry config is allowed to be set only once")
```
RonFed, line 127: en(telemetry1.Metrics) > 0 && len(telemetry2.Metrics) > 0

we would return an error? is this expected?

common/config/config.go
```diff
@@ -51,3 +53,89 @@ type Pipeline struct {
...
+		mergedConfig.Service.Telemetry, err = mergeTelemetry(mergedConfig.Service.Telemetry, config.Service.Telemetry)
+		if err != nil {
+			return Config{}, err
+		}
+	}
+	return mergedConfig, nil
+}
+
+func mergeExtensions(extensions1 []string, extensions2 []string) []string {
+	// TODO: check for duplicates and return an error
```
RonFed, line 96: should we do this?
duplication will probably crash the collector?

autoscaler/controllers/nodecollector/collectorconfig/common.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/odigos-io/odigos/k8sutils/pkg/env"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+)
+
+const (
+	BatchProcessorName           = "batch"
+	MemoryLimiterProcessorName   = "memory_limiter"
+	NodeNameProcessorName        = "resource/node-name"
+	ClusterCollectorExporterName = "otlp/out-cluster-collector"
+	OTLPInReceiverName           = "otlp/in"
```
RonFed, line 19: should these be exported? seems like they are private to node collector config map package

common/config/config.go
blumamir: I don't know what should populate the "Telemetry" part of the pipeline. currently only own-metrics cares about it. if it ever encounters I don't want to try to merge them here, since not sure what the usecase is and how to check for conflicts. so now just assuming only one will set it

common/config/config.go
blumamir: we should - yes :)
but I choose to postpone it for later

autoscaler/controllers/nodecollector/collectorconfig/common.go
blumamir: They start with capital letter since they are needed for the pipeline building in the configmap.go.

Once we move more config sections into this package, we can consider making them private, but I think it's still nice to document them as "consumed in other config builders"

PR #3551, opened 2025-09-22
chore: improve text when there are no running pods

blumamir:
Fixes: CORE-272

Before:

image

After:

image


blumamir:
> @blumamir nit but did you mean to change these from "runtime detection" -> "runtime inspection"?

@damemi 

We always mix the names:

- Runtime detection
- Runtime Inspection
- Language detection

It wasn't intentional, but good opportunity to pick one and use it consistently here and going forward (also maybe rename other places).
What do you think works best? I like `runtime detection` but `runtime inspection` also sounds good

damemi:
@blumamir I also like `runtime detection`

PR #3552, opened 2025-09-22
refactor: extract trace related config to it's own file

blumamir:
Fixes: CORE-271

extract the "traces" related config to it's own file and use it generically from the autoscaler nodecollector configmap calculation

### DIFF
--- a/autoscaler/controllers/nodecollector/collectorconfig/common.go
+++ b/autoscaler/controllers/nodecollector/collectorconfig/common.go
@@ -12,11 +12,12 @@ import (
 )
 
 const (
-	BatchProcessorName           = "batch"
-	MemoryLimiterProcessorName   = "memory_limiter"
-	NodeNameProcessorName        = "resource/node-name"
-	ClusterCollectorExporterName = "otlp/out-cluster-collector"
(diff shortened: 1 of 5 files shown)


autoscaler/controllers/nodecollector/collectorconfig/common.go
```diff
@@ -34,6 +35,19 @@ func commonProcessors(nodeCG *odigosv1.CollectorsGroup) config.GenericMap {
 	memoryLimiterConfig := commonconf.GetMemoryLimiterConfig(nodeCG.Spec.ResourcesSettings)
 	allProcessors[MemoryLimiterProcessorName] = memoryLimiterConfig
 
+	var detectors []string
+	// This is a workaround to avoid adding the gcp detector if not running on a gke environment
+	// once https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/1026 is resolved, we can always put the gcp detector
+	if runningOnGKE {
+		detectors = []string{"gcp"}
```
blumamir, line 42: note:  small behavior change. if running on "gke" then populate just the "gcp" detector and avoid adding the "ec2" and "azure". make sense to me but it's a change from current implementation

PR #3556, opened 2025-09-23
chore: do spanmetrics in node collector instead of cluster collector

blumamir:
Fixes: CORE-271

This is so that span metrics calculations are scoped only to the pods in the same node. today we calculate time-series for the same metric and pod in each instance of cluster collector.

This will help:
- making sure that each metric is generated only once
- over all do less work and use less memory since less metrics computations in pipeline

The downside is that work is now done in the node collector instead of the cluster collector (more sensitive to memory and cpu resource and cannot scale).

### DIFF
--- a/autoscaler/controllers/nodecollector/configmap.go
+++ b/autoscaler/controllers/nodecollector/configmap.go
@@ -306,7 +306,7 @@ func calculateConfigMapData(
 
 	// traces
 	tracesEnabledInClusterCollector := slices.Contains(clusterCollectorSignals, odigoscommon.TracesObservabilitySignal)
-	if len(additionalTraceExporters) > 0 {
+	if len(additionalTraceExporters) > 0 || tracesEnabledInClusterCollector {
 		tracesConfig := collectorconfig.TracesConfig(nodeCG, odigosNamespace, tracesProcessors, additionalTraceExporters, tracesEnabledInClusterCollector)
 		activeConfigDomains = append(activeConfigDomains, tracesConfig)
 	}


autoscaler/controllers/nodecollector/collectorconfig/metrics.go
```diff
@@ -99,5 +171,5 @@ func MetricsConfig(nodeCG *odigosv1.CollectorsGroup, odigosNamespace string, man
 				},
 			},
 		},
-	}
+	}, connectorNames
```
RonFed, line 174: nit: why do we need to return the connectors an another parameter if it is part of the config (the first returned value)?

autoscaler/controllers/nodecollector/configmap.go
```diff
@@ -158,20 +158,20 @@ func noopConfigMap() (string, error) {
...
 	var cmData string
 
-	if cg == nil || len(signals) == 0 {
+	if cg == nil || len(clusterCollectorSignals) == 0 {
 		// if collectors group is not created yet, or there are no signals to collect, return a no-op configmap
 		// this can happen if no sources are instrumented yet or no destinations are added.
 		cmData, err = noopConfigMap()
 	} else {
-		cmData, err = calculateConfigMapData(cg, sources, signals, processors, commonconf.ControllerConfig.OnGKE)
+		cmData, err = calculateConfigMapData(cg, sources, clusterCollectorSignals, processors, commonconf.ControllerConfig.OnGKE)
```
RonFed, line 174: I know it was done in pervious PR and yet: why onGKE is passed here as parameter? did we always had it that way? can't the internal function just use `commonconf.ControllerConfig.OnGKE`?

autoscaler/controllers/nodecollector/configmap.go
```diff
@@ -294,18 +294,21 @@ func calculateConfigMapData(
...
-	metricsEnabled := slices.Contains(signals, odigoscommon.MetricsObservabilitySignal)
+	metricsEnabled := slices.Contains(clusterCollectorSignals, odigoscommon.MetricsObservabilitySignal)
 	metricsConfigSettings := nodeCG.Spec.Metrics
+	var additionalTraceExporters []string
 	if metricsEnabled && metricsConfigSettings != nil {
-		metricsConfig := collectorconfig.MetricsConfig(nodeCG, odigosNamespace, metricsProcessors, metricsConfigSettings)
+		metricsConfig, metricsAdditionalTraceExporters := collectorconfig.MetricsConfig(nodeCG, odigosNamespace, metricsProcessors, metricsConfigSettings)
 		activeConfigDomains = append(activeConfigDomains, metricsConfig)
+		additionalTraceExporters = append(additionalTraceExporters, metricsAdditionalTraceExporters...)
+	}
```
RonFed, line 305: can you please add comment here as to why traces exporteds are under metrics config - just to make it easier to understand what's going on

autoscaler/controllers/nodecollector/sync.go
```diff
@@ -78,13 +79,21 @@ func (b *nodeCollectorBaseReconciler) syncDataCollection(ctx context.Context, so
...
 
-	err = b.SyncConfigMap(ctx, sources, signals, processors, dataCollection)
+	err = b.SyncConfigMap(ctx, sources, clusterCollectorSignals, processors, dataCollection)
 	if err != nil {
 		logger.Error(err, "Failed to sync config map")
 		return err
 	}
 
-	dm.RunSyncDaemonSetWithDelayAndSkipNewCalls(time.Duration(env.GetSyncDaemonSetDelay())*time.Second, syncDaemonsetRetry, signals, dataCollection, ctx, b.Client)
+	enabledSignals := clusterCollectorSignals
```
RonFed, line 88: can you please add a comment here saying what's the difference between `clusterCollectorSignals` and `enabledSignals` ?

autoscaler/controllers/nodecollector/collectorconfig/metrics.go
blumamir: It's semantic - these are actually "additionalTraceExporters" which just happen to be the same as the connector names

autoscaler/controllers/nodecollector/configmap.go
blumamir: It can, but that requires more effort in tests, and has an hidden side effect (reading global value).

I wantted to make this function pure (depend only on it's parameters) so it's more obvious what affects it's output.

autoscaler/controllers/nodecollector/configmap.go
blumamir: sure adding it in next PR

PR #3566, opened 2025-09-25
fix: handle sources deletion if exists multiple sources for workload

AvihuHenya:
## Description
This PR introduces the first-phase bug fix for a race condition in the source validation webhook cache during source creation. Due to this race condition, multiple sources could be created for the same workload, which then could not be deleted.

When attempting to delete such sources, they became stuck in the "Terminating" state.
This issue occurred because the uniqueness validation in the GetSources function was designed to only allow a single source per workload.

To fix this, I updated the code so that the uniqueness constraints in the source validation webhook do not run on deletion events, while still maintaining deterministic behavior in cases where multiple sources exist.

I updated the code so that the uniqueness constraints in the sources validation webhook do not run on deletion events. This ensures that deletion works correctly and maintains deterministic behavior even in cases where multiple sources exist.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/api/odigos/v1alpha1/source_types.go
+++ b/api/odigos/v1alpha1/source_types.go
@@ -19,6 +19,7 @@ package v1alpha1
 import (
 	"context"
 	"errors"
+	"sort"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -129,11 +130,22 @@ func GetSources(ctx context.Context, kubeClient client.Client, pw k8sconsts.PodW
 		if err != nil {
 			return nil, err
 		}
-		if len(sourceList.Items) > 1 {
-			return nil, ErrorTooManySources
+
+		// Filter out sources that are being deleted (have deletionTimestamp)
+		var activeSources []Source
+		for _, source := range sourceList.Items {
+			if source.DeletionTimestamp == nil || source.DeletionTimestamp.IsZero() {
+				activeSources = append(activeSources, source)
+			}
 		}
-		if len(sourceList.Items) == 1 {
-			workloadSources.Workload = &sourceList.Items[0]
+
+		sort.Slice(activeSources, func(i, j int) bool {
+			return activeSources[i].CreationTimestamp.Before(&activeSources[j].CreationTimestamp)
+		})
+
+		// Allow multiple active sources - just take the first one
+		if len(activeSources) >= 1 {
+			workloadSources.Workload = &activeSources[0]
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


blumamir (approved):

instrumentor/controllers/sources_webhooks.go
```diff
@@ -228,14 +228,17 @@ func (s *SourcesValidator) validateSourceFields(ctx context.Context, source *v1a
...
+	// Skip uniqueness validation if source is being deleted
+	if source.DeletionTimestamp == nil || source.DeletionTimestamp.IsZero() {
+		err := s.validateSourceUniqueness(ctx, source)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(
+				field.NewPath("spec").Child("workload"),
+				source.Spec.Workload,
+				err.Error(),
+			))
+		}
```
damemi, line 241: we could pull this check out of `validateSourceFields()` and just call `validateSourceUniqueness()` in the webhook's `ValidateCreate()` function

api/odigos/v1alpha1/source_types.go
```diff
@@ -129,12 +130,28 @@ func GetSources(ctx context.Context, kubeClient client.Client, pw k8sconsts.PodW
...
+		if activeCount > 1 {
+			sort.Slice(activeSources, func(i, j int) bool {
+				return activeSources[i].CreationTimestamp.Before(&activeSources[j].CreationTimestamp)
+			})
 		}
-		if len(sourceList.Items) == 1 {
-			workloadSources.Workload = &sourceList.Items[0]
+
+		// Only assign if there are active sources
+		if activeCount >= 1 {
```
damemi, line 154: isn't this still a bug? This could lead to some behavior that's very tough to debug if some other issue pops up that allows multiple sources for a workload

AvihuHenya: yes, but it’s still better than the current state where we are not instrumenting the source when there is more than one, and we aren't viewing the source in the ui

instrumentor/controllers/sources_webhooks.go
AvihuHenya: don’t you think it’s better to also cover the edge cases that may arise from ValidateUpdate(), and to have the same behavior for `create` and `update` validations?

api/odigos/v1alpha1/source_types.go
damemi: my concern is if this will mask the source of a bug in the future if someone doesn't know there are two sources (such as, "I am trying to disable my source but nothing is changing" since we're always returning `sources[0]`)

Perhaps we should still return an error here that can be propagated to the UI/controller saying "multiple sources detected, this is bad, but here's the first source we found"

I'm fine using "oldest source" as the source of truth, I think that could actually even be a good way for us to even *allow* multiple sources that stack (say app team source config + sre source config), but that's just an idea

instrumentor/controllers/sources_webhooks.go
damemi: ah you're right, misread sorry

api/odigos/v1alpha1/source_types.go
AvihuHenya: I agree with you about the future plans and the need to visualize an error somewhere, but for now to recover from this bug — do you think we need to add anything else here?

damemi: Correct me if I misunderstood, but I think now that this PR updates the checks to filter out terminating sources and skip the uniqueness webhook check for terminating sources, the original issue should be fixed right? In that case, if we hit >1 source at this new point in the code, there is some other new bug. Could we still return an error in that case?

PR #3570, opened 2025-09-25
feat: config options to control span metrics resource attributes to exclude

blumamir:
Fixes: CORE-271

this PR adds a pipeline in node collector which mutate spans before they are sent into the spanmetrics connector.

it will add a `resource` processor that will delete some resource attributes, effectively removing them as dimensions from the auto generate metrics.

also removed 2 `telemetery.sdk` attributes which looks useless, and removed few `process.*` attributes by default, with an opt-in config option to enable them

### DIFF
--- a/autoscaler/controllers/nodecollector/collectorconfig/spanmetrics.go
+++ b/autoscaler/controllers/nodecollector/collectorconfig/spanmetrics.go
@@ -1,0 +1,159 @@
+package collectorconfig
+
+import (
+	"slices"
+
+	"github.com/odigos-io/odigos/common"
+	"github.com/odigos-io/odigos/common/config"
+	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
+)
+
+var (
+	spanMetricsConnectorName                         = "spanmetrics"
+	spanMetricsTracesInConnectorName                 = "forward/trace/spanmetrics"
+	spanMetricsPipelineName                          = "traces/spanmetrics"
+	spanMetricsResourceRemoveDimensionsProcessorName = "resource/spanmetrics/remove-dimensions"
+)
+
+func getSpanMetricsConnectorConfig(spanMetricsConfig common.MetricsSourceSpanMetricsConfiguration) config.GenericMap {
+	histogramConfig := config.GenericMap{}
+
+	if spanMetricsConfig.HistogramDisabled {
+		histogramConfig["disabled"] = true
+	} else {
+		if spanMetricsConfig.ExplicitHistogramBuckets != nil {
+			histogramConfig["explicit"] = config.GenericMap{
+				"buckets": spanMetricsConfig.ExplicitHistogramBuckets,
+			}
+		}
+	}
+
+	dimensionsAttributeNames := []string{
+		"http.method",
+		"http.request.method",
+		"http.status_code",
+		"http.response.status_code",
+		"http.route",
+	}
+	if spanMetricsConfig.AdditionalDimensions != nil {
+		dimensionsAttributeNames = append(dimensionsAttributeNames, spanMetricsConfig.AdditionalDimensions...)
+	}
+
+	dimensions := make([]config.GenericMap, len(dimensionsAttributeNames))
+	for i, dimensionAttributeName := range dimensionsAttributeNames {
+		dimensions[i] = config.GenericMap{
+			"name": dimensionAttributeName,
+		}
+	}
+
+	return config.GenericMap{
+		"histogram": histogramConfig,
+		// Taking into account changes in the semantic conventions, to support a range of instrumentation libraries
+		"dimensions": dimensions,
+		"exemplars": config.GenericMap{
+			"enabled": true,
+		},
+		"exclude_dimensions":              []string{"status.code"},
+		"dimensions_cache_size":           1000,
+		"aggregation_temporality":         "AGGREGATION_TEMPORALITY_CUMULATIVE",
+		"metrics_flush_interval":          spanMetricsConfig.Interval,
+		"metrics_expiration":              spanMetricsConfig.MetricsExpiration,
+		"resource_metrics_key_attributes": []string{"service.name", "telemetry.sdk.language", "telemetry.sdk.name"},
+		"events": config.GenericMap{
+			"enabled": true,
+			"dimensions": []config.GenericMap{
+				{
+					"name": "exception.type",
+				},
+				{
+					"name": "exception.message",
+				},
+			},
+		},
+	}
+}
+
+func getSpanMetricsConnectors(spanMetricsConfig common.MetricsSourceSpanMetricsConfiguration) config.GenericMap {
+	return config.GenericMap{
+		spanMetricsConnectorName:         getSpanMetricsConnectorConfig(spanMetricsConfig),
+		spanMetricsTracesInConnectorName: &config.GenericMap{},
+	}
+}
+
+func getSpanMetricsPipelineProcessors(spanMetricsConfig common.MetricsSourceSpanMetricsConfiguration) (config.GenericMap, []string) {
+
+	processors := config.GenericMap{}
+	processorNames := []string{}
+	resourceAttrToExclude := []string{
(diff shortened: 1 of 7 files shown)


common/odigos_config.go
```diff
@@ -189,6 +189,21 @@ type MetricsSourceSpanMetricsConfiguration struct {
...
+	// to have more granularity, you can set this to true.
+	// this will create more dimensions for metrics, including process info,
+	// but with cost of more metrics series and costs.
+	IncludedProcessInDimensions *bool `json:"includedProcessInDimensions,omitempty"`
+
+	// exclude resource attributes from being added to span metrics.
+	// for example - if you don't care about the process granularity,
+	// and prefer the metrics to be aggregated for all processes in a pod container,
+	// you can list all "process.*" attributes here to exclude them from being added to span metrics.
+	// any other resource attribute can be set, either for senetation or to reduce dimenssions for generate metrics.
```
damemi:
```suggestion
	// any other resource attribute can be set, either for sanitation or to reduce dimenssions for generate metrics.
```

common/odigos_config.go
```diff
@@ -189,6 +189,21 @@ type MetricsSourceSpanMetricsConfiguration struct {
 	// 		["2ms", "4ms", "6ms", "8ms", "10ms", "50ms", "100ms", "200ms", "400ms", "800ms", "1s", "1400ms", "2s", "5s", "10s", "15s"]
 	// notice that more granular buckets are recommended for better precision but costs more since more metric series are produced.
 	ExplicitHistogramBuckets []string `json:"histogramBuckets,omitempty"`
+
+	// by default, odigos will aggregate base on pod container.
```
damemi:
```suggestion
	// by default, Odigos will aggregate base on pod container.
```

common/odigos_config.go
```diff
@@ -189,6 +189,21 @@ type MetricsSourceSpanMetricsConfiguration struct {
...
+	// if you have multiple processes in the same container,
+	// you will get each metrics aggregatint all processes measurements.
+	// to have more granularity, you can set this to true.
+	// this will create more dimensions for metrics, including process info,
+	// but with cost of more metrics series and costs.
```
damemi: This option could be a little clearer if it's called something like "IncludeProcessLabels", I haven't heard metric labels referred to very much as dimensions, usually I hear dimensionality used as just the description of how many labels there are

Then this comment could be reworded to be a little clearer on what it's doing, maybe something like

```
// By default, Odigos does not include process labels meaning
// metrics will be aggregated by container as the lowest level.
// This means that multiple processes running in the same container
// will be aggregated into the same time series.
// For more granular metrics, set this option to true.
// This will include process-specific labels on metrics,
// which will cause more unique time series to be created.
```

helm/odigos/values.yaml
```diff
@@ -539,6 +539,15 @@ metricsSources:
     ## format is duration string (1us, 2ms, 3s, 4m, 5h, 6d, etc)
     # explicitHistogramBuckets: ['2ms', '4ms', '6ms', '8ms', '10ms', '50ms', '100ms', '200ms', '400ms', '800ms', '1s', '1400ms', '2s', '5s', '10s', '15s']
 
+    ## set if process level dimenssions will be included in span metrics
```
damemi:
```suggestion
    ## set if process level dimensions will be included in span metrics
```

autoscaler/controllers/nodecollector/configmap.go
```diff
@@ -218,13 +218,18 @@ func calculateConfigMapData(
...
-		activeConfigDomains = append(activeConfigDomains, metricsConfig)
 
-		// due to spanmetrics connector, adding a metrics pipeline can result in needing to also collect
-		// traces which are fed into the collector and generate metrics from spans.
-		// additionalTraceExporters will be populated if traces need to be exportered to any additional place.
-		additionalTraceExporters = append(additionalTraceExporters, metricsAdditionalTraceExporters...)
+		// span metrics
+		additionalMetricsRecivers := []string{}
+		if metricsConfigSettings.SpanMetrics != nil {
+			spanMetricsConfig, additionalSpanMetricsTraceExporters, additionalSpanMetricsMetricsRecivers := collectorconfig.GetSpanMetricsConfig(*metricsConfigSettings.SpanMetrics)
```
damemi:
```suggestion
			spanMetricsConfig, additionalSpanMetricsTraceExporters, additionalSpanMetricsMetricsReceivers := collectorconfig.GetSpanMetricsConfig(*metricsConfigSettings.SpanMetrics)
```

autoscaler/controllers/nodecollector/configmap.go
```diff
@@ -218,13 +218,18 @@ func calculateConfigMapData(
...
+			additionalTraceExporters = append(additionalTraceExporters, additionalSpanMetricsTraceExporters...)
+			additionalMetricsRecivers = append(additionalMetricsRecivers, additionalSpanMetricsMetricsRecivers...)
+			activeConfigDomains = append(activeConfigDomains, spanMetricsConfig)
+		}
+
+		metricsConfig := collectorconfig.MetricsConfig(nodeCG, odigosNamespace, additionalMetricsProcessors, additionalMetricsRecivers, metricsConfigSettings)
```
damemi, line 231:
```suggestion
		metricsConfig := collectorconfig.MetricsConfig(nodeCG, odigosNamespace, additionalMetricsProcessors, additionalMetricsReceivers, metricsConfigSettings)
```

common/odigos_config.go
damemi: is IncludedProcessInDimensions just the inverse of this setting? What if that is set to true but some process fields are set here, does one override the other?

autoscaler/controllers/nodecollector/collectorconfig/spanmetrics.go
```diff
@@ -0,0 +1,159 @@
...
+func getSpanMetricsPipelineProcessors(spanMetricsConfig common.MetricsSourceSpanMetricsConfiguration) (config.GenericMap, []string) {
+
+	processors := config.GenericMap{}
+	processorNames := []string{}
+	resourceAttrToExclude := []string{
+		// always delete these two attributes, as they are just noise in span metrics
+		// TODO: consider making it an opt-in configuration option one day
+		string(semconv.TelemetrySDKNameKey),
+		string(semconv.TelemetrySDKVersionKey),
+	}
```
damemi, line 92: I'm not sure about excluding these by default, I think sdk.name is kind of nominally important in otel, though I don't know exactly how that relates to span metrics. if these are causing issues I'm fine with it but just thinking users can exclude them anyway if it's just annoying

damemi (approved):
overall lgtm just some nits on typos and wording mainly (sorry I know that's annoying). thanks for the work @blumamir 👍

autoscaler/controllers/nodecollector/collectorconfig/spanmetrics.go
blumamir: It's just saying 

"telemetry.sdk.name": "opentelemetry" and
"telemetry.sdk.version": "1.2.3" (version of opentelemetry that we use).

we still have the `telemetry.distro.name` and `telemetry.distro.version` which are odigos versions, but I guess most users would not want these either.

Users like to have as metrics dimensions just what they actually use and not tons of data "just in case" and don't see how anyone would want the sdk name and version which are meaningless most of the time. but we can have it as opt-in option later on.

common/odigos_config.go
blumamir: Thanks - changed to your suggested text.

Regarding the property name - 
I think 'labels' are the way that promethues calls it, but metrics in odigos are otel metrics and these things are not called labels, but resource attributes and metric attributes?
Only when they are exported to promethus they are transformed into labels.

We should probably make it more consistent in our internal naming and how we refer to these in our api and docs. for now let's merge like this and consider an effort to align naming to convention later on?

damemi: Attributes or Labels is fine I just wouldn't call it Dimensions, the option itself is phrased a bit awkwardly as "IncludedProcessInDimensions"

PR #3577, opened 2025-09-28
Allow exposing central UI on any interface

TheErikNaz:
## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

emergency-ticket id: EMR-290

### DIFF
--- a/cli/cmd/pro.go
+++ b/cli/cmd/pro.go
@@ -528,15 +528,6 @@ var portForwardCentralCmd = &cobra.Command{
 		}
 		startPortForward(&wg, ctx, uiPod, client, k8sconsts.CentralUIPort, "UI", localAddress)
 
-		keycloakPod, err := findPodWithAppLabel(ctx, client, proNamespaceFlag, k8sconsts.KeycloakAppName)
-		if err != nil {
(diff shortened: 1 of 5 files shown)


cli/cmd/pro.go
```diff
@@ -492,10 +493,10 @@ var portForwardCentralCmd = &cobra.Command{
 			wg.Wait()
 			os.Exit(1)
 		}
-		startPortForward(&wg, ctx, keycloakPod, client, fmt.Sprintf("%d", k8sconsts.KeycloakPort), "Keycloak")
+		startPortForward(&wg, ctx, keycloakPod, client, fmt.Sprintf("%d", k8sconsts.KeycloakPort), "Keycloak", localAddress)
```
TheErikNaz: @alonbraymok Not sure if we should expose the keyclock port outwards or not.

TheErikNaz: Assuming that only the backend interacts with it, we can leave it only as ClusterIP (Same as with redis)

alonkeyval (approved):

alonkeyval (approved):

PR #3595, opened 2025-10-01
CORE-190: Remove memory-intensive k8sattributes processor config

damemi:
## Description

This addresses high memory usage of the k8sattributes processor by removing unnecessary attributes collection and shifting those attributes to the instrumentor's pod webhook.

### Background

The [k8sattributes](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/1d59fdeaec6db001bfa1f26681e6b33357d38d38/processor/k8sattributesprocessor) processor inserts Kubernetes resource data as resource attributes into spans, metrics, and logs. It works by caching all of the resources in a cluster (pods, deployments, namespaces, replicasets, etc). Users (us) define pod "associations" (ways to identify which pod a signal came from) and the processor injects the data from that pod's deployment/replicaset/namespace/etc.

For example, we currently define an association with pod name + pod namespace, and ask the processor to insert that pod's replicaset name and uid. When the processor sees a span with the right pod name/namespace, it looks in its cache for that pod and grabs the replicaset associated with it to get the name and uid.

More importantly, this processor allows parsing custom *kubernetes labels* or *kubernetes attributes* into *otlp* attributes. This means that you can set a label `my-foo-label` on a Node, and when the processor sees a span from a pod on that node it can insert your custom attribute into the span.

### Memory issues with this processor

In large clusters, the initial informer list (and subsequent event updates) can be memory intensive with lots of irrelevant objects and events. The main issue is with objects that are not node-scoped, such as Deployments, which will be listed for the entire cluster even if the processor is run in a [daemonset "agent" mode](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/1d59fdeaec6db001bfa1f26681e6b33357d38d38/processor/k8sattributesprocessor#as-an-agent). Since we only use this processor on the node collectors, which are in the Odiglet pod and not recommended for users to export to, we can scope this down to just the relevant apps that are instrumented by Odigos.

### Our use case

We use this processor to add the following to spans:

* Custom labels/Attributes from Pod or Namespace
* Workload name/UID
* Ownerreference (replicaset/statefulset/job) name/uid, specifically replicaset
* Container/image attributes

Since we are only parsing custom labels/attributes from Pods (which are node scoped) and Namespaces (which are much less numerous than other resources), we don't need the processor to run informers for other objects like Deployments.

Additionally, we already set [Workload name](https://github.com/odigos-io/odigos/blob/39260eb325869cccbaaf4130e7119442f0cca749/instrumentor/controllers/agentenabled/podswebhook/otelresource.go#L37-L38) in the pod webhook via `OTEL_RESOURCE_ATTRIBUTES` (even though this option is actually [disabled](https://github.com/odigos-io/odigos/blob/39260eb325869cccbaaf4130e7119442f0cca749/autoscaler/controllers/actions/k8sattributesresolver_controller.go#L243) in the action). So it's unnecessary for us to set this in the processor.

### What this PR does

This PR updates the pod webhook to also inject the ownerreference attributes and removes these config settings from the k8sattributes processor. Since the processor [does not initialize and informer if these attributes are not enabled](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1d59fdeaec6db001bfa1f26681e6b33357d38d38/processor/k8sattributesprocessor/internal/kube/client.go#L246-L284), this should show an immediate performance improvement.

For users who don't want to export these attributes (ie, the equivalent of disabling "workload attributes"), they can easily be filtered with a simple processor.

This also updates the Odiglet to parse this env var for ebpf distros like Go. This is a nice bonus because it adds support for the standard OTEL_RESOURCE_ATTRIBUTES env var in these distros.

### Alternatives

We considered 2 alternatives for this:

1. Rewrite the processor from scratch to only do what we need -- for pod associations, look up the relevant PodWorkload on a span and inject the workload name based on that.
2. Fork the upstream processor and replace all informers with a single InstrumentationConfig informer

For option 1, this was the simplest solution but it is ultimately redundant by what can be achieved with existing config. We are already looking up the relevant podworkload in the pod webhook, so we would now be doing that twice in the same pipeline.

For option 2, it was overkill for what we actually need right now. This is an option for the future if we ever want to add support for Deployment/Node/etc custom labels+attributes without the performance overhead of a cluster-wide resource informer, since the InstrumentationConfig informer will scope the processor to only Odigos apps.

## How Has This Been Tested?

Collector memory usage with deployment attributes (1000 deployments/30 nodes):

```
otelcol_process_runtime_heap_alloc_bytes{otel_scope_name="go.opentelemetry.io/collector/service",otel_scope_schema_url="",otel_scope_version=""} 1.9817e+07
# HELP otelcol_process_runtime_total_alloc_bytes_total Cumulative bytes allocated for heap objects (see 'go doc runtime.MemStats.TotalAlloc') [alpha]
# TYPE otelcol_process_runtime_total_alloc_bytes_total counter
otelcol_process_runtime_total_alloc_bytes_total{otel_scope_name="go.opentelemetry.io/collector/service",otel_scope_schema_url="",otel_scope_version=""} 1.88940312e+08
# HELP otelcol_process_runtime_total_sys_memory_bytes Total bytes of memory obtained from the OS (see 'go doc runtime.MemStats.Sys') [alpha]
# TYPE otelcol_process_runtime_total_sys_memory_bytes gauge
otelcol_process_runtime_total_sys_memory_bytes{otel_scope_name="go.opentelemetry.io/collector/service",otel_scope_schema_url="",otel_scope_version=""} 6.1166856e+07
# HELP otelcol_process_uptime_seconds_total Uptime of the process [alpha]
# TYPE otelcol_process_uptime_seconds_total counter
otelcol_process_uptime_seconds_total{otel_scope_name="go.opentelemetry.io/collector/service",otel_scope_schema_url="",otel_scope_version=""} 259.928988012
```

* otelcol_process_runtime_heap_alloc_bytes = 1.9817e+07 = 19MB
* otelcol_process_runtime_total_alloc_bytes_total = 1.88940312e+08 = 188MB
* otelcol_process_runtime_total_sys_memory_bytes 6.1166856e+07 = 61MB

Without:

```
# HELP otelcol_process_runtime_heap_alloc_bytes Bytes of allocated heap objects (see 'go doc runtime.MemStats.HeapAlloc') [alpha]
# TYPE otelcol_process_runtime_heap_alloc_bytes gauge
otelcol_process_runtime_heap_alloc_bytes{otel_scope_name="go.opentelemetry.io/collector/service",otel_scope_schema_url="",otel_scope_version=""} 1.9855952e+07
# HELP otelcol_process_runtime_total_alloc_bytes_total Cumulative bytes allocated for heap objects (see 'go doc runtime.MemStats.TotalAlloc') [alpha]
# TYPE otelcol_process_runtime_total_alloc_bytes_total counter
otelcol_process_runtime_total_alloc_bytes_total{otel_scope_name="go.opentelemetry.io/collector/service",otel_scope_schema_url="",otel_scope_version=""} 5.204148e+07
# HELP otelcol_process_runtime_total_sys_memory_bytes Total bytes of memory obtained from the OS (see 'go doc runtime.MemStats.Sys') [alpha]
# TYPE otelcol_process_runtime_total_sys_memory_bytes gauge
otelcol_process_runtime_total_sys_memory_bytes{otel_scope_name="go.opentelemetry.io/collector/service",otel_scope_schema_url="",otel_scope_version=""} 4.0195336e+07
# HELP otelcol_process_uptime_seconds_total Uptime of the process [alpha]
# TYPE otelcol_process_uptime_seconds_total counter
otelcol_process_uptime_seconds_total{otel_scope_name="go.opentelemetry.io/collector/service",otel_scope_schema_url="",otel_scope_version=""} 225.246217102
```

* otelcol_process_runtime_heap_alloc_bytes = 1.9855952e+07 = 19MB
* otelcol_process_runtime_total_alloc_bytes_total = 5.204148e+07 = 50MB
* otelcol_process_runtime_total_sys_memory_bytes 4.0195336e+07 = 40MB

Total alloc bytes is heap bytes allocated since the start of the process. Notably this dropped from 188MB -> 50MB in this test (73% reduction). Obviously this number continues to climb through the life of the process, so both were taken at ~250s of uptime.

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

### DIFF
--- a/frontend/graph/generated.go
+++ b/frontend/graph/generated.go
@@ -387,10 +387,12 @@ type ComplexityRoot struct {
 	K8sAnnotationAttribute struct {
 		AnnotationKey func(childComplexity int) int
 		AttributeKey  func(childComplexity int) int
+		From          func(childComplexity int) int
 	}
 
 	K8sLabelAttribute struct {
 		AttributeKey func(childComplexity int) int
+		From         func(childComplexity int) int
 		LabelKey     func(childComplexity int) int
 	}
 
@@ -2358,12 +2360,26 @@ func (e *executableSchema) Complexity(typeName, field string, childComplexity in
 
 		return e.complexity.K8sAnnotationAttribute.AttributeKey(childComplexity), true
 
+	case "K8sAnnotationAttribute.from":
+		if e.complexity.K8sAnnotationAttribute.From == nil {
+			break
+		}
+
+		return e.complexity.K8sAnnotationAttribute.From(childComplexity), true
+
 	case "K8sLabelAttribute.attributeKey":
 		if e.complexity.K8sLabelAttribute.AttributeKey == nil {
 			break
 		}
 
 		return e.complexity.K8sLabelAttribute.AttributeKey(childComplexity), true
+
+	case "K8sLabelAttribute.from":
+		if e.complexity.K8sLabelAttribute.From == nil {
+			break
+		}
+
+		return e.complexity.K8sLabelAttribute.From(childComplexity), true
 
 	case "K8sLabelAttribute.labelKey":
 		if e.complexity.K8sLabelAttribute.LabelKey == nil {
@@ -6060,6 +6076,8 @@ func (ec *executionContext) fieldContext_ActionFields_labelsAttributes(_ context
 				return ec.fieldContext_K8sLabelAttribute_labelKey(ctx, field)
 			case "attributeKey":
 				return ec.fieldContext_K8sLabelAttribute_attributeKey(ctx, field)
+			case "from":
+				return ec.fieldContext_K8sLabelAttribute_from(ctx, field)
 			}
 			return nil, fmt.Errorf("no field named %q was found under type K8sLabelAttribute", field.Name)
 		},
@@ -6107,6 +6125,8 @@ func (ec *executionContext) fieldContext_ActionFields_annotationsAttributes(_ co
 				return ec.fieldContext_K8sAnnotationAttribute_annotationKey(ctx, field)
 			case "attributeKey":
 				return ec.fieldContext_K8sAnnotationAttribute_attributeKey(ctx, field)
+			case "from":
+				return ec.fieldContext_K8sAnnotationAttribute_from(ctx, field)
 			}
 			return nil, fmt.Errorf("no field named %q was found under type K8sAnnotationAttribute", field.Name)
 		},
@@ -15058,6 +15078,47 @@ func (ec *executionContext) fieldContext_K8sAnnotationAttribute_attributeKey(_ c
 	return fc, nil
 }
 
+func (ec *executionContext) _K8sAnnotationAttribute_from(ctx context.Context, field graphql.CollectedField, obj *model.K8sAnnotationAttribute) (ret graphql.Marshaler) {
+	fc, err := ec.fieldContext_K8sAnnotationAttribute_from(ctx, field)
+	if err != nil {
+		return graphql.Null
+	}
+	ctx = graphql.WithFieldContext(ctx, fc)
+	defer func() {
+		if r := recover(); r != nil {
+			ec.Error(ctx, ec.Recover(ctx, r))
+			ret = graphql.Null
+		}
+	}()
+	resTmp, err := ec.ResolverMiddleware(ctx, func(rctx context.Context) (any, error) {
+		ctx = rctx // use context from middleware stack in children
+		return obj.From, nil
+	})
+	if err != nil {
+		ec.Error(ctx, err)
+		return graphql.Null
+	}
+	if resTmp == nil {
+		return graphql.Null
+	}
+	res := resTmp.(*model.K8sAttributesFrom)
+	fc.Result = res
+	return ec.marshalOK8sAttributesFrom2ᚖgithubᚗcomᚋodigosᚑioᚋodigosᚋfrontendᚋgraphᚋmodelᚐK8sAttributesFrom(ctx, field.Selections, res)
+}
+
+func (ec *executionContext) fieldContext_K8sAnnotationAttribute_from(_ context.Context, field graphql.CollectedField) (fc *graphql.FieldContext, err error) {
+	fc = &graphql.FieldContext{
+		Object:     "K8sAnnotationAttribute",
+		Field:      field,
+		IsMethod:   false,
+		IsResolver: false,
+		Child: func(ctx context.Context, field graphql.CollectedField) (*graphql.FieldContext, error) {
+			return nil, errors.New("field of type K8sAttributesFrom does not have child fields")
+		},
+	}
+	return fc, nil
+}
+
 func (ec *executionContext) _K8sLabelAttribute_labelKey(ctx context.Context, field graphql.CollectedField, obj *model.K8sLabelAttribute) (ret graphql.Marshaler) {
 	fc, err := ec.fieldContext_K8sLabelAttribute_labelKey(ctx, field)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


instrumentor/controllers/agentenabled/podswebhook/otelresource.go
```diff
@@ -38,6 +39,74 @@ func getResourceAttributes(podWorkload k8sconsts.PodWorkload, containerName stri
...
 	}
+	resourceAttributes = append(resourceAttributes, getOwnerReferenceAttributes(ownerReferences)...)
+	return resourceAttributes
+}
+
+func getOwnerReferenceAttributes(ownerReferences []metav1.OwnerReference) []resourceAttribute {
+	resourceAttributes := []resourceAttribute{}
+	for _, ownerReference := range ownerReferences {
+		switch ownerReference.Kind {
+		case "Deployment":
```
RonFed: I don't think it is possible for a Deployment to be an owner of a pod - it will always be ReplicaSet in that case, no?

instrumentor/controllers/agentenabled/podswebhook/otelresource.go
```diff
@@ -38,6 +39,74 @@ func getResourceAttributes(podWorkload k8sconsts.PodWorkload, containerName stri
...
+				resourceAttribute{
+					Key:   semconv.K8SDeploymentUIDKey,
+					Value: string(ownerReference.UID),
+				},
+			)
+		case "StatefulSet":
+			resourceAttributes = append(resourceAttributes,
+				resourceAttribute{
+					Key:   semconv.K8SStatefulSetNameKey,
+					Value: ownerReference.Name,
```
RonFed: I think this already done by `getResourceAttributes` In
```
		{
			Key:   workloadKindKey,
			Value: podWorkload.Name,
		},
```

instrumentor/controllers/agentenabled/podswebhook/otelresource.go
```diff
@@ -38,6 +39,74 @@ func getResourceAttributes(podWorkload k8sconsts.PodWorkload, containerName stri
...
+					Key:   semconv.K8SStatefulSetUIDKey,
+					Value: string(ownerReference.UID),
+				},
+			)
+		case "DaemonSet":
+			resourceAttributes = append(resourceAttributes,
+				resourceAttribute{
+					Key:   semconv.K8SDaemonSetNameKey,
+					Value: ownerReference.Name,
+				},
```
RonFed: same as above comment. believe this is already handled

odiglet/pkg/ebpf/settings_getter.go
```diff
@@ -86,6 +119,12 @@ func getResourceAttributes(podWorkload *k8sconsts.PodWorkload, podName string, p
 			attrs = append(attrs, semconv.K8SContainerName(containerName))
 		}
 
+		// Parse OTEL_RESOURCE_ATTRIBUTES environment variable
+		if otelResourceAttrs, ok := envs[k8sconsts.OtelResourceAttributesEnvVar]; ok {
+			parsedAttrs := parseOtelResourceAttributes(otelResourceAttrs)
+			attrs = append(attrs, parsedAttrs...)
+		}
```
RonFed, line 162: I think this code needs to handle possible duplicated keys in the `attrs` slice - to make it more robust - since having duplicate keys can lead to some non deterministic behaviors.
For eaxmple, this function adds `K8SDeploymentName` `K8SPodName`  and maybe more that will be duplicate - so we need to clean this up I think before merging this - and in addition handle duplicate keys

odiglet/pkg/ebpf/settings_getter.go
```diff
@@ -86,6 +137,12 @@ func getResourceAttributes(podWorkload *k8sconsts.PodWorkload, podName string, p
 			attrs = append(attrs, semconv.K8SContainerName(containerName))
 		}
 
+		// Parse OTEL_RESOURCE_ATTRIBUTES environment variable
+		if otelResourceAttrs, ok := envs[k8sconsts.OtelResourceAttributesEnvVar]; ok {
```
RonFed, line 156: I think that we can remove the first part of this function since 
`semconv.K8SNamespaceName`
`semconv.K8SPodName`
`semconv.K8SDeploymentName`
`semconv.K8SStatefulSetName` etc'
Are now all handled in the webhook - so better to keep that as the single source of truth I guess

odiglet/pkg/ebpf/settings_getter.go
```diff
@@ -59,6 +59,57 @@ func (ksg *k8sSettingsGetter) instrumentationSDKConfig(ctx context.Context, kd K
 	return nil, "", fmt.Errorf("no sdk config found for language %s", lang)
 }
 
+// parseOtelResourceAttributes parses the OTEL_RESOURCE_ATTRIBUTES environment variable
+// which is in the format "key1=value1,key2=value2" and returns a slice of attribute.KeyValue
+func parseOtelResourceAttributes(envValue string) []attribute.KeyValue {
```
RonFed: nit: I wonder if we should have some error handling here - in case the format is wrong or something to have a log line or to fallback to adding the resource attributes that we can know here without the env var value? (instead of the `continue` statements)

odiglet/pkg/ebpf/settings_getter.go
```diff
@@ -59,6 +59,57 @@ func (ksg *k8sSettingsGetter) instrumentationSDKConfig(ctx context.Context, kd K
...
+		if key != "" && value != "" {
+			attrs = append(attrs, attribute.String(key, value))
+		}
+	}
+
+	return attrs
+}
+
+// appendUniqueAttributes appends new attributes to the existing slice, skipping any that already exist
+func appendUniqueAttributes(existing []attribute.KeyValue, new []attribute.KeyValue) []attribute.KeyValue {
```
RonFed, line 108: nit: similar to above comment - if we have duplicate keys - which one should we take? should we log this if the values are different - since this is kind an un-defined behavior.

instrumentor/controllers/agentenabled/podswebhook/otelresource.go
```diff
@@ -38,6 +39,41 @@ func getResourceAttributes(podWorkload k8sconsts.PodWorkload, containerName stri
 			Value: podWorkload.Name,
 		},
 	}
+	resourceAttributes = append(resourceAttributes, getOwnerReferenceAttributes(ownerReferences)...)
```
RonFed, line 44: can you please add a comment here about the reasoning for doing that here?

odiglet/pkg/ebpf/settings_getter.go
damemi: I did think it looked redundant that we had them in both places. I guess for languages like Go where it wasn't being set in the webhook it made sense. but at that point, I think it just makes more sense to do what we're doing with this and set them for every language

RonFed: Do you mean it makes sense to remove it from here? or to keep it? This path is indeed common to all eBPF supported distributions but it is redundant after the webhook change in this PR, right?

damemi: Yeah sorry, I'm agreeing with you. I think now after this change it makes sense to remove

damemi: On second thought, with your feedback below https://github.com/odigos-io/odigos/pull/3595#discussion_r2411148496 maybe keeping these as the fallback in case the resource attributes fail for some reason makes sense. I can add a comment to explain that at least -- we're pre-initializing these here. wdyt?

RonFed: Yeah having them here as backup/fallback sounds good to me. I agree worth explaining this in comments

damemi:
I realize now this change will impact user upgrades because any existing pods that *don't* have these environment variables injected will not report the attributes that were removed from the collector. Users will need to roll out any deployments that rely on collecting these attributes, so that the new pods get the environment variable injected by the webhook. 

As a bridge we could provide a standalone processor config yaml that includes the removed attributes, then when the user is ready they can remove that processor to get the performance boost. @RonFed wdyt?

PR #3624, opened 2025-10-14
chore: bump odiglet base image to use go1.25

RonFed:
Update the base image of odiglet to use go1.25.3.
In addition using the `trixie` variant of the image to update to the latest stable version of Debian.
Updating CI to publish the base image also to GCP.
In a follow-up PR I plan to update the odiglet to use this new image - and update all the CI and go.mod to use go1.25.

emergency-ticket id: EMR-326


edeNFed (approved):

damemi (approved):
@RonFed is there a specific reason for the bump in the odiglet, or why not all components in this PR? Just curious and to document why if so

RonFed:
@damemi yes, mainly because I want to test the CI for the base image after merging this - assuming that will work I will continue to bump everything.
Other components don't have a dedicated base image also.

PR #3626, opened 2025-10-14
Add InstrumentationLibraries to manager Load()

damemi:
## Description

This makes it possible for more eBPF instrumentations to return the libraries they have instrumented, see initial implementation for Go with

https://github.com/odigos-io/enterprise-go-instrumentation/pull/1753
https://github.com/odigos-io/ui-kit/pull/421

This can be expanded to show failed libraries and more information about the libraries

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [x] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

emergency-ticket id: EMR-324

### DIFF
--- a/instrumentation/factory.go
+++ b/instrumentation/factory.go
@@ -49,7 +49,7 @@ type Instrumentation interface {
 	// for the instrumentation to be ready to run.
 	// For eBPF, this will load the probes into the kernel
 	// In case of a failure, an error will be returned and all the resources will be cleaned up.
-	Load(ctx context.Context) ([]InstrumentationLibrary, error)
+	Load(ctx context.Context) (Status, error)
 
 	// Run will start reading events from the probes and export them.
 	// It is a blocking call, and will return only when the instrumentation is stopped.
@@ -65,9 +65,19 @@ type Instrumentation interface {
 	ApplyConfig(ctx context.Context, config Config) error
 }
 
-// InstrumentationLibrary is used to identify a specific instrumentation library.
-type InstrumentationLibrary struct {
-	// Name is the name of the instrumentation library.
-	// For example, "net/http", "@opentelemetry/instrumentation-redis".
-	Name string
+// Status is used to identify the status of an instrumentation and its libraries.
+type Status struct {
+	// Status is the status of the instrumentation.
+	Status string
+
+	// Libraries is a map of library names to their status.
+	Libraries map[string]LibraryStatus
+}
+
+type LibraryStatus struct {
+	// Healthy indicates if the library was successfully loaded.
+	Healthy bool
+
+	// Message is the message of the library, if applicable.
+	Message string
 }
--- a/instrumentation/factory.go
+++ b/instrumentation/factory.go
@@ -49,7 +49,7 @@ type Instrumentation interface {
 	// for the instrumentation to be ready to run.
 	// For eBPF, this will load the probes into the kernel
 	// In case of a failure, an error will be returned and all the resources will be cleaned up.
-	Load(ctx context.Context) error
+	Load(ctx context.Context) ([]InstrumentationLibrary, error)
 
 	// Run will start reading events from the probes and export them.
 	// It is a blocking call, and will return only when the instrumentation is stopped.
@@ -63,4 +63,11 @@ type Instrumentation interface {
 
 	// ApplyConfig will send a configuration update to the instrumentation.
 	ApplyConfig(ctx context.Context, config Config) error
+}
+
+// InstrumentationLibrary is used to identify a specific instrumentation library.
+type InstrumentationLibrary struct {
+	// Name is the name of the instrumentation library.
+	// For example, "net/http", "@opentelemetry/instrumentation-redis".
+	Name string
 }
--- a/instrumentation/factory.go
+++ b/instrumentation/factory.go
@@ -50,6 +50,7 @@ type Instrumentation interface {
 	// For eBPF, this will load the probes into the kernel
 	// In case of a failure, an error will be returned and all the resources will be cleaned up.
 	// Returns the overall status of the instrumentation, or an error if the instrumentation failed to load.
+	// Status is only valid if the error is nil.
 	Load(ctx context.Context) (Status, error)
 
 	// Run will start reading events from the probes and export them.
@@ -68,9 +69,6 @@ type Instrumentation interface {
 
 // Status is used to identify the status of an instrumentation and its libraries.
 type Status struct {
-	// Status is the status of the instrumentation.
-	Status string
-
-	// Libraries is a map of library names to their status.
-	Libraries map[string]error
+	// Components is a map of component names (such as instrumentation library names) to their status.
+	Components map[string]error
 }
--- a/instrumentation/factory.go
+++ b/instrumentation/factory.go
@@ -49,6 +49,7 @@ type Instrumentation interface {
 	// for the instrumentation to be ready to run.
 	// For eBPF, this will load the probes into the kernel
 	// In case of a failure, an error will be returned and all the resources will be cleaned up.
+	// Returns the overall status of the instrumentation, or an error if the instrumentation failed to load.
 	Load(ctx context.Context) (Status, error)
 
 	// Run will start reading events from the probes and export them.
@@ -71,13 +72,5 @@ type Status struct {
 	Status string
 
 	// Libraries is a map of library names to their status.
-	Libraries map[string]LibraryStatus
-}
-
-type LibraryStatus struct {
-	// Healthy indicates if the library was successfully loaded.
-	Healthy bool
-
-	// Message is the message of the library, if applicable.
-	Message string
+	Libraries map[string]error
 }
--- a/instrumentation/manager.go
+++ b/instrumentation/manager.go
@@ -374,8 +374,8 @@ func (m *manager[ProcessDetails, ConfigGroup]) tryInstrument(ctx context.Context
 		return initErr
 	}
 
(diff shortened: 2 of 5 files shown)


instrumentation/factory.go
RonFed, line 1: FYI changes here will break the vm-agent and require it to update accordingly
cc @TheErikNaz

instrumentation/factory.go
```diff
@@ -49,7 +49,7 @@ type Instrumentation interface {
 	// for the instrumentation to be ready to run.
 	// For eBPF, this will load the probes into the kernel
 	// In case of a failure, an error will be returned and all the resources will be cleaned up.
-	Load(ctx context.Context) error
+	Load(ctx context.Context) ([]InstrumentationLibrary, error)
```
RonFed: WDYT about having a struct type similar to
```Go
type Status struct {
      ...
}

Load(ctx context.Context) (Status, error)
```

Inside it we can have a per lib status (or error)
this will allow us to make the API more extensible and not break it every time we add status reports options.
We can also make use of this `Status` in the `ApplyConfig` call - since it can also change the status of some of the loaded instrumentations (we can do this in a follow up),
Also, since this is inside the `instrumentation` package I think we can make the names shorter to avoid stater - for example `instrumentation.Library` instead of `instrumentation.InstrumentationLibrary` and `instrumentation.Status` instead of `instrumentation.InstrumentationStatus` etc'.

instrumentation/manager.go
```diff
@@ -381,8 +381,8 @@ func (m *manager[ProcessDetails, ConfigGroup]) tryInstrument(ctx context.Context
 		return initErr
 	}
 
-	loadErr := inst.Load(ctx)
-	reporterErr = m.handler.Reporter.OnLoad(ctx, e.PID, loadErr, pd)
+	libraries, loadErr := inst.Load(ctx)
```
aviadhahami: wouldn't it be clearer for the reader if getting the libs names would come from its own func, like `GetInstrumentationLibraries` ?

instrumentation/factory.go
damemi: good news is I'll be working with @TheErikNaz on the vm agent soon :)

instrumentation/factory.go
```diff
@@ -64,3 +64,20 @@ type Instrumentation interface {
...
+// Status is used to identify the status of an instrumentation and its libraries.
+type Status struct {
+	// Status is the status of the instrumentation.
+	Status string
+
+	// Libraries is a map of library names to their status.
+	Libraries map[string]LibraryStatus
+}
+
+type LibraryStatus struct {
```
RonFed: I guess we could just have an error type instead of this struct. 
I understand that this definition aligns with the instrumentation instance object and more specifically with the conditions structure.
But I think the definitions here should not "favor" k8s. 
I.e using go errors seems like it can do the trick here and keep it generic. 
wdyt?

instrumentation/types.go
```diff
@@ -56,7 +56,7 @@ type Reporter[processDetails ProcessDetails] interface {
 
 	// OnLoad is called after an instrumentation is loaded successfully or failed to load.
 	// The error parameter will be nil if the instrumentation was loaded successfully.
-	OnLoad(ctx context.Context, pid int, err error, pg processDetails) error
+	OnLoad(ctx context.Context, pid int, err error, pg processDetails, status Status) error
```
RonFed, line 59: Think we should also update the comment above

instrumentation/factory.go
damemi: tbh, this just happens to align with the k8s structs by coincidence. This is just meant to report the overall status of the instrumentation, starting out with whether each library was healthy. That's not meant to favor k8s, it's just the most basic status of what we're doing. 

Not sure I understand why use Go errors though, these aren't all meant to be errors (most of them should actually not be errors when Libs load successfully), so I don't think successful libraries should be represented as an error interface or the major benefit of it over a generic struct with a map

instrumentation/factory.go
TheErikNaz: I'm currently finishing up the go 1.25 shenanigans, from there it'll be a pretty easy upgrade

instrumentation/factory.go
RonFed: since this struct currently has a boolean healthy field and a ,message it really sounds equivalent to an error type to me. where err == nil implies an "all good" status.
Do you see advantages to this struct type?
The message in case of a non error is not something specific to the instrumentations I guess - it is similar to the "eBPF probe loaded" message we have today, right? It is a generic success message that is common to all and should not be returned by all the instrumentations is my point there.

damemi: Ah, I think I understand now. so if the overall return is `(Status, error)`, that `Status` is a map of libraries with their own `error`s, where a nil is healthy

tbh I think it's pretty equivalent here to do a custom error impl vs a struct, so I'm not picky but maybe we can chat about it if I'm missing something

instrumentation/factory.go
```diff
@@ -49,7 +49,8 @@ type Instrumentation interface {
 	// for the instrumentation to be ready to run.
 	// For eBPF, this will load the probes into the kernel
 	// In case of a failure, an error will be returned and all the resources will be cleaned up.
-	Load(ctx context.Context) error
+	// Returns the overall status of the instrumentation, or an error if the instrumentation failed to load.
```
RonFed, line 52:
```suggestion
	// Returns the overall status of the instrumentation, or an error if the instrumentation failed to load.
	// Status is only valid if the error is nil.
```

odiglet/pkg/ebpf/reporter.go
```diff
@@ -117,10 +117,25 @@ func (r *k8sReporter) OnExit(ctx context.Context, pid int, e K8sProcessDetails)
...
+			Name:           name,
+			Type:           odigosv1.InstrumentationLibraryTypeInstrumentation,
+			LastStatusTime: metav1.Now(),
+			Healthy:        &libraryHealthy,
+		}
+		if libraryStatus != nil {
+			libStatus.Message = libraryStatus.Error()
+		}
+		components = append(components, libStatus)
+	}
```
RonFed: status.Status is not used right? if so, we can remove it.
Also, just want to make sure we keep the same behavior here where if the everything is doing well, we'll have the success message on the instrumentation instance with the "loaded eBPF.... " message - this success message should be generic (as it is today) and not instrumentation specific.

instrumentation/factory.go
```diff
@@ -64,3 +65,12 @@ type Instrumentation interface {
...
 	ApplyConfig(ctx context.Context, config Config) error
 }
+
+// Status is used to identify the status of an instrumentation and its libraries.
+type Status struct {
+	// Status is the status of the instrumentation.
+	Status string
+
+	// Libraries is a map of library names to their status.
+	Libraries map[string]error
```
RonFed: nit: saw we use the `component` terminology in the instance type
wdyt about using it here as well?

damemi: sgtm keeps it abstracted in the instrumentation package

odiglet/pkg/ebpf/reporter.go
damemi: Looks good to me:

```
apiVersion: v1
items:
- apiVersion: odigos.io/v1alpha1
  kind: InstrumentationInstance
  metadata:
    creationTimestamp: "2025-11-17T21:55:02Z"
    generation: 1
    labels:
      instrumented-app: deployment-gin-middleware-demo
      ownerPodName: gin-middleware-demo-9c9b4497-rgpzc
    name: gin-middleware-demo-9c9b4497-rgpzc-56381
    namespace: default
    ownerReferences:
    - apiVersion: v1
      blockOwnerDeletion: true
      controller: true
      kind: Pod
      name: gin-middleware-demo-9c9b4497-rgpzc
      uid: a2eabcc7-f70f-426f-afda-96d97f68b3c3
    resourceVersion: "86718"
    uid: 3dadcf8a-7e54-444b-aa74-842e976ab6b0
  spec:
    containerName: gin-middleware-demo
  status:
    components:
    - healthy: true
      lastStatusTime: "2025-11-18T15:06:24Z"
      name: go.opentelemetry.io/otel/internal/global/client
      type: instrumentation
    - healthy: true
      lastStatusTime: "2025-11-18T15:06:24Z"
      name: google.golang.org/grpc/client
      type: instrumentation
    - healthy: true
      lastStatusTime: "2025-11-18T15:06:24Z"
      name: github.com/gin-gonic/gin/server
      type: instrumentation
    - healthy: true
      lastStatusTime: "2025-11-18T15:06:24Z"
      name: net/http/server
      type: instrumentation
    - healthy: true
      lastStatusTime: "2025-11-18T15:06:24Z"
      name: github.com/odigos-io/odigos/hooks/go/internal
      type: instrumentation
    - healthy: true
      lastStatusTime: "2025-11-18T15:06:24Z"
      name: google.golang.org/grpc/server
      type: instrumentation
    - healthy: true
      lastStatusTime: "2025-11-18T15:06:24Z"
      name: go.opentelemetry.io/otel/trace/client
      type: instrumentation
    - healthy: true
      lastStatusTime: "2025-11-18T15:06:24Z"
      name: go.opentelemetry.io/auto/client
      type: instrumentation
    healthy: true
    lastStatusTime: "2025-11-18T15:06:24Z"
    message: 'Successfully loaded eBPF probes to pod: gin-middleware-demo-9c9b4497-rgpzc
      container: gin-middleware-demo'
    reason: LoadedSuccessfully
kind: List
metadata:
  resourceVersion: ""
```

RonFed (approved):
🚀

PR #3650, opened 2025-10-20
chore: run release preflight checks on prerelease PRs

damemi:
## Description

Currently we check for release blockers, open dependency syncers, and open Go offsets PRs after the prerelease PR has merged. This means the tag is already created, even if the release will just fail.

This moves the checks to run as a presubmit on the prerelease PR itself, allowing us to catch this before the release actually starts. It's a simple copy and drag right now but can be expanded later to target specific branches and more repos.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

emergency-ticket id: EMR-349

### DIFF
--- a/.github/workflows/create-release-candidate.yml
+++ b/.github/workflows/create-release-candidate.yml
@@ -151,6 +151,7 @@ jobs:
                 --base $RELEASE_BRANCH \
                 --head $PR_BRANCH \
                 --title "$PR_TITLE" \
+                --label "prerelease" \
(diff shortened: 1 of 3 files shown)


damemi:
bump @blumamir @tamirdavid1 if you think this would be a good change to have

RonFed (approved):

blumamir (approved):
👍

PR #3651, opened 2025-10-21
chore: Bump operator to v1.8.0

damemi:
## Description

This updates the operator along with some changes for compatibility since this last operator release:

* Add permission to DeleteCollection for uninstall calls https://github.com/odigos-io/odigos/pull/3651/commits/f230431d3fad4321b2cddf5382dbc106e69d4f28 https://github.com/odigos-io/odigos/pull/3651/commits/24516fef126a2b7df8882ba4af6da5ca63638c9e#diff-bf75d41206005954349fa60826c9cc3bd1a976c9e5c8884de0cb59ee68a13fa8 (used in [uninstall](https://github.com/odigos-io/odigos/blob/cbb90d34081062e12ca65d909eb07a9d1cb0b60e/cli/pkg/kube/client.go#L158)
* Update the DataCollection rolebinding to use the Odiglet role https://github.com/odigos-io/odigos/pull/3651/commits/ba445e4cf368fd7df4e240cebceee67b147e096f (data collection role no longer exists now that it's bundled into the odiglet)
* Update the autoscaler deployment controller to use the node collector *configmap* name instead of *collectorgroup* https://github.com/odigos-io/odigos/pull/3651/commits/174718a8dd2bc7d7703452545be70469d998ee72 (this is effectively a no-op since they're the same but makes it clearer since this controller actually reconciles the configmap)
* Removes the agentenvinjection kubebuilder default https://github.com/odigos-io/odigos/pull/3651/commits/fd0c69c5c938ba891c80e9c65cebc58dceb1f612 (this was updating the operator CR and causing a no-op sync loop -- moved defaulting logic to controller)
* Don't overwrite operator CR values (similar to above reason) and also uses `Patch()` instead of `Update()` to set finalizers, because `Update()` triggers a reconcile that bypasses the `GenerationChangedPredicate` https://github.com/odigos-io/odigos/pull/3651/commits/b3b45f1ee4e77eb30b05787a20d6ce60d377ae79
* Add finalizers permissions to scheduler and autoscaler https://github.com/odigos-io/odigos/pull/3651/commits/176f7bb0a1c381a871dad8ccd4d3c2370feb636c (similar to introduced in https://github.com/odigos-io/odigos/pull/2996, this is an openshift requirement for setting finalizers and is due to the changes around how data collection configmap is created by autoscaler instead of at install)

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

emergency-ticket id: EMR-361

### DIFF
--- a/cli/cmd/resources/autoscaler.go
+++ b/cli/cmd/resources/autoscaler.go
@@ -171,16 +171,23 @@ func NewAutoscalerRoleBinding(ns string) *rbacv1.RoleBinding {
 func NewAutoscalerClusterRole(ownerPermissionEnforcement bool) *rbacv1.ClusterRole {
 	finalizersUpdate := []rbacv1.PolicyRule{}
 	if ownerPermissionEnforcement {
-		finalizersUpdate = append(finalizersUpdate, rbacv1.PolicyRule{
-			// Required for OwnerReferencesPermissionEnforcement (on by default in OpenShift)
-			// When we create a collector COnfigMap, we set the OwnerReference to the collectorsgroups.
-			// Controller-runtime sets BlockDeletion: true. So with this Admission Plugin we need permission to
-			// update finalizers on the collectorsgroup so that they can block deletion.
-			// seehttps://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement
-			APIGroups: []string{"odigos.io"},
-			Resources: []string{"collectorsgroups/finalizers"},
-			Verbs:     []string{"update"},
-		})
+		finalizersUpdate = append(finalizersUpdate,
+			rbacv1.PolicyRule{
+				// Required for OwnerReferencesPermissionEnforcement (on by default in OpenShift)
+				// When we create a collector COnfigMap, we set the OwnerReference to the collectorsgroups.
+				// Controller-runtime sets BlockDeletion: true. So with this Admission Plugin we need permission to
+				// update finalizers on the collectorsgroup so that they can block deletion.
+				// seehttps://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement
+				APIGroups: []string{"odigos.io"},
+				Resources: []string{"collectorsgroups/finalizers"},
+				Verbs:     []string{"update"},
+			},
+			rbacv1.PolicyRule{
+				APIGroups: []string{"apps"},
+				Resources: []string{"deployments/finalizers"},
+				Verbs:     []string{"update"},
+			},
+		)
 	}
 
 	return &rbacv1.ClusterRole{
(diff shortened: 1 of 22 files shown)


edeNFed (approved):

cli/cmd/resources/autoscaler.go
```diff
@@ -171,16 +181,18 @@ func NewAutoscalerRoleBinding(ns string) *rbacv1.RoleBinding {
...
+				// Required for OwnerReferencesPermissionEnforcement (on by default in OpenShift)
+				// When we create a collector COnfigMap, we set the OwnerReference to the collectorsgroups.
+				// Controller-runtime sets BlockDeletion: true. So with this Admission Plugin we need permission to
+				// update finalizers on the collectorsgroup so that they can block deletion.
+				// seehttps://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement
+				APIGroups: []string{"odigos.io"},
+				Resources: []string{"collectorsgroups/finalizers"},
+				Verbs:     []string{"update"},
+			},
+		)
```
RonFed, line 200: Do we need this now that the owner was changed to the autoscaler deployment?

cli/cmd/resources/scheduler.go
```diff
@@ -55,7 +55,72 @@ func NewSchedulerLeaderElectionRoleBinding(ns string) *rbacv1.RoleBinding {
...
+		},
+		{ // for calculating collectors group config based on the active destinations
+			APIGroups: []string{"odigos.io"},
+			Resources: []string{"destinations"},
+			Verbs:     []string{"get", "list", "watch"},
+		},
+	}
+
+	if openshiftEnabled {
+		rules = append(rules, rbacv1.PolicyRule{
```
RonFed, line 122: can you pleas add a comment as to why we need this (for which object)

cli/cmd/resources/autoscaler.go
```diff
@@ -32,7 +32,115 @@ func NewAutoscalerServiceAccount(ns string) *corev1.ServiceAccount {
...
+		},
+		{ // Update conditions of the action after transforming it to a processor
+			APIGroups: []string{"odigos.io"},
+			Resources: []string{"actions/status"},
+			Verbs:     []string{"get", "patch", "update"},
+		},
+	}
+
+	if openshiftEnabled {
+		rules = append(rules, rbacv1.PolicyRule{
```
RonFed, line 142: can you pleas add a comment as to why we need this (for which object)

RonFed (approved):
left a few comments about permissions

cli/cmd/resources/autoscaler.go
damemi: This is still needed for the `odigos-gateway` configmap which has the collectorsgroups as its ownerreference

RonFed (approved):

PR #3696, opened 2025-10-27
feat: list pods from odiglet and gateway

alonkeyval:
this depend on https://github.com/odigos-io/odigos/pull/3674 to be merged first

emergency-ticket id: EMR-395


frontend/services/collectors/pods.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	return "Unknown"
+}
+
+func sumRestarts(statuses []corev1.ContainerStatus) int {
+	sum := 0
+	for _, cs := range statuses {
+		sum += int(cs.RestartCount)
+	}
+	return sum
```
blumamir, line 86: restart count is for a container, not for pod.

if you see that `c1`, `c2` and `c3` restarted in total 10 times. is `c1` healthy? there is no way to tell...

frontend/services/collectors/pods.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+		if cs.State.Terminated != nil && cs.State.Terminated.Reason != "" {
+			return cs.State.Terminated.Reason
+		}
+	}
+	if string(pod.Status.Phase) != "" {
+		return string(pod.Status.Phase)
+	}
+	return "Unknown"
+}
```
blumamir: also here - if the endpoint show "terminated", there is no way to know which container is creating it and if collector is healthy or not

frontend/services/collectors/pods.go
```diff
@@ -0,0 +1,88 @@
...
+	for _, cs := range pod.Status.ContainerStatuses {
+		if cs.State.Waiting != nil && cs.State.Waiting.Reason != "" {
+			return cs.State.Waiting.Reason
+		}
+		if cs.State.Terminated != nil && cs.State.Terminated.Reason != "" {
+			return cs.State.Terminated.Reason
+		}
+	}
+	if string(pod.Status.Phase) != "" {
+		return string(pod.Status.Phase)
```
blumamir: this one is tricky, as the pod phase can show "Running" even when pods are starting or not healthy. I suggest to validate or remove it, as the phase can be confusing or simply incorrect

frontend/services/collectors/pods.go
```diff
@@ -0,0 +1,88 @@
...
+		return "0/0"
+	}
+	total := len(statuses)
+	ready := 0
+	for _, cs := range statuses {
+		if cs.Ready {
+			ready++
+		}
+	}
+	return fmt.Sprintf("%d/%d", ready, total)
```
blumamir, line 63: This is not very useful - if it says `1/3` - is the collector container ready? we cannot tell....

frontend/services/collectors/pods.go
```diff
@@ -0,0 +1,88 @@
...
+
+	return podsInfo, nil
+}
+
+func podToInfo(pod *corev1.Pod) *model.PodInfo {
+	ready := formatReady(pod.Status.ContainerStatuses)
+	status := deriveStatus(pod)
+	restarts := sumRestarts(pod.Status.ContainerStatuses)
+	node := pod.Spec.NodeName
+	age := strings.ToLower(pod.CreationTimestamp.Time.Format(time.RFC3339))
```
blumamir: age is `1 hour`
the content of this value (`2025-11-18T15:04:05Z`) is the creation timestamp and not age

frontend/graph/collectors.graphqls
```diff
@@ -68,10 +68,30 @@ type CollectorDaemonSetInfo {
...
+# =====================
+
+type PodInfo {
+  name: String!
+  ready: String!
+  status: String!
+  restarts: Int!
+  nodeName: String!
+  age: String!
+  image: String!
```
blumamir, line 85: consider commenting on these fields. 
for example 
- what is `status`? can a user that want to consume this api tell what this means without looking at the code?
- `ready` sounds like a boolean value, but it's a string which isn't trivial.

frontend/services/collectors/pods.go
```diff
@@ -0,0 +1,88 @@
...
+			return cs.State.Waiting.Reason
+		}
+		if cs.State.Terminated != nil && cs.State.Terminated.Reason != "" {
+			return cs.State.Terminated.Reason
+		}
+	}
+	if string(pod.Status.Phase) != "" {
+		return string(pod.Status.Phase)
+	}
+	return "Unknown"
```
blumamir: instead of passing "Unknown", you can make this field not required, and only populate it if we can actually resolve it.

frontend/graph/collectors.graphqls
```diff
@@ -68,10 +68,30 @@ type CollectorDaemonSetInfo {
...
 
+# =====================
+# Gateway Pods List API
+# =====================
+
+type PodInfo {
+  name: String!
+  ready: String!
+  status: String!
+  restarts: Int!
```
blumamir: consider naming it 

```suggestion
  restartsCount: Int!
```

so it is clear what the number means (count), and also to align with k8s

frontend/services/manifest.go
```diff
@@ -0,0 +1,48 @@
...
+	// this can be extended to support other kinds of resources in the future.
+	switch kind {
+	case model.K8sResourceKindDeployment:
+		obj, err := kube.DefaultClient.AppsV1().Deployments(namespace).Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return "", err
+		}
+		obj.ObjectMeta.ManagedFields = nil
+		jb, _ := json.Marshal(obj)
+		yb, _ := syaml.JSONToYAML(jb)
```
blumamir: can we marshal it directly to yaml without json first? looks wasteful to convert twice

frontend/services/manifest.go
```diff
@@ -0,0 +1,48 @@
...
+		return string(yb), nil
+	case model.K8sResourceKindDaemonSet:
```
blumamir, line 29: nit: consider adding space between each case - having all these lines in a big chunk is less readable

blumamir (approved):

frontend/services/collectors/pods.go
alonkeyval: this data is display in the pod drawer

PR #3727, opened 2025-11-02
chore: migrate datastreams router connector to use k8s cache instead of cm

blumamir:
Fixes: CORE-348

We are currently populating the sources in the gateway collector configmap, which the odigos router connector uses to send each workload to the right pipelines.

This PR migrates this logic so the collector "watch" the k8s api, and store small internal cache with this data. 
Main motivation is to reduce the number of config reloads we need to run.

This way: 
- the autoscaler does not need to be aware of sources (instrumentation configs)
- no need to reload gateway config on each potential change to sources.
- shorter and cleaner configmap for the gateways (if there are hundreds of sources, they do not fill the config content with repetitive data)


collector/extension/odigosk8sresourcesexention/odigosk8sresources.go
```diff
@@ -0,0 +1,192 @@
...
+	logger := set.Logger
+	k8sClusterConfig, err := rest.InClusterConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get in-cluster config: %w", err)
+	}
+	// the `HTTPS_PROXY` we allow to set is used for exporting to https destinations.
+	// we don't want to use it for accessing k8s api.
+	k8sClusterConfig.Proxy = func(req *http.Request) (*url.URL, error) {
+		return nil, nil
+	}
```
damemi, line 55: @blumamir where is HTTPS_PROXY set/used?

collector/extension/odigosk8sresourcesexention/odigosk8sresources.go
```diff
@@ -0,0 +1,192 @@
...
+
+func (c *OdigosKsResources) GetDatastreamsForWorkload(workloadKey WorkloadKey) ([]DatastreamName, bool) {
+	c.m.RLock()
+	defer c.m.RUnlock()
+	datastreams, ok := c.workloadToDatastreamsCache[workloadKey]
+	return datastreams, ok
+}
+
+// Wait for cache to sync
+func (c *OdigosKsResources) WaitForCacheSync(ctx context.Context) bool {
```
damemi, line 190: @blumamir why did you need to implement WaitForCacheSync?

collector/extension/odigosk8sresourcesexention/odigosk8sresources.go
blumamir: https://github.com/odigos-io/odigos/pull/3296

blumamir:
fixed in another PR

PR #3831, opened 2025-11-20
chore: build cli image also to arm

tamirdavid1:
## Description

Cli image should be built also for arm nodes.

## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

emergency-ticket id: EMR-546

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -190,7 +190,7 @@ jobs:
           SHORT_COMMIT: ${{ steps.vars.outputs.short_commit }}
           DATE: ${{ steps.vars.outputs.date }}
         run: |
-          ko build --bare --tags latest --tags ${{ env.TAG }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -190,7 +190,7 @@ jobs:
           SHORT_COMMIT: ${{ steps.vars.outputs.short_commit }}
           DATE: ${{ steps.vars.outputs.date }}
         run: |
-          ko build --bare --tags latest --tags ${{ env.TAG }}
+          ko build --bare --tags latest --tags ${{ env.TAG }} --platform=linux/amd64,linux/arm64
```
blumamir, line 193: nit: should we consolidate this configuration into the ko config file?

https://github.com/odigos-io/odigos/blob/main/cli/.ko.yaml

blumamir (approved):

PR #3854, opened 2025-11-25
fix: admin readonly access

alonkeyval:
emergency-ticket id: EMR-565


RonFed (approved):

damemi (approved):
operator rbac no op, approved :)

PR #3871, opened 2025-12-01
chore: adding a metric for total ebpf read and reuse protomarshler

tamirdavid1:
## Description

Add a new metric calls `ebpf_total_bytes_read` to track the amount of data in bytes we read from the ebpf map.
also, reuse `protoUnmarshaler` and not create it in every read operation.
## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

emergency-ticket id: EMR-592

### DIFF
--- a/collector/receivers/odigosebpfreceiver/documentation.md
+++ b/collector/receivers/odigosebpfreceiver/documentation.md
@@ -24,7 +24,7 @@ Total time spent waiting due to memory pressure. Can be compared with otelcol_pr
 
 ### otelcol_ebpf_total_bytes_read
 
-Total number of bytes read from the eBPF ring buffer.
+Total number of bytes read from the eBPF buffer (perf or ring).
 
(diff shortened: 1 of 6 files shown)


collector/receivers/odigosebpfreceiver/documentation.md
```diff
@@ -21,3 +21,11 @@ Total time spent waiting due to memory pressure. Can be compared with otelcol_pr
 | Unit | Metric Type | Value Type | Monotonic |
 | ---- | ----------- | ---------- | --------- |
 | ms | Sum | Int | true |
+
+### otelcol_ebpf_total_bytes_read
+
+Total number of bytes read from the eBPF ring buffer.
```
blumamir: is it only for ring buffer? for all languages or only java?

blumamir (approved):

PR #3880, opened 2025-12-02
feat: add data collection pods is drop data indicator

alonkeyval:
## How Has This Been Tested?

- [ ] Added Unit Tests
- [ ] Updated e2e Tests
- [ ] Manual Testing
- [ ] Manual Load Test

## Kubernetes Checklist

## User Facing Changes

emergency-ticket id: EMR-603


frontend/services/metrics/collector_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+		namespace = env.GetCurrentNamespace()
+	}
+	if len(podNames) == 0 {
+		return map[string]PodRates{}, nil
+	}
+
+	podRegex := buildPodRegex(podNames)
+	now := time.Now()
+
+	qAccepted := rateSumByPod("otelcol_receiver_accepted_spans_total", namespace, podRegex, window)
```
tamirdavid1: nit: can we create consts for metrics names?

frontend/services/metrics/collector_metrics.go
```diff
@@ -0,0 +1,100 @@
...
+	MetricsDroppedRps  float64
+	ExporterSuccessRps float64
+	ExporterFailedRps  float64
+	Window             string
+	LastScrape         time.Time
+}
+
+// GetOdigletCollectorMetrics queries own-metrics (VictoriaMetrics via Prometheus-compatible API)
+// and aggregates per-pod rates for accepted/dropped metrics and exporter success/failures.
+func GetOdigletCollectorMetrics(ctx context.Context, api v1.API, namespace string, podNames []string, window string) (map[string]PodRates, error) {
```
tamirdavid1: We're talking about `data-collection`/ `node-collector` metrics? if so maybe lets change the name of the functions

frontend/services/metrics/client.go
```diff
@@ -0,0 +1,26 @@
...
+		Address:      baseURL,
+		RoundTripper: promapi.DefaultRoundTripper,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return v1.NewAPI(client), nil
+}
+
+func DefaultMetricsWindow() string {
```
tamirdavid1: This can be a const? why we need func for it

frontend/main.go
```diff
@@ -311,8 +314,16 @@ func main() {
 		log.Fatalf("Error starting watchers: %s", err)
 	}
 
+	var promAPI v1.API
+	metricsURL := fmt.Sprintf("http://odigos-victoriametrics.%s.svc:8428", flags.Namespace)
```
tamirdavid1: can we set const for `odigos-victoriametrics` as victoriametricssvcname or something?

frontend/services/metrics/collector_metrics.go
tamirdavid1: also, we we need to buildPodRegex?
i believe there is a way doing that with label=value
e.g k8s_pod_name = podName
is it because we want to do it in 1 shot ?

tamirdavid1 (approved):

PR #3909, opened 2025-12-07
feat: remote config from the central

alonkeyval:
add option the override odigos config from odigos central (remote)

emergency-ticket id: EMR-629


scheduler/controllers/odigosconfiguration/odigosconfiguration_controller.go
```diff
@@ -133,6 +145,53 @@ func (r *odigosConfigurationController) getOdigosConfigMap(ctx context.Context)
...
+// configuration managed by the central-backend. This config takes precedence
+// over helm-managed configuration for supported fields.
+func (r *odigosConfigurationController) getRemoteConfig(ctx context.Context) (*common.OdigosConfiguration, error) {
+	var configMap corev1.ConfigMap
+	odigosNs := env.GetCurrentNamespace()
+
+	err := r.Client.Get(ctx, types.NamespacedName{Namespace: odigosNs, Name: consts.OdigosRemoteConfigName}, &configMap)
+	if err != nil {
+		return nil, err
+	}
```
blumamir, line 163: what if the remote config does not exists? do we want to err in this case?

scheduler/controllers/odigosconfiguration/odigosconfiguration_controller.go
```diff
@@ -57,6 +59,16 @@ func (r *odigosConfigurationController) Reconcile(ctx context.Context, _ ctrl.Re
 		return ctrl.Result{}, err
 	}
 
+	// Read and merge remote config (from central-backend) if it exists
+	// Remote config takes precedence over helm-managed config
+	remoteConfig, err := r.getRemoteConfig(ctx)
+	if err != nil {
+		logger.V(1).Info("No remote config found, using only helm-managed config", "error", err)
```
blumamir: most odigos deployment will not contain the remote config. should we avoid this log line when it's not relevant?

blumamir: `err` can also be related to other issues, not only `NotFound` consider better handling of not found vs real error

blumamir (approved):

PR #3920, opened 2025-12-09
chore: Run prerelease checks on patch and relevant branches

damemi:
https://github.com/odigos-io/odigos/pull/3650 misconfigured the labels check for the prerelease CI, so it was just comparing `[Array] = prerelease`. This switches it to `contains()` to actually check for the label

It also now runs the prerelease checks on patch releases. This updates the very-dependencies and verify-release-blockers checks to only look at the relevant release branch. So for example, if we are running a patch release, we don't need to verify that there aren't any release blocks or dependency bots on main in odigos-enterprise.

emergency-ticket id: EMR-643

### DIFF
--- a/.github/workflows/create-patch.yml
+++ b/.github/workflows/create-patch.yml
@@ -108,6 +108,7 @@ jobs:
                 --base $RELEASE_BRANCH \
                 --head $PR_BRANCH \
                 --title "$PR_TITLE" \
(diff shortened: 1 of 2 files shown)


aviadhahami (approved):
Lgtm

BenElferink (approved):

PR #3978, opened 2025-12-24
Improve central UI command

tamirdavid1:
Currently, if either central-ui or central-backend restarts, the CLI command breaks and must be re-run.
This change adds a retry and reconnection mechanism to the new pod, so the command no longer needs to be re-executed.

emergency-ticket id: EMR-750


cli/cmd/ui.go
```diff
@@ -90,3 +83,67 @@ func init() {
 	uiCmd.Flags().Int("remote-port", defaultPort, "Port to forward to the remote UI")
 	uiCmd.Flags().String("address", "localhost", "Address to serve the UI on")
 }
+
+func startResilientUIPortForward(ctx context.Context, client *kube.Client, localPort, remotePort, localAddress, namespace string) {
```
dan505512: Should we maybe unify the two functions?

dan505512 (approved):

PR #3979, opened 2025-12-24
Core 467/support argo rollouts

natyharary:
Support Argo Rollouts CRDs in the odigos.
This includes:
1. Support instrumentation and de-instrumentation of rollouts in the instrumentor
2. UI and CLI support for these new objects
3. Updated RBAC for the odigos controller and operator to be able to list and interact with rollouts
4. Support for both Argo Rollout startegies - green/blue and canary

Tests done:
🆗 Cluster without argo rollout controller works as usual - instrumentation, de-instrumentation with the demo app, restarts
🆗 Cluster with argo rollouts controller + namespace - deploying a blue/green argo rollout deployment with the python test app
🆗 For previous deployment, promote to have two replicasets: one blue, one green
🆗 Instrumentation of said argo rollout caused restart of both blue and green sets + instrumentation, without switching blue to green
🆗 De-istrumentation of said argo rollout caused restart of both blue and green sets + instrumentation, without switching blue to green
🆗 Work with the UI and CLI with argo rollouts proved satisfactory
🆗 Instrumentation + de-instrumentation of canary rollouts - start, end and mid-rollout
🆗 Disabling auto rollout and seeing that instrumentation and de-instrumentation of argo rollouts doesn't trigger a restart

emergency-ticket id: EMR-749


api/k8sconsts/podworkload.go
```diff
@@ -12,6 +12,7 @@ const (
 	WorkloadKindCronJob          WorkloadKind = "CronJob"
 	WorkloadKindJob              WorkloadKind = "Job"
 	WorkloadKindDeploymentConfig WorkloadKind = "DeploymentConfig"
+	WorkloadKindRollout          WorkloadKind = "Rollout"
```
RonFed: nit: (for clarity)

```suggestion
	WorkloadKindArgoRollout          WorkloadKind = "ArgoRollout"
```

cli/cmd/resources/instrumentor.go
```diff
@@ -256,6 +256,11 @@ func NewInstrumentorClusterRole(openshiftEnabled bool) *rbacv1.ClusterRole {
 				ResourceNames: []string{k8sconsts.InstrumentorSourceValidatingWebhookName},
 				Verbs:         []string{"update"},
 			},
+			{
+				APIGroups: []string{"argoproj.io"},
+				Resources: []string{"rollouts"},
+				Verbs:     []string{"get", "list", "watch", "patch"},
+			},
```
RonFed, line 263: I think this part of the code is deprecated, not sure if we want to add stuff here
cc @tamirdavid1

cli/cmd/resources/ui.go
RonFed, line 1: same as other comment. not sure if this code is still maintained

cli/cmd/describe.go
RonFed, line 1: similar to other comments - specifically the describe command is also not maintained - did it work? not sure what is our plan for this command
cc @blumamir

k8sutils/pkg/workload/ownerreference.go
```diff
@@ -68,6 +71,16 @@ func GetWorkloadNameAndKind(ownerName, ownerKind string) (string, k8sconsts.Work
 	}
 }
 
+// ReplicaSets can be created from either Deployment or (Argo) Rollout kinds, so determine which one is that
+func determineReplicaSetOwner(ownerName string, pod *corev1.Pod) (string, k8sconsts.WorkloadKind, error) {
+	// If we find a label associated with Argo rollouts, it is an Rollout kind
+	if _, ok := pod.Labels[argorolloutsv1alpha1.DefaultRolloutUniqueLabelKey]; ok {
```
RonFed: since we only use the labels here, consider passing only the labels down the function chain

RonFed, line 77: is this label present on all argo rollouts? since it is called `Default...` is there an option in argo that will override it?

opampserver/pkg/sdkconfig/configresolvers/resourceattributes.go
```diff
@@ -65,6 +65,9 @@ func CalculateServerAttributes(k8sAttributes *K8sResourceAttributes, nodeName st
 		objectNameKey = string(semconv.K8SJobNameKey)
 	case "CronJob":
 		objectNameKey = string(semconv.K8SCronJobNameKey)
+	case "Rollout":
+		// Argo Rollout - use custom key with argoproj prefix since it's an Argo-specific resource
+		objectNameKey = "k8s.argoproj.rollout.name"
```
RonFed: better to put this attribute key in a const in a global location since it used also in the autoscaler, odiglat and ui

opampserver/go.mod
```diff
@@ -16,7 +16,7 @@ require (
 )
 
 require (
-	github.com/evanphx/json-patch v4.9.0+incompatible // indirect
```
RonFed, line 19: a bit weird that it went away

helm/odigos/templates/ui/clusterrole.yaml
```diff
@@ -52,6 +52,17 @@ rules:
       - patch
       {{- end }}
 {{- end }}
+  - apiGroups:
+      - argoproj.io
```
RonFed, line 57: if argo is not installed in the cluster - will this be fine? since the apiserver will not be aware of this type.

instrumentor/controllers/sourceinstrumentation/instrumentationconfig_controller.go
```diff
@@ -83,6 +83,12 @@ func getObjectByOwnerReference(ctx context.Context, k8sClient client.Client, own
 		return dc, err
 	}
 
+	if ownerRef.Kind == "Rollout" {
```
RonFed: use the enum instead of hardcoded string

instrumentor/controllers/sourceinstrumentation/manager.go
```diff
@@ -32,6 +33,21 @@ func isDeploymentConfigAvailable(mgr ctrl.Manager) bool {
 	return err == nil
 }
 
+// isRolloutAvailable checks if the Rollout resource is available in the cluster
+// using the RESTMapper to avoid permission errors on non-Rollout clusters
+func isRolloutAvailable(mgr ctrl.Manager) bool {
```
RonFed:
```suggestion
func isArgoRolloutAvailable(mgr ctrl.Manager) bool {
```

instrumentor/controllers/sourceinstrumentation/manager.go
```diff
@@ -32,6 +33,21 @@ func isDeploymentConfigAvailable(mgr ctrl.Manager) bool {
...
+	gvk := schema.GroupVersionKind{
+		Group:   "argoproj.io",
+		Version: "v1alpha1",
+		Kind:    "Rollout",
+	}
+
+	// Try to get the REST mapping for DeploymentConfig
```
RonFed:
```suggestion
	// Try to get the REST mapping for Argo Rollout
```

instrumentor/controllers/sourceinstrumentation/manager.go
RonFed: if its the same logic as `isDeploymentConfigAvailable` - we can create one function and pass the resource group and name as parmas

instrumentor/controllers/sourceinstrumentation/common.go
```diff
@@ -103,6 +105,14 @@ func syncNamespaceWorkloads(
 					Kind:      k8sconsts.WorkloadKindDeploymentConfig,
 				})
 			}
+		case *argorolloutsv1alpha1.RolloutList:
+			for _, dc := range obj.Items {
```
RonFed:
```suggestion
			for _, ar := range obj.Items {
```

api/k8sconsts/podworkload.go
natyharary: I agree, but I only changed the variable name tho, the kind is still `Rollout` according to the argo rollouts yaml spec:
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Rollout
```

When changing the kind to ArgoRollout stuff started to unexpectedly break within Odigos, since these consts are used as both an internal odigos use and as a parser for yamls. If we want to have `ArgoRollout` as an internal kind it'll require some workarounds which I prefer not to do...

cli/cmd/describe.go
natyharary: I didn't know if it was maintained or not so I implemented argo rollouts for it, just like instrumentationconfigs.
At any rate, it worked - output in the next comment:

helm/odigos/templates/ui/clusterrole.yaml
natyharary: How do you suggest testing this?
I tried `make helm-install` with a new odigos version and deployed this branch's code , the UI operated as usual without errors.

cli/cmd/describe.go
natyharary:
```
odigos-new describe source rollouts rollout-bluegreen
Name: rollout-bluegreen
Kind: Rollout
Namespace: default
Source Custom Resources:
  Instrumented: true
  Workload: instrumented
  Namespace: unset
  DecisionText: Workload is instrumented because the workload source is present and enabled

Runtime Inspection Details:
  Detected Containers:
  - Container Name: rollouts-demo
    Programming Language: python
    Runtime Version: 3.11.13
    Relevant Environment Variables:
    - PYTHONPATH: /app

Instrumentation Config:
  Created: created
  create time: 2025-12-28 11:22:06 +0200 IST
  Containers:
  - Container Name: rollouts-demo
    Agent Enabled: true
    Otel Distro Name: python-community

Pods (Total 1, Running 1):

- Pod Name: rollout-bluegreen-5c758cfcfc-wzx6t
  Node Name: kind-control-plane
  Odigos Agent Injected: true
  Phase: Running
  Containers:
  - Container Name: rollouts-demo
    Actual Devices: [generic]
    Started: true
    Ready: true

    Instrumentation Instances:
    - Healthy: true
      Identifying Attributes:
        service.instance.id: 06950f6b-fb56-7b36-8000-6db1bc1dce0d
        process.vpid: 1
        telemetry.sdk.language: python
        k8s.namespace.name: default
        k8s.container.name: rollouts-demo
        k8s.pod.name: rollout-bluegreen-5c758cfcfc-wzx6t
```

natyharary: (it is called `odigos-new` since I installed the new CLI temporarily as `odigos-new`)

opampserver/go.mod
natyharary: Agreed. In general I don't get how `go mod tidy` decides to throw away versions but I can't modify it otherwise.
Should I manually re-add it? WDYT?

k8sutils/pkg/workload/ownerreference.go
natyharary: It looks pretty ubiquitous throughout the argo rollouts code.
All argo rollouts (canary and bluegreen) use replicasets, and this label is always present in such replicaset

instrumentor/controllers/sourceinstrumentation/instrumentationconfig_controller.go
natyharary: Good idea.

instrumentor/controllers/sourceinstrumentation/manager.go
natyharary: Oopsie poopsie.

instrumentor/controllers/sourceinstrumentation/manager.go
natyharary: Done. I've moved the `GroupVersionKind` objects to k8sconsts like so:
```go
// GroupVersionKind constants for external workload types.
// These can be used to check resource availability in clusters.
var (
	DeploymentConfigGVK = schema.GroupVersionKind{
		Group:   "apps.openshift.io",
		Version: "v1",
		Kind:    string(WorkloadKindDeploymentConfig),
	}
	ArgoRolloutGVK = schema.GroupVersionKind{
		Group:   "argoproj.io",
		Version: "v1alpha1",
		Kind:    string(WorkloadKindArgoRollout),
	}
)
```
under `api/k8sconsts/podworkload.go`

instrumentor/controllers/sourceinstrumentation/common.go
natyharary: Right you are

opampserver/pkg/sdkconfig/configresolvers/resourceattributes.go
blumamir: +1

frontend/kube/client.go
```diff
@@ -122,3 +128,35 @@ func IsDeploymentConfigAvailable() bool {
...
+
+	listOptions := metav1.ListOptions{Limit: 0}
+	_, err := DefaultClient.DynamicClient.Resource(gvr).List(context.Background(), listOptions)
+
+	if err != nil {
+		// Resource doesn't exist or we don't have permission to access it
+		rolloutAvailable = false
+		rolloutAvailabilityChecked = true
+		return false
+	}
```
blumamir: nit: there can be transient errors which are not related to permissions problems.

consider refining and adding log so we can know if there are any issues

operator/config/rbac/role.yaml
```diff
@@ -189,6 +189,18 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - argoproj.io
+  resources:
+  - rollouts
+  verbs:
+  - create
+  - delete
```
blumamir: I don't think we need to create or delete argo rollouts in odigos. probably safer to skip these permissions

instrumentor/controllers/sourceinstrumentation/common.go
```diff
@@ -44,6 +45,7 @@ func syncNamespaceWorkloads(
 		k8sconsts.WorkloadKindStatefulSet,
 		k8sconsts.WorkloadKindCronJob,
 		k8sconsts.WorkloadKindDeploymentConfig,
+		k8sconsts.WorkloadKindRollout,
 	} {
 		workloadObjects := workload.ClientListObjectFromWorkloadKind(kind)
 		err := k8sClient.List(ctx, workloadObjects, client.InNamespace(namespace))
```
blumamir, line 52: check if this `List` works correctly when argo rollout is not found

cli/cmd/resources/instrumentor.go
natyharary: Tamir and I talked - we still want this code to be maintained as it is still partially used by the UI.

cli/cmd/resources/ui.go
natyharary: ditto

instrumentor/controllers/sourceinstrumentation/common.go
natyharary: Yes - the code below:
```go
if err != nil {
			// Ignore "no matches for kind" and "forbidden" errors for DeploymentConfig
			// This happens on non-OpenShift clusters where:
			// - The DeploymentConfig resource doesn't exist (NoMatchError)
			// - RBAC permissions aren't granted (Forbidden)
			if kind == k8sconsts.WorkloadKindDeploymentConfig && (meta.IsNoMatchError(err) || apierrors.IsForbidden(err)) {
				continue
			}
			// For other errors or other workload kinds, collect the error
			if !meta.IsNoMatchError(err) {
				errs = errors.Join(errs, err)
			}
			continue
		}
```

Catches any errors and logs them, in a nonblocking way.
Either way I added a case for the argo rollouts as well so there will not be any error logging if this kind is not in the cluster.

instrumentor/controllers/sourceinstrumentation/manager.go
natyharary: Done

opampserver/pkg/sdkconfig/configresolvers/resourceattributes.go
natyharary: Done

operator/config/rbac/role.yaml
natyharary: Agreed - When I was testing I thought I saw some error in instrumenting the rollouts but now I see this has ntohing to do with the create/delete permissions. Removed

frontend/kube/client.go
natyharary: This is a bit of a pickle.
`IsArgoRolloutsAvailable` checks this only once per FE call, and then caches the response (whether we have or don't have argo rollouts available).

Maybe this just means that the cluster does not install argo rollouts. Maybe it means that they lack permissions and have installed it.

The middle way that I can suggest is recording an error if argo rollouts are installed but without enough permissions to access them (list for example), and recording a debug message if argo rollouts were found at all.

This way we can know if we have permission problems as errors or we just don't have argo rollouts.

WDYT?

common/consts/consts.go
```diff
@@ -62,6 +62,10 @@ const (
 	// This is needed to distinguish between workloads that share the same semconv key
 	// (e.g., DeploymentConfig uses k8s.deployment.name)
 	OdigosWorkloadKindAttribute = "odigos.workload.kind"
+
+	// Custom resource attribute for Argo Rollouts workload name.
+	// There is no semconv key for Argo Rollouts, so we use this custom key with argoproj prefix.
+	K8SArgoRolloutNameAttribute = "k8s.argoproj.rollout.name"
```
blumamir: move to api module. common is shared with vm

k8sutils/pkg/utils/rollout.go
```diff
@@ -176,6 +180,41 @@ func isDeploymentConfigRolloutDone(dc *openshiftappsv1.DeploymentConfig) bool {
...
 
+func isArgoRolloutRolloutDone(rollout *argorolloutsv1alpha1.Rollout) bool {
+	// Yes, this name is ridiculous. The function returns whether the rollout of an Argo rollout is done.
+
+	// A paused rollout (e.g., during a canary pause step) is considered "done" for instrumentation purposes.
+	// This is because the rollout is in a stable state where all current pods are running and healthy,
+	// just waiting for user input to continue. Odigos should be able to instrument pods in this state.
+	if rollout.Status.Phase == argorolloutsv1alpha1.RolloutPhasePaused {
+		return true
+	}
```
blumamir: can we use the phase always? is it paused different from other states?

k8sutils/pkg/utils/rollout.go
```diff
@@ -176,6 +180,41 @@ func isDeploymentConfigRolloutDone(dc *openshiftappsv1.DeploymentConfig) bool {
...
+	observedGen, err := strconv.ParseInt(rollout.Status.ObservedGeneration, 10, 64)
+	if err != nil {
+		// If we can't parse it, assume the spec hasn't been observed yet
+		return false
+	}
+	if rollout.Generation > observedGen {
+		// Waiting for rollout spec update to be observed
+		return false
+	}
+	if rollout.Spec.Replicas != nil && rollout.Status.UpdatedReplicas < *rollout.Spec.Replicas {
```
blumamir: should we calculate replicas from spec?

k8sutils/pkg/workload/workload.go
```diff
@@ -24,6 +25,7 @@ var _ Workload = &StatefulSetWorkload{}
 var _ Workload = &CronJobWorkloadV1{}
 var _ Workload = &CronJobWorkloadBeta{}
 var _ Workload = &DeploymentConfigWorkload{}
+var _ Workload = &RolloutWorkload{}
```
blumamir: ArgoRollout

api/k8sconsts/podworkload.go
```diff
@@ -12,6 +29,11 @@ const (
 	WorkloadKindCronJob          WorkloadKind = "CronJob"
 	WorkloadKindJob              WorkloadKind = "Job"
 	WorkloadKindDeploymentConfig WorkloadKind = "DeploymentConfig"
+	// Note: The actual Kubernetes resource has kind "Rollout" (not "ArgoRollout"):
+	//   apiVersion: argoproj.io/v1alpha1
+	//   kind: Rollout
+	// We use "ArgoRollout" in the variable name to distinguish it from other rollout concepts.
```
RonFed, line 35: like

api/k8sconsts/podworkload.go
```diff
@@ -1,5 +1,22 @@
 package k8sconsts
 
+import "k8s.io/apimachinery/pkg/runtime/schema"
```
RonFed, line 3: kind of a nit, not sure if this this the right place for this? maybe should be in `k8sutils`? not sure about it

k8sutils/pkg/utils/rollout.go
```diff
@@ -176,6 +180,27 @@ func isDeploymentConfigRolloutDone(dc *openshiftappsv1.DeploymentConfig) bool {
...
+	case argorolloutsv1alpha1.RolloutPhaseDegraded, argorolloutsv1alpha1.RolloutPhaseProgressing:
+		return false
+	}
+
+	// Check if the spec has been observed yet
+	// ObservedGeneration in Argo Rollouts is a string, so we need to parse it
+	observedGen, err := strconv.ParseInt(rollout.Status.ObservedGeneration, 10, 64)
+	if err != nil || rollout.Generation > observedGen {
+		return false
+	}
```
blumamir: check the observed generation before

operator/bundle/manifests/odigos-operator.clusterserviceversion.yaml
```diff
@@ -198,6 +198,7 @@ spec:
           resources:
           - pods
           verbs:
+          - delete
```
blumamir: remove

operator/bundle/manifests/odigos-operator.clusterserviceversion.yaml
```diff
@@ -320,6 +321,18 @@ spec:
           - patch
           - update
           - watch
+        - apiGroups:
+          - argoproj.io
+          resources:
+          - rollouts
+          verbs:
+          - create
+          - delete
```
blumamir: no create and delete

blumamir (approved):

opampserver/go.mod
RonFed: if its the result of go mod tidy, I guess its fine

RonFed (approved):
🔥 
I would consider moving the argo tests to the existing sources one

api/k8sconsts/podworkload.go
natyharary: Perhaps in a future refactor when we move `WorkloadKind` entirely. I kind of just went with what worked with `WorkloadKindDeploymentConfig`

cli/cmd/resources/instrumentor.go
damemi: +1 thanks @natyharary, Tamir talked to me too-- it's really just the operator that's still dependent on this so I appreciate everyone's patience while I find time to migrate that over to the new helm backed cli deploy method

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -238,6 +239,11 @@ func rolloutRestartWorkload(ctx context.Context, workload client.Object, c clien
 		return c.Patch(ctx, obj, client.RawPatch(types.MergePatchType, patch))
 	case *openshiftappsv1.DeploymentConfig:
 		return c.Patch(ctx, obj, client.RawPatch(types.MergePatchType, patch))
+	case *argorolloutsv1alpha1.Rollout:
+		// Rollouts use a different field (spec.restartAt) for restarting, so we need to patch it differently
+		// https://github.com/argoproj/argo-rollouts/blob/cb1c33df7a2c2b1c2ed31b1ee0aa22621ef5577c/utils/replicaset/replicaset.go#L223-L232
+		rolloutPatch := []byte(fmt.Sprintf(`{"spec":{"restartAt":"%s"}}`, ts.Format(time.RFC3339)))
```
damemi, line 245: tbh, I wonder if we can switch to using a better annotation for all resource types here

afaik we just use `kubectl.kubernetes.io/restartedAt` because it's how kubectl restarts a deployment. but really, any annotation should work right? I think an odigos-specific annotation would be really helpful for showing what time *we* restarted the workloads wdyt about that @blumamir 

But still, in this case Argo rollouts don't just restart on an annotation update @natyharary ?

k8sutils/pkg/client/client.go
```diff
@@ -97,3 +99,10 @@ func GetK8sClientset() (*kubernetes.Clientset, error) {
 	}
 	return clientset, nil
 }
+
+// IsResourceAvailable checks if the external resource is available in the cluster
+// using the RESTMapper to avoid permission errors on clusters that don't have the resource installed.
+func IsResourceAvailable(mapper meta.RESTMapper, gvk schema.GroupVersionKind) bool {
```
damemi, line 105: awesome

instrumentor/controllers/agentenabled/rollout/rollout.go
natyharary: Argo rollouts work a bit differently here when it comes to `restartedAt` - instead of  `kubectl.kubernetes.io/restartedAt`, it restarts upon `spec.restartedAt`.

I might be missing something here regarding changing which annotation triggers which action, isn't it something rigid that cannot be changed? So Argo rollouts keep restarting upon `spec.restartedAt` and the rest of the workloads upon `kubectl.kubernetes.io/restartedAt`

> But still, in this case Argo rollouts don't just restart on an annotation update @natyharary ?

@damemi  When I checked, Argo rollouts didn't restart upon `kubectl.kubernetes.io/restartedAt`

PR #3996, opened 2025-12-29
feat: send clusterName to UI

yarden-lang:
emergency-ticket id: EMR-773

### DIFF
--- a/frontend/services/config.go
+++ b/frontend/services/config.go
@@ -3,6 +3,7 @@ package services
 import (
 	"context"
 	"errors"
+	"fmt"
 	"log"
 
 	"github.com/odigos-io/odigos/api/k8sconsts"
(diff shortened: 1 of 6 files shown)


frontend/services/config.go
```diff
@@ -59,22 +59,44 @@ func GetConfig(ctx context.Context) model.GetConfigResponse {
...
+	if err != nil {
 		return false
 	}
 
-	return odigosConfiguration.UiMode == common.UiModeReadonly
+	return config.UiMode == common.UiModeReadonly
+}
+
+func GetClusterName(ctx context.Context) string {
+	config, err := getOdigosConfigForConfigFile(ctx)
```
alonkeyval: If GetClusterName returns "", the API returns "".

GraphQL optional fields should usually distinguish between:

not configured - null

configured but empty - "" 


suggestion: return *string

frontend/services/config.go
```diff
@@ -59,22 +59,44 @@ func GetConfig(ctx context.Context) model.GetConfigResponse {
 	return response
 }
 
-func IsReadonlyMode(ctx context.Context) bool {
+func getOdigosConfigForConfigFile(ctx context.Context) (*common.OdigosConfiguration, error) {
```
alonkeyval: this name is little bit verbose, ForConfigFile” doesn’t add meaning.

suggestion: getOdigosConfiguration

frontend/services/config.go
```diff
@@ -59,22 +59,44 @@ func GetConfig(ctx context.Context) model.GetConfigResponse {
...
 	ns := env.GetCurrentNamespace()
 
 	configMap, err := kube.DefaultClient.CoreV1().ConfigMaps(ns).Get(ctx, consts.OdigosEffectiveConfigName, metav1.GetOptions{})
 	if err != nil {
-		log.Printf("Error getting config maps: %v\n", err)
-		return false
+		log.Printf("Error getting effective-config: %v\n", err)
```
alonkeyval: you print this error but fallback to odigos configuration file. this log may look alarming while everything works

frontend/services/config.go
yarden-lang: fixed

alonkeyval (approved):

PR #3998, opened 2025-12-30
feat: added isConnectedToCentralBackend to sent to client

yarden-lang:
emergency-ticket id: EMR-779


frontend/webapp/utils/constants/routes.tsx
```diff
@@ -24,7 +24,7 @@ const IS_DEV = process.env.NODE_ENV === 'development';
 const DEFAULT_BACKEND_HTTP_ORIGIN = 'http://localhost:8085';
 
 // TODO: improve this
-const BACKEND_HTTP_ORIGIN = typeof window !== 'undefined' ? (IS_INGRESSED_DOMAIN ? `${PROTOCOL}//${HOSTNAME}` : window.location.origin) : 'http://localhost:3000';
+const BACKEND_HTTP_ORIGIN = typeof window !== 'undefined' ? (IS_INGRESSED_DOMAIN ? `${PROTOCOL}//${HOSTNAME}` : DEFAULT_BACKEND_HTTP_ORIGIN) : 'http://localhost:3000';
```
alonkeyval: revert

frontend/services/config.go
```diff
@@ -86,6 +93,42 @@ func IsReadonlyMode(ctx context.Context) bool {
 	return config.UiMode == common.UiModeReadonly
 }
 
+func isConnectedToCentralBackend(ctx context.Context, clusterName *string) (bool, error) {
+	config, err := GetOdigosConfiguration(ctx)
```
alonkeyval: dont we already read the config in GetClusterName?
that means:
- two cm reads
- two yaml unmarshals

i think is better to pas it down as prop

frontend/services/config.go
```diff
@@ -86,6 +93,42 @@ func IsReadonlyMode(ctx context.Context) bool {
...
+	if clusterName == nil || *clusterName == "" {
+		return false, nil
+	}
+
+	if config.CentralBackendURL == "" {
+		return false, nil
+	}
+
+	ns := env.GetCurrentNamespace()
+	tier, err := utils.GetCurrentOdigosTier(ctx, ns, kube.DefaultClient.Interface.(*kubernetes.Clientset))
```
alonkeyval: this is the high priority check, if the user is not enterprise tier we dont even need to check the config.

also, you already have it in the response object above, you can use it instead of get it again

frontend/services/config.go
```diff
@@ -86,6 +93,42 @@ func IsReadonlyMode(ctx context.Context) bool {
 	return config.UiMode == common.UiModeReadonly
 }
 
+func isConnectedToCentralBackend(ctx context.Context, clusterName *string) (bool, error) {
```
alonkeyval: the name of the func suggest the this cluster connected to the central backend, but from the implementation it check the central proxy health. also, to connect the cluster to the central we also need to check the ws connection health.

maybe we should think on other name

frontend/services/config.go
```diff
@@ -86,6 +93,42 @@ func IsReadonlyMode(ctx context.Context) bool {
...
+	if tier != common.OnPremOdigosTier {
+		return false, nil
+	}
+
+	deployment, err := kube.DefaultClient.AppsV1().Deployments(ns).Get(ctx, k8sconsts.CentralProxyDeploymentName, metav1.GetOptions{})
+	if err != nil {
+		log.Printf("Central proxy deployment not found: %v\n", err)
+		return false, nil
+	}
+	if deployment.Status.AvailableReplicas == 0 {
```
alonkeyval, line 128: what is this check mean? if the replicas are available it mean they ready?

frontend/services/config.go
```diff
@@ -59,6 +60,24 @@ func GetConfig(ctx context.Context) model.GetConfigResponse {
...
+	response.Readonly = config.UiMode == common.UiModeReadonly
+	response.PlatformType = model.ComputePlatformTypeK8s
+	response.Tier = model.Tier(deploymentData[k8sconsts.OdigosDeploymentConfigMapTierKey])
+	response.OdigosVersion = deploymentData[k8sconsts.OdigosDeploymentConfigMapVersionKey]
+	response.InstallationMethod = string(deploymentData[k8sconsts.OdigosDeploymentConfigMapInstallationMethodKey])
+	response.ClusterName = &config.ClusterName
+
+	isConnected, err := isCentralProxyRunning(ctx, &config.ClusterName, config)
+	if err != nil {
+		log.Printf("Error checking if connected to central backend: %v\n", err)
```
alonkeyval: we need to match the comment to the logic

frontend/services/config.go
```diff
@@ -86,13 +105,80 @@ func IsReadonlyMode(ctx context.Context) bool {
 	return config.UiMode == common.UiModeReadonly
 }
 
-func GetClusterName(ctx context.Context) *string {
-	config, err := GetOdigosConfiguration(ctx)
+func isCentralProxyRunning(ctx context.Context, clusterName *string, config *common.OdigosConfiguration) (bool, error) {
+	ns := env.GetCurrentNamespace()
+	tier, err := utils.GetCurrentOdigosTier(ctx, ns, kube.DefaultClient.Interface.(*kubernetes.Clientset))
```
alonkeyval: we already have the tier in the parent func

frontend/graph/schema.graphqls
```diff
@@ -392,6 +392,7 @@ type GetConfigResponse {
   installationMethod: String!
   installationStatus: InstallationStatus!
   clusterName: String
+  isConnectedToCentralBackend: Boolean
```
alonkeyval: we need to match the name to the new naming and logic

frontend/services/config.go
```diff
@@ -86,13 +105,80 @@ func IsReadonlyMode(ctx context.Context) bool {
...
+}
+
+func isCentralProxyDeploymentHealthy(deployment *appsv1.Deployment) bool {
+	// Check if we have at least one available replica
+	if deployment.Status.AvailableReplicas == 0 {
+		return false
+	}
+
+	// Check deployment conditions for a more robust health check
+	var availableCondition, progressingCondition *appsv1.DeploymentCondition
```
alonkeyval: we dont need all those logic, the previous was fine (AvailableReplicas > 0)

frontend/services/config.go
```diff
@@ -41,21 +39,40 @@ func GetConfig(ctx context.Context) model.GetConfigResponse {
...
+	response.PlatformType = model.ComputePlatformTypeK8s
 	response.Tier = model.Tier(deploymentData[k8sconsts.OdigosDeploymentConfigMapTierKey])
 	response.OdigosVersion = deploymentData[k8sconsts.OdigosDeploymentConfigMapVersionKey]
 	response.InstallationMethod = string(deploymentData[k8sconsts.OdigosDeploymentConfigMapInstallationMethodKey])
-	clusterName := GetClusterName(ctx)
-	response.ClusterName = clusterName
+	response.ClusterName = &config.ClusterName
+	isConnected, err := isCentralProxyRunning(ctx)
+	if err != nil {
+		log.Printf("Error checking if central proxt connected: %v\n", err)
```
alonkeyval: proxt -> proxy

frontend/services/config.go
yarden-lang: reverted

alonkeyval (approved):

PR #4017, opened 2026-01-04
feat: create the custom metrics apiservice using helm

tamirdavid1:
Due to a bug of deleting odigos-system ns once autoscaler deleted, apiservice creation move to helm so it will be responsible for deleting it.

emergency-ticket id: EMR-790


autoscaler/controllers/metricshandler/custom_metrics_handler.go
```diff
@@ -25,6 +25,14 @@ import (
 
 const (
 	gatewayRejectionMetricName = "odigos_gateway_memory_limiter_rejections_total"
+
+	// Custom Metrics API constants
+	customMetricsAPIGroup     = "odigos-custom-metrics.k8s.io"
```
damemi: Maybe we should namespace it under `odigos.io` instead of `k8s.io` like `custom-metrics.odigos.io`? Don’t know if that’s the convention here or not

RonFed: +1
not sure what the naming restriction are

operator/config/rbac/role.yaml
```diff
@@ -140,6 +140,7 @@ rules:
   - apiservices
   verbs:
   - create
```
RonFed, line 142: can we remove the create verb?
can we make the delete specific to the legacy resource?

RonFed (commented):
Can you please add to `assert_odigos_uninstalled.sh` deletion of the odigos ns namespace and assertion that is removed?

operator/config/rbac/role.yaml
tamirdavid1: create + delete permissions removed

PR #4027, opened 2026-01-05
fix:  admin token not found after upgrade central

alonkeyval:
The Helm chart required the Keycloak admin password (--set auth.adminPassword=xxx) to be provided on every upgrade. This caused the Keycloak credentials secret to be re-applied on each upgrade, which triggered:
- Keycloak pod restart - Any change to the secret triggered a rolling update of the Keycloak deployment
- Potential password mismatch - If persistence was enabled and a different password was provided (even accidentally), the secret would update but Keycloak's internal password remained unchanged. This caused the central-backend to fail authentication against Keycloak with a 401 error ("Failed to get admin token: 401 Unauthorized")

Fix:
I updated both the Helm chart and CLI to:

- Make admin password optional - No longer required on every upgrade
- Preserve existing password - On upgrades, the system now reads the existing password from the secret instead of re-creating it
- Auto-generate on first install - If no password is provided, a secure random password is generated

emergency-ticket id: EMR-801


cli/cmd/resources/centralodigos/keycloak.go
```diff
@@ -38,17 +40,49 @@ func (m *keycloakResourceManager) Name() string { return k8sconsts.KeycloakResou
...
+// 3. Otherwise, generate a random UUID as password
+func (m *keycloakResourceManager) resolveAdminPassword(ctx context.Context) (string, error) {
+	existingSecret, err := m.client.CoreV1().Secrets(m.ns).Get(ctx, k8sconsts.KeycloakSecretName, metav1.GetOptions{})
+	if err == nil {
+		if password, ok := existingSecret.Data[k8sconsts.KeycloakAdminPasswordKey]; ok {
+			return string(password), nil
+		}
+	} else if !apierrors.IsNotFound(err) {
+		return "", err
+	}
```
BenElferink, line 76: I think we can have better error control with this:


```suggestion
	existingSecret, err := m.client.CoreV1().Secrets(m.ns).Get(ctx, k8sconsts.KeycloakSecretName, metav1.GetOptions{})
	if err != nil && !apierrors.IsNotFound(err) {
		return "", err
	}
	if password, ok := existingSecret.Data[k8sconsts.KeycloakAdminPasswordKey]; ok {
		return string(password), nil
	}
```

cli/cmd/resources/centralodigos/keycloak.go
```diff
@@ -38,17 +40,49 @@ func (m *keycloakResourceManager) Name() string { return k8sconsts.KeycloakResou
...
+
+	if m.config.AdminPassword != "" {
+		return m.config.AdminPassword, nil
+	}
+
+	// Generate a random password using UUID
+	return uuid.New().String(), nil
```
BenElferink, line 83: If from flag, or auto-generated... Then we should save to secret for next time :)

```suggestion
	var newPass string
	if m.config.AdminPassword != "" {
	    // New password given by user
	    newPass = m.config.AdminPassword
	} else {
	    // Generate a random password using UUID
	    newPass = uuid.New().String()
	}
	
	// Save new password
	_, err := m.client.CoreV1().Secrets(m.ns).Create(ctx, &corev1.Secret{
	    ObjectMeta: metav1.ObjectMeta{
	        Name: k8sconsts.KeycloakSecretName,
	    },
	    Data: map[string][]byte{
	        k8sconsts.KeycloakAdminPasswordKey: []byte(newPass),
	    },
	}, metav1.CreateOptions{})
	
	if err != nil {
	    return "", err
	}
	return newPass, nil
```

BenElferink (approved):

cli/cmd/resources/centralodigos/keycloak.go
alonkeyval: we create it on resources apply

cli/cmd/resources/centralodigos/keycloak.go
alonkeyval: this have issue here when the secret is not exist.
in this case:
existingSecret = nil -> 
`
existingSecret, err := m.client.CoreV1().Secrets(m.ns).Get(ctx, k8sconsts.KeycloakSecretName, metav1.GetOptions{})
// existingSecret = nil 
	if err != nil && !apierrors.IsNotFound(err) { // return false -> !apierrors.IsNotFound(err) is false
		return "", err
	}
//then this part will panic existingSecret.Data[k8sconsts.KeycloakAdminPasswordKey];
	if password, ok := existingSecret.Data[k8sconsts.KeycloakAdminPasswordKey]; ok {
		return string(password), nil
	}
`

PR #4028, opened 2026-01-05
CORE279: add StartupProbe to gateway

natyharary:
Adds a StartupProbe to gateway, which allows up to 5 minutes of delay in startup using a health check endpoint.

Tests done:
Manually delayed startup for 1 minute with a sleep command:
```sh
kubectl patch deployment odigos-gateway -n odigos-system --type='json' -p='[
  {
    "op": "add",
    "path": "/spec/template/spec/initContainers",
    "value": [
      {
        "name": "slow-startup-simulator",
        "image": "busybox",
        "command": ["sh", "-c", "echo Starting slow init... && sleep 60 && echo Init complete"]
      }
    ]
  }
]'
```

Watching the deployment showed that gateway did not restart (startup probe delayed the liveness probe from acting)

I didn't add more e2e tests or unit tests since those will just check that k8s startup probes work as expected, and it is redundant to test the functionaility of an external lib

emergency-ticket id: EMR-800

### DIFF
--- a/autoscaler/controllers/clustercollector/deployment.go
+++ b/autoscaler/controllers/clustercollector/deployment.go
@@ -236,6 +236,16 @@ func getDesiredDeployment(ctx context.Context, c client.Client, enabledDests *od
 								SuccessThreshold: 1,
 								TimeoutSeconds:   5,
 							},
+							StartupProbe: &corev1.Probe{
+								ProbeHandler: corev1.ProbeHandler{
+									HTTPGet: &corev1.HTTPGetAction{
+										Path: "/",
+										Port: intstr.FromInt(13133),
+									},
+								},
+								FailureThreshold: 30,
+								PeriodSeconds:    10,
+							},
 							Resources: corev1.ResourceRequirements{
 								Requests: corev1.ResourceList{
 									corev1.ResourceMemory: requestMemoryQuantity,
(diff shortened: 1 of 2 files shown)


damemi (approved):

tamirdavid1:
what is the motivation for this PR?
i'm not aware of any related issues with the gateway

natyharary:
> what is the motivation for this PR? i'm not aware of any related issues with the gateway

@tamirdavid1  Motivation is described in the linear, https://linear.app/odigos/issue/CORE-279/add-startup-probes-to-all-components, quote here:

> The liveness probe is a recurring issue we see amongst our services. We've tried changing the timeout values but we still see it happening from time to time.
Another possible solution, which we've tried [here](https://github.com/odigos-io/odigos/pull/3522), is to add a startup probe that will indicate that the pod is up and ready for its first request

As far as I understood, we are getting timeouts since our services sometimes take time to boot up and become healthy. Therefore the liveness probes are not enough, we need startup probes to delay the indication for health.
I took the value of `FailureThreshold: 30, PeriodSeconds: 10,` from the odiglet, which roughly translates to 5 minutes grace for the gateway to boot.

natyharary:
Won't do for now - we suspect that startup probes are not the way to solve this at that moment.

PR #4048, opened 2026-01-07
CORE513 add odigoslogsresourceattrsprocessor for extracting servicename (and more!) for logs

natyharary:
Adds a collector called `odigoslogsresourceattrsprocessor` to process attributes for logs.
Mainly `serviceName`, but also `serviceNamespace`, the pod's name and namespace, and the owner's kind.

Tests:
* Unit tests for the processor.
* Manual test of streaming logs from kv-mall apps to a clickhouse installed on a local kind cluster. Checking that these logs contain the new attributes.

emergency-ticket id: EMR-817


collector/processors/odigospartialk8sattrsprocessor/factory.go
```diff
@@ -0,0 +1,60 @@
...
+
+	proc, err := newServiceNameProcessor(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create service name processor: %w", err)
+	}
+
+	// Start the informer
+	if err := proc.start(ctx); err != nil {
+		return nil, fmt.Errorf("failed to start pod metadata informer: %w", err)
+	}
```
damemi: you can use [`exporterhelper.WithStart()`](https://pkg.go.dev/go.opentelemetry.io/collector/exporter/exporterhelper#WithStart) and define a Start function that matches the signature for stuff like this. Also good for getting the inclusterconfig there too

k8sattributesprocessor does this for example: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/7367f5d258f5f59c94eba0125f8831a3046d470e/processor/k8sattributesprocessor/factory.go#L104

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/7367f5d258f5f59c94eba0125f8831a3046d470e/processor/k8sattributesprocessor/processor.go#L60

collector/processors/odigospartialk8sattrsprocessor/factory.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func createLogsProcessor(
+	ctx context.Context,
+	set processor.Settings,
+	cfg component.Config,
+	nextConsumer consumer.Logs) (processor.Logs, error) {
+
+	// Get in-cluster config
+	config, err := rest.InClusterConfig()
```
damemi: I would just initialize this in `newServiceNameProcessor`

collector/processors/odigospartialk8sattrsprocessor/processor.go
```diff
@@ -0,0 +1,98 @@
...
+		podMetadataClient: client,
+	}, nil
+}
+
+func (p *serviceNameProcesser) start(ctx context.Context) error {
+	stopCh := make(chan struct{})
+	go func() {
+		<-ctx.Done()
+		close(stopCh)
+	}()
```
damemi: do you need this?

collector/processors/odigospartialk8sattrsprocessor/processor.go
```diff
@@ -0,0 +1,98 @@
...
+	return p.podMetadataClient.Start(stopCh)
+}
+
+// ownerKindToSemconvKey maps Kubernetes owner kinds to their semconv attribute keys
+var ownerKindToSemconvKey = map[string]string{
+	"ReplicaSet":  string(semconv.K8SReplicaSetNameKey),
+	"DaemonSet":   string(semconv.K8SDaemonSetNameKey),
+	"StatefulSet": string(semconv.K8SStatefulSetNameKey),
+	"Job":         string(semconv.K8SJobNameKey),
+	"CronJob":     string(semconv.K8SCronJobNameKey),
```
damemi: we have consts for the owner kind in k8sconsts: https://github.com/odigos-io/odigos/blob/2ff40d08092c504435dfc685e3d5eaed327116bc/api/k8sconsts/podworkload.go#L25-L31

collector/processors/odigospartialk8sattrsprocessor/processor.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (processor *serviceNameProcesser) processResource(resource pcommon.Resource) {
+	attrs := resource.Attributes()
+	podUIDVal, isPodUIDExist := attrs.Get("k8s.pod.uid")
```
damemi: is there a semconv const for `k8s.pod.uid`?

collector/processors/odigospartialk8sattrsprocessor/internal/kube/client.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create metadata client: %w", err)
+	}
+
+	c := &PodMetadataClient{
+		pods: map[types.UID]*PartialPodMetadata{},
+	}
+
+	factory := metadatainformer.NewSharedInformerFactory(metadataClient, 0)
+	c.informer = factory.ForResource(podGVR).Informer()
```
damemi: I think you can do `factory.Core().V1().Pods().Informer()`. It would be good if we can also filter this to only pods on this collector's node (this is going in the node collector right?)

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
```diff
@@ -0,0 +1,198 @@
...
+	serviceName := extractServiceName(podMeta.OwnerReferences)
+	ownerName, ownerKind := extractOwnerInfo(podMeta.OwnerReferences)
+	c.pods[podMeta.UID] = &PartialPodMetadata{
+		ServiceName: serviceName,
+		Name:        podMeta.Name,
+		Namespace:   podMeta.Namespace,
+		OwnerName:   ownerName,
+		OwnerKind:   ownerKind,
+	}
+}
```
damemi: this is the same as `handlePodAdd`.

but also, a pod's ownerreferences should never update right? But this will still update the map for every update event on the pod. I would just ignore pod update events if there's nothing that can change from it that we care about

collector/processors/odigospartialk8sattrsprocessor/internal/kube/client.go
```diff
@@ -0,0 +1,198 @@
...
+	_, err = c.informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		// Cache functions require "any"
+		AddFunc: func(obj any) {
+			if podMeta := extractPartialMetadata(obj); podMeta != nil {
+				c.handlePodAdd(podMeta)
+			}
+		},
+		UpdateFunc: func(oldObj, newObj any) {
+			if podMeta := extractPartialMetadata(newObj); podMeta != nil {
+				c.handlePodUpdate(podMeta)
```
damemi: I think we can just ignore pod updates see my other comment

collector/processors/odigospartialk8sattrsprocessor/internal/kube/client.go
```diff
@@ -0,0 +1,198 @@
...
+func (c *PodMetadataClient) Start(stopCh <-chan struct{}) error {
+	c.stopCh = make(chan struct{})
+	go c.informer.Run(c.stopCh)
+
+	if stopCh != nil {
+		go func() {
+			<-stopCh
+			c.Stop()
+		}()
+	}
```
damemi: I don't think you need this stopCh loop either

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
```diff
@@ -0,0 +1,198 @@
...
+// extractOwnerInfo extracts the owner name and kind from owner references.
+// Returns empty strings if owner info cannot be determined.
+func extractOwnerInfo(ownerRefs []metav1.OwnerReference) (name, kind string) {
+	if len(ownerRefs) != 1 {
+		return "", ""
+	}
+	return ownerRefs[0].Name, ownerRefs[0].Kind
+}
+
+func (c *PodMetadataClient) handlePodAdd(podMeta *metav1.PartialObjectMetadata) {
```
damemi: maybe we only track pods that are from instrumented owners. You could use our API client to look up if there is an instrumentationconfig for that pod's ownerreference

collector/processors/odigoslogsresourceattrsprocessor/processor.go
```diff
@@ -0,0 +1,98 @@
...
+	if !isPodUIDExist {
+		return
+	}
+
+	podUID := podUIDVal.AsString()
+
+	// Look up pod metadata from the cache (populated by the informer)
+	podMeta, found := processor.podMetadataClient.GetPodMetadata(types.UID(podUID))
+	if !found {
+		// Pod not in cache yet - this can happen briefly after pod creation
```
damemi: there could be some important edge cases here, like a pod is created and quickly crashes, we would want to process those logs. wdyt? wondering if there's a way to handle this race

damemi (changes requested):
good start @natyharary! if you haven't already I would check out the k8sattributes processor code upstream: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor

collector/processors/odigospartialk8sattrsprocessor/factory.go
natyharary: Absolutely

collector/processors/odigospartialk8sattrsprocessor/factory.go
natyharary: Done. I do want to understand the logic in adding this to the `Start` function vs the `newServiceNameProcessor` function - why does it make more sense to put those in the former than the latter?
Is it because we can use the `exporterhelper.WithStart()`?

collector/processors/odigospartialk8sattrsprocessor/processor.go
natyharary: Actually I don't, I can use the context passed here. This was mainly used for aborting the client when needed.

collector/processors/odigospartialk8sattrsprocessor/internal/kube/client.go
natyharary: Yes this is going in the node collector, so I added a filter.
However, since I'm creating a metadata informer (and not a regular informer) I can't use `factory.Core().V1().Pods().Informer()`

BTW, I thought that in general, the kube client I'm using has knowledge scoped down to wherever the component running it.
So I thought that it might be scoped down to the node the collector is running in, by default.
I understand now that's not the case, so it means that this kube client (I'm referring to the metadata client) can get data about the entire cluster?

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
natyharary: Good idea, we really shouldn't track pods that aren't instrumented.
But, if I use our API client, it means that I'll be calling the API each time I'd want to see if there's an instrumentationconfig for that pod owner...
How about adding another informer which will watch for instrumentationconfigs? I am a bit worried about sharing info between two informer instances tho. WDYT?

collector/processors/odigospartialk8sattrsprocessor/internal/kube/client.go
natyharary: Commented on your other comment 😁

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
natyharary: I'm not sure I understood this comment. `handlePodAdd` calls `extractOwnerInfo` to get the ownerreference, where's the logic duplication?

> but also, a pod's ownerreferences should never update right? But this will still update the map for every update event on the pod. I would just ignore pod update events if there's nothing that can change from it that we care about

Got it. I didn't know if owner reference can change or not, apparently it won't 👍
Aren't there any other pod update events that we are interested in? Like, pod going down for example?
If we aren't interested in any pod update events then we can safely remove `handlePodUpdate` completely. WDYT?

collector/processors/odigospartialk8sattrsprocessor/internal/kube/client.go
natyharary: I'm not sure I understand :(
Where did you see a stopCh loop here?

collector/processors/odigospartialk8sattrsprocessor/processor.go
natyharary: I tried that and had some dependency resolve problems, and now it seems that I don't get them. Must have been from some other import.
Added that + used an existing function for determining the owner kind (`GetWorkloadNameAndKind()`).

This does however enlarges the data collector's docker file since those packages need to be copied and be made available for that image. Is that terrible?

collector/processors/odigospartialk8sattrsprocessor/processor.go
natyharary: Good catch - added the semconv

collector/processors/odigoslogsresourceattrsprocessor/processor.go
natyharary: Hacky solution: add some artifical delay (like, 1 minute) to the `handlePodDelete`. Not sleep, since this will stall the thread, but probably some other thing.
This would prevent such pods from being removed too quickly from the cache and allow them to get processed like any other pods.
WDYT?

natyharary (commented):
Thanks for all the comments!!! Replied to them + added some questions to understand some stuff better.

collector/processors/odigospartialk8sattrsprocessor/processor.go
damemi: if you haven't already, check out setting up Go workspaces for your local dev: https://github.com/odigos-io/odigos/blob/main/DEVELOPMENT.md#go-workspace-setup

that can both cause and fix import issues 😛

collector/processors/odigospartialk8sattrsprocessor/factory.go
damemi: it's better ordering and error handling when the collector framework actually initializes the components. By default every component has a nil Start() function, but by doing your real initialization in there any errors can be handled at that time by the collector

collector/processors/odigospartialk8sattrsprocessor/internal/kube/client.go
damemi: Ah I see, sounds good then

And yeah there's nothing scoping the k8s client by default besides the RBAC on the pod that's running it.

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
damemi: `handlePodAdd` and `handlePodDelete` are both calling `extractOwnerInfo` right? But yeah, I don't think there is anything in the pod updates that we'll need to track here, it will just be a lot of noisy events.

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
damemi: the two informers would be fine, they are watching different objects

to keep from doing an api lookup each time (which btw should actually just be doing a cached lookup -- k8s informers build an in memory cache), you could store a map of pods+instrumentationconfigs. That's basically how the upstream k8sattributesprocessor works. Might be a little more performant than the cache client calls

collector/processors/odigoslogsresourceattrsprocessor/processor.go
damemi: hm yeah I think we should avoid a hacky solution too. this is probably fine as is but I wanted to raise the question

damemi (commented):
replied to your comments and resolved ones that were good

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
natyharary: I've reverted back after playing around with two informers, went for an API call.
Adding an odigos informer breaks the library dependencies so I kept it simple. Also I am happy enough with this, since the API call is cached.

collector/processors/odigoslogsresourceattrsprocessor/processor.go
natyharary: If I were instrumenting some pod and it was crashlooping and giving me some logs I'd be interested in knowing about it. I'll add a 30 second delay before removing pods from the cache

collector/processors/odigospartialk8sattrsprocessor/processor.go
natyharary: lol that indeed happened to me! after some issues I had with it I gave up but it does make sense to come back to it.

Other than that, what do you think about enlarging the data collector's docker file? Terrible or not? @damemi

natyharary (commented):
🏓

damemi:
> Adding an odigos informer breaks the library dependencies so I kept it simple. Also I am happy enough with this, since the API call is cached.

can you share more what you mean by this? adding another informer shouldn't break any dependencies, what's the specific error?

> what do you think about enlarging the data collector's docker file?

what do you mean?

collector/processors/odigospartialk8sattrsprocessor/internal/kube/client.go
```diff
@@ -0,0 +1,249 @@
...
+	metadataClient, err := metadata.NewForConfig(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create metadata client: %w", err)
+	}
+
+	// Create dynamic client for InstrumentationConfig access (avoids Odigos API dependency)
+	dynClient, err := dynamic.NewForConfig(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create dynamic client: %w", err)
+	}
```
damemi: we don't actually need the instrumentationconfig client now right? since the logs collection is scoped by collector config to instrumented pods anyway

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
```diff
@@ -0,0 +1,249 @@
...
+	defer c.m.Unlock()
+	c.pods[podMeta.UID] = &PartialPodMetadata{
+		Name:         podMeta.Name,
+		Namespace:    podMeta.Namespace,
+		WorkloadName: workloadName,
+		WorkloadKind: workloadKind,
+	}
+}
+
+func (c *PodMetadataClient) handlePodUpdate(podMeta *metav1.PartialObjectMetadata) {
```
damemi: also still wondering if we actually need to do anything on pod update. I don't think there's any relevant events here but it will trigger on a lot of irrelevant events

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/consts.go
```diff
@@ -0,0 +1,22 @@
...
+	WorkloadKindArgoRollout      WorkloadKind = "Rollout"
+)
+
+// K8SArgoRolloutNameAttribute is the attribute key for Argo Rollout name
+const K8SArgoRolloutNameAttribute = "k8s.argoproj.rollout.name"
+
+// argoRolloutUniqueLabelKey is the default key of the selector that is added
+// to rollout pods. This is used to detect if a pod belongs to an Argo Rollout.
+// Value from: github.com/argoproj/argo-rollouts/pkg/apis/rollouts/v1alpha1
+const argoRolloutUniqueLabelKey = "rollouts-pod-template-hash"
```
damemi, line 24: was importing these from k8sconsts causing the dependency issue?

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/workload.go
```diff
@@ -0,0 +1,56 @@
+package kube
```
damemi, line 1: these should also be available from k8s utils if possible?

damemi (approved):

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/consts.go
natyharary: A comment for future maintainers:
After struggling with importing k8sconsts and k8sutils into this processor (and by that, into the odigos collector), it seems that the problem is that the contrib parts we are using (k8sattributes, prometheus exporter) rely on older `k8s.io` package, and k8sconsts and k8sutils use a new one.
The collector aggregates all dependencies into a single `go.mod` and we can't have two different versions at the same time.

We'll stay with this ugly code duplication until we solve this dep conflict of contrib vs. k8sconsts && k8sutils

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/workload.go
natyharary: A comment for future maintainers:
After struggling with importing k8sconsts and k8sutils into this processor (and by that, into the odigos collector), it seems that the problem is that the contrib parts we are using (k8sattributes, prometheus exporter) rely on older `k8s.io` package, and k8sconsts and k8sutils use a new one.
The collector aggregates all dependencies into a single `go.mod` and we can't have two different versions at the same time.

We'll stay with this ugly code duplication until we solve this dep conflict of contrib vs. k8sconsts && k8sutils

collector/processors/odigospartialk8sattrsprocessor/internal/kube/client.go
natyharary: Absolutely - removed it.

autoscaler/controllers/nodecollector/collectorconfig/common.go
```diff
@@ -27,6 +27,7 @@ const (
 	clusterCollectorMetricsExporterName = "otlp/out-cluster-collector-metrics"
 	clusterCollectorLogsExporterName    = "otlp/out-cluster-collector-logs"
 	resourceDetectionProcessorName      = "resourcedetection"
+	odigosPartialK8sAttrsProcessorName  = "odigospartialk8sattrsprocessor"
```
natyharary: Comments from Amir:
The name does not describe it correctly - it does not related to k8s directly. It has more to do with logs and resource attributes. Change it please

autoscaler/controllers/nodecollector/collectorconfig/ownmetrics-ui.go
```diff
@@ -47,6 +47,7 @@ func init() {
 				consts.OdigosWorkloadNameAttribute,
 			},
 		},
+		odigosPartialK8sAttrsProcessorName: config.GenericMap{},
```
natyharary: Amir: is this neccessary?

collector/Dockerfile
```diff
@@ -2,6 +2,9 @@ FROM --platform=$BUILDPLATFORM golang:1.25-trixie AS builder
 COPY ./collector/ /go/src/collector
 # Need to copy the common folder to the collector for custom connector that uses it
 COPY ./common/ /go/src/common
+# Need to copy api and k8sutils for the odigospartialk8sattrsprocessor
+COPY ./api/ /go/src/api
```
natyharary: Amir: This needs to be removed

collector/builder-config.yaml
```diff
@@ -127,8 +128,11 @@ replaces:
   - github.com/open-telemetry/opentelemetry-collector-contrib/odigos/exporter/mockdestinationexporter => ../exporters/mockdestinationexporter
   - github.com/open-telemetry/opentelemetry-collector-contrib/odigos/processor/odigostrafficmetrics => ../processors/odigostrafficmetrics
   - github.com/odigos-io/odigos/collector/processor/odigosurltemplateprocessor => ../processors/odigosurltemplateprocessor
+  - github.com/odigos-io/odigos/collector/processor/odigospartialk8sattrsprocessor => ../processors/odigospartialk8sattrsprocessor
   - go.opentelemetry.io/collector/confmap/provider/odigosk8scmprovider => ../providers/odigosk8scmprovider
   - github.com/odigos-io/odigos/collector/connectors/odigosrouterconnector => ../connectors/odigosrouterconnector
   - github.com/odigos-io/odigos/collector/receivers/odigosebpfreceiver => ../receivers/odigosebpfreceiver
   - github.com/odigos-io/odigos/common => ../../common
+  - github.com/odigos-io/odigos/api => ../../api
+  - github.com/odigos-io/odigos/k8sutils => ../../k8sutils
```
natyharary: Amir: this also shouldn't be here, we don't use the k8sutils in this processor

collector/odigosotelcol/go.mod
```diff
@@ -756,6 +759,10 @@ replace github.com/odigos-io/odigos/collector/receivers/odigosebpfreceiver => ..
 
 replace github.com/odigos-io/odigos/common => ../../common
 
+replace github.com/odigos-io/odigos/api => ../../api
```
natyharary: Amir: is this needed?

natyharary (commented):
Partial comments from Amir from my user

collector/processors/odigospartialk8sattrsprocessor/processor.go
```diff
@@ -0,0 +1,85 @@
...
+		attrs.PutStr(string(semconv.K8SNamespaceNameKey), podMeta.Namespace)
+	}
+
+	if podMeta.WorkloadName != "" && podMeta.WorkloadKind != "" {
+		if semconvKey, ok := workloadKindToSemconvKey[podMeta.WorkloadKind]; ok {
+			attrs.PutStr(semconvKey, podMeta.WorkloadName)
+		}
+
+		attrs.PutStr(string(semconv.ServiceNameKey), podMeta.WorkloadName)
+		attrs.PutStr(string(semconv.ServiceNamespaceKey), podMeta.Namespace)
```
blumamir: I don't think the k8s namespace belongs in the `service.namespace` attribute.
we don't do it anywhere else at the moment, so I suggest not to add at at this point

collector/processors/odigospartialk8sattrsprocessor/processor.go
```diff
@@ -0,0 +1,85 @@
...
+
+// workloadKindToSemconvKey maps Kubernetes workload kinds to their attribute keys.
+// Note: Argo Rollout uses a custom attribute since there's no semconv key for it.
+var workloadKindToSemconvKey = map[kube.WorkloadKind]string{
+	kube.WorkloadKindDeployment:  string(semconv.K8SDeploymentNameKey),
+	kube.WorkloadKindDaemonSet:   string(semconv.K8SDaemonSetNameKey),
+	kube.WorkloadKindStatefulSet: string(semconv.K8SStatefulSetNameKey),
+	kube.WorkloadKindJob:         string(semconv.K8SJobNameKey),
+	kube.WorkloadKindCronJob:     string(semconv.K8SCronJobNameKey),
+	kube.WorkloadKindArgoRollout: kube.K8SArgoRolloutNameAttribute,
```
blumamir: missing static pod here

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
```diff
@@ -0,0 +1,199 @@
...
+	if podMeta, ok := obj.(*metav1.PartialObjectMetadata); ok {
+		return podMeta
+	}
+	if deleted, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		if podMeta, ok := deleted.Obj.(*metav1.PartialObjectMetadata); ok {
+			return podMeta
+		}
+	}
+	return nil
+}
```
blumamir: What is the purpose of this step?
why aren't we working directly with `corev1.Pod`?

collector/processors/odigospartialk8sattrsprocessor/internal/kube/client.go
```diff
@@ -0,0 +1,199 @@
...
+	_, err = c.podInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		// Cache functions require "any"
+		AddFunc: func(obj any) {
+			if podMeta := extractPartialMetadata(obj); podMeta != nil {
+				c.handlePodAdd(podMeta)
+			}
+		},
+		UpdateFunc: func(oldObj, newObj any) {
+			if podMeta := extractPartialMetadata(newObj); podMeta != nil {
+				c.handlePodUpdate(podMeta)
```
blumamir: I too think we have no action on pod update

collector/processors/odigospartialk8sattrsprocessor/internal/kube/client.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// extractWorkloadInfo resolves the workload name and kind from owner references.
+// Handles ReplicaSet → Deployment/ArgoRollout resolution.
+func extractWorkloadInfo(podMeta *metav1.PartialObjectMetadata) (name string, kind WorkloadKind) {
+	if len(podMeta.OwnerReferences) != 1 {
+		return "", ""
+	}
+
+	ownerRef := podMeta.OwnerReferences[0]
```
blumamir: Odiglet checks all owners, and not assume one owner per pod.

This should be supported by some documentation / code.

collector/processors/odigospartialk8sattrsprocessor/processor.go
natyharary: Right, added it!

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
natyharary: I try to work with the `corev1.PartialObjectMetadata` to conserve memory space, instead of bringing the `corev1.Pod` object which I was under the impression that it takes a larger size in memory.

The `Pod` struct has the same `objectMeta` field as `PartialObjectMetadata`, but also brings things like TypeMeta, PodSpec and PodStatus which we don't need.

collector/processors/odigospartialk8sattrsprocessor/internal/kube/client.go
natyharary: Alright. It is plausible that name-namespace-workloadname-workloadkind don't change during the pod's life. Removed it

collector/processors/odigospartialk8sattrsprocessor/internal/kube/client.go
natyharary: I changed this to work the same as in the Odiglet.

autoscaler/controllers/nodecollector/collectorconfig/ownmetrics-ui.go
natyharary: It isn't!

autoscaler/controllers/nodecollector/collectorconfig/logs.go
```diff
@@ -13,129 +13,6 @@ const (
...
-// 						{"from": "resource_attribute", "name": string(semconv.K8SPodNameKey)},
-// 						{"from": "resource_attribute", "name": string(semconv.K8SNamespaceNameKey)},
-// 					},
-// 				},
-// 			},
-// 		}
-// 	}
-
-// 	return nil
-// }
```
blumamir, line 137: 🥳

autoscaler/controllers/nodecollector/collectorconfig/common.go
```diff
@@ -16,17 +16,18 @@ const (
...
+	pprofExtensionName                   = "pprof"
+	batchProcessorName                   = "batch"
+	memoryLimiterProcessorName           = "memory_limiter"
+	balancerName                         = "round_robin"
+	nodeNameProcessorName                = "resource/node-name"
+	clusterCollectorTracesExporterName   = "otlp/out-cluster-collector-traces"
+	clusterCollectorMetricsExporterName  = "otlp/out-cluster-collector-metrics"
+	clusterCollectorLogsExporterName     = "otlp/out-cluster-collector-logs"
+	resourceDetectionProcessorName       = "resourcedetection"
+	odigosLogsResourceAttrsProcessorName = "odigoslogsresourceattrsprocessor"
```
blumamir: this is specific to the logs, consider declaring it in `logs.go`

autoscaler/controllers/nodecollector/collectorconfig/logs.go
```diff
@@ -193,6 +70,7 @@ func LogsConfig(nodeCG *odigosv1.CollectorsGroup, odigosNamespace string, manife
 		memoryLimiterProcessorName,
 		nodeNameProcessorName,
 		resourceDetectionProcessorName,
+		odigosLogsResourceAttrsProcessorName,
```
blumamir, line 74: maybe I am missing something, but this adds the name to the pipeline. we also need to declare this processor under the `processors:` list in the config yaml, right?

collector/.claude/settings.local.json
blumamir, line 1: was that meant to be added?

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
```diff
@@ -0,0 +1,190 @@
...
+		Namespace:    podMeta.Namespace,
+		WorkloadName: workloadName,
+		WorkloadKind: workloadKind,
+	}
+}
+
+func (c *PodMetadataClient) handlePodDelete(podMeta *metav1.PartialObjectMetadata) {
+	c.m.Lock()
+	defer c.m.Unlock()
+	c.deleteQueue = append(c.deleteQueue, deleteRequest{podUID: podMeta.UID, timestamp: time.Now()})
```
blumamir: here we add items to this list, but where are we removing them and reclaiming the memory?

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
```diff
@@ -0,0 +1,190 @@
...
+	Start(ctx context.Context) error
+}
+
+type deleteRequest struct {
+	podUID    types.UID
+	timestamp time.Time
+}
+
+type PodMetadataClient struct {
+	m           sync.RWMutex
```
blumamir: you have `deleteMut` below, so can be nice to give this one also a meaningful name, to make it more obvious the scope it locks.

also, It's a matter of personal taste, but wondering what you think about calling it `Mutex`, 2 more characters but much more readable ;) and also inline with the type that follows (`sync.Mutex`)

autoscaler/controllers/nodecollector/collectorconfig/common.go
natyharary: I went along with where the rest of the processors are defined - but I can move it to logs as part of tidying up. Is that what you were aiming for?

autoscaler/controllers/nodecollector/collectorconfig/logs.go
natyharary: De-facto it worked without declaring this processer under any yaml.
Moreover, most YAMLs in our codebase are for test data anyway.
I didn't track the code all the way but it does feel like the odigos config eventually piles up to become a YAML used by various components. Doesn't it?

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
natyharary: Makes perfect sense. Wrote it as `m` long before `deleteMut` was added.
In general I am confused on the shorthand variable names in Go as I personally prefer long descriptive names vs short ones but it doesn't seem to be the Go meta

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
natyharary: Removed it :)

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
natyharary: Good catch. I remember adding the queue but not adding the logic that handles it - now in

collector/.claude/settings.local.json
natyharary: Absolutely not.

autoscaler/controllers/nodecollector/collectorconfig/common.go
```diff
@@ -113,7 +114,8 @@ func getCommonExporters(otlpExporterConfiguration *common.OtlpExporterConfigurat
 func init() {
 
 	staticProcessors = config.GenericMap{
-		batchProcessorName: config.GenericMap{},
+		batchProcessorName:                   config.GenericMap{},
+		odigosLogsResourceAttrsProcessorName: config.GenericMap{},
```
blumamir, line 117: only relevant for logs, add it in `logs.go` and not in common (so if there is only traces/metrics, it does not get added)

autoscaler/controllers/nodecollector/collectorconfig/logs.go
blumamir: it's from the `staticProcessors` above (not sure if I missed that in original review or was it added after)

collector/processors/odigoslogsresourceattrsprocessor/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+| Attribute | Description |
+|-----------|-------------|
+| `service.name` | Resolved workload name (e.g., Deployment name) |
+| `k8s.pod.name` | Pod name |
+| `k8s.namespace.name` | Namespace name |
+| `k8s.deployment.name` | Deployment name (if applicable) |
+| `k8s.daemonset.name` | DaemonSet name (if applicable) |
+| `k8s.statefulset.name` | StatefulSet name (if applicable) |
+| `k8s.job.name` | Job name (if applicable) |
```
blumamir, line 19: do we populate job name? or only cron job?

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
blumamir: but obj is already the entire pod in memory, right? the memory has already been allocated so we don't conserve anything here

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
```diff
@@ -0,0 +1,218 @@
...
+
+	c.deleteQueue = remaining
+	c.deleteMutex.Unlock()
+
+	if len(toDelete) > 0 {
+		c.addMutex.Lock()
+		for _, uid := range toDelete {
+			delete(c.pods, uid)
+		}
+		c.addMutex.Unlock()
```
blumamir: the name `addMutex` is not very accurate here :)

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
```diff
@@ -0,0 +1,218 @@
...
+	for _, request := range c.deleteQueue {
+		if now.Sub(request.timestamp) >= podDeletionDelay {
+			toDelete = append(toDelete, request.podUID)
+		} else {
+			remaining = append(remaining, request)
+		}
+	}
+
+	c.deleteQueue = remaining
+	c.deleteMutex.Unlock()
```
blumamir: nit: I think this should ideally be located at the top, right after, the `Lock`, which guarantees that the lock is always released regardless of where and how the function exits

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	// Create filtered informer factory to only watch pods on this node
+	tweakListOptions := func(options *metav1.ListOptions) {
+		options.FieldSelector = fields.OneTermEqualSelector("spec.nodeName", nodeName).String()
+	}
+	factory := metadatainformer.NewFilteredSharedInformerFactory(metadataClient, 0, metav1.NamespaceAll, tweakListOptions)
+	c.podInformer = factory.ForResource(podGVR).Informer()
+
+	_, err = c.podInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
blumamir, line 82: Not sure exactly how it works, but I think the informer populates the cache with objects, which are then sent on the event handler. so in order not to overwhelm the process memory, it is important that we don't store the whole pod in the cache.

Looks like opentelemetry k8sattributeprocessor are [using a `SetTransform` on the informer to do that pre cache storing](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/6b51098ea19a3e60bb4a2807d3fe2b2868c96d64/processor/k8sattributesprocessor/internal/kube/client.go#L233)


```go
	// The TransformFunc is called for each object which is about to be stored.
	//
	// This function is intended for you to take the opportunity to
	// remove, transform, or normalize fields. One use case is to strip unused
	// metadata fields out of objects to save on RAM cost.
	//
	// Must be set before starting the informer.
	//
	// Please see the comment on TransformFunc for more details.
	SetTransform(handler TransformFunc) error
```

should we consider it here as well?

autoscaler/controllers/nodecollector/collectorconfig/common.go
natyharary: @blumamir what do you say?

autoscaler/controllers/nodecollector/collectorconfig/common.go
natyharary: here too, what do you say?

autoscaler/controllers/nodecollector/collectorconfig/logs.go
natyharary: I'm not sure I understood your comment :(
Where is the config yaml you are talking about?

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
natyharary: You are right. Change it to podsMapMutex, and the other mutex to be deleteQueueMutex

collector/processors/odigoslogsresourceattrsprocessor/README.md
natyharary: This is a copy from:

https://github.com/natyharary/odigos/blob/a2d4b0ed95bdaa9a09ddd63bbbbd1cb2269fc40e/k8sutils/pkg/workload/ownerreference.go#L61-L77

But now that I come to look at it again, I think the original code has a bug. The kind case-switch from "Job" leads to `WorkloadKindCronJob`:

```go

case "Job":
		return extractInfoWithSuffix(ownerName, k8sconsts.WorkloadKindCronJob)
		
```
Which does not cover cronjob, and also gives it the wrong kind.

Do you know if this makes sense or not?

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
natyharary: Done.
I do want to say that this isn't the ideal solution in terms of memory coservation YET.
Informer has its own cache, but if we use it, we will lose delayed pod deletion that happens here:

https://github.com/natyharary/odigos/blob/a2d4b0ed95bdaa9a09ddd63bbbbd1cb2269fc40e/collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go#L220-L226

So what I suggest is applying the transform, saving only the partial metadata in the informer's cache, while also keeping an indexed copy as the pods map.
That's why it isn't ideal memory-wise.

The tradeoff here is losing the deletion delay vs. saving some memory in this processor.
For this struct:

```go

type PartialPodMetadata struct {
	UID          types.UID
	Name         string
	Namespace    string
	WorkloadName string       // The resolved workload name (e.g., "my-app" for a Deployment)
	WorkloadKind WorkloadKind // The resolved workload kind (e.g., "Deployment", "DaemonSet")
}

```

I'd estimate, for 1000 objects, that'll be about 80KB for the struct objects themselves and another 330KB (~66bytes per field, * 5), it'll be around 410KB. Not terrible I guess?

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
natyharary: I've applied the transform function before saving to the cache, see previous comment on discussion

collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
natyharary: I added this as a defer to the line after the `.Lock()`, so we know we release the lock regardless on how this function exits

natyharary (commented):
🏓

autoscaler/controllers/nodecollector/collectorconfig/common.go
blumamir: Yes, we should avoid adding this processor for all signals, and scope it only to logs

collector/Dockerfile
```diff
@@ -2,6 +2,7 @@ FROM --platform=$BUILDPLATFORM golang:1.25-trixie AS builder
 COPY ./collector/ /go/src/collector
 # Need to copy the common folder to the collector for custom connector that uses it
 COPY ./common/ /go/src/common
+COPY ./k8sutils/ /go/src/k8sutils
```
blumamir: we ended up not using k8sutils directly. is it still needed?

blumamir (approved):

PR #4073, opened 2026-01-13
feat: added wait for keycloak conditionally

yarden-lang:
emergency-ticket id: EMR-852

### DIFF
--- a/cli/cmd/resources/centralodigos/centralbackend.go
+++ b/cli/cmd/resources/centralodigos/centralbackend.go
@@ -23,7 +23,12 @@ import (
 )
 
 type CentralBackendConfig struct {
-	MaxMessageSize string
+	MaxMessageSize  string
+	WaitForKeycloak WaitForKeycloakConfig
+}
+
+type WaitForKeycloakConfig struct {
+	Enabled bool
 }
 
 type centralBackendResourceManager struct {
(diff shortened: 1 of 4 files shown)


cli/cmd/resources/centralodigos/centralbackend.go
```diff
@@ -97,6 +102,49 @@ func NewCentralBackendDeployment(ns, imagePrefix, imageName, version string, ima
 		})
 	}
 
+	// Keycloak is required for central-backend auth bootstrapping.
+	// Optionally gate startup with an initContainer (useful for installation ordering).
+	// Timeout is intentionally hard-coded (60s) to keep CLI behavior simple and predictable.
+	const (
+		waitForKeycloakImage         = "busybox:1.36.1"
```
alonkeyval: what is this image? look like the deployment use different image: 
KeycloakImage               = "quay.io/keycloak/keycloak:24.0.3"

cli/cmd/resources/centralodigos/centralbackend.go
```diff
@@ -97,6 +102,49 @@ func NewCentralBackendDeployment(ns, imagePrefix, imageName, version string, ima
...
+	// Optionally gate startup with an initContainer (useful for installation ordering).
+	// Timeout is intentionally hard-coded (60s) to keep CLI behavior simple and predictable.
+	const (
+		waitForKeycloakImage        = "busybox:1.36.1"
+		waitForKeycloakTimeoutSecs  = 60
+		waitForKeycloakPollInterval = 2
+	)
+	keycloakURL := fmt.Sprintf("http://%s:%d/realms/master", k8sconsts.KeycloakServiceName, k8sconsts.KeycloakPort)
+
+	waitForKeycloakScript := fmt.Sprintf(`#!/bin/sh
```
alonkeyval, line 115: move this to const file

cli/cmd/resources/centralodigos/centralbackend.go
```diff
@@ -23,7 +23,12 @@ import (
...
 
 type CentralBackendConfig struct {
-	MaxMessageSize string
+	MaxMessageSize  string
+	WaitForKeycloak WaitForKeycloakConfig
+}
+
+type WaitForKeycloakConfig struct {
+	Enabled bool
 }
```
alonkeyval, line 32: looking at the code, I think using startup probe is a better approach here: 
1. we now have 3 places that chack the same logic for keyclock health 
2. we controll the central backend code 
3. using initContainer will make us pull another image (busybox) 
4. we already have the logic on the central backend 
5. using the init container is less visibillty for my opintion, it will make the central pod in init status with out any additional logs

cli/cmd/resources/centralodigos/centralbackend.go
yarden-lang: this is a linux small container for running sh commands

PR #4120, opened 2026-01-22
chore: changing the value of MetricValue to be 1024

tamirdavid1:
Refactor eBPF metrics processing to read resource attributes from map key=0 instead of processKey parameter. Implemented two-phase processing where attributes are extracted from key=0 first, then metrics are processed from remaining keys, with proper separation of concerns between reusable attribute extraction and JVM-specific metric handling.

emergency-ticket id: EMR-906


aviadhahami:
we've decided to go another way; closing

PR #4131, opened 2026-01-25
CORE542: refactor the rollout reconciliation flow + concurrent rollouts limiting

natyharary:
1. Refactor the rollout function so we check for configuration related things at the beginning of `Do()`, making it more readable
2. Aggregate rollout conditions to a single constants file
3. Add a concurrent rollout limiter for rollouts which is configurable via the Odigos configuration and the Helm `values.yaml`
4. A heap-ton of tests

Important notes:
1. The ticket asks refactoring the rollback/autoheal - this PR is huge already so I'll add that in a follow-up
2. The ticket asks for persistent instrumentation (in case the instrumentor is not up) - not in this PR, will be in a follow up

emergency-ticket id: EMR-918


helm/odigos/values.yaml
```diff
@@ -572,6 +572,11 @@ rollout:
 # will still have the agent injected, regardless of this setting.
 # When set to true, all additional configurations related to automated rollouts or rollbacks are ignored.
   automaticRolloutDisabled: false
+  # ConcurrentRollouts is the maximum number of concurrent rollouts allowed.
+  # The default of 5 is selected for a cluster of 200 workloads with an average of 5 minute rollout time.
+  # This means instrumentation of the entire cluster will take 3.33 hours.
+  isConcurrentRolloutsLimiterEnabled: false
```
tamirdavid1: I think better to keep it as 1 commented `concurrentRollouts` 
when it is nil the default behavior is "without limit" if user uncomment he can play with the concurrentRollouts.
Also maybe worth adding in the comment that the rollout is post instrumentation?

tamirdavid1: after discussion f2f, maybe better to keep it with 0 which means no limit and and number will set the limit?
by that you wont need the NoLimit Const

instrumentor/controllers/agentenabled/manager.go
```diff
@@ -8,11 +8,15 @@ import (
...
 	odigosv1 "github.com/odigos-io/odigos/api/odigos/v1alpha1"
 	"github.com/odigos-io/odigos/distros"
+	"github.com/odigos-io/odigos/instrumentor/controllers/agentenabled/rollout"
 	instrumentorpredicate "github.com/odigos-io/odigos/instrumentor/controllers/utils/predicates"
 	odigospredicate "github.com/odigos-io/odigos/k8sutils/pkg/predicate"
 )
 
 func SetupWithManager(mgr ctrl.Manager, dp *distros.Provider) error {
+	// Create the limiter - it will be initialized with config on first use in Do()
+	rolloutConcurrencyLimiter := rollout.NewRolloutConcurrencyLimiter()
```
tamirdavid1: better to put this in the reconcile all? it will make this file cleaner and also you will have the configuration so maybe you wont need to set default to 0 and then change it?

tests/e2e/instrumentation-lifecycle/assert-uninstrumented-rollouts.yaml
```diff
@@ -0,0 +1,50 @@
+# Workload Uninstrumented Assertions
```
tamirdavid1: Delete this file

instrumentor/controllers/agentenabled/rollout/constants.go
```diff
@@ -0,0 +1,99 @@
...
+		Type:    odigosv1alpha1.WorkloadRolloutStatusConditionType,
+		Status:  metav1.ConditionTrue,
+		Reason:  string(odigosv1alpha1.WorkloadRolloutReasonTriggeredSuccessfully),
+		Message: message,
+	}
+}
+
+// ConditionWaitingInQueue is used when the workload is waiting for other rollouts to complete
+// due to rate limiting of concurrent reconciliations
+var ConditionWaitingInQueue = metav1.Condition{
```
blumamir: consider placing it up with other constants

helm/odigos/values.yaml
```diff
@@ -572,6 +572,11 @@ rollout:
...
+  concurrentRollouts: 5
```
blumamir: As discussed in person, I wonder if this way is most straight forward to users to configure how to do rollout.

Since we have 3 states: no rollout (disabled), concurent and all-in-once, we can consider making it explicit in the user facing value, instead of having it implicit from 2 boolean sets

helm/odigos/values.yaml
```diff
@@ -572,6 +572,11 @@ rollout:
...
```
blumamir, line 966: consider: `maxConcurrentRollouts`, to make it read a bit better

instrumentor/controllers/agentenabled/rollout/conditions.go
```diff
@@ -0,0 +1,99 @@
+package rollout
```
blumamir, line 1: this file is actually about conditions, and contains constants but also functions.
consider calling it `conditions.go`

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
```diff
@@ -0,0 +1,116 @@
...
+	mutex              sync.Mutex
+	logger             logr.Logger
+	concurrentRollouts int
+	// set of workload keys currently rolling out - we use a map for fast lookup,
+	// and a struct{} for the value to avoid allocating memory for the value
+	inFlightRollouts map[string]struct{}
+}
+
+func NewRolloutConcurrencyLimiter() *RolloutConcurrencyLimiter {
+	logger := log.FromContext(context.Background()).WithName("RolloutConcurrencyLimiter")
```
blumamir: Should we pass the logger as parameter here? since we don't actually extract it from the context

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (r *RolloutConcurrencyLimiter) ApplyConfig(conf *common.OdigosConfiguration) {
+	if conf != nil && conf.Rollout != nil && conf.Rollout.IsConcurrentRolloutsLimiterEnabled != nil && *conf.Rollout.IsConcurrentRolloutsLimiterEnabled {
+		if conf.Rollout.ConcurrentRollouts > 0 {
+			r.concurrentRollouts = conf.Rollout.ConcurrentRollouts
+		} else {
+			r.concurrentRollouts = DefaultConcurrentRollouts
+		}
+	}
+}
```
blumamir: This function can be called from multiple go routines, should we syncrhonize the access to the shared memory?

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
```diff
@@ -0,0 +1,116 @@
...
+	logger := log.FromContext(context.Background()).WithName("RolloutConcurrencyLimiter")
+	return &RolloutConcurrencyLimiter{
+		logger:             logger,
+		concurrentRollouts: NoConcurrencyLimiting,
+		inFlightRollouts:   make(map[string]struct{}),
+	}
+}
+
+func (r *RolloutConcurrencyLimiter) ApplyConfig(conf *common.OdigosConfiguration) {
+	if conf != nil && conf.Rollout != nil && conf.Rollout.IsConcurrentRolloutsLimiterEnabled != nil && *conf.Rollout.IsConcurrentRolloutsLimiterEnabled {
```
blumamir: should we check if if the rollout is disabled?

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
```diff
@@ -0,0 +1,116 @@
...
+		return true
+	}
+
+	// Already has a slot for this workload (from a previous reconcile that triggered rollout)
+	if _, exists := r.inFlightRollouts[workloadKey]; exists {
+		return true
+	}
+
+	// Check if under limit
+	if len(r.inFlightRollouts) < r.concurrentRollouts {
```
blumamir: The property `concurrentRollouts` make it sound as if it's the current count of active rollouts.
from the code I understand this is this the maximum allowed. consider givin it a name to make it obvious.

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
```diff
@@ -0,0 +1,116 @@
...
+		r.logger.V(2).Info("Acquired rollout slot", "workload", workloadKey, "inFlight", len(r.inFlightRollouts), "limit", r.concurrentRollouts)
+		return true
+	}
+
+	r.logger.V(2).Info("Rollout slot denied - at capacity", "workload", workloadKey, "inFlight", len(r.inFlightRollouts), "limit", r.concurrentRollouts)
+	return false
+}
+
+// HasSlot returns true if the workload currently holds a slot
+func (r *RolloutConcurrencyLimiter) HasSlot(workloadKey string) bool {
```
blumamir: `HasSlot` can be ambigious.

the property is `inFlightRollouts` so you can call it `IsWorkloadRolloutInFlight`

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -38,141 +49,124 @@ const requeueWaitingForWorkloadRollout = 10 * time.Second
...
 	newRolloutHash := ic.Spec.AgentsMetaHash
 	if savedRolloutHash == newRolloutHash {
 		rolloutDone := utils.IsWorkloadRolloutDone(workloadObj)
-		if !rolloutDone && !rollbackDisabled {
+		if rolloutDone {
+			// Rollout is complete - release the slot if we had one
+			if rolloutConcurrencyLimiter.HasSlot(workloadKey) {
+				logger.V(3).Info("rollout complete, releasing slot", "workload", pw.Name, "namespace", pw.Namespace)
+				rolloutConcurrencyLimiter.Release(workloadKey)
+			}
```
blumamir: There is a race condition here, from the time of the check (`HasSlot`) to time of act `Release`.
While it's not major in this specific case, it's an anti pattern, and better to have `Release` return a boolean to denote if it was actually deleted or not with one mutex lock

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
```diff
@@ -0,0 +1,116 @@
...
+	NoConcurrencyLimiting     = 0
+)
+
+type RolloutConcurrencyLimiter struct {
+	mutex              sync.Mutex
+	logger             logr.Logger
+	concurrentRollouts int
+	// set of workload keys currently rolling out - we use a map for fast lookup,
+	// and a struct{} for the value to avoid allocating memory for the value
+	inFlightRollouts map[string]struct{}
```
blumamir, line 18: The content of this map is internal to the class, and there is no visibility for which sources are currently rolling out.
For us troubleshooting issues, and for visibility - please consider how we can show this list in UI

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func NewRolloutConcurrencyLimiter() *RolloutConcurrencyLimiter {
+	logger := log.FromContext(context.Background()).WithName("RolloutConcurrencyLimiter")
+	return &RolloutConcurrencyLimiter{
+		logger:             logger,
+		concurrentRollouts: NoConcurrencyLimiting,
+		inFlightRollouts:   make(map[string]struct{}),
```
blumamir: What if the instrumentor is restarted and there are currently in-flight rollouts in the cluster? I would expect this to be set to the actual number in this case

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -38,141 +49,124 @@ const requeueWaitingForWorkloadRollout = 10 * time.Second
 // If a rollout is triggered the status of the instrumentation config is updated with the new rollout hash
 // and a corresponding condition is set.
 //
-// Returns a boolean indicating if the status of the instrumentation config has changed, a ctrl.Result and an error.
-func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.InstrumentationConfig, pw k8sconsts.PodWorkload, conf *common.OdigosConfiguration, distroProvider *distros.Provider) (bool, ctrl.Result, error) {
+// Returns a RolloutResult and an error.
+func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.InstrumentationConfig, pw k8sconsts.PodWorkload, conf *common.OdigosConfiguration, distroProvider *distros.Provider, rolloutConcurrencyLimiter *RolloutConcurrencyLimiter) (RolloutResult, error) {
+	rolloutConcurrencyLimiter.ApplyConfig(conf)
```
blumamir: Instead of calling it on every reconcile, we can consider only when the config actually changes:

https://github.com/odigos-io/odigos/blob/de2df355a657ae757a738ddd3a61bc0abcc2714b/instrumentor/controllers/agentenabled/effectiveconfig_controller.go#L16

instrumentor/controllers/agentenabled/rollout/constants.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	odigosv1alpha1 "github.com/odigos-io/odigos/api/odigos/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Predefined rollout condition states - these represent all discrete states for workload rollout
+var (
+	// ConditionNotRequired is used when the selected instrumentation distributions do not require application restart
+	ConditionNotRequired = metav1.Condition{
```
blumamir: those are private to the package, right?
consider having them as lowercase

instrumentor/controllers/agentenabled/rollout/rollout_config_helpers.go
```diff
@@ -0,0 +1,68 @@
...
+		d := distroProvider.GetDistroByName(containerConfig.OtelDistroName)
+		if d == nil {
+			continue
+		}
+		if distro.IsRestartRequired(d, conf) {
+			hasDistributionThatRequiresRollout = true
+		}
+	}
+	return hasDistributionThatRequiresRollout
+}
```
blumamir: this code has been modified in #3995
please rebase

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -38,141 +49,124 @@ const requeueWaitingForWorkloadRollout = 10 * time.Second
...
-		return statusChanged, ctrl.Result{}, nil
+		changed := meta.SetStatusCondition(&ic.Status.Conditions, ConditionWaitingForJobTrigger)
+		return RolloutResult{StatusChanged: changed}, nil
+	}
+
+	// Check if the distribution requires a rollout (only when ic is not nil)
+	if ic != nil && !isRolloutDistro(ic, distroProvider, conf) {
+		changed := meta.SetStatusCondition(&ic.Status.Conditions, ConditionNotRequired)
+		return RolloutResult{StatusChanged: changed}, nil
 	}
```
blumamir: Why is this here? I see it also below

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -38,141 +49,124 @@ const requeueWaitingForWorkloadRollout = 10 * time.Second
...
 		rolloutDone := utils.IsWorkloadRolloutDone(workloadObj)
-		if !rolloutDone && !rollbackDisabled {
+		if rolloutDone {
+			// Rollout is complete - release the slot if we had one
+			if rolloutConcurrencyLimiter.HasSlot(workloadKey) {
+				logger.V(3).Info("rollout complete, releasing slot", "workload", pw.Name, "namespace", pw.Namespace)
+				rolloutConcurrencyLimiter.Release(workloadKey)
+			}
+			statusChanged := meta.SetStatusCondition(&ic.Status.Conditions, rolloutCondition(nil))
+			return RolloutResult{StatusChanged: statusChanged}, nil
```
blumamir: don't we need to requeue here to test for auto-heal?
the code prior to your changes requeued in this case and now we do not anymore

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
natyharary: Good idea.

helm/odigos/values.yaml
natyharary: Changed it to only be an int, which will inferred as unlimited rollouts (=no limit) or limited rollouts

helm/odigos/values.yaml
natyharary: Agreed - moved to a single boolean and an int

helm/odigos/values.yaml
natyharary: Done

instrumentor/controllers/agentenabled/rollout/constants.go
natyharary: 👍

instrumentor/controllers/agentenabled/rollout/conditions.go
natyharary: Agreed

tests/e2e/instrumentation-lifecycle/assert-uninstrumented-rollouts.yaml
natyharary: 👍

instrumentor/controllers/agentenabled/rollout/rollout.go
natyharary: Got it - it didn't click for me that the odigos configuration becomes the effective config! So it makes perfect sense to move it there.

instrumentor/controllers/agentenabled/manager.go
natyharary: Done.

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
natyharary: Good idea.

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
natyharary: The `Do()` function checks for this configuration before we call the limiter, so we never get to "use" the limiter if automatic rollouts are disabled.
Either way, the limiter decides if we reached the amount of concurrent rollouts, not whether we even rollout at all.

Moreso, there's no neat way to add this check here - what would the config be, if we have automatic rollouts disabled?
By default it is 0 (=unlimited rollouts), will we set it to... a negative integer? Raise an error?

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
natyharary: Now that we call this function from the `reconcileAll()` in `sync.go`, is this still relevant? I don't understand how multiple goroutines call it

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
natyharary: In this case, we would probably have an overload of in-flight rollouts with the maximum value of 2 * `maxConcurrentRollouts` (if we had the maximum amount of rollouts before in-flight). I also expect we'll get a higher multiplier if the instrumentor keeps restarting.
Is this a plausible situation?
The only way to mitigate this is to store this as some sort of state, and I'm not sure how to persist this since we don't have any volumes available to the instrumentor AFAIK.
WDYT?

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
natyharary: I'm not sure I understood - this line is indeedly extracting it from the context.. or I don't get something in this syntax

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
natyharary: The string represented here is built here (`rollout.go`):

```go

// WorkloadKey generates a unique key for rate limiting purposes
func WorkloadKey(pw k8sconsts.PodWorkload) string {
	return fmt.Sprintf("%s/%s/%s", pw.Namespace, pw.Kind, pw.Name)
}

```

So it would be like `mynamespaces/deployment/myname`
Do you think that's descriptive enough for the UI? It looks OK to me

instrumentor/controllers/agentenabled/rollout/rollout.go
natyharary: This was like this in the code before:

```go

	if ic.Spec.PodManifestInjectionOptional {
		// all distributions used by this workload do not require a restart
		// thus, no rollout is needed
		statusChanged := meta.SetStatusCondition(&ic.Status.Conditions, metav1.Condition{
			Type: odigosv1alpha1.WorkloadRolloutStatusConditionType,
			// currently we interpret False as an error state, so we use True here to indicate a healthy state
			// it might be confusing since rollout is not actually done, but this is the closest match given the k8s condition semantics.
			Status:  metav1.ConditionTrue,
			Reason:  string(odigosv1alpha1.WorkloadRolloutReasonNotRequired),
			Message: "The selected instrumentation distributions do not require application restart",
		})
		return statusChanged, ctrl.Result{}, nil
	}

```

it moved a bit up because I moved the configuration checks to the first line of `Do()`.

I'll also change the condition name, it is a bit misleading.

Where do you see this below too?

instrumentor/controllers/agentenabled/rollout/rollout.go
natyharary: After talking F2F - I ge it. We don't even need the check (HasSlot)

instrumentor/controllers/agentenabled/rollout/rollout.go
natyharary: If the rollout is complete then we don't need to test for auto-heal, so we don't need to requeue.

HOWEVER
I do think that this check is redundant here now, since if we don't enter `if !rolloutDone && !rollBackOptions.IsRollbackDisabled ` (I added `!rolloutDone` for clarity - as it was implicitly else-ed), we'll get to the code you've added below:

```go

		// at this point, we know that rollout happened.
		// make sure to cleanup any stale condition from previous rollout attempts.
		// example:
		// - workload is instrumented successfully.
		// - rollout regardless of odigos
		// - agent disabled -> status is changed to "waiting for previous rollout to finish"
		// - agent enabled -> no need for a rollout, but the status needs to be updated.
		statusChanged := false
		if rolloutDone {
			statusChanged = meta.SetStatusCondition(&ic.Status.Conditions, rolloutCondition(nil))
		}
		return RolloutResult{StatusChanged: statusChanged, Result: ctrl.Result{}}, nil

```

So I'll just add the release slot logic there.

instrumentor/controllers/agentenabled/rollout/rollout_config_helpers.go
natyharary: got it, removed

instrumentor/controllers/agentenabled/rollout/conditions.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// ConditionWaitingInQueue is used when the workload is waiting for other rollouts to complete
+// due to rate limiting of concurrent reconciliations
+var conditionWaitingInQueue = metav1.Condition{
+	Type:    odigosv1alpha1.WorkloadRolloutStatusConditionType,
+	Status:  metav1.ConditionTrue,
+	Reason:  string(odigosv1alpha1.WorkloadRolloutReasonWaitingInQueue),
+	Message: "Waiting for other workload rollouts to complete",
+}
```
blumamir: consider moving it up to other constants

instrumentor/controllers/agentenabled/rollout/conditions.go
```diff
@@ -0,0 +1,99 @@
...
+// NewConditionTriggeredWithMessage creates a triggered successfully condition with a custom message.
+// Used for rollback scenarios where the message contains backoff details.
+func NewConditionTriggeredWithMessage(message string) metav1.Condition {
+	return metav1.Condition{
+		Type:    odigosv1alpha1.WorkloadRolloutStatusConditionType,
+		Status:  metav1.ConditionTrue,
+		Reason:  string(odigosv1alpha1.WorkloadRolloutReasonTriggeredSuccessfully),
+		Message: message,
+	}
+}
```
blumamir: these (and the function below) should also start with lower case I think, as they are internal to the packahe

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
blumamir: When I look in the UI - how can I tell which are the 5 (or something else) that are currently being rolled out? how can we troubleshoot this if there are any issues?

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
blumamir: you use `context.Background()` so we don't actually extract anything from the context

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
blumamir: Yes, the `sync.go` function can be run from multiple go routines - for example from multiple controllers. I think it should be synced (or we should pass the limit to `TryAcquire` as parameter)

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
```diff
@@ -0,0 +1,104 @@
...
+		return true
+	}
+
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+
+	// No rate limiting configured
+	if r.maxConcurrentRollouts == NoConcurrencyLimiting {
+		return true
+	}
```
blumamir: do we need to lock if the `maxConcurrentRollouts` is 0?

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
```diff
@@ -0,0 +1,104 @@
...
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+
+	if _, exists := r.inFlightRollouts[workloadKey]; !exists {
+		if r.maxConcurrentRollouts == NoConcurrencyLimiting {
+			// Not an error - workload may not have had a slot (e.g., no rate limiting)
+			r.logger.V(2).Info("Workload does not have a slot - rate limiting is disabled", "workload", workloadKey)
+		} else {
+			// This should not happen, but we log it for debugging purposes
+			r.logger.V(2).Info("Workload does not have a slot - this should not happen, rate limiting is enabled", "workload", workloadKey)
```
blumamir: should this be an error then? if should not happen, or at least `V(0)`?

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
```diff
@@ -0,0 +1,104 @@
...
+		return
+	}
+
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+
+	if _, exists := r.inFlightRollouts[workloadKey]; !exists {
+		if r.maxConcurrentRollouts == NoConcurrencyLimiting {
+			// Not an error - workload may not have had a slot (e.g., no rate limiting)
+			r.logger.V(2).Info("Workload does not have a slot - rate limiting is disabled", "workload", workloadKey)
```
blumamir: does this log give real value?

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -230,20 +200,28 @@ func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.Instrumentation
 		statusChanged := false
 		if rolloutDone {
 			statusChanged = meta.SetStatusCondition(&ic.Status.Conditions, rolloutCondition(nil))
+			// Rollout is complete - release the slot if we had one
+			rolloutConcurrencyLimiter.ReleaseWorkloadRolloutSlot(workloadKey)
```
blumamir, line 172: This part is extremely important (without it we would leak the the available slots and get into deadlock).

Are we sure that all cases are covered? what if someone uninstrument while we are activly rolling out? what if the rollout hash changes? what if the agent becomes disabled during a rollout?

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
blumamir: We can store it on the `instrumentationconfig` object in api server

instrumentor/controllers/agentenabled/rollout/rollout.go
blumamir: It was twice in the relevant commit, but I see now it's refactored to happen only once

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
natyharary: Yeah actually it makes sense to lower it, it is an important enough message

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
natyharary: It is there for debugging, that's why it is in V(2). Helped me to debug and see the limit happening, might be useful in the future as well

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
natyharary: ~Will add a rollout status~

We've talked about it - we've decided to keep the updates that'll flow to the UI via the condition updates

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
natyharary: I'm not sure. On the one hand, I agree that it is redundant since we immediately return, but if we don't lock, can't the `limit` change?

I've read that mutex locking/unlocking in Go is pretty performant, and there's some discussion here about how performant:
https://stackoverflow.com/questions/57562606/why-does-sync-mutex-largely-drop-performance-when-goroutine-contention-is-more-t

Around 2400 goroutines we get a performance of ~46ns per operation and on 4800 goroutines it is ~317 ns per operation.
I don't believe we encounter this volume on a single instrumentor at the same time and even on the slower performance, we still won't delay the reconcile very much. WDYT?

instrumentor/controllers/agentenabled/rollout/rollout.go
natyharary: We've talked about it and reviewed the code f2f - basically we release the slots on rollbacks and on succesful rollout. If you see more cases where this is not treated please let me know.

frontend/graph/status/rollout.go
```diff
@@ -21,6 +21,8 @@ func workloadRolloutStatusCondition(reason *string) model.DesiredStateProgress {
 		return model.DesiredStateProgressUnknown // rollout is disabled in config, or agent was not enabled etc. need to refine those cases in the future.
 	case v1alpha1.WorkloadRolloutReasonWaitingForRestart:
 		return model.DesiredStateProgressIrrelevant // rollout state is irrelevant in this case (no rollout for cronjob for example)
+	case v1alpha1.WorkloadRolloutReasonWaitingInQueue:
```
blumamir, line 28: make sure this is synchronized with new condition

instrumentor/controllers/agentenabled/manager.go
```diff
@@ -25,8 +28,9 @@ func SetupWithManager(mgr ctrl.Manager, dp *distros.Provider) error {
 			),
 		)).
 		Complete(&CollectorsGroupReconciler{
-			Client:          mgr.GetClient(),
-			DistrosProvider: dp,
+			Client:                    mgr.GetClient(),
+			DistrosProvider:           dp,
+			RolloutConcurrencyLimiter: rolloutConcurrencyLimiter,
```
blumamir, line 33: should be added in all controllers

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// Already has a slot for this workload (from a previous reconcile that triggered rollout)
+	if _, exists := r.inFlightRollouts[workloadKey]; exists {
+		return true
+	}
+
+	// Check if under limit
+	if len(r.inFlightRollouts) < limit {
+		r.inFlightRollouts[workloadKey] = struct{}{}
+		r.logger.V(2).Info("Acquired rollout slot", "workload", workloadKey, "inFlight", len(r.inFlightRollouts), "limit", r.maxConcurrentRollouts)
```
blumamir: replace `maxConcurrentRollouts` with `limit`

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+const (
+	NoConcurrencyLimiting = 0
+)
+
+type RolloutConcurrencyLimiter struct {
+	mutex                 sync.Mutex
+	logger                logr.Logger
+	maxConcurrentRollouts int
```
blumamir: remove

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// TryAcquire attempts to acquire a slot for the given workload to allow rollout to proceed.
+func (r *RolloutConcurrencyLimiter) TryAcquire(workloadKey string, limit int) bool {
+	if r == nil {
+		return true
+	}
+
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
```
blumamir: can be moved down to only lock after `limit == NoConcurrencyLimitin`

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -27,7 +26,20 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-const requeueWaitingForWorkloadRollout = 10 * time.Second
+const RequeueWaitingForWorkloadRollout = 10 * time.Second
+
+type RolloutResult struct {
+	StatusChanged bool
+	// Used for acquiring lock on the rollout concurrency limiter
+	WorkloadKey string
```
blumamir: seems unused, check if we should keep it

instrumentor/controllers/agentenabled/rollout/rollout_config_helpers.go
```diff
@@ -0,0 +1,58 @@
...
+
+// GetRolloutAndRollbackOptions extracts rollout and rollback configuration from OdigosConfiguration.
+// Returns an error if the configuration contains invalid duration strings.
+func getRolloutAndRollbackOptions(conf *common.OdigosConfiguration) (isAutomaticRolloutDisabled bool, rollBackOptions RollBackOptions, err error) {
+	isAutomaticRolloutDisabled = conf.Rollout != nil && conf.Rollout.AutomaticRolloutDisabled != nil && *conf.Rollout.AutomaticRolloutDisabled
+
+	isRollbackDisabled := conf.RollbackDisabled != nil && *conf.RollbackDisabled
+
+	defaultRollbackGraceTime, _ := time.ParseDuration(consts.DefaultAutoRollbackGraceTime)
```
blumamir: check if we can define the const as duration

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -38,186 +50,110 @@ const requeueWaitingForWorkloadRollout = 10 * time.Second
...
 
 		// instrumentation config is deleted, trigger a rollout for the associated workload
 		// this should happen once per workload, as the instrumentation config is deleted
 		// and we want to rollout the workload to remove the instrumentation
+		// Note: uninstrumentation rollouts are not rate limited since we can't track completion
+		// (the IC is deleted so we won't get subsequent reconciles)
+		logger.V(2).Info("proceeding with uninstrumentation rollout",
+			"workload", pw.Name,
+			"namespace", pw.Namespace)
 		rolloutErr := rolloutRestartWorkload(ctx, workloadObj, c, time.Now())
```
blumamir, line 110: we should make sure we release the slot for concurrent here, to avoid deadlocking

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -38,186 +50,110 @@ const requeueWaitingForWorkloadRollout = 10 * time.Second
...
-		}
-	}
-
-	rollbackStabilityWindow, _ := time.ParseDuration(consts.DefaultAutoRollbackStabilityWindow)
-	if conf.RollbackStabilityWindow != "" {
-		rollbackStabilityWindow, err = time.ParseDuration(conf.RollbackStabilityWindow)
-		if err != nil {
-			return false, ctrl.Result{}, fmt.Errorf("invalid duration %q: %w", rollbackGraceTime, err)
-		}
+		changed := meta.SetStatusCondition(&ic.Status.Conditions, conditionRestartNotRequiredForDistro)
```
blumamir, line 118: should release the slot here, in case it's locked

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -38,186 +50,110 @@ const requeueWaitingForWorkloadRollout = 10 * time.Second
...
 		// and signal that the process is considered completed.
 		// If manual rollout is required, we can mention this for better UX.
-		statusChanged := meta.SetStatusCondition(&ic.Status.Conditions, metav1.Condition{
-			Type:    odigosv1alpha1.WorkloadRolloutStatusConditionType,
-			Status:  metav1.ConditionTrue, // this might not be a success, need to refine into multiple discrete states
-			Reason:  string(odigosv1alpha1.WorkloadRolloutReasonDisabled),
-			Message: "odigos automatic rollout is disabled",
-		})
-		return statusChanged, ctrl.Result{}, nil
+		changed := meta.SetStatusCondition(&ic.Status.Conditions, conditionRolloutDisabled)
```
blumamir, line 128: check if we need to release the slots

frontend/graph/status/rollout.go
natyharary: Done. What do you think about adding 

```go

case v1alpha1.WorkloadRolloutReasonNotRequired:
		return model.DesiredStateProgressSuccess // rollout is not required

```

It wasn't there in the first place, but I added it since the condition didn't reflect here

instrumentor/controllers/agentenabled/manager.go
natyharary: Done

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
natyharary: Done

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
natyharary: Done

instrumentor/controllers/agentenabled/rollout/rollout_concurrency_limiter.go
natyharary: Done

instrumentor/controllers/agentenabled/rollout/rollout.go
natyharary: Unused indeed. Removed it

instrumentor/controllers/agentenabled/rollout/rollout_config_helpers.go
natyharary: Done

instrumentor/controllers/agentenabled/rollout/rollout.go
natyharary: Done

instrumentor/controllers/agentenabled/rollout/rollout.go
natyharary: Done

instrumentor/controllers/agentenabled/rollout/rollout.go
natyharary: Adding this comment for explanation:
It depends; if `isAutomaticRolloutDisabled` was set before starting instrumentation of any of the pods, then it is redundant to release the slots since we never get to `TryAcquire()`.

The only situation where this is needed is if `isAutomaticRolloutDisabled` was set mid-rollout (maybe in order to stop stuff rolling out like crazy?) so we don't choke existing automatic rollouts in the queue.
However, in this situation, all the slots would be released pretty much immediately, rendering the queue useless.

I think it is better to release everything and letting things be operational vs. not allowing existing automatic rollouts to advance, even in an expediated rate.
(@blumamir feel free to share you thhoughts on this)

api/odigos/v1alpha1/instrumentationconfig_types.go
```diff
@@ -114,17 +114,19 @@ const (
 	AgentEnabledReasonImagePullBackOff AgentEnabledReason = "ImagePullBackOff"
 )
 
-// +kubebuilder:validation:Enum=RolloutTriggeredSuccessfully;FailedToPatch;PreviousRolloutOngoing;Disabled;WaitingForRestart;WorkloadNotSupporting
+// +kubebuilder:validation:Enum=RolloutTriggeredSuccessfully;FailedToPatch;PreviousRolloutOngoing;Disabled;WaitingForRestart;WorkloadNotSupporting;NotRequired;WaitingInQueue
```
blumamir:
```suggestion
// +kubebuilder:validation:Enum=RolloutTriggeredSuccessfully;FailedToPatch;PreviousRolloutOngoing;Disabled;WaitingForRestart;WorkloadNotSupporting;NotRequired;WaitingInQueue;RolloutFinished
```

common/consts/consts.go
```diff
@@ -143,9 +144,9 @@ const (
 )
 
 // Auto rollback related consts
-const (
-	DefaultAutoRollbackGraceTime       = "5m"
-	DefaultAutoRollbackStabilityWindow = "1h"
+var (
+	DefaultAutoRollbackGraceTime       = 5 * time.Minute
+	DefaultAutoRollbackStabilityWindow = 1 * time.Hour
 )
```
blumamir, line 150: 🤩 much better like this

frontend/graph/status/rollout.go
```diff
@@ -17,11 +17,18 @@ func workloadRolloutStatusCondition(reason *string) model.DesiredStateProgress {
 		return model.DesiredStateProgressFailure
 	case v1alpha1.WorkloadRolloutReasonPreviousRolloutOngoing:
 		return model.DesiredStateProgressWaiting // short period until the rollout is complete
+	case v1alpha1.WorkloadRolloutReasonRolloutFinished:
+		return model.DesiredStateProgressSuccess // rollout is finished
 	case v1alpha1.WorkloadRolloutReasonDisabled:
 		return model.DesiredStateProgressUnknown // rollout is disabled in config, or agent was not enabled etc. need to refine those cases in the future.
+	case v1alpha1.WorkloadRolloutReasonNotRequired:
+		return model.DesiredStateProgressSuccess // rollout is not required
```
blumamir: consider
```suggestion
		return model.DesiredStateProgressIrrelevant
```

api/odigos/v1alpha1/instrumentationconfig_types.go
natyharary: Done

common/consts/consts.go
natyharary: woohoooo

frontend/graph/status/rollout.go
natyharary: Done

blumamir (approved):

PR #4161, opened 2026-01-29
fix: duplicate signals

BenElferink:
emergency-ticket id: EMR-959


frontend/services/action.go
```diff
@@ -636,8 +635,14 @@ func convertActionToModel(action *v1alpha1.Action) (*model.Action, error) {
...
-		signals = append(signals, model.SignalType(strings.ToUpper(string(signal))))
+	seen := make(map[model.SignalType]bool)
+	for _, s := range action.Spec.Signals {
+		signal := model.SignalType(s)
+		// Deduplicate: only add if not already seen
+		if !seen[signal] {
+			seen[signal] = true
+			signals = append(signals, signal)
+		}
 	}
```
alonkeyval, line 646: look like you already filter them in getSpecFromInput

alonkeyval (approved):

frontend/services/action.go
BenElferink: right, one is for write-data (input), and the other is for read-data (in case the customer already has a bugged CRD in cluster)

PR #4162, opened 2026-01-29
Yarden/plat 698 verify on prem token before updating it

yarden-lang:
emergency-ticket id: EMR-958


odigosauth/odigosauth.go
```diff
@@ -0,0 +1,132 @@
...
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/golang-jwt/jwt/v5"
+)
+
+// RsaPublicKeyString is the RSA public key used to verify Odigos enterprise tokens
+const RsaPublicKeyString = `-----BEGIN PUBLIC KEY-----
```
edeNFed: I think a secret in the helm chart might be a better place for this

odigosauth/odigosauth.go
```diff
@@ -0,0 +1,132 @@
...
+MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAqjGWuv6zbpSmSA5Ygaw5
+7ZLIO7U9bNjyWYzdVyToeayRT7sfKY313W0vbRgZySxEOwyeiHYrF/vvKn2MQx4e
+V9d7oNDnL8otNzNSp7S6BsXxZCALlk8YtVPQOV2a7uzrtNK4AS/EB6NuqZHJwvH9
+DmqEr0dkIbvoBjtWUo70ez6lBE1sQEShoORb11CVnGQwDBaeIXTL6+ajCFzyea0D
+l9NYI4XJhhdZGedb44mhNNiigJ09z+5KHemhwUHwrLvnZKrAKkUtZlNvu8JR9eU8
+daEeTbWsOQsnXuxSqxTlEaRddXbDNZjRyP7KtlMIrTII+MYEoigxHkO2nZmCBYhP
```
alonkeyval: I think we might be over-validating the on-prem token here.
Given that the token is issued by Odigos and stored internally, full signature verification + issuer/audience checks feel unnecessary and add operational overhead

alonkeyval: I’d suggest simplifying validation to:
- JWT structure
- expiration check

odigosauth/odigosauth.go
```diff
@@ -0,0 +1,102 @@
...
+
+func ValidateToken(onpremToken string) error {
+	if onpremToken == "" {
+		return fmt.Errorf("missing Odigos Pro token")
+	}
+
+	trimmedOnpremToken := strings.TrimSpace(onpremToken)
+	aud, err := checkTokenAttributes(trimmedOnpremToken)
+	if err != nil {
+		return fmt.Errorf("Failed to verify onprem token")
```
alonkeyval:
```suggestion
return fmt.Errorf("failed to verify onprem token: %w", err)
```

odigosauth/odigosauth.go
```diff
@@ -0,0 +1,102 @@
...
+
+	aud, ok := claims["aud"].(string)
+	if !ok || aud == "" {
+		return "", fmt.Errorf("missing aud claim")
+	}
+
+	return aud, nil
+}
+
+func ValidateToken(onpremToken string) error {
```
alonkeyval: the name is some kind misleading, we only check the name and the claims. so change he name or add comment

odigosauth/odigosauth.go
```diff
@@ -0,0 +1,102 @@
...
+		return fmt.Errorf("missing Odigos Pro token")
+	}
+
+	trimmedOnpremToken := strings.TrimSpace(onpremToken)
+	aud, err := checkTokenAttributes(trimmedOnpremToken)
+	if err != nil {
+		return fmt.Errorf("Failed to verify onprem token")
+	}
+
+	fmt.Println("Odigos onprem token verified", "audience", aud)
```
alonkeyval: remove

odigosauth/odigosauth.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	if expTime == nil {
+		return "", fmt.Errorf("missing exp claim")
+	}
+	if time.Now().After(expTime.Time) {
+		expirationDuration := time.Since(expTime.Time)
+		roundedDuration := expirationDuration.Round(time.Minute)
+		return "", fmt.Errorf("token is expired for %v, contact Odigos support to issue a new one", roundedDuration)
+	}
+
```
alonkeyval, line 34: use consts:
const (
    expectedIssuer = "https://odigos.io"
    expectedSubject = "https://odigos.io/onprem"
)

odigosauth/odigosauth.go
```diff
@@ -0,0 +1,102 @@
...
+	iss, ok := claims["iss"].(string)
+	if !ok || iss != "https://odigos.io" {
+		return "", fmt.Errorf("invalid iss")
```
alonkeyval, line 38: all your error message are inconsistent. 
fix all to:
"failed to verify onprem token"

odigosauth/odigosauth.go
```diff
@@ -0,0 +1,102 @@
...
+	return aud, nil
+}
+
+func ValidateToken(onpremToken string) error {
+	if onpremToken == "" {
+		return fmt.Errorf("missing Odigos Pro token")
+	}
+
+	trimmedOnpremToken := strings.TrimSpace(onpremToken)
+	aud, err := checkTokenAttributes(trimmedOnpremToken)
```
alonkeyval: what are we doing with this "aud"?

k8sutils/pkg/pro/common.go
```diff
@@ -11,9 +11,13 @@ import (
 
 	"github.com/odigos-io/odigos/api/k8sconsts"
 	odigosconsts "github.com/odigos-io/odigos/common/consts"
+	"github.com/odigos-io/odigos/odigosauth"
 )
 
 func UpdateOdigosToken(ctx context.Context, client kubernetes.Interface, namespace string, onPremToken string) error {
+	if err := odigosauth.ValidateToken(onPremToken); err != nil {
```
alonkeyval: what are this func do?

frontend/graph/schema.resolvers.go
```diff
@@ -410,7 +423,10 @@ func (r *k8sActualSourceResolver) WorkloadOdigosHealthStatus(ctx context.Context
 func (r *mutationResolver) UpdateAPIToken(ctx context.Context, token string) (bool, error) {
 	ns := env.GetCurrentNamespace()
 	err := pro.UpdateOdigosToken(ctx, kube.DefaultClient, ns, token)
-	return err == nil, nil
+	if err != nil {
+		return false, fmt.Errorf("failed to update odigos token: %v", err)
```
alonkeyval: replace %v with %w from error chaining

frontend/graph/schema.resolvers.go
```diff
@@ -47,20 +47,33 @@ func (r *computePlatformResolver) APITokens(ctx context.Context, obj *model.Comp
 	}
 
 	token := string(secret.Data[k8sconsts.OdigosOnpremTokenSecretKey])
+	err = odigosauth.ValidateToken(token)
```
alonkeyval: i think we need to rethink on the struct of this func.

instead of validate the token first, we can try to extract the payload first, if it failed, we can return generic error, otherwise we can validate the token and in case of an error we dont lose the data

frontend/graph/schema.resolvers.go
```diff
@@ -47,20 +47,33 @@ func (r *computePlatformResolver) APITokens(ctx context.Context, obj *model.Comp
...
+	if err != nil {
+		msg := err.Error()
+		return []*model.APIToken{
```
alonkeyval, line 57: not sure return "ERROR" as the name of the token is relevent in case of an error
```suggestion
    return []*model.APIToken{{Token: token, Name: "", Message: &msg}}, ni
```

odigosauth/odigosauth.go
yarden-lang: fixed

odigosauth/odigosauth.go
yarden-lang: added

odigosauth/odigosauth.go
yarden-lang: removed

odigosauth/odigosauth.go
yarden-lang: fixed

odigosauth/odigosauth.go
yarden-lang: the w will be "invalid iss" so i want a error log like this : ailed to verify onprem token, invalid iss

odigosauth/odigosauth.go
yarden-lang: printing , so i removed it from previous comment

k8sutils/pkg/pro/common.go
yarden-lang: removed

frontend/graph/schema.resolvers.go
yarden-lang: fixed

frontend/graph/schema.resolvers.go
yarden-lang: why ? why is it helpful ? extracting the payload means we are "validating" because if the payload is malformed so the verification is failed , so basically it will do the same.

frontend/graph/schema.resolvers.go
yarden-lang: i agree , but there is a line in the ui that represents the name of the token 
so suddenly we won't see anything in the line so it might be bad UX

alonkeyval (approved):

PR #4224, opened 2026-02-10
fix: skip Homebrew tap update when releasing older patch versions

alonkeyval:
Adds a semver comparison check before GoReleaser publishes the Homebrew formula. This prevents older patch releases (e.g. v1.16.9) from overwriting a newer version (e.g. v1.17.0) in the homebrew-odigos-cli tap.

emergency-ticket id: EMR-1016

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -168,6 +168,43 @@ jobs:
         with:
           node-version: 18
 
+      - name: Check if Homebrew tap should be updated
+        run: |
+          # Pre-releases always update their respective tap (homebrew-odigos-cli-rc)
+          if [ "${{ env.IS_PRERELEASE }}" = "true" ]; then
+            echo "Pre-release detected, will update Homebrew RC tap"
+            echo "SKIP_HOMEBREW_UPDATE=false" >> $GITHUB_ENV
+            exit 0
+          fi
+
+          # For stable releases, check if this version is newer than what's currently in the tap.
+          # This prevents older patch releases (e.g. v1.16.9) from overwriting a newer version (e.g. v1.17.0).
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -168,6 +168,43 @@ jobs:
...
+            | grep 'version "' | head -1 | sed 's/.*version "\(.*\)"/\1/')
+
+          if [ -z "$CURRENT_TAP_VERSION" ]; then
+            echo "Could not fetch current tap version, will update Homebrew tap"
+            echo "SKIP_HOMEBREW_UPDATE=false" >> $GITHUB_ENV
+            exit 0
+          fi
+
+          # Strip 'v' prefix from release tag for comparison
+          RELEASE_VERSION="${TAG#v}"
```
BenElferink, line 192: Does this work with `-rc0` & `-pre0` ?

BenElferink (approved):

.goreleaser.yaml
```diff
@@ -45,6 +45,7 @@ brews:
       owner: odigos-io
       name: '{{ if eq .Env.IS_PRERELEASE "true" }}homebrew-odigos-cli-rc{{ else }}homebrew-odigos-cli{{ end }}'
       token: "{{ .Env.HOMEBREW_GITHUB_API_TOKEN }}"
+    skip_upload: '{{ .Env.SKIP_HOMEBREW_UPDATE }}'
```
RonFed, line 48: We do want to upload but not overwrite the latest version. Is this what this is doing - or is skipping the upload completely?

tamirdavid1: @alonbraymok the same thing I told u

RonFed (approved):

PR #4228, opened 2026-02-11
fix: nav-icons in sidebar, rollout-restart warning

BenElferink:
- include all nav-icons in sidebar (between V1 & V2 designs)
- add warning before triggering rollout-restart function (kit: https://github.com/odigos-io/ui-kit/pull/663)

emergency-ticket id: EMR-1034

### DIFF
--- a/frontend/webapp/app/(overview)/layout.tsx
+++ b/frontend/webapp/app/(overview)/layout.tsx
@@ -1,16 +1,15 @@
 'use client';
 
-import React, { CSSProperties, useCallback, useMemo, type PropsWithChildren } from 'react';
+import React, { type CSSProperties, useMemo, type PropsWithChildren } from 'react';
 import { usePathname, useRouter } from 'next/navigation';
 import styled from 'styled-components';
 import { EntityTypes } from '@odigos/ui-kit/types';
 import { OdigosProvider } from '@odigos/ui-kit/contexts';
-import { DATA_FLOW_HEIGHT, MENU_BAR_HEIGHT, ROUTES } from '@/utils';
-import { ErrorBoundary, FlexColumn } from '@odigos/ui-kit/components';
 import { OverviewHeader, OverviewModalsAndDrawers } from '@/components';
-import { PipelineCollectorIcon, ServiceMapIcon } from '@odigos/ui-kit/icons';
+import { DataFlowActionsMenu, ToastList } from '@odigos/ui-kit/containers';
+import { ErrorBoundary, FlexColumn, IconsNav } from '@odigos/ui-kit/components';
 import { useConfig, useDataStreamsCRUD, useSSE, useTokenTracker } from '@/hooks';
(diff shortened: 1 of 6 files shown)


frontend/webapp/app/(overview)/layout.tsx
```diff
@@ -33,50 +32,16 @@ const ContentUnderActions = styled.div`
...
-const serviceMapId = 'service-map';
-const serviceMapDisplayName = 'Service Map';
-
-const pipelineCollectorsId = 'pipeline-collectors';
-const pipelineCollectorsDisplayName = 'Pipeline Collectors';
-
 const getEntityType = (pathname: string) => {
   return pathname.includes(ROUTES.SOURCES)
     ? EntityTypes.Source
     : pathname.includes(ROUTES.DESTINATIONS)
```
alonkeyval: is not cleaner you use here switch case instead of this long conditinal rendering?

frontend/webapp/utils/functions/navigation.ts
```diff
@@ -0,0 +1,29 @@
...
+const getPayloadForIcon = (router: AppRouterInstance, currentPath: string, targetPath: string, icon: SVG): NavbarProps['icons'][number] => {
+  return {
+    id: targetPath,
+    icon,
+    selected: currentPath === targetPath,
+    onClick: () => router.push(targetPath),
+  };
+};
+
+export const getNavbarIcons = (router: AppRouterInstance, currentPath: string) => {
```
alonkeyval, line 16: this architcute is a bit weird to me. 

why not just create the const array of nav icons (it not dynamic), and use an adpter func as getNavbarIcon:
`
export const NAV_ITEMS = [
  { route: ROUTES.OVERVIEW, icon: OverviewIcon },
  { route: ROUTES.SOURCES, icon: SourceIcon },
  { route: ROUTES.DESTINATIONS, icon: DestinationIcon },
  { route: ROUTES.ACTIONS, icon: ActionIcon },
  { route: ROUTES.INSTRUMENTATION_RULES, icon: InstrumentationRuleIcon },
  { route: ROUTES.SERVICE_MAP, icon: ServiceMapIcon },
  { route: ROUTES.PIPELINE_COLLECTORS, icon: PipelineCollectorIcon },
];

const navbarIcons = useMemo(() => {
  return NAV_ITEMS.map(item => ({
    id: item.route,
    icon: item.icon,
    selected: pathname.startsWith(item.route),
    onClick: () => router.push(item.route),
  }));
}, [pathname, router]);

`

it also keep it clear without the router prop

alonkeyval (approved):

frontend/webapp/utils/functions/navigation.ts
BenElferink: true, was just keeping it consistent with central, and there we do use conditional/dynamic

PR #4231, opened 2026-02-11
feat: auto-change mount method

BenElferink:
## Root Cause
Pods crashlooped on **Bottlerocket** because the webhook injected:

- `PHP_INI_SCAN_DIR=/var/odigos/php/8.2`

…but `/var/odigos` was never accessible inside the container.

With the default **`k8s-virtual-device`** method, kubelet is supposed to bind-mount host `/var/odigos` (populated by Odiglet) into the pod. On Bottlerocket this mount path became unreadable - likely **SELinux** or a **silent device-plugin mount failure**.

## Fix
We avoided the host-mount path entirely by auto-switching to **`k8s-init-container`**, which copies agent files into an `emptyDir` volume.

### 3 override layers
1. **Namespace annotation (manual):** `odigos.io/mount-method=k8s-init-container`  
2. **PSA detection:** `baseline` / `restricted` → force init-container  
3. **Bottlerocket detection:** Odiglet labels nodes → webhook switches cluster-wide (60s cache)

Init-container was also hardened to be **PSA-restricted compliant**.

## Result
Bottlerocket clusters now self-heal automatically, and support can force the safe method instantly via namespace annotation.

---

#### Changelog entry:
```release-note
NONE
```

emergency-ticket id: EMR-1032

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -98,6 +98,8 @@ jobs:
           - "env-injection"
           - "data-streams"
           - "mount-method"
+          - "mount-method-override"
+          - "psa-baseline"
           - "argo-rollouts"
         include:
           - kube-version: "1.21.12"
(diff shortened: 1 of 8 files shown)


BenElferink:
Drafting this for now, we were unable to reproduce this issue, we found that the actual issue had to do with a "toleration" applied externally:

```yaml
spec:
  tolerations:
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 300

  # The root cause:
  - effect: NoSchedule
    key: instrumentation.odigos.io/generic
    operator: Exists
```

PR #4254, opened 2026-02-16
fix: remove RHEL image check from release workflow

alonkeyval:
RHEL images are now built via a separate manual workflow (publish-modules-rhel.yml) and should not block the main release flow. This check was left over from #4195 when the RHEL build was separated.

emergency-ticket id: EMR-1057


RonFed (commented):
comment for future:
we remove this following https://github.com/odigos-io/odigos/pull/4195/changes

RonFed (approved):

PR #4260, opened 2026-02-18
CORE605: prevent data collector from crashing when enabling OwnTelemetry

natyharary:

OwnTelemetry is flag (set using helm via `--set ownTelemetry.metricsStore.disabled=false`) which enables reporting telemetry on Odigos pods and services.
This has two "flavors" - the first streams this telemetry to a VictoriaMetrics app, and the second streams it to a custom metrics destination (e.g. prometheus). Both can be enabled.

This PR solves a problem that when changing the configmap of the data collection, which sends the collected own telemetry to VictoriaMetrics, it crashes.
This is done by sending this telemetry to the gateway, which in turn can send the own telemetry data to VictoriaMetrics and/or the custom metrics destination.

```release-note
Odigos OwnTelemetry now flows through the gateway pod instead of through the data collection.
It can be directed to the Odigos metrics store and/or custom metrics destinations.
```

emergency-ticket id: EMR-1068


autoscaler/controllers/clustercollector/ownmetrics.go
```diff
@@ -0,0 +1,78 @@
...
+			"tls": config.GenericMap{
+				"insecure": true,
+			},
+		},
+	}
+}
+
+// addOwnMetricsPipeline integrates own-metrics collection into the gateway config.
+func addOwnMetricsPipeline(c *config.Config, ownMetricsConfig *odigosv1.OdigosOwnMetricsSettings, odigosNamespace string, ownTelemetryPort int32, destinationPipelineNames []string) error {
+	for name, receiverConfig := range receiversConfigForOwnMetricsPrometheus() {
```
tamirdavid1: why need range?

autoscaler/controllers/clustercollector/ownmetrics.go
```diff
@@ -0,0 +1,78 @@
...
+		c.Receivers[name] = receiverConfig
+	}
+
+	exporters := []string{}
+
+	if ownMetricsConfig.SendToOdigosMetricsStore {
+		for name, exporterConfig := range victoriaMetricsExporter(odigosNamespace) {
```
tamirdavid1: why need range?

autoscaler/controllers/clustercollector/ownmetrics.go
```diff
@@ -0,0 +1,78 @@
...
+	ownMetricsStorePipelineName        = "metrics/own-metrics"
+	odigosVictoriametricsExporterName  = "otlphttp/odigos-victoriametrics"
+)
+
+func receiversConfigForOwnMetricsPrometheus() config.GenericMap {
+	return config.GenericMap{
+		odigosOwnTelemetryOtlpReceiverName: config.GenericMap{
+			"protocols": config.GenericMap{
+				"grpc": config.GenericMap{
+					"endpoint": "0.0.0.0:44317",
```
tamirdavid1: is this needed if we send in http?

autoscaler/controllers/clustercollector/service.go
```diff
@@ -81,6 +81,18 @@ func updateGatewaySvc(svc *v1.Service, collectorsGroup *odigosv1.CollectorsGroup
 			Port:       4318,
 			TargetPort: intstr.FromInt(4318),
 		},
+		{
+			Name:       "own-metrics-grpc",
```
tamirdavid1: is this needed if we send in http?

natyharary: Removed it

autoscaler/controllers/clustercollector/ownmetrics.go
natyharary: Removed it

autoscaler/controllers/clustercollector/ownmetrics.go
natyharary: Removed it

autoscaler/controllers/clustercollector/ownmetrics.go
natyharary: Removed it

autoscaler/controllers/clustercollector/configmap.go
```diff
@@ -160,7 +160,20 @@ func syncConfigMap(enabledDests *odigosv1.DestinationList, allProcessors *odigos
 		common.ToExporterConfigurerArray(enabledDests),
 		common.ToProcessorConfigurerArray(processors),
 		func(c *config.Config, destinationPipelineNames []string, signalsRootPipelines []string) error {
-			return addSelfTelemetryPipeline(c, gateway.Spec.CollectorOwnMetricsPort, destinationPipelineNames, signalsRootPipelines)
+			// Creating a metric pipeline (throughput metrics) for the gateway to be sent to the UI
+			if err := addSelfTelemetryPipeline(c, gateway.Spec.CollectorOwnMetricsPort, destinationPipelineNames, signalsRootPipelines); err != nil {
+				return err
+			}
+			// Create pipeline for metrics from other Odigos components that are sending to the gateway
```
tamirdavid1:
```suggestion
			// Creating a metric pipeline for Odigos components metrics
```

autoscaler/controllers/nodecollector/configmap.go
```diff
@@ -227,6 +230,38 @@ func calculateCollectorConfigDomains(
 	return configDomains, string(mergedConfigYaml), nil
 }
 
+func ownMetricsTelemetryConfig(ownMetricsConfig *odigosv1.OdigosOwnMetricsSettings, odigosNamespace string) (config.Config, error) {
+	duration, err := time.ParseDuration(ownMetricsConfig.Interval)
+	if err != nil {
+		return config.Config{}, fmt.Errorf("failed to parse own metrics interval %q: %w", ownMetricsConfig.Interval, err)
```
tamirdavid1: do not fail the config creation due to this, lets out the default (from a const maybe)

tamirdavid1 (approved):

PR #4302, opened 2026-02-25
feat: add recover from rollback UI for crash-protected workloads

alonkeyval:
- Adds `rollbackOccurred: Boolean!` field to the `K8sWorkload` GraphQL type, resolved from `ic.Status.RollbackOccurred`
- Adds `rollbackOccurred` to the `GET_WORKLOADS` query and flows it through to the Source entity store
- Wires the existing `recoverFromRollbackForWorkload` mutation via a new `RECOVER_FROM_ROLLBACK` GraphQL mutation and `recoverFromRollback` hook function
- Passes `recoverFromRollback` to the SourceDrawer component so the "Recover" button can trigger recovery

```release-note
Add "Recover" action in the source drawer for workloads whose instrumentation was rolled back due to a crash
```

emergency-ticket id: EMR-1103


frontend/webapp/graphql/mutations/source.ts
```diff
@@ -23,3 +23,9 @@ export const RESTART_POD = gql`
     restartPod(namespace: $namespace, name: $name)
   }
 `;
+
+export const RECOVER_FROM_ROLLBACK = gql`
```
natyharary, line 27: The GQL query I've used is a bit different;
```graphql

"query": "mutation RecoverFromRollback($sourceId: K8sSourceId!) { recoverFromRollbackForWorkload(sourceId: $sourceId) }",

```

Do you see how the mutation is called different? I haven't used GQL for a while but is this important?

natyharary (commented):
small question about the graphql query

BenElferink (approved):

frontend/webapp/graphql/mutations/source.ts
alonkeyval: this is apollo syntax (client side), base on the schema this is the mutation:
```
recoverFromRollbackForWorkload(sourceId: K8sSourceId!): Boolean!
```

PR #4306, opened 2026-02-26
fix: another schema validations issues

tamirdavid1:

The Helm chart's JSON schema had additionalProperties: false on several fields that accept freeform user-defined data, causing schema validation to reject valid configurations. This affected affinity (all 6 components), images, userInstrumentationEnvs.languages.*.env (all 6 languages), and the global nodeSelector. The fix adds # additionalProperties: true annotations to each of these fields in values.yaml so that make helm-schema generates the correct schema, allowing users to set Kubernetes affinity rules, custom image overrides, arbitrary OTEL environment variables, and additional node selector labels without hitting validation errors.

```release-note
NONE
```

emergency-ticket id: EMR-1118


damemi (approved):
xref https://github.com/odigos-io/odigos/pull/4304 
falls under same release note for ^^ just linking here for future reference

PR #4312, opened 2026-03-01
fix: include source workloads from specified namespaces in DiagnoseGraphQL

BenElferink:
```release-note
fix: include source workloads from specified namespaces in DiagnoseGraphQL
```

emergency-ticket id: EMR-1128


frontend/services/diagnose.go
```diff
@@ -53,6 +53,10 @@ func DiagnoseGraphQL(
 			includeSourceWorkloads = *input.IncludeSourceWorkloads
 		}
 		sourceWorkloadNamespaces = input.SourceWorkloadNamespaces
+		if len(sourceWorkloadNamespaces) > 0 {
+			// If specific namespaces were selected, include source workloads from those namespaces
```
alonkeyval, line 57: is this fix: 
"Odigos Diagnose -> If you select specific namespaces, the download still includes all namespaces that have been instrumented" 
@BenElferink

alonkeyval (approved):

frontend/services/diagnose.go
damemi: @alonkeyval @BenElferink not sure what the original bug was from the description, but the `source workloads` means the actual deployment yaml, not the source CRD. 

All source CRDs are always collected from every namespace no matter what

SourceWorkloads(+SourceWorkloadNamespaces) adds the workload YAML (deployment, daemonset, etc)

PR #4374, opened 2026-03-07
chore: remove unused config value psp

blumamir:

psp is pod security policy, and it is effectively unused in code for anything, so I think it is safe to remove it.
originally it was added for one user, which had since upgrade his k8s version.

[PodSecurityPolicy was [deprecated](https://kubernetes.io/blog/2021/04/08/kubernetes-1-21-release-announcement/#podsecuritypolicy-deprecation) in Kubernetes v1.21, and removed from Kubernetes in v1.25.](https://kubernetes.io/docs/concepts/security/pod-security-policy/?utm_source=chatgpt.com)

```release-note
remove unused config option "psp" (PodSecurityPolicy)
```

### DIFF
--- a/operator/bundle/manifests/odigos-operator.clusterserviceversion.yaml
+++ b/operator/bundle/manifests/odigos-operator.clusterserviceversion.yaml
@@ -20,7 +20,7 @@ metadata:
     capabilities: Basic Install
     categories: Logging & Tracing
     containerImage: registry.connect.redhat.com/odigos/odigos-operator-certified:v1.8.7
-    createdAt: "2026-01-22T17:18:22Z"
+    createdAt: "2026-03-07T01:46:34Z"
     description: Odigos enables automatic distributed tracing with OpenTelemetry and
       eBPF.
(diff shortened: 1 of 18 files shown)


operator/bundle/manifests/odigos-operator.clusterserviceversion.yaml
```diff
@@ -426,6 +421,18 @@ spec:
...
+          resources:
+          - sampling
+          verbs:
+          - create
+          - delete
+          - get
+          - list
+          - patch
+          - update
+          - watch
```
blumamir, line 435: this came via the run of `make bundler` in operator. I guess it was unsynced prior to this PR. @damemi is this checked on CI? is this change ok?

damemi: yeah this is probably just something that wasn't checked. I don't think we have an active CI check for it right now since operator is in maintenance mode.

Did this come up in a CI check for you? If so we can probably remove it or shift it to a nightly

PR #4418, opened 2026-03-15
CORE-609 | e2e tests for URL templatization

shivanshuraj1333:

Depends on https://github.com/odigos-io/odigos/pull/4415 and https://github.com/odigos-io/odigos/pull/4324

```release-note
none
```


RonFed (changes requested):
I think we should not add an entire e2e foe this
Better to add to an existing test IMO, or find some test this in a way that has less overhead than complete e2e

shivanshuraj1333:
Yes, makes sense!

PR #4449, opened 2026-03-20
CORE-675 interface for sharing collector components with vmagent

natyharary:

Adds an interface to share actions/collector components with VM agent

I'm a draft! Don't take me so seriously!

```release-note

```

### DIFF
--- a/common/datacollectorcfg/spanmetrics.go
+++ b/common/datacollectorcfg/spanmetrics.go
@@ -1,0 +1,184 @@
+package datacollectorcfg
+
+import (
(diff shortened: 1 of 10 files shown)


common/datacollectorcfg/spanmetrics.go
```diff
@@ -0,0 +1,184 @@
+package datacollectorcfg
```
tamirdavid1, line 1: Better to call the package `nodecollector`

natyharary:
decided not be done atm

PR #4465, opened 2026-03-23
chore: bump client-go to v0.35.0

blumamir:

This is to support the [streaming lists](https://kubernetes.io/blog/2025/05/09/kubernetes-v1-33-streaming-list-responses/) new api which reduces memory spikes at init while syncing the informer cache

```release-note
ui and instrumentor cache is now using streaming list if api server supports it, to reduce memory spikes on init, and perform efficient lists in api server and etcd
```

emergency-ticket id: EMR-1295


blumamir:
Fixed in #4525

PR #4468, opened 2026-03-23
CORE-680: Add OBI distro

damemi:
This adds `opentelemetry-ebpf-instrumentation` (OBI) as a supported workload instrumentation distro.

The benefit of this is to provide additional coverage and offer users a way to deploy OBI within our platform if they'd like to.

It relies on features we added upstream to OBI in https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1321 and https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pull/1388 to support dynamic PID selection (as opposed to OBI's static service/exe config approach).

The OBI `Instrumenter` is a single long-lived routine that handles process events, filters/matches them based on config criteria (ie, which PIDs we want), and attaches shared eBPF programs to those processes. The `DynamicSelector` we added upstream provides a hook into the `Instrumenter` to update the filter/matching criteria during runtime without needing to restart OBI.

This adds OBI as an sdk/distro that can be used with any language. The basic flow of the new distro is:
1. `NewOBIInstrumentationFactory()`: Initialize the OBI config and DynamicSelector, but does not start the OBI `Instrumenter` routine yet. This is to prevent the `Instrumenter` from running and using resources if nothing is actually instrumented by OBI.
2. `factory.CreateInstrumentation(ctx, pid)`: If the OBI `Instrumenter` is not started yet, this starts it and returns an `obiInstrumentation` handle that stores that DynamicSelector and the instrumented PID. This does not attach the PID to OBI yet.
3. `obiInstrumentation.Load()`: Uses the stored PID in the `o` instrumentation object to call `DynamicSelector.AddPIDs(pid)` which updates the running OBI manager to include/attach this process.
4. `obiInstrumentation.Close()`: Removes the stored PID in `o` with `DynamicSelect.RemovePIDs(pid)` to update the OBI manager to exclude/detach this process.

A caveat to this design is that it's difficult to _stop_ the Instrumenter once it's started (ie, all OBI processes get uninstrumented and we no longer need that routine) without complex async management. We could do something like this on `Close()`:

```go
func Close() {
  f.Selector.RemovePIDs(pid)
  if len(f.Selector.GetPIDs()) == 0 {
    cancelObiCtx()
  }
}
```

But that would require a mutex, and it's possible that another OBI app could come along in the meantime waiting to be instrumented while that mutex is held, and the Instrumenter gets cancelled before the new app gets added. Which would be very confusing to debug. I tried a lot of different approaches and they were all messy. I think some more upstream OBI changes around an actual handle on the instrumenter would help, along with other use cases for holding a reference to the instrumenter itself.

Overall:

* OBI Factory -- Called once at odiglet startup, runs nothing
* OBI CreateInstrumentation -- Starts OBI (if necessary) as a singleton and creates the per-process `instrumentation` used by Odiglet
* Instrumentation Load -- Attaches the PID to the OBI singleton

Another possible upstream contribution would be a signal from OBI for when it's actually started/running/ready to accept new PIDs which Load() could wait on

This whole approach is slightly different from languages like Go, where the *Odiglet* is the long lived process, so each `CreateInstrumentation()` in Go calls the go-auto framework to load, attach, and manage. In the case of OBI, the OBI Instrumenter handles loading, attaching, and managing so Odiglet is a wrapped layer on top of that for integration with our control plane.

This also adds OBI ebpf generation to the odiglet Dockerfile using the upstream [obi-generator[(https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/pkgs/container/obi-generator)

The OBI distro is added as its own module so it can be imported into enterprise easier: https://github.com/odigos-io/odigos-enterprise/pull/2577 (see that PR for details on OBI distro module)

UI Changes here: https://github.com/odigos-io/ui-kit/pull/748

New C++ app in simple-demo for an e2e test here: https://github.com/odigos-io/simple-demo/pull/67

```release-note
feat: Support opentelemetry-ebpf-instrumentation (OBI) for workload instrumentation
```


api/k8sconsts/device.go
```diff
@@ -6,4 +6,8 @@ const (
 	// it would be more ideal to only mount what is needed,
 	// but it's not desirable to have tons of different devices for each case.
 	OdigosGenericDeviceName = "instrumentation.odigos.io/generic"
+
+	// OdigosDistroNameOBI is the distro name for OpenTelemetry eBPF Instrumentation (OBI).
+	// This distro supports any language and is not enabled by default.
+	OdigosDistroNameOBI = "opentelemetry-ebpf-instrumentation"
```
RonFed, line 12: I think this is not the correct place for this since this is not a device - in addition the distro yaml should be the single source of truth with regards to the distro metadata

distros/distro/oteldistribution.go
```diff
@@ -17,9 +18,19 @@ type RuntimeEnvironment struct {
 	Name string `yaml:"name"`
 
 	// semconv range of the runtime versions supported by this distribution.
+	// Use SupportedVersionsWildcard ("*") to accept any runtime version (not a hashicorp/go-version constraint).
 	SupportedVersions string `yaml:"supportedVersions,omitempty"`
 }
```
RonFed, line 23: I think a more accurate change would be something like this in the language field - since the obi distro is basically "any language" not "any language version"

i.e in here:
```Go
	// the programming language this distribution targets.
	// each distribution must target a single language.
	Language common.ProgrammingLanguage `yaml:"language"`
```

distros/yamls/opentelemetry-ebpf-instrumentation.yaml
RonFed, line 1: we should declare this distro as not requiring restart - `noRestartRequired: true` under `runtimeAgent`

distros/yamls/opentelemetry-ebpf-instrumentation.yaml
```diff
@@ -0,0 +1,16 @@
...
+  language: go
+  runtimeEnvironments:
+    - name: multi-runtime
+      supportedVersions: '*'
+  displayName: OpenTelemetry eBPF Instrumentation (OBI)
+  description: |
+    Language-agnostic eBPF-based instrumentation using the OpenTelemetry eBPF Instrumentation (OBI) SDK.
+    Supports any language without code changes. Not enabled by default.
+  runtimeAgent:
+    device: 'instrumentation.odigos.io/generic'
```
RonFed, line 16: do we need the device? I don't think there are any specific files we need to mount, right?

distros/yamls/opentelemetry-ebpf-instrumentation.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: internal.odigos.io/v1beta1
+kind: OtelDistribution
+metadata:
+  name: opentelemetry-ebpf-instrumentation
+spec:
+  name: opentelemetry-ebpf-instrumentation
+  language: go
```
RonFed, line 7: this distro doesn't target go, it targets all the languages

Makefile
```diff
@@ -117,7 +118,8 @@ build-image/%:
 	--build-arg SUMMARY="$(SUMMARY)" \
 	--build-arg DESCRIPTION="$(DESCRIPTION)" \
 	--build-arg LD_FLAGS="$(LD_FLAGS)" \
-	--build-arg RHEL="$(RHEL)"
+	--build-arg RHEL="$(RHEL)" \
+	--no-cache
```
RonFed, line 122: remove the no-cache

Makefile
```diff
@@ -107,8 +107,9 @@ $(HELM_SCHEMA_BIN):
 helm-schema-clean:
 	rm -f $(HELM_SCHEMA_BIN)
 
+# Pass DOCKER_BUILD_OPTS=--no-cache to force a clean build (e.g. when go.mod replace changes and cache is stale).
 build-image/%:
-	docker build $(TARGET_FLAG) \
+	docker build $(DOCKER_BUILD_OPTS) $(TARGET_FLAG) \
```
RonFed, line 112: we already have
```
# If you made a change and it's not being reflected in the image,
# it might be because the cache is being used and not invalidated properly.
# Use NO_CACHE=--no-cache to force a rebuild, or make build-odiglet-with-otel_agents NO_CACHE=--no-cache
NO_CACHE?=
```

think we can remove the `DOCKER_BUILD_OPTS`

odiglet/Dockerfile
```diff
@@ -119,12 +119,46 @@ COPY distros/ distros/
...
+    --mount=type=cache,target=/go/pkg \
+    sh -c 'go mod download && OBI_DIR=`go list -m -f "{{.Dir}}" go.opentelemetry.io/obi` && chmod -R +w "$OBI_DIR" && cp -r /tmp/obi-generated/* "$OBI_DIR/"'
+
 ARG TARGETARCH
 ARG LD_FLAGS
 ARG RHEL=false
 RUN --mount=type=cache,target=/root/.cache/go-build \
     --mount=type=cache,target=/go/pkg \
-    GOOS=linux GOARCH=$TARGETARCH LD_FLAGS="${LD_FLAGS}" make build-odiglet
+    GOOS=linux GOARCH=$TARGETARCH LD_FLAGS="${LD_FLAGS}" SKIP_OBI_GENERATE=1 make build-odiglet
```
RonFed, line 161: can we have the obi build as part of `make build-odiglet` keeping the Dockerfile in a single step for odiglet build?
The approach here have a `builder` and `odiglet-builder` - ideally we should just have a builder I thin.

Can we use the exact same method we use to geenerate the files as we do in our repos - same as done in the Makefile - just adding obi to the libs list - than `go mod download`, generate in the downloaded mod - it seems very similar here, just duplicated.

odiglet/pkg/ebpf/sdks/obi/obi.go
```diff
@@ -0,0 +1,89 @@
...
+	f.startOnce.Do(func() {
+		go func() {
+			err := instrumenter.Run(f.ctx, f.obiCfg, instrumenter.WithDynamicPIDSelector(f.selector))
+			if err != nil && f.ctx.Err() == nil {
+				obiLog.Error("OBI instrumenter exited with error", "err", err)
+			}
+		}()
+	})
+	return &obiInstrumentation{selector: f.selector, pid: pid}, nil
+}
```
RonFed, line 63: Since we create instrumentation, and clean them in an [event loop pattern](https://github.com/odigos-io/odigos/blob/main/instrumentation/manager.go#L243) there is no real concern for concurrency here.

If that helps, we can simplify this a bit, and add the check you mentioned in the `Close` function - if we got to 0 selected PIDs, we can close the obi routine completely.

We can document here that we assume no concurrency, but this assumption is built into the event loop pattern and all the other instrumentations do not have concurrency guarantees.

helm/odigos/templates/odiglet/clusterrole.yaml
```diff
@@ -9,6 +9,7 @@ rules:
       - ''
     resources:
       - pods
+      - services
```
RonFed, line 12: why is this required?

instrumentor/controllers/agentenabled/sync.go
```diff
@@ -600,31 +600,34 @@ func calculateContainerInstrumentationConfig(containerName string,
...
-		}
-		if !constraint.Check(detectedVersion) {
-			return odigosv1.ContainerAgentConfig{
-				ContainerName:       containerName,
-				AgentEnabled:        false,
-				AgentEnabledReason:  odigosv1.AgentEnabledReasonUnsupportedRuntimeVersion,
-				AgentEnabledMessage: fmt.Sprintf("%s runtime not supported by OpenTelemetry. supported versions: '%s', found: %s", d.RuntimeEnvironments[0].Name, constraint, detectedVersion),
-			}, nil
+		sv := d.RuntimeEnvironments[0].SupportedVersions
+		if !distro.IsSupportedVersionsWildcard(sv) {
```
RonFed, line 604: see comments on the distro definitions - think the wildcard should be on the language - and not on the version

instrumentor/controllers/agentenabled/sync.go
```diff
@@ -822,6 +825,11 @@ func resolveContainerDistro(
 			}
 		}
 
+		// OBI is language-agnostic (eBPF); skip language match when user explicitly selected it.
+		if overwriteDistroName == k8sconsts.OdigosDistroNameOBI {
+			return distro, nil
+		}
```
RonFed, line 831: instead of checking obi specifically - a more robust approach is to check "is this distro language agnostivc"?

odiglet/pkg/ebpf/process_details.go
```diff
@@ -104,6 +104,12 @@ func (kd *K8sProcessDetails) Distribution(ctx context.Context) (*distro.OtelDist
 		}
 	}
 
+	// OBI (opentelemetry-ebpf-instrumentation) supports any language; skip language verification
+	if kd.Distro.Name == k8sconsts.OdigosDistroNameOBI {
+		kd.langVerified = &trueVal
+		return kd.Distro, nil
+	}
```
RonFed, line 111: instead of checking obi specifically, a more robust approach is to check "is the distro language agnostic?"

distros/yamls/opentelemetry-ebpf-instrumentation.yaml
damemi: we don't need to mount anything, but doesn't this also control scheduling to make sure the pod goes onto an odiglet node?

helm/odigos/templates/odiglet/clusterrole.yaml
damemi: OBI uses service discovery out of the box, not sure if it's specifically for trace decoration or just cluster wide discovery. I'll check if it's configurable, otherwise the odiglet logs a bunch of permission errors from OBI

Makefile
damemi: I don't think we do have that in this repo, or it was removed at some point?

odiglet/Dockerfile
damemi: I'm looking into other ways to do this cleanly, and I think it's a priority in the OBI sig to improve consumability (especially with the goal of moving it to the collector) so I think it will get better in the future.

But right now, [OBI's `go generate` commands](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/81871be619bbea1509edd4db720178fae94601b4/pkg/internal/ebpf/logger/logger.go#L20) are a little different than [go-auto](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/01e1defb272e3ad0db5d43d80343a13dfe34b910/internal/pkg/instrumentation/bpf/net/http/client/probe.go#L31) or [runtime-detector](https://github.com/odigos-io/runtime-detector/blob/b2be7069fcf6681101bb4f29e996ca9041b2cc15/internal/probe/probe.go#L23).

In each of the current libs, we can just call `make generate` because their generate flags call `go run github.com/cilium/ebpf/cmd/bpf2go` directly, but OBI [sets up environment variables](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/81871be619bbea1509edd4db720178fae94601b4/Makefile#L228-L246) that depend on an [`internal/tools` directory](https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/81871be619bbea1509edd4db720178fae94601b4/Makefile#L90-L96) that outputs to a local `.tools/bpf2go` with `go tool`. That `internal/tools` isn't available in the Go modcache checkout.

Other parts of the OBI `make generate` also assume an initialized Git. They don't seem to directly affect the generate, but it leads to weird messages in the build like `fata: not a git repository`.

In Beyla, they currently import [OBI as a submodule](https://github.com/grafana/beyla/blob/c948015cb73650112a6a8e4742f7b9ab97b27219/.gitmodules#L1-L3) (technically their fork but same thing) and use it that way. They also [vendor obi](https://github.com/grafana/beyla/blob/c948015cb73650112a6a8e4742f7b9ab97b27219/Makefile#L225-L226) and run generate commands from there.

I'm still trying different things but in a lot of cases where I'm trying to generate in the go mod cache directly it seems like the Go ebpf files just aren't being generated. Will update with whatever I find but something like what's here now seems to be the simplest approach due to just how OBI is currently set up

damemi: @RonFed ok this was horrible but I think I figured out the problem. Basically, OBI does not commit the generated `*_bpfel.go` files (but go-auto and runtime detector do). All we're really generating in those libraries are the `*.o` binary files.

The `go` cmd apparently builds an index of the go mod cache the first time it's run. So I think if there is any `go` command that interacts with the OBI module, that module gets double secret cached. If that happens before we generate the `bpfel.go` files, then when we compile the Odiglet those bpfel.go won't be there and we get a bunch of undefined variables.

This isn't a problem for the .o binary files because they aren't Go files. they're embedded and the go compiler pulls them in by their raw path (relevant to the mod cache) to satisfy the embed

You can disable the goindex (which is what I'm doing now) with the tradeoff of slower builds. It didn't have much effect in my builds with no-cache (~3 minutes for odiglet).

I'm not entirely sure why the old approach here with multi-stage even worked at all. I think because it was generating _outside_ the mod cache and then copying it in during a different stage before the final `go build` happens. I wrote (cursor wrote) a summary here which we don't have to commit but might be useful https://github.com/damemi/odigos/blob/4216c2e21847a30b9bcb1a5e23e04ddceb3b64c7/odiglet/docs/obi-bpf2go-module-cache.md

So I think the options are:
1. find out why OBI isn't committing the generated Go files
2. use `goindex=0` (pushed now ptal)
3. use the "generate outside and copy in" approach

odiglet/pkg/ebpf/sdks/obi/obi.go
```diff
@@ -0,0 +1,131 @@
...
+	"go.opentelemetry.io/obi/pkg/config"
+	"go.opentelemetry.io/obi/pkg/instrumenter"
+	"go.opentelemetry.io/obi/pkg/obi"
+)
+
+// obiLogger must be called after commonlogger.Init (e.g. from CreateInstrumentation), not at package
+// init: LoggerCompat() returns a nop logger until Init runs, and main imports this package before Init.
+func obiLogger() *commonlogger.OdigosLogger {
+	return commonlogger.LoggerCompat().With("subsystem", "ebpfobi")
+}
```
RonFed: think this will create a new logger instance on every call?
maybe worth to consider putting the logger in the factory

In addition some of the logging might be too verbose and in general it is better to let the caller (the instrumentation manager) handle any logging for errors or meaningful events if possible

odiglet/pkg/ebpf/sdks/obi/obi.go
```diff
@@ -0,0 +1,131 @@
...
+	var lv slog.LevelVar
+	lv.Set(slog.LevelInfo)
+	if s := strings.TrimSpace(os.Getenv("OTEL_LOG_LEVEL")); s != "" {
+		var parsed slog.Level
+		if err := parsed.UnmarshalText([]byte(strings.ToUpper(s))); err == nil {
+			lv.Set(parsed)
+		}
+	}
+	h := slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: &lv})
+	slog.SetDefault(slog.New(h))
```
RonFed: this is a global operation, no? it will affect any logger in the odiglet not just the obi once I think

odiglet/pkg/ebpf/sdks/obi/obi.go
```diff
@@ -0,0 +1,131 @@
...
+	// Export traces to the node collector (same node as odiglet). Use http scheme for insecure gRPC.
+	// Protocol is inferred from port (4317 -> gRPC) by OBI.
+	cfg.Traces.TracesEndpoint = fmt.Sprintf("http://localhost:%d", consts.OTLPPort)
+	return &cfg
+}
+
+// CreateInstrumentation starts the OBI instrumenter if it is not already running
+// and returns an obiInstrumentation that allows adding/removing this PID using the dynamic selector.
+func (f *OBIInstrumentationFactory) CreateInstrumentation(ctx context.Context, pid int, settings instrumentation.Settings) (instrumentation.Instrumentation, error) {
+	obiLogger().Info("Creating OBI instrumentation", "pid", pid, "serviceName", settings.ServiceName)
```
RonFed: remove

odiglet/pkg/ebpf/sdks/obi/obi.go
```diff
@@ -0,0 +1,131 @@
...
+	if f.obiCtx == nil {
+		obiCtx, obiCtxCancel := context.WithCancel(f.ctx)
+		f.obiCtx = obiCtx
+		f.obiCtxCancel = obiCtxCancel
+
+		go func() {
+			setObiSlogDefault()
+			obiLogger().Info("Starting OBI instrumenter", "pid", pid, "serviceName", settings.ServiceName)
```
RonFed: remove

odiglet/pkg/ebpf/sdks/obi/obi.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (o *obiInstrumentation) Run(ctx context.Context) error {
+	<-ctx.Done()
+	return nil
+}
+
+func (o *obiInstrumentation) Close(_ context.Context) error {
+	o.selector.RemovePIDs(uint32(o.pid))
+	obiLogger().Info("Removed PID", "pid", o.pid)
```
RonFed: remove

odiglet/pkg/ebpf/sdks/obi/obi.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	return &obiInstrumentation{selector: f.selector, pid: pid, factory: f}, nil
+}
+
+// obiInstrumentation implements instrumentation.Instrumentation; it only holds the factory's
+// selector so Close(ctx) can call RemovePIDs for the PID.
+type obiInstrumentation struct {
+	selector *discover.DynamicPIDSelector
+	pid      int
+	factory  *OBIInstrumentationFactory
```
RonFed, line 73: I think it would be cleaner to either:
1. have the factory pass a cleanup function
2. have the factory pass a context and cancel function

Instead of each instrumentation object having a pointer to the factory.

odiglet/pkg/ebpf/sdks/obi/go.mod
```diff
@@ -0,0 +1,228 @@
+module github.com/odigos-io/odigos/odiglet/pkg/ebpf/sdks/obi
+
+go 1.26.2
+
+require (
+	github.com/odigos-io/odigos/common v0.0.0
+	github.com/odigos-io/odigos/instrumentation v0.0.0-00010101000000-000000000000
```
RonFed: should be v0.0.0?

distros/yamls/opentelemetry-ebpf-instrumentation.yaml
RonFed: if we support instrumenting without a restart this doesn't have a real meaning I guess - since we'd only be able to instrument where there is an odiglet

odiglet/go.mod
```diff
@@ -3,129 +3,251 @@ module github.com/odigos-io/odigos/odiglet
 go 1.26.2
 
 require (
-	github.com/cilium/ebpf v0.20.0
+	github.com/cilium/ebpf v0.21.0
```
RonFed: revert - cilium/ebpf v0.21 has a known bug in https://github.com/cilium/ebpf/issues/1964
upgrading that lib can have a lot of side affects and should be done separatly

odiglet/go.mod
```diff
@@ -3,129 +3,251 @@ module github.com/odigos-io/odigos/odiglet
...
 	golang.org/x/sys v0.42.0
-	google.golang.org/grpc v1.79.3
+	google.golang.org/grpc v1.80.0
 	google.golang.org/protobuf v1.36.11
-	k8s.io/api v0.35.2
-	k8s.io/apimachinery v0.35.2
-	k8s.io/client-go v0.35.2
+	k8s.io/api v0.35.3
+	k8s.io/apimachinery v0.35.3
+	k8s.io/client-go v0.35.3
```
RonFed, line 30: I thought the point of making obi a separate module was so it won't make us align to its dependencies - if that's the case I'd expect to not have all these changes in this PR - or if they are un avoidable. - to remove the separate module for obi

damemi: the point of the separate module is so we can import into enterprise without getting the ambiguous go-auto import

odiglet/pkg/ebpf/sdks/obi/obi.go
damemi: I tried that but we need to be able to write back to the factory somehow when an Instrumentation object cancels the OBI instrumenter so that the factory knows to restart the OBI instrumenter on the next instrumentation

odiglet/go.mod
damemi: good catch, not sure how that got pulled in. OBI is on 0.20, so I must have done a go get somewhere. Downgraded back

distros/yamls/opentelemetry-ebpf-instrumentation.yaml
damemi: oh yeah, right if we're not restarting then scheduling doesn't even matter

odiglet/pkg/ebpf/sdks/obi/obi.go
damemi: yeah this was my attempt to control OBI logs but it is too broad. They have an ExternalLogger() func that the comment also seems to imply could be a problem with how it affects the default logger: https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation/blob/53b500fb6a5a5256e2a140c42517c7ce3e57b565/pkg/obi/config.go#L750-L763

it would be a good contribution upstream to be able to wire in a custom logger but I'll just remove this for now

common/lang_detection.go
```diff
@@ -38,6 +41,11 @@ const (
 	UnknownProgrammingLanguage ProgrammingLanguage = "unknown"
 )
 
+// IsProgrammingLanguageWildcard reports whether lang is the distro wildcard meaning "any language".
+func IsProgrammingLanguageWildcard(lang ProgrammingLanguage) bool {
+	return strings.TrimSpace(string(lang)) == string(ProgrammingLanguageWildcard)
+}
```
blumamir, line 47: for other languages we compare directly with `==` instead of adding a a util function. why is it different for the `*` language?

distros/yamls/opentelemetry-ebpf-instrumentation.yaml
```diff
@@ -0,0 +1,17 @@
...
+kind: OtelDistribution
+metadata:
+  name: opentelemetry-ebpf-instrumentation
+spec:
+  name: opentelemetry-ebpf-instrumentation
+  language: '*'
+  isEbpf: true
+  runtimeEnvironments:
+    - name: multi-runtime
+      supportedVersions: '*'
```
blumamir, line 11: consider not populating this array. it can have zero length which should mean - skip runtime env checks

distros/distro/oteldistribution.go
```diff
@@ -16,7 +16,8 @@ type RuntimeEnvironment struct {
 	// while java-script can run in both nodejs and browser, the distribution should specify where it is intended to run.
 	Name string `yaml:"name"`
 
-	// semconv range of the runtime versions supported by this distribution.
+	// semconv range of the runtime versions supported by this distribution (hashicorp/go-version constraint).
+	// May be '*' in YAML for distros that skip runtime version checks in the instrumentor (e.g. wildcard language).
```
blumamir, line 20: as I wrote in another comment, consider just not populating the list in this case

distros/yamls/opentelemetry-ebpf-instrumentation.yaml
blumamir, line 1: we also have `java-ebpf-instrumentations` in enterprise. should we use the same here (instrumentations with an s at the end) to keep things consistent?

also, since the mechanism is a bit different, I wonder if this can cause confusion for users

instrumentor/controllers/agentenabled/distroresolver/distroresolver.go
```diff
@@ -79,6 +84,9 @@ func CalculateDefaultDistroPerLanguage(defaultDistros map[common.ProgrammingLang
 			if distro == nil {
 				continue
 			}
+			if common.IsProgrammingLanguageWildcard(distro.Language) {
+				continue
+			}
```
blumamir, line 89: could be nice to add support for this in the rule someday in the duture

RonFed (commented):
lgtm, still think we need to checkout consuming the archive OBI publishes thay you shared with me before merging with the current Dockerfile/Makefile changes

odiglet/pkg/ebpf/sdks/obi/obi.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+// OBIInstrumentationFactory creates instrumentations that add/remove PIDs on a shared
+// DynamicPIDSelector while a single OBI instrumenter runs in the background.
+// Requires OBI with DynamicPIDSelector support (e.g. go.opentelemetry.io/obi from main after PR 1388).
+type OBIInstrumentationFactory struct {
+	ctx          context.Context
+	logger       *commonlogger.OdigosLogger
+	obiCtx       context.Context
+	obiCtxCancel context.CancelFunc
```
blumamir, line 23: why do we need both `ctx` and `obiCtx` here?

storing ctx as a property on a long-living object feels not right

odiglet/go.mod
```diff
@@ -11,121 +11,243 @@ require (
...
+	github.com/aws/aws-sdk-go-v2/internal/configsources v1.4.19 // indirect
+	github.com/aws/aws-sdk-go-v2/internal/endpoints/v2 v2.7.19 // indirect
+	github.com/aws/aws-sdk-go-v2/internal/ini v1.8.5 // indirect
+	github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding v1.13.6 // indirect
+	github.com/aws/aws-sdk-go-v2/service/internal/presigned-url v1.13.19 // indirect
+	github.com/aws/aws-sdk-go-v2/service/signin v1.0.7 // indirect
+	github.com/aws/aws-sdk-go-v2/service/sso v1.30.12 // indirect
+	github.com/aws/aws-sdk-go-v2/service/ssooidc v1.35.16 // indirect
+	github.com/aws/aws-sdk-go-v2/service/sts v1.41.8 // indirect
+	github.com/aws/smithy-go v1.24.2 // indirect
```
blumamir, line 57: could be nice not to pull these. I guess obi is a bundle of everything and cannot be composed of just what needed?

odiglet/Makefile
```diff
@@ -1,26 +1,63 @@
...
 LIBRARIES := go.opentelemetry.io/auto \
 			github.com/odigos-io/runtime-detector
 
 build-odiglet: generate
-	CGO_ENABLED=0 go build -ldflags="$(LD_FLAGS)" -o odiglet cmd/main.go
+	$(GOBUILD) -ldflags="$(LD_FLAGS)" -o odiglet cmd/main.go
 
 debug-build-odiglet: generate
-	go build -o odiglet -gcflags "all=-N -l" cmd/main.go
+	CGO_ENABLED=0 go build -a -o odiglet -gcflags "all=-N -l" cmd/main.go
```
blumamir:
```suggestion
	$(GOBUILD) -o odiglet -gcflags "all=-N -l" cmd/main.go
```

odiglet/pkg/ebpf/sdks/obi/obi.go
blumamir: @damemi why is it not possible with a cleanup function?

blumamir (approved):
added few comments/suggestions, but overall looks good. 
would love to chat about it before merging

distros/yamls/opentelemetry-ebpf-instrumentation.yaml
damemi: I don't think so, I named this literally after OBI "opentelemetry-ebpf-instrumentation" the project itself for clarity https://github.com/open-telemetry/opentelemetry-ebpf-instrumentation

odiglet/pkg/ebpf/sdks/obi/obi.go
damemi: @blumamir the cleanup function would still need to have a reference back to the factory anyway (to signal that the next instrumentation should re-start the instrumenter), so it's essentially just a closure with the logic that's already here

odiglet/go.mod
damemi: not today unfortunately

RonFed (approved):

PR #4484, opened 2026-03-25
feat: settings page

BenElferink:
```release-note
feat: settings page
```

emergency-ticket id: EMR-1325


frontend/graph/configs.graphqls
```diff
@@ -178,59 +222,59 @@ type MetricsSourceConfig {
...
+  requestMemoryMiB: ReconciledInt
+  limitMemoryMiB: ReconciledInt
 }
 
 type OdigosOwnTelemetryConfig {
-  metricsStoreDisabled: Boolean
+  metricsStoreDisabled: ReconciledBoolean
 }
 
 type EffectiveConfig {
```
blumamir: I was thinking, what if we could just send the config yaml which we read from the config maps as texts. then whenever we add new configs, it's one less place to change.

We tend to add config values very often, and since this object is small anyway, we don't gain a lot of value from using gql here.

aviadhahami (approved):

PR #4506, opened 2026-03-26
Odiglet base fixes

blumamir:

Small fixes after #4504 to address:

- the automated PR fails the release notes check
- add `-j` to the make command for a bit faster build

Internal build, not user facing

```release-note
NONE
```

emergency-ticket id: EMR-1342


damemi (approved):
We should be able to skip the release note check for bots

blumamir:
> We should be able to skip the release note check for bots

I actually want the user to think if this change carries any release notes. for example, if this drops support for some version or something. we have a chance and place to write it down

PR #4544, opened 2026-03-31
CORE-863: add heuristics for runtime selection for java and python

natyharary:
* add heuristics for runtime selection for java and python
* add determinism of the languages found, by sorting the result list

```release-note
Adds runtime selection improvements for multi-runtime containers (specifically for python and java)
```

### DIFF
--- a/odiglet/pkg/kube/runtime_details/inspection.go
+++ b/odiglet/pkg/kube/runtime_details/inspection.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/odigos-io/odigos/procdiscovery/pkg/libc"
@@ -16,16 +17,22 @@ import (
 	"github.com/odigos-io/odigos/common"
 	"github.com/odigos-io/odigos/common/consts"
 	"github.com/odigos-io/odigos/common/envOverwrite"
+	commonlogger "github.com/odigos-io/odigos/common/logger"
 	criwrapper "github.com/odigos-io/odigos/k8sutils/pkg/cri"
 	"github.com/odigos-io/odigos/k8sutils/pkg/env"
 	"github.com/odigos-io/odigos/k8sutils/pkg/workload"
-	commonlogger "github.com/odigos-io/odigos/common/logger"
 	"github.com/odigos-io/odigos/procdiscovery/pkg/inspectors"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+var LowPriorityLanguages = []common.ProgrammingLanguage{
+	common.CPlusPlusProgrammingLanguage, // can be a wrapper around the real application
+	common.NginxProgrammingLanguage,     // can run as a sidecar process in the same container
+	common.PythonProgrammingLanguage,    // can be a ML/AI serving runtime alongside a server
+}
 
 var errNoKnownLanguageDetected = errors.New("no known programming language detected in the container")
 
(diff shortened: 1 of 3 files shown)


odiglet/pkg/kube/runtime_details/inspection.go
```diff
@@ -117,6 +100,29 @@ func runtimeInspection(ctx context.Context, pods []corev1.Pod, criClient *criwra
...
+	case 1:
+		return uniqueKnownLanguages[0], nil
+	case 2:
+		switch {
+		case uniqueKnownLanguages[0] == common.JavaProgrammingLanguage || uniqueKnownLanguages[1] == common.JavaProgrammingLanguage:
+			return common.JavaProgrammingLanguage, nil
+		case slices.Contains(LowPriorityLanguages, uniqueKnownLanguages[0]):
+			return uniqueKnownLanguages[1], nil
+		case slices.Contains(LowPriorityLanguages, uniqueKnownLanguages[1]):
+			return uniqueKnownLanguages[0], nil
```
RonFed, line 119: why do we need the first case?
since Python is in LowPriorityLanguages and Java not, it should be handled by the other two, no?

RonFed, line 119: what would happen if we have 2 languages from the low priority ones? will it be ordered by A-Z and we would prefer C++?

odiglet/pkg/kube/runtime_details/inspection_test.go
```diff
@@ -0,0 +1,125 @@
...
+	languages := []common.ProgrammingLanguage{common.GoProgrammingLanguage}
+
+	result, selectionError := SelectContainerMainLanguage(languages)
+
+	assert.NoError(t, selectionError)
+	assert.Equal(t, common.GoProgrammingLanguage, result)
+}
+
+func Test_SelectMainLanguage_JavaWithHighPriority_JavaSelected(t *testing.T) {
+	// Java paired with another high-priority language — Java always wins
```
RonFed, line 36: I don't think that is a correct behavior honestly

odiglet/pkg/kube/runtime_details/inspection_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+	assert.NoError(t, selectionError)
+	assert.Equal(t, common.PythonProgrammingLanguage, result)
+}
+
+func Test_SelectMainLanguage_TwoHighPriorityNonJava_ReturnsError(t *testing.T) {
+	// two high-priority non-Java languages — ambiguous, should error
+	languages := []common.ProgrammingLanguage{common.GoProgrammingLanguage, common.DotNetProgrammingLanguage}
+
+	result, selectionError := SelectContainerMainLanguage(languages)
```
RonFed, line 99: since we are sorting in the code, which is done before calling `SelectContainerMainLanguage` I think these tests not fully represent the actual code - should we move the sorting into `SelectContainerMainLanguage`?

RonFed (changes requested):

odiglet/pkg/kube/runtime_details/inspection.go
natyharary: Exactly. That's the determinism per what @tamirdavid1 talked about

natyharary: Python + Java is just one possible combo, we could have Python + Rust (low + "regular"), or Python + Ruby (low + "regular")

We could also have Java + Javascript or Java + C (high prioirty + "regular") and then we'll also select Java - and then we would need the first case

odiglet/pkg/kube/runtime_details/inspection_test.go
natyharary: That's what @tamirdavid1 and I talked about and what I understood from the linear ticket (https://linear.app/odigos/issue/CORE-825/handle-multi-language-containers-deterministic-selection-or-explicit)

If we don't want Java to be selected over all other languages, we can remove that.
But in general, those are heuristics and a catch-all which naturally can be wrong.

What did you see usually when in multi runtime envs? Do we even have a higher priority language that always should be selected?

odiglet/pkg/kube/runtime_details/inspection_test.go
natyharary: Yeah, I actually added the sorting after writing those tests so I agree, they do not represent the actual code. Moved it into `SelectContainerMainLanguage()`

natyharary:
decided not be done atm

PR #4548, opened 2026-04-01
CORE-778 | feat(profiling): OTLP profile ingest and in-memory slot store for UI

shivanshuraj1333:
### Description
When a user enables profiling via Helm:

```
profiling:
  enabled: true
```
and runs helm upgrade, the UI pod should begin accepting OTLP profiles without requiring a restart. This PR adds the missing plumbing in the UI to support that flow.

Changes:
- Shared OTLP receiver on port 4317:
    - The UI runs a single OTLP gRPC receiver that handles metrics and profiles.
- Dynamic profiles ingest (no restart required):
    - Profiles ingest is gated by profiling.enabled from the effective-config ConfigMap. A Kubernetes cache watcher tracks changes and flips ingest on/off at runtime.
    
- In-memory profile store:
  - Introduces a purpose-built in-memory store for profiles with sensible defaults, configurable via environment variables (limits/TTL/cleanup interval).

```release-note
None
```


frontend/services/collector_profiles/flamegraph/pyroscope_convert.go
```diff
@@ -0,0 +1,299 @@
+// Pyroscope OTLP conversion: github.com/grafana/pyroscope/pkg/ingester/otlp.ConvertOtelToGoogle
+// (same code path as Grafana Pyroscope ingest). Used only via SamplesFromOTLPChunk.
+package flamegraph
+
+import (
```
shivanshuraj1333: please add some unit tests on some test data, so that we can verify things happening end to end

frontend/services/collector_metrics/collector_metrics.go
```diff
@@ -188,8 +190,16 @@ func NewOdigosMetrics() *OdigosMetricsConsumer {
 	}
 }
 
-// Run starts the OTLP receiver and the notifications loop for receiving and processing the metrics from different Odigos collectors
-func (c *OdigosMetricsConsumer) Run(ctx context.Context, odigosNS string) {
+// Run serves OTLP gRPC (metrics; profiles when profilingEnabled) on otlpGrpcPort, or consts.OTLPPort if unset.
+// The OTLP receiver factory default HTTP server remains enabled unless configured elsewhere.
+func (c *OdigosMetricsConsumer) Run(ctx context.Context, odigosNS string, nextProfiles xconsumer.Profiles, profilingEnabled bool, otlpGrpcPort int) {
```
damemi: maybe we should rename `OdigosMetricsConsumer` (and this package) to something like `OdigosTelemetryConsumer`


or perhaps separate consumer objects for Metrics and Profiles that run separately instead of coupled into one, then you don't need the separate conditionals in here too for profiling

frontend/main.go
```diff
@@ -323,12 +333,30 @@ func main() {
...
 
+	recvOn, otlpGrpcPort, maxSlots, ttlSec, slotMaxBytes, cleanupInt, profCfgErr := services.ResolveProfilingFromEffectiveConfig(ctx, k8sCacheClient)
+	if profCfgErr != nil {
+		log.Info("profiling: could not load effective config; receiver will stay disabled", "err", profCfgErr)
+	}
+	profileStore := collectorprofiles.NewProfileStore(maxSlots, ttlSec, slotMaxBytes, cleanupInt)
+	profileStore.RunCleanup(ctx)
+	defer profileStore.StopCleanup()
+	var profConsumer xconsumer.Profiles
+	if recvOn {
```
damemi: should the NewProfileStore and profileStore.RunCleanup() also be gated in this check if the receiver is on?

frontend/services/profiling_startup.go
```diff
@@ -0,0 +1,78 @@
...
+//     overrides maxSlots, slot TTL, and slotMaxBytes via applyProfilingUiOverrides (same keys as Helm
+//     writes into odigos-configuration).
+//  3. If effective config does not enable profiling, fall back to Helm deployment config
+//     (GetHelmDeploymentConfig) so the UI can still receive profiles before the effective ConfigMap catches up.
+//  4. receiverOn is true only when the chosen config (effective or Helm fallback) has profiling enabled
+//     per ProfilingPipelineActive / ProfilingEnabled (explicit enabled: true).
+//
+// Env vars remain useful for tests and operators who do not patch odigos-configuration; they are not
+// required when Helm + effective config define profiling.ui.
+func ResolveProfilingFromEffectiveConfig(ctx context.Context, c client.Client) (receiverOn bool, otlpGrpcPort int, maxSlots, ttlSec, slotMaxBytes int, cleanupInterval time.Duration, err error) {
```
damemi: lot of return values, could the profiling config struct be used directly?

frontend/services/collector_profiles/utils.go
```diff
@@ -0,0 +1,119 @@
...
+	odigosconsts "github.com/odigos-io/odigos/common/consts"
+	"github.com/odigos-io/odigos/frontend/services/common"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
+)
+
+// NormalizeWorkloadKind maps API/UI strings to canonical WorkloadKind values for source keys.
+// This is Odigos-specific (not from Pyroscope): GraphQL and resource attributes may use mixed casing
+// or synonyms; keys must match k8sconsts and the semconv branches in SourceKeyFromResource.
+func NormalizeWorkloadKind(kindStr string) k8sconsts.WorkloadKind {
```
damemi: did you look at using this already? https://github.com/odigos-io/odigos/blob/main/k8sutils/pkg/workload/workloadkinds.go#L93

frontend/services/common/profiling_store_ref.go
```diff
@@ -0,0 +1,11 @@
+package common
+
+// ProfileStoreRef is the subset of the in-memory profiling store used by GraphQL and OTLP ingestion.
+type ProfileStoreRef interface {
+	StartViewing(sourceKey string)
+	RemoveSlot(sourceKey string)
+	GetProfileData(sourceKey string) [][]byte
+	MaxSlots() int
+	DebugSlots() (activeKeys []string, keysWithData []string)
+	MemoryStats() (totalBytes int, maxSlots int, slotMaxBytes int, ttlSeconds int)
```
damemi: a struct here might be easier to manage in the future for the memory stats return values

frontend/services/collector_profiles/defaults.go
```diff
@@ -0,0 +1,16 @@
...
+
+// Canonical defaults for the UI in-memory profiling store.
+// Keep in sync with helm/odigos values.yaml profiling.ui and odigos-configuration-cm.yaml template defaults.
+// Store ticker / env-only knobs that are not in Helm also belong here so config.go and store.go stay aligned.
+const (
+	DefaultProfilingMaxSlots       = 24
+	DefaultProfilingSlotMaxBytes   = 8 * 1024 * 1024 // 8 MiB
+	DefaultProfilingSlotTTLSeconds = 120             // seconds
+	// DefaultProfilingCleanupIntervalSeconds is the ProfileStore TTL sweep ticker period (pod-local only).
+	DefaultProfilingCleanupIntervalSeconds = 15
```
damemi: these seem like a lot of things to keep in sync with helm/cm settings. I wonder do these even really need to be configurable from a user perspective? That could at least clean up the helm requirement

frontend/services/profiles/store.go
```diff
@@ -0,0 +1,208 @@
...
+		maxSlots:        maxSlots,
+		ttlSeconds:      ttlSeconds,
+		slotMaxBytes:    slotMaxBytes,
+		cleanupInterval: cleanupInterval,
+	}
+	return s
+}
+
+// StartViewing ensures a slot exists for the given source key and refreshes LastRequestAt.
+// If the store is full, the slot with the oldest LastRequestAt is evicted first.
```
damemi: when is this called?

frontend/services/collector_profiles/consumer.go
```diff
@@ -0,0 +1,61 @@
...
+			return nil
+		}
+
+		for idx := 0; idx < numResources; idx++ {
+			attrs := resourceProfiles.At(idx).Resource().Attributes()
+			sourceKey, ok := SourceKeyFromResource(attrs)
+			if !ok {
+				continue
+			}
+			if !store.IsActive(sourceKey) {
```
damemi: maybe I'm missing it but how does a source key get created so it's active?

frontend/services/collector_profiles/store.go
```diff
@@ -0,0 +1,208 @@
...
+
+	if slot, ok := s.slots[sourceKey]; ok {
+		slot.LastRequestAt = now
+		return
+	}
+
+	if len(s.slots) >= s.maxSlots {
+		var oldestKey string
+		var oldestTime time.Time
+		first := true
```
damemi: instead of the `first` flag could you just initialize `oldestTime=now`?

frontend/services/profiles/store.go
```diff
@@ -0,0 +1,208 @@
...
+// IsActive returns true if the source has a slot (and is within TTL if cleanup has run).
+func (s *ProfileStore) IsActive(sourceKey string) bool {
+	s.mu.RLock()
+	_, ok := s.slots[sourceKey]
+	s.mu.RUnlock()
+	return ok
+}
+
+// DebugSlots returns active source keys and which have non-empty buffers (for debugging).
+func (s *ProfileStore) DebugSlots() (activeKeys []string, keysWithData []string) {
```
damemi: do you have an example of what this is used for in debugging?

frontend/services/collector_metrics/collector_metrics.go
tamirdavid1: Agree about separation of consumers

frontend/services/collector_metrics/collector_metrics.go
```diff
@@ -188,8 +190,16 @@ func NewOdigosMetrics() *OdigosMetricsConsumer {
...
+	lg := commonlogger.LoggerCompat().With("subsystem", "collector-metrics", "component", "otlp-receiver")
+
+	// In-cluster collectors send OTLP to this process; the port must match gateway/collector config (default consts.OTLPPort).
+	listenPort := otlpGrpcPort
+	if listenPort <= 0 {
```
tamirdavid1: I think this is a redundant call?
ResolveProfilingFromEffectiveConfig already checking it?
also I think even not needed in `ResolveProfilingFromEffectiveConfig` as it can be validated and set the default value in the effective config creation (scheduler)

tamirdavid1: and then we can use directly `otlpGrpcPort`

frontend/services/collector_metrics/collector_metrics.go
```diff
@@ -222,23 +232,56 @@ func (c *OdigosMetricsConsumer) Run(ctx context.Context, odigosNS string) {
...
+	host := componenttest.NewNopHost()
+	set := receivertest.NewNopSettings(f.Type())
+
+	// Register profiles on the shared OTLP receiver before metrics so the same otlpReceiver
+	// instance has nextProfiles set before CreateMetrics; both must be registered before Start.
+	var profilesReceiver xreceiver.Profiles
+	if profilingEnabled && nextProfiles != nil {
+		xf, xok := f.(xreceiver.Factory)
+		if !xok {
+			lg.Info("OTLP receiver factory does not support profiles; continuing with metrics only")
```
tamirdavid1: it is valid that our collector wont support it? i dont think so.
also, lets try to keep logging in standard of log.Info or whatever you choose (never saw lg)

frontend/services/collector_metrics/collector_metrics.go
```diff
@@ -222,23 +232,56 @@ func (c *OdigosMetricsConsumer) Run(ctx context.Context, odigosNS string) {
...
+	if err := metricsReceiver.Start(ctx, host); err != nil {
+		lg.Error("failed to start OTLP metrics receiver", "err", err)
+	}
+	if profilesReceiver != nil {
+		if err := profilesReceiver.Start(ctx, host); err != nil {
+			lg.Error("failed to start OTLP profiles receiver", "err", err)
+		}
+	}
+	if profilingEnabled && nextProfiles != nil && profilesReceiver == nil {
+		lg.Info("OTLP profiles receiver was not created; profiling ingestion disabled until fixed (gateway may log Unimplemented on ProfilesService)")
```
tamirdavid1: sounds like a warning ?

frontend/services/collector_profiles/config.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	sec := DefaultProfilingCleanupIntervalSeconds
+	if s := os.Getenv(envCleanupIntervalSeconds); s != "" {
+		if v, err := strconv.Atoi(s); err == nil && v >= 0 {
+			sec = v
+		}
+	}
+	cleanupInterval = time.Duration(sec) * time.Second
+	return maxSlots, ttlSeconds, slotMaxBytes, cleanupInterval
+}
```
tamirdavid1:
```suggestion
package collectorprofiles

import (
	"os"
	"strconv"
	"time"
)

// Optional env overrides for the in-memory profile store (local dev / ops). In-cluster defaults usually
// come from odigos-configuration (profiling.ui) via ResolveProfilingFromEffectiveConfig, not from the UI pod env.
const (
	envSlotTTLSeconds         = "PROFILES_SLOT_TTL_SECONDS"
	envMaxSlots               = "PROFILES_MAX_SLOTS"
	envSlotMaxBytes           = "PROFILES_SLOT_MAX_BYTES"
	envCleanupIntervalSeconds = "PROFILES_CLEANUP_INTERVAL_SECONDS"
)

func intEnvOrDefault(key string, defaultVal int) int {
	if s := os.Getenv(key); s != "" {
		if v, err := strconv.Atoi(s); err == nil && v >= 0 {
			return v
		}
	}
	return defaultVal
}

// StoreConfigFromEnv reads profiling store limits from the environment (defaults on unset/invalid).
func StoreConfigFromEnv() (maxSlots int, ttlSeconds int, slotMaxBytes int, cleanupInterval time.Duration) {
	maxSlots = intEnvOrDefault(envMaxSlots, DefaultProfilingMaxSlots)
	ttlSeconds = intEnvOrDefault(envSlotTTLSeconds, DefaultProfilingSlotTTLSeconds)
	slotMaxBytes = intEnvOrDefault(envSlotMaxBytes, DefaultProfilingSlotMaxBytes)
	cleanupInterval = time.Duration(intEnvOrDefault(envCleanupIntervalSeconds, DefaultProfilingCleanupIntervalSeconds)) * time.Second
	return maxSlots, ttlSeconds, slotMaxBytes, cleanupInterval
}
}
```

frontend/services/collector_profiles/consumer.go
```diff
@@ -0,0 +1,61 @@
...
+	singleResourceChunk := pprofile.NewProfiles()
+	incomingBatch.Dictionary().CopyTo(singleResourceChunk.Dictionary())
+	resourceProfile.CopyTo(singleResourceChunk.ResourceProfiles().AppendEmpty())
+	chunkBytes, marshalErr := jsonMarshaler.MarshalProfiles(singleResourceChunk)
+	if marshalErr != nil {
+		logger.Warn("store_chunk", "sourceKey", sourceKey, "err", marshalErr)
+		return
+	}
+	store.AddProfileData(sourceKey, chunkBytes)
+}
```
tamirdavid1:
```suggestion
package collectorprofiles

import (
	"context"

	commonlogger "github.com/odigos-io/odigos/common/logger"
	"go.opentelemetry.io/collector/consumer"
	"go.opentelemetry.io/collector/consumer/xconsumer"
	"go.opentelemetry.io/collector/pdata/pprofile"
)

var jsonMarshaler pprofile.JSONMarshaler

// NewProfilesConsumer ingests OTLP profiles: each resource becomes one JSON chunk in the store
// when its source key is active (profiling slot open).
func NewProfilesConsumer(store *ProfileStore) (xconsumer.Profiles, error) {
	return xconsumer.NewProfiles(func(ctx context.Context, incomingBatch pprofile.Profiles) error {
		resourceProfiles := incomingBatch.ResourceProfiles()
		numResources := resourceProfiles.Len()
		if numResources == 0 {
			return nil
		}

		for idx := 0; idx < numResources; idx++ {
			attrs := resourceProfiles.At(idx).Resource().Attributes()
			sourceKey, ok := SourceKeyFromResource(attrs)
			if !ok || !store.IsActive(sourceKey) {
				continue
			}
			appendResourceProfileJSONChunk(store, sourceKey, incomingBatch, resourceProfiles, idx)
		}
		return nil
	}, consumer.WithCapabilities(consumer.Capabilities{MutatesData: false}))
}

// appendResourceProfileJSONChunk extracts one resource from the batch, marshals it as OTLP Profiles JSON,
// and appends the bytes to the store slot for that resource's source key (caller ensures slot is active).
func appendResourceProfileJSONChunk(store *ProfileStore, sourceKey string, incomingBatch pprofile.Profiles, resourceProfiles pprofile.ResourceProfilesSlice, resourceIndex int) {
	logger := commonlogger.LoggerCompat().With("subsystem", "backend-profiling")
	singleResourceChunk := pprofile.NewProfiles()
	incomingBatch.Dictionary().CopyTo(singleResourceChunk.Dictionary())
	resourceProfiles.At(resourceIndex).CopyTo(singleResourceChunk.ResourceProfiles().AppendEmpty())
	chunkBytes, err := jsonMarshaler.MarshalProfiles(singleResourceChunk)
	if err != nil {
		logger.Warn("store_chunk", "sourceKey", sourceKey, "err", err)
		return
	}
	store.AddProfileData(sourceKey, chunkBytes)
}

```

frontend/services/profiles/store.go
```diff
@@ -0,0 +1,208 @@
...
+// RemoveSlot deletes the slot and buffered OTLP data for sourceKey (e.g. user closed the profiling UI).
+// When no slots remain, profile buffer memory returns to ~zero aside from store metadata.
+func (s *ProfileStore) RemoveSlot(sourceKey string) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	delete(s.slots, sourceKey)
+}
+
+// MaxSlots returns the maximum number of concurrent profiling slots (services).
+func (s *ProfileStore) MaxSlots() int {
```
tamirdavid1, line 106: in use ?

frontend/services/profiles/store.go
```diff
@@ -0,0 +1,208 @@
...
+
+	s.slots[sourceKey] = &Slot{
+		LastRequestAt: now,
+		Buffer:        NewBoundedBuffer(s.slotMaxBytes),
+	}
+}
+
+// RemoveSlot deletes the slot and buffered OTLP data for sourceKey (e.g. user closed the profiling UI).
+// When no slots remain, profile buffer memory returns to ~zero aside from store metadata.
+func (s *ProfileStore) RemoveSlot(sourceKey string) {
```
tamirdavid1, line 78: in use ?

frontend/services/collector_profiles/store.go
```diff
@@ -0,0 +1,208 @@
...
+
+// MaxSlots returns the maximum number of concurrent profiling slots (services).
+func (s *ProfileStore) MaxSlots() int {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return s.maxSlots
+}
+
+// MemoryStats returns total bytes buffered across slots and the configured limits (for UI / debugging).
+func (s *ProfileStore) MemoryStats() (totalBytes int, maxSlots int, slotMaxBytes int, ttlSeconds int) {
```
tamirdavid1: in use ?

frontend/services/profiles/store.go
```diff
@@ -0,0 +1,208 @@
...
+		return
+	}
+	buf.Add(chunk)
+}
+
+// GetProfileData returns a snapshot of the buffer for the given source key.
+// Returns nil if the source has no slot.
+// Holds a single write lock for the full operation to prevent the cleanup
+// goroutine from evicting the slot between the existence check and the refresh.
+func (s *ProfileStore) GetProfileData(sourceKey string) [][]byte {
```
tamirdavid1, line 151: in use?

frontend/services/profiles/store.go
tamirdavid1: also this is used somewhere ?

frontend/services/collector_profiles/utils.go
```diff
@@ -0,0 +1,119 @@
...
+	case "staticpod":
+		return k8sconsts.WorkloadKindStaticPod
+	default:
+		return k8sconsts.WorkloadKind(kindStr)
+	}
+}
+
+// SourceKeyFromSourceID returns a stable string key for the given SourceID.
+// Format: "namespace/kind/name" so it matches keys derived from profile resource attributes.
+func SourceKeyFromSourceID(id common.SourceID) string {
```
tamirdavid1: in use?
please go over all functions and make sure we dont have funs that not in use

frontend/services/collector_profiles/utils.go
```diff
@@ -0,0 +1,119 @@
...
+func getStr(attrs pcommon.Map, key string) (string, bool) {
+	v, ok := attrs.Get(key)
+	if !ok {
+		return "", false
+	}
+	return v.Str(), true
+}
+
+// allNamesArePlaceholders reports whether every frame name is synthetic (no resolved symbols).
+func allNamesArePlaceholders(names []string) bool {
```
tamirdavid1: in use ?

frontend/services/profiling_startup.go
```diff
@@ -0,0 +1,78 @@
...
+	maxSlots, ttlSec, slotMaxBytes, cleanup := collectorprofiles.StoreConfigFromEnv()
+	otlpGrpcPort = odigosconsts.OTLPPort
+
+	cfg, err := GetEffectiveConfig(ctx, c)
+	if err != nil {
+		return false, otlpGrpcPort, maxSlots, ttlSec, slotMaxBytes, cleanup, err
+	}
+
+	var useCfg *common.OdigosConfiguration
```
tamirdavid1: I didn't understood this concept, the Idea of the effective config that it is created/updated based on the order why this section is needed.

frontend/services/collector_profiles/store.go
```diff
@@ -0,0 +1,208 @@
...
+// IsActive returns true if the source has a slot (and is within TTL if cleanup has run).
+func (s *ProfileStore) IsActive(sourceKey string) bool {
+	s.mu.RLock()
+	_, ok := s.slots[sourceKey]
+	s.mu.RUnlock()
+	return ok
+}
+
+// DebugSlots returns active source keys and which have non-empty buffers (for debugging).
+func (s *ProfileStore) DebugSlots() (activeKeys []string, keysWithData []string) {
```
shivanshuraj1333: Used by GraphQL `profilingSlots` for observability/debug (active keys + keys with buffered data). Helpful when validating slot activation and TTL cleanup behavior.

frontend/services/collector_profiles/utils.go
shivanshuraj1333: `SourceKeyFromSourceID` is used by the profiling service (`EnableProfilingForSource`, `ReleaseProfilingForSource`, `GetProfilingForSource`) in the GraphQL branch. I can also run a final dead-code pass before merge once both backend + GraphQL splits are together.

frontend/services/profiling_startup.go
shivanshuraj1333: Good question. This section handles config timing precedence: we prefer effective config when available, but fall back to Helm deployment config so profiling can still come up before effective config is materialized/synced. Also, if effective config explicitly sets `profiling.enabled=false`, we do not fall back to Helm to avoid re-enabling it accidentally.

frontend/services/profiling_startup.go
shivanshuraj1333: Agreed. Returning a dedicated runtime config struct would be cleaner than multiple return values. I left it as-is to keep this split minimal, but I can take this as a focused follow-up refactor.

frontend/services/collector_profiles/utils.go
shivanshuraj1333: https://github.com/odigos-io/odigos/pull/4555/changes#diff-764fc3111c7aef297fcb84ca189be9ac135fc46f9f1456ca87a5e0367ff3fe44R23

frontend/services/collector_profiles/utils.go
shivanshuraj1333: https://github.com/odigos-io/odigos/blob/87b0891aff25d09a2e2b722f3cdc282e1dfa41cf/frontend/services/collector_profiles/profiling_service.go#L34

frontend/services/profiles/store.go
shivanshuraj1333: https://github.com/odigos-io/odigos/blob/87b0891aff25d09a2e2b722f3cdc282e1dfa41cf/frontend/services/collector_profiles/profiling_service.go#L81

frontend/services/collector_profiles/store.go
shivanshuraj1333: Yes these are used in later PRs, let me do cleanup if there's any

frontend/services/profiles/store.go
shivanshuraj1333: https://github.com/odigos-io/odigos/blob/87b0891aff25d09a2e2b722f3cdc282e1dfa41cf/frontend/services/collector_profiles/profiling_service.go#L58

frontend/services/profiles/store.go
shivanshuraj1333: https://github.com/odigos-io/odigos/blob/87b0891aff25d09a2e2b722f3cdc282e1dfa41cf/frontend/services/collector_profiles/profiling_service.go#L40

frontend/services/collector_metrics/collector_metrics.go
shivanshuraj1333: Version mismatch can probably cause this, and yes improving the naming convention for the logger etc

frontend/services/collector_metrics/collector_metrics.go
shivanshuraj1333: Yeah I think port was hardocded and I was thinking if it makes sense to make frontend code more cleaner and expose such variables as configuration, but then I reverted the change to open a separate PR for such things. This left as a cleanup, let me fix it.

frontend/services/collector_profiles/defaults.go
shivanshuraj1333: We have default values setup, a user just needs to do
```yaml
profiling:
    enabled: true
```
The remaining settings are for them to fine tune the memory consumption of profiling, say there's a user running frontend with lower limits, then our default of 8MB *24 = ~200MB would be too much, or say if they want to increase the number of services that they can see profiling data, then they might not want to use the default settings.

frontend/services/collector_profiles/consumer.go
shivanshuraj1333: the graphQL side changes are present in the later PR, and this is how it is being used https://www.notion.so/keyval/Profiling-GraphQL-Examples-33d5ac7207b18051ae89cc6d65fbbe2f?source=copy_link

shivanshuraj1333: https://github.com/odigos-io/odigos/pull/4555/changes/BASE..87b0891aff25d09a2e2b722f3cdc282e1dfa41cf#diff-e8839ed48fb77894e1449aa2a29e9b775a274ef41ab9da74887164ea63b227b4R80

https://github.com/odigos-io/odigos/pull/4555/changes/BASE..87b0891aff25d09a2e2b722f3cdc282e1dfa41cf#diff-603b76b777cb7aa8474dc7bb7a1ef22a996afe958daff319b96b26a5f90eec5bR76

frontend/services/collector_profiles/store.go
shivanshuraj1333: https://github.com/odigos-io/odigos/pull/4555/changes#diff-603b76b777cb7aa8474dc7bb7a1ef22a996afe958daff319b96b26a5f90eec5bR63

frontend/services/profiles/store.go
shivanshuraj1333: https://github.com/odigos-io/odigos/pull/4555/changes/BASE..87b0891aff25d09a2e2b722f3cdc282e1dfa41cf#diff-e8839ed48fb77894e1449aa2a29e9b775a274ef41ab9da74887164ea63b227b4R80

in the next PR that we have

frontend/services/profiles/store.go
shivanshuraj1333: https://github.com/odigos-io/odigos/pull/4555/changes#diff-603b76b777cb7aa8474dc7bb7a1ef22a996afe958daff319b96b26a5f90eec5bR62

frontend/services/collector_profiles/defaults.go
damemi: > say there's a user running frontend with lower limits, then our default of 8MB *24 = ~200MB would be too much, or say if they want to increase the number of services that they can see profiling data, then they might not want to use the default settings.

this is what I mean, do we know that users will want to configure the frontend limits? Could this be something that's just baked into the sizing presets we have? It's a lot of settings and tbh I don't picture users coming in and wanting to precisely configure how many MB their profiling store holds, that seems very advanced

frontend/services/profiles/store.go
shivanshuraj1333: eg of debugSlots (this helps avoid a lot of debugging steps etc) https://www.notion.so/keyval/Profiling-GraphQL-Examples-33d5ac7207b18051ae89cc6d65fbbe2f?source=copy_link

frontend/services/collector_profiles/config.go
shivanshuraj1333: fixed

frontend/services/collector_profiles/consumer.go
shivanshuraj1333: fixed

frontend/services/profiling_startup.go
shivanshuraj1333: fixed it

frontend/services/collector_profiles/defaults.go
shivanshuraj1333: makes sense, removing the helm and just keeping the cache settings behind a feature flag.

frontend/services/collector_metrics/collector_metrics.go
shivanshuraj1333: Done, implemented two separate consumers for metrics and profiling

common/consts/consts.go
```diff
@@ -156,15 +155,3 @@ var (
 const (
 	ExchangeDir = "/var/exchange"
 )
-
-// UiOtlpGrpcEndpoint returns the OTLP gRPC host:port for the UI Service in the given namespace.
-// Used by the gateway profiles-to-UI exporter and the node collector own-metrics-to-UI exporter.
-func UiOtlpGrpcEndpoint(namespace string) string {
```
shivanshuraj1333, line 162: cleanup, not being used any more, right now we are using the const.OTLP port for starting OTLP receiver for metrics and profiler, if we want to change that behaviour that can be taken up in a separate PR

frontend/services/collector_metrics/collector_metrics.go
```diff
@@ -207,38 +204,10 @@ func (c *OdigosMetricsConsumer) Run(ctx context.Context, odigosNS string) {
 			deleteNotifications: c.deletedChan,
 		})
 		if err != nil {
-			log.Printf("Collector metrics: Error running delete watcher: %v\n", err)
+			log.Error("Error running delete watcher", "err", err)
 		}
 	}()
 
-	// setup the OTLP receiver
-	f := otlpreceiver.NewFactory()
```
shivanshuraj1333, line 215: now setting up shared OTLP receiver (frontend/services/otlp/consumers.go)

helm/odigos/templates/odigos-configuration-cm.yaml
tamirdavid1: use profiling.enabled maybe ??

frontend/services/collector_metrics/node_collector.go
```diff
@@ -198,46 +199,29 @@ func metricAttributesToSourceID(attrs pcommon.Map) (common.SourceID, error) {
...
-			kind = k8sconsts.WorkloadKindCronJob
-			name = cjName
-		} else if jobName, ok := attrs.Get(K8SJobNameKey); ok {
-			kind = k8sconsts.WorkloadKindJob
-			name = jobName
-		} else if rolloutName, ok := attrs.Get(K8SRolloutNameKey); ok {
-			kind = k8sconsts.WorkloadKindArgoRollout
-			name = rolloutName
-		} else {
+		// Fallback: standard OTLP k8s workload name keys (see workload.OTLPWorkloadNameAttrKindPairs).
```
tamirdavid1: I think the prev was move clear then this

frontend/services/collector_metrics/node_collector.go
```diff
@@ -198,46 +199,29 @@ func metricAttributesToSourceID(attrs pcommon.Map) (common.SourceID, error) {
...
-				name, found = attrs.Get(K8SStatefulSetNameKey)
-			case k8sconsts.WorkloadKindDaemonSet:
-				name, found = attrs.Get(K8SDaemonSetNameKey)
-			case k8sconsts.WorkloadKindCronJob:
-				name, found = attrs.Get(K8SCronJobNameKey)
-			case k8sconsts.WorkloadKindJob:
-				name, found = attrs.Get(K8SJobNameKey)
-			case k8sconsts.WorkloadKindArgoRollout:
-				name, found = attrs.Get(K8SRolloutNameKey)
+			for _, pair := range workload.OTLPWorkloadNameAttrKindPairs {
```
tamirdavid1: also here, the prev change was more clear in my opinion

frontend/services/otlp/consumers.go
```diff
@@ -0,0 +1,96 @@
...
+	Metrics  *collectormetrics.OdigosMetricsConsumer
+	Profiles *profiles.OdigosProfilesConsumer
+}
+
+func NewConsumers(metrics *collectormetrics.OdigosMetricsConsumer, profiles *profiles.OdigosProfilesConsumer) *Consumers {
+	return &Consumers{Metrics: metrics, Profiles: profiles}
+}
+
+// Run creates and starts the OTLP receiver until ctx is canceled.
+func (c *Consumers) Run(ctx context.Context, port int) {
```
damemi: This is basically just the same approach as before but wrapped in another layer

I meant actually split the consumers into 2 types: `MetricsConsumer.Run()` and `ProfilesConsumer.Run()`. Then each run function only needs to handle settings for that specific consumer and can handle errors for that one specifically

damemi (dismissed):

frontend/services/profiles/utils.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	return sID.Namespace + "/" + string(sID.Kind) + "/" + sID.Name, true
+}
+
+func intFromEnvOrDefault(key string, def int) int {
+	if v, err := strconv.Atoi(env.GetEnvVarOrDefault(key, strconv.Itoa(def))); err == nil && v > 0 {
+		return v
+	}
+	return def
+}
```
blumamir: should this be in env.go along with the other env utils?

frontend/services/common/otlp_resource_attributes.go
```diff
@@ -0,0 +1,81 @@
...
+// (see odigostrafficmetrics res_attributes_keys and metricAttributesToSourceID history).
+func ResourceAttributesToSourceID(attrs pcommon.Map) (SourceID, error) {
+	ns, ok := attrs.Get(string(semconv.K8SNamespaceNameKey))
+	if !ok {
+		return SourceID{}, errors.New("namespace not found")
+	}
+
+	var kind k8sconsts.WorkloadKind
+	var name pcommon.Value
+	var found bool
```
blumamir:
```suggestion
	var workloadNameFound bool
```

frontend/services/common/otlp_resource_attributes.go
```diff
@@ -0,0 +1,81 @@
...
+// attributes using the same rules as the node collector traffic metrics processor
+// (see odigostrafficmetrics res_attributes_keys and metricAttributesToSourceID history).
+func ResourceAttributesToSourceID(attrs pcommon.Map) (SourceID, error) {
+	ns, ok := attrs.Get(string(semconv.K8SNamespaceNameKey))
+	if !ok {
+		return SourceID{}, errors.New("namespace not found")
+	}
+
+	var kind k8sconsts.WorkloadKind
+	var name pcommon.Value
```
blumamir:
```suggestion
	var workloadName pcommon.Value
```

frontend/services/common/otlp_resource_attributes.go
```diff
@@ -0,0 +1,81 @@
...
+		name, found = odigosWorkloadName, true
+	}
+
+	if odigosKind, ok := attrs.Get(odigosconsts.OdigosWorkloadKindAttribute); ok {
+		kind = k8sconsts.WorkloadKind(odigosKind.Str())
+
+		if !found {
+			switch kind {
+			case k8sconsts.WorkloadKindDeployment:
+				name, found = attrs.Get(string(semconv.K8SDeploymentNameKey))
```
blumamir: try to avoid overloading variants, as it's very hard to read and understand, and it's an invite for bugs.

Use `workloadNameFound` and `workloadKindFound`

frontend/services/common/otlp_resource_attributes.go
```diff
@@ -0,0 +1,81 @@
...
+		} else if cjName, ok := attrs.Get(string(semconv.K8SCronJobNameKey)); ok {
+			kind = k8sconsts.WorkloadKindCronJob
+			name = cjName
+		} else if jobName, ok := attrs.Get(string(semconv.K8SJobNameKey)); ok {
+			kind = k8sconsts.WorkloadKindJob
+			name = jobName
+		} else if rolloutName, ok := attrs.Get(k8sconsts.K8SArgoRolloutNameAttribute); ok {
+			kind = k8sconsts.WorkloadKindArgoRollout
+			name = rolloutName
+		} else {
```
blumamir, line 122: missing deploymentConfig, and StaticPod. what will happen in these cases?

frontend/services/profiles/buffer.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Add appends a full chunk, then evicts whole oldest chunks until total size is at most maxBytes.
+func (b *BoundedBuffer) Add(chunk []byte) {
+	if len(chunk) == 0 {
+		return
+	}
+	if b.maxBytes > 0 && len(chunk) > b.maxBytes {
+		return
+	}
```
blumamir: nit: consider first appending and only then check and drop

frontend/services/profiles/consumer.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// appendResourceProfileChunk extracts one resource from the batch,
+// marshals it as OTLP Profiles protobuf (ExportProfilesServiceRequest wire, pdata ProtoMarshaler),
+// and appends to the store slot.
+func appendResourceProfileChunk(store *ProfileStore, incomingBatch pprofile.Profiles, resourceProfiles pprofile.ResourceProfilesSlice, resourceIndex int) {
+	log := commonlogger.LoggerCompat().With("subsystem", "backend-profiling")
+	resourceProfile := resourceProfiles.At(resourceIndex)
+	attrs := resourceProfile.Resource().Attributes()
+	sourceKey, hasKey := SourceKeyFromResource(attrs)
```
blumamir: We already computed it in the caller function. pass as argument

shivanshuraj1333:
Note: helm upgrade example (values.yaml) for upgrading cache settings via env variable

frontend/services/collector_metrics/node_collector.go
```diff
@@ -30,7 +27,7 @@ func newSourcesMetrics() sourcesMetrics {
 }
 
 func (sm *sourcesMetrics) updateSourceMetrics(dp pmetric.NumberDataPoint, metricName string, nodeCollectorID string) {
-	sID, err := metricAttributesToSourceID(dp.Attributes())
+	sID, err := common.ResourceAttributesToSourceID(dp.Attributes())
```
blumamir: nit: these are not the resource attributes in this case :)

blumamir (approved):

PR #4555, opened 2026-04-04
CORE-778 | feat(profiling): Store and Flamegraph APIs and OTLP to Flamegraph conversion pipeline

shivanshuraj1333:

Ingest to cache when key is active was done here https://github.com/odigos-io/odigos/pull/4548
and from cache to API responses is being done in this PR.

PR flow is converting OTLP bytes stored in cache keys to flamegraph json response, and also store some cache stats for debugging, all exposes via Graph queries and mutations.

Cache Buffer ([][]byte) → OTLP/proto decode → [famegraph building pipeline using third party libraries]→ profileJson response

```release-note
None
```

cherry-pick: v1.25


damemi:
xref https://github.com/odigos-io/odigos/pull/4548

frontend/services/collector_profiles/profile_builder.go
```diff
@@ -0,0 +1,68 @@
...
+		Metadata:    meta,
+		Timeline:    pyroscopeTimeline(fb.NumTicks, startTimeSec),
+		Groups:      nil,
+		Heatmap:     nil,
+		Symbols:     nil,
+	}
+}
+
+// Pyroscope web UI expects this metadata shape for single CPU-style profiles (historical JSON contract).
+const (
```
tamirdavid1: lets create a const file for it, also probably more things can go there

tamirdavid1 (commented):
In general what missing to me is good readme.md or comments on top of each strcuts file (tree/flamebearer) that explain the high level struct of flamegraph why we need to convert it to pyroscope etc.
the AI comment above each func say nothing for someone who reads it once (most of them can be delete IMO)

frontend/services/profiles/profiling_service.go
```diff
@@ -0,0 +1,99 @@
...
+
+// EnableProfilingForSource opens or refreshes the profiling slot for a workload.
+func EnableProfilingForSource(store common.ProfileStoreRef, namespace, kindStr, name string) (*EnableProfilingOutput, error) {
+	id, err := SourceIDFromStrings(namespace, kindStr, name)
+	if err != nil {
+		return nil, err
+	}
+	key := SourceKeyFromSourceID(id)
+	store.StartViewing(key)
+	activeKeys, _ := store.DebugSlots()
```
damemi, line 36: Ah so this is what I asked about in https://github.com/odigos-io/odigos/pull/4548#discussion_r3047640986, nit on the name but it's being used for more than just debug

frontend/go.mod
```diff
@@ -218,10 +360,14 @@ require (
...
+	// Match github.com/grafana/pyroscope: fork includes APIs required by grafana/dskit (NodeState, etc.)
+	github.com/hashicorp/memberlist => github.com/grafana/memberlist v0.3.1-0.20251126142931-6f9f62ab6f86
 	github.com/odigos-io/odigos/api => ../api
 	github.com/odigos-io/odigos/common => ../common
 	github.com/odigos-io/odigos/config => ../config
 	github.com/odigos-io/odigos/destinations => ../destinations
 	github.com/odigos-io/odigos/k8sutils => ../k8sutils
 	github.com/odigos-io/odigos/odigosauth => ../odigosauth
+	// Pyroscope expects patched yaml.v3 (DecodeWithOptions); see grafana/pyroscope go.mod
+	gopkg.in/yaml.v3 => github.com/colega/go-yaml-yaml v0.0.0-20220720105220-255a8d16d094
```
damemi: pinning these to grafana forks is interesting, are these functions (DecodeWithOptions and grafana/dskit) actually needed here are are they just conflicted through the broad `grafana/pyroscope` dependency?

frontend/services/profiles/chunk_time.go
```diff
@@ -0,0 +1,32 @@
...
+						minNano = nano
+					}
+				}
+			}
+		}
+	}
+	if minNano == 0 {
+		return 0
+	}
+	return minNano / 1e9
```
damemi, line 33: you can just `return minNano / 1e9` (`0/1e9=0`)

frontend/services/profiles/profile_builder.go
```diff
@@ -0,0 +1,60 @@
...
+		Flamebearer: fb,
+		Metadata:    meta,
+		Timeline:    flamebearerTimeline(fb.NumTicks, startTimeSec),
+		Groups:      nil,
+		Heatmap:     nil,
+		Symbols:     nil,
+	}
+}
+
+func flamebearerMetadata(_ int64) flamegraph.FlamebearerMetadata {
```
damemi: what's the unused int64 param?

frontend/services/profiles/profile_builder.go
```diff
@@ -0,0 +1,60 @@
...
+		meta.SymbolsHint = "Symbols unavailable. Ensure the collector sends full OTLP profile dictionaries."
+	}
+	return flamegraph.FlamebearerProfile{
+		Version:     flamebearerJSONVersion,
+		Flamebearer: fb,
+		Metadata:    meta,
+		Timeline:    flamebearerTimeline(fb.NumTicks, startTimeSec),
+		Groups:      nil,
+		Heatmap:     nil,
+		Symbols:     nil,
```
damemi, line 33: are groups/heatmap/symbols used anywhere?

frontend/services/profiles/flamegraph/otlp_profile_convert.go
```diff
@@ -0,0 +1,320 @@
...
+const (
+	// failReasonProfileExtractPrefix is prefixed to ConvertOtelToGoogle extract failures for logging.
+	failReasonProfileExtractPrefix = "otlp_convert_profile_extract_failed:"
+	failReasonNoSamplesFromConvert = "convert_otel_to_google_yielded_no_samples"
+)
+
+// decodeOTLPChunkToSamples unmarshals one OTLP protobuf chunk, runs ConvertOtelToGoogle per
+// profile, and maps the result to Sample slices. ok is false when the payload is invalid or conversion
+// produces no extractable stacks (failReason is for diagnostics/logging).
+func decodeOTLPChunkToSamples(chunk []byte) (samples []Sample, ok bool, failReason string) {
```
damemi: might be better to define failReason as an error type then instead of `ok` and `failReason` you can just check if the err != nil and log that instead

damemi: also doesn't look like failReason is even handled in chunk_samples right now (is it used anywhere else?) it would be useful to wire failures through. you can define the error consts for ones that don't need to be fatal so you can ignore them easier

frontend/services/profiles/flamegraph/otlp_profile_convert.go
```diff
@@ -0,0 +1,320 @@
...
+	otelResource "go.opentelemetry.io/proto/otlp/resource/v1"
+
+	googleProfile "github.com/grafana/pyroscope/api/gen/proto/go/google/v1"
+	"github.com/grafana/pyroscope/pkg/ingester/otlp"
+)
+
+const (
+	// failReasonProfileExtractPrefix is prefixed to ConvertOtelToGoogle extract failures for logging.
+	failReasonProfileExtractPrefix = "otlp_convert_profile_extract_failed:"
+	failReasonNoSamplesFromConvert = "convert_otel_to_google_yielded_no_samples"
```
damemi: do these need to be camel case strings or would full messages here be more useful? also like I mentioned these could be defined errors

frontend/services/profiles/flamegraph/otlp_profile_convert.go
```diff
@@ -0,0 +1,320 @@
...
+		}
+	}
+	if extractedAny {
+		return out, true, ""
+	}
+	return nil, false, fmt.Sprintf("converted_profiles=%d", len(converted))
+}
+
+// extractGoogleProfile reads a pprof Profile from the ingester conversion output struct.
+// Field layout may vary by dependency version; we locate by type first.
```
damemi: could you explain more in this comment what this function needs to do? What is the dependency version that could vary, is it our import of googleProfile.Profile or something that could vary on the incoming otlp?

also just in general why do we need to extract the google profile from the otlp.ConvertOtelToGoogle? I see some mention of unexported field reflection, what are the fields we need and why isn't the converted google profile from that function usable?

frontend/graph/profiling.graphqls
```diff
@@ -0,0 +1,71 @@
...
+    kind: String!
+    name: String!
+  ): ReleaseProfilingResult!
+}
+
+type ProfilingSlots {
+  activeKeys: [String!]!
+  keysWithData: [String!]!
+  """Total bytes currently buffered across all active slots (rolling OTLP chunks)."""
+  totalBytesUsed: Int!
```
blumamir, line 45: Consider this name instead:

```suggestion
  """Total bytes currently buffered across all active slots (rolling OTLP chunks)."""
  totalBytesInUse: Int!
```

"Used" sound like it's the total since started

frontend/graph/profiling.graphqls
```diff
@@ -0,0 +1,71 @@
...
+}
+
+type ReleaseProfilingResult {
+  status: String!
+  sourceKey: String!
+  activeSlots: Int!
+}
+
+type SourceProfilingResult {
+  profileJson: String!
```
blumamir, line 70: add a comment about what this field contains and the format used, otherwise one need to dig the code to figure out how to use the api

frontend/graph/profiling.graphqls
```diff
@@ -0,0 +1,71 @@
...
+  sourceProfiling(
+    namespace: String!
+    kind: String!
+    name: String!
+  ): SourceProfilingResult!
+}
+
+extend type Mutation {
+  """
+  Enable or refresh the profiling slot for a workload. Must be called before sourceProfiling returns data.
```
blumamir, line 23: What do you mean it must be called before? what happens if we call it after?

frontend/graph/profiling.graphqls
```diff
@@ -0,0 +1,71 @@
...
+  releaseSourceProfiling(
+    namespace: String!
+    kind: String!
+    name: String!
+  ): ReleaseProfilingResult!
+}
+
+type ProfilingSlots {
+  activeKeys: [String!]!
+  keysWithData: [String!]!
```
blumamir, line 43: It's not clear from the schema what "keys" mean. all other places we use ns + kind + name.

Is it only for debugging? or is it actually used in the UI?

frontend/graph/profiling.resolvers.go
```diff
@@ -0,0 +1,91 @@
...
+		return &model.ProfilingSlots{
+			ActiveKeys:          []string{},
+			KeysWithData:        []string{},
+			TotalBytesUsed:      0,
+			SlotMaxBytes:        0,
+			MaxSlots:            0,
+			MaxTotalBytesBudget: 0,
+			SlotTTLSeconds:      0,
+		}, nil
+	}
```
blumamir, line 61: nit: Instead of returning empty result, make the property optional and return `nil`. it will communicate that profile is not actually enabled, rather then reporting zero values, and reduce maintenance in any future changes

frontend/graph/profiling.graphqls
```diff
@@ -0,0 +1,71 @@
...
+    kind: String!
+    name: String!
+  ): SourceProfilingResult!
+}
+
+extend type Mutation {
+  """
+  Enable or refresh the profiling slot for a workload. Must be called before sourceProfiling returns data.
+  """
+  enableSourceProfiling(
```
blumamir, line 25: it's called "enable", but it's also for a refreshing. "enable" sounds like a one-time action.

frontend/graph/profiling.graphqls
blumamir, line 23: How is it being used? when should the caller "refresh"? what happens if we don't refresh?

please add to the comment how this mutation is expected to work

frontend/graph/profiling.resolvers.go
```diff
@@ -0,0 +1,91 @@
...
+	result, err := profiles.EnableProfilingForSource(r.ProfileStore, namespace, kind, name)
+	if err != nil {
+		return nil, err
+	}
+	return &model.EnableProfilingResult{
+		Status:      result.Status,
+		SourceKey:   result.SourceKey,
+		MaxSlots:    result.MaxSlots,
+		ActiveSlots: result.ActiveSlots,
+	}, nil
```
blumamir, line 30: can we do
```suggestion
	return profiles.EnableProfilingForSource(r.ProfileStore, namespace, kind, name)
```

?

frontend/graph/profiling.graphqls
```diff
@@ -0,0 +1,71 @@
...
+
+  """
+  Aggregated Pyroscope-shaped profile for a source (profileJson is JSON-encoded FlamebearerProfile).
+  """
+  sourceProfiling(
+    namespace: String!
+    kind: String!
+    name: String!
+  ): SourceProfilingResult!
+}
```
blumamir, line 19: we already have a type for source/workload, please add the profiling under it and do not create a new query

frontend/services/profiles/profile_builder.go
```diff
@@ -0,0 +1,68 @@
+package profiles
+
+import (
+	"github.com/odigos-io/odigos/frontend/services/profiles/flamegraph"
+)
+
+// BuildPyroscopeProfileFromChunks builds a Pyroscope-shaped profile from OTLP JSON chunks.
+func BuildPyroscopeProfileFromChunks(chunks [][]byte) flamegraph.FlamebearerProfile {
```
blumamir, line 8: this is private to the package

```suggestion
func buildPyroscopeProfileFromChunks(chunks [][]byte) flamegraph.FlamebearerProfile {
```

frontend/services/profiles/flamegraph/tree.go
```diff
@@ -0,0 +1,60 @@
...
+type Tree struct {
+	root []*node
+}
+
+type node struct {
+	parent   *node
+	children []*node
+	self     int64
+	total    int64
+	name     string
```
blumamir, line 26: I know these names are coming from pyroscope, but it's not trivial to understand from the code what the expected content of `name` and what `self` means, and what `node` actually represents in this context, or what is the int value stored in `total` represents.

consider adding some more info here to help increase readability and maintainability

frontend/services/profiles/flamegraph/tree.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// InsertStack merges a stack (root-first frame names) with the given value into the tree.
+func (t *Tree) InsertStack(value int64, stack ...string) {
+	if value <= 0 || len(stack) == 0 {
+		return
+	}
+	r := &node{children: t.root}
+	n := r
+	for _, name := range stack {
```
blumamir, line 41: use `stackFrame` or `functionName`, so it's clear what the object we are iterating means

frontend/graph/profiling.graphqls
```diff
@@ -0,0 +1,71 @@
...
+  ): EnableProfilingResult!
+
+  """
+  Drop the profiling slot and buffered OTLP data for this workload (e.g. user closed the profiling panel).
+  """
+  releaseSourceProfiling(
+    namespace: String!
+    kind: String!
+    name: String!
+  ): ReleaseProfilingResult!
```
blumamir, line 38: if the first one is called "enabled" should the reverse operation be "disable"? 

release is the implementation aspect of what is going on under the hood, not the api verb IMO. 

how about "register" and "unregister"? or "watch" and "unwatch"? or "subscribe" and "removeSubscription"?

blumamir (commented):
Thank you @shivanshuraj1333 ,

Overall looks really good, added few nits and suggestions. let's discuss them when you are around

frontend/services/profiles/flamegraph/otlp_profile_convert.go
shivanshuraj1333: for later steps in the pipeline

blumamir (approved):

PR #4573, opened 2026-04-07
fix: strip managedFields from informer caches to reduce data-collection memory

edeNFed:
## Summary

- Strip `managedFields` from the `odigoslogsresourceattrsprocessor` pod metadata informer cache via `SetTransform` — this is the number one in data-collection pods (~20MB / 40% of heap)
- Strip `managedFields` from the `odigosconfigk8sextension` InstrumentationConfig dynamic informer cache
- Add unit tests and benchmarks proving the fix

## Context

Heap profiles from a 116-node production cluster show `FieldsV1.Unmarshal` consuming 12-21MB (30-40% of heap) across **all** nodes. The call stack confirms the source is `metadatainformer.NewFilteredMetadataInformer` in the `odigoslogsresourceattrsprocessor` — NOT the k8sattributes processor (which already strips managedFields via `removeUnnecessaryPodData()`).

The metadata informer caches `PartialObjectMetadata` objects including full `ObjectMeta` with `ManagedFields` (server-side-apply field ownership data). This data is completely unused by the processor but consumes ~40KB per pod. At ~500 pods/node, that's ~20MB wasted per node, pushing data-collection pods past their 128Mi memory limit and causing OOMKills on 19/116 nodes.

## Benchmark Results

```
BenchmarkManagedFieldsMemorySavings/WithManagedFields     20,960,112 B/op  (~20MB)
BenchmarkManagedFieldsMemorySavings/WithoutManagedFields     360,010 B/op  (~0.36MB)
BenchmarkStripManagedFields                                        0 B/op  (zero-alloc transform)
```

## Expected Impact

| Metric | Before | After |
|--------|--------|-------|
| Heap (FieldsV1) | 12-21MB | ~0 |
| Total heap | ~52MB | ~32MB |
| OOMKilled nodes | 19/116 | 0 |

```release-note
Strip managedFields from informer caches in data-collection pods, reducing heap usage by ~20MB per node and eliminating OOMKills at scale.
```

emergency-ticket id: EMR-1439

### DIFF
--- a/collector/extension/odigosconfigk8sextension/informer.go
+++ b/collector/extension/odigosconfigk8sextension/informer.go
@@ -10,6 +10,7 @@ import (
 
 	commonapi "github.com/odigos-io/odigos/common/api"
 
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -50,6 +51,18 @@ func (o *OdigosWorkloadConfig) startInformer(ctx context.Context) error {
 
 	factory := dynamicinformer.NewDynamicSharedInformerFactory(client, resyncPeriod)
 	informer := factory.ForResource(instrumentationConfigGVR).Informer()
+
+	// Strip managedFields from cached objects to reduce memory usage. Mirrors
+	// sigs.k8s.io/controller-runtime/pkg/cache.TransformStripManagedFields. The
+	// nil-check guards against kubernetes/kubernetes#124337.
+	if err = informer.SetTransform(func(obj any) (any, error) {
+		if accessor, err := meta.Accessor(obj); err == nil && accessor.GetManagedFields() != nil {
+			accessor.SetManagedFields(nil)
+		}
+		return obj, nil
+	}); err != nil {
+		return err
+	}
 
 	_, err = informer.AddEventHandler(k8scache.ResourceEventHandlerFuncs{
 		AddFunc:    o.handleInstrumentationConfig,
--- a/collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
+++ b/collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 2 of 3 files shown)


collector/processors/odigoslogsresourceattrsprocessor/internal/kube/client.go
```diff
@@ -102,6 +109,18 @@ func NewMetadataClient(config *rest.Config) (Client, error) {
 	return c, nil
 }
 
+// stripManagedFields removes managedFields from PartialObjectMetadata objects
+// before they are stored in the informer cache. ManagedFields contain
+// server-side-apply field ownership metadata that is unused by this processor
+// but consumes significant memory (~40KB per pod at scale).
+func stripManagedFields(obj any) (any, error) {
```
RonFed, line 121: controller-runtime has the following (which we use in all of our components)
```Go
// TransformStripManagedFields strips the managed fields of an object before it is committed to the cache.
// If you are not explicitly accessing managedFields from your code, setting this as `DefaultTransform`
// on the cache can lead to a significant reduction in memory usage.
func TransformStripManagedFields() toolscache.TransformFunc {
	return func(in any) (any, error) {
		// Nilcheck managed fields to avoid hitting https://github.com/kubernetes/kubernetes/issues/124337
		if obj, err := meta.Accessor(in); err == nil && obj.GetManagedFields() != nil {
			obj.SetManagedFields(nil)
		}

		return in, nil
	}
}
```

should we use this? or just copy it? it seems to handle some edge cases that I'm not sure this approach handels

collector/extension/odigosconfigk8sextension/informer_test.go
RonFed, line 1: don't think this is necessary to commit
it tests `unstructured.RemoveNestedField` not really our code

collector/extension/odigosconfigk8sextension/informer.go
```diff
@@ -51,6 +52,18 @@ func (o *OdigosWorkloadConfig) startInformer(ctx context.Context) error {
...
+	// Strip managedFields from cached objects to reduce memory usage. Mirrors
+	// sigs.k8s.io/controller-runtime/pkg/cache.TransformStripManagedFields. The
+	// nil-check guards against kubernetes/kubernetes#124337.
+	if err = informer.SetTransform(func(obj any) (any, error) {
+		if accessor, err := meta.Accessor(obj); err == nil && accessor.GetManagedFields() != nil {
+			accessor.SetManagedFields(nil)
+		}
+		return obj, nil
+	}); err != nil {
+		return err
```
RonFed, line 64: why not use stripManagedFields function

RonFed (approved):

PR #4593, opened 2026-04-12
CORE-795: common/logger: Implement log files rotation in the common logger

liranm-odigos:
What this PR does / why we need it:
Adds optional log file rotation support to the shared common/logger package using [lumberjack](https://github.com/natefinsh/lumberjack.v2). This is a backward-compatible extension — existing callers of logger.Init(level, component) are unaffected.

Two new functional options are introduced:

WithRotation(cfg) — directs log output to a rotated file instead of stdout, with configurable max size, backup count, retention age, and gzip compression.
WithStdout(true) — when used alongside WithRotation, tees logs to both the rotated file and stdout.
The Init signature changes from Init(levelStr, component string) to Init(levelStr, component string, opts ...Option), which is fully backward-compatible since the variadic options default to the current stdout-only behavior.

This facility is needed by the vm-agent, which runs as a long-lived host process where persisting debug-level logs to disk is critical for post-hoc troubleshooting on customer machines. The actual usage will be in the vm-agent repo.

#### Changelog entry:

```release-note
NONE
```


TheErikNaz (approved):
lgtm

PR #4605, opened 2026-04-13
feat: noisy operations in tail sampling

blumamir:

Fixes: CORE-633

Add "noisy operations" sampling to the tail sampling processor

```release-note
Add "noisy operations" sampling to the tail sampling processor
```


collector/Dockerfile
```diff
@@ -6,7 +6,9 @@ WORKDIR /go/src/collector
 
 ARG TARGETARCH
 ARG RHEL=false
-RUN GOOS=linux GOARCH=$TARGETARCH make build-odigoscol
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    GOOS=linux GOARCH=$TARGETARCH make build-odigoscol
```
RonFed, line 11: 👍

collector/processors/odigostailsamplingprocessor/category/noisyoperations.go
RonFed, line 1: what is the motivation for the category sub directory? what is the planned structure?

collector/processors/odigostailsamplingprocessor/utils.go
```diff
@@ -3,8 +3,70 @@ package odigostailsamplingprocessor
...
+
+					if ruleName != "" {
+						span.Attributes().PutStr("odigos.sampling.trace.deciding_rule.name", ruleName)
+					}
+				}
+
+				if dryRun {
+					span.Attributes().PutBool("odigos.sampling.dry_run", dryRun)
+					span.Attributes().PutBool("odigos.sampling.trace.kept", kept) // can be false to indicate this trace would have been dropped.
+				}
```
RonFed, line 66: should we put these attributes keys in a `const` block? might be useful if we also gonna use some of them for metrics

collector/processors/odigostailsamplingprocessor/processor.go
```diff
@@ -40,12 +45,53 @@ func (p *tailSamplingProcessor) processTraces(ctx context.Context, td ptrace.Tra
...
+		percentageAtMost := 0.0
+		if noisyOperationRule.PercentageAtMost != nil {
+			percentageAtMost = *noisyOperationRule.PercentageAtMost
+		}
+
+		// either drop it, or keep it and add relevant sampling attributes to all spans.
+		keepTrace := tracePercentage <= percentageAtMost
+
+		if keepTrace || p.config.DryRun {
+			enrichSpansWithSamplingAttributes(td, "noisy", noisyOperationRule.Id, noisyOperationRule.Name, percentageAtMost, p.config.DryRun, keepTrace, p.config.SpanSamplingAttributes)
```
RonFed: nit: `enrichSpansWithSamplingAttributes` can be a method of `tailSamplingProcessor` making the call here more minimal since we won't need to pass `p.config.DryRun` and `p.config.SpanSamplingAttributes`

collector/processors/odigostailsamplingprocessor/processor.go
```diff
@@ -40,12 +45,53 @@ func (p *tailSamplingProcessor) processTraces(ctx context.Context, td ptrace.Tra
...
+		// as the rules are evaluated only on the root span.
+		return false, nil
+	}
+
+	tailSamplingConfig, ok := p.getTailSamplingConfig(resource)
+	if !ok {
+		// the tail sampling config is set only if there are actually any rules.
+		// this source is not relevant for noisy operations category.
+		return false, nil
+	}
```
RonFed, line 87: its a bit confusing.
noisy operation is associated with head sampling, so checking tail sampling config here is a bit confusing.

collector/processors/odigostailsamplingprocessor/utils.go
```diff
@@ -3,8 +3,70 @@ package odigostailsamplingprocessor
...
+					span.Attributes().PutStr("odigos.sampling.trace.deciding_rule.id", ruleId)
+					span.Attributes().PutDouble("odigos.sampling.trace.deciding_rule.keep_percentage", keepPercentage)
+
+					if ruleName != "" {
+						span.Attributes().PutStr("odigos.sampling.trace.deciding_rule.name", ruleName)
+					}
+				}
+
+				if dryRun {
+					span.Attributes().PutBool("odigos.sampling.dry_run", dryRun)
```
RonFed: `odigos.sampling.dry_run` will always be true or not existent? it can't be false according to this, right?

collector/processors/odigostailsamplingprocessor/processor.go
RonFed: I would create a "enum" of categories instead of passing "noisy"

collector/processors/odigostailsamplingprocessor/matchers/headsampling.go
```diff
@@ -0,0 +1,70 @@
...
+
+	if operation.ServerAddress != "" && !matchServerAddress(span, operation.ServerAddress) {
+		return false
+	}
+
+	if (operation.TemplatedPath != "" || operation.TemplatedPathPrefix != "") && !matchTemplatedPath(span, operation.TemplatedPath, operation.TemplatedPathPrefix) {
+		return false
+	}
+
+	return true
```
RonFed: can use switch-case pattern if you think it looks better:
```Go
httpMethod, found := getHttpMethod(span)
switch {
case !found:
	return false
case operation.Method != "" && !compareHttpMethod(httpMethod, operation.Method):
	return false
case operation.ServerAddress != "" && !matchServerAddress(span, operation.ServerAddress):
	return false
case (operation.TemplatedPath != "" || operation.TemplatedPathPrefix != "") &&
	!matchTemplatedPath(span, operation.TemplatedPath, operation.TemplatedPathPrefix):
	return false
default:
	return true
}
```

RonFed (approved):

collector/processors/odigostailsamplingprocessor/category/noisyoperations.go
blumamir: There is going to be a lot of code per category, specifically for highly-relevance, and cost-reduction. that is why there is a directory which now has just one file but that is soon to change

collector/processors/odigostailsamplingprocessor/processor.go
blumamir: noisy operations is a category of traces, that can be executed as head or tail depending on the distro support for head sampling and general configuration (tail sampling disabled).

the terminology "noisy-operation" is intended - to decouple the implementation from the required behavior

collector/processors/odigostailsamplingprocessor/utils.go
blumamir: right, if `dryRun` is set, we add this attribute, but if it's unset - this attribute is just noise and we avoid setting it. so it will either be unset or `true`

collector/processors/odigostailsamplingprocessor/processor.go
blumamir: This logic is going to expend with the additional categories which are more complex, and with the metrics collection. that is why I separated it to a directory with a common interface for all 3 categories. 

I can show you the final PR tomorrow so you can understand the big picture

collector/processors/odigostailsamplingprocessor/matchers/headsampling.go
blumamir: like

collector/processors/odigostailsamplingprocessor/category/noisyoperations.go
RonFed: we could have a directory for each category, not sure if that is what you mean but something like
`odigossamplingprocessor/noisyoperation/*.go`
`odigossamplingprocessor/costreduction/*.go`

than, when we use it in code we can have calls like
`noisyoperation.match` or something like that - it gives the function contexts based on their package name which is the more idimoatic way and avoid very long names, but just a suggestion

PR #4606, opened 2026-04-13
refactor(ebpf-logs): replace ext attributes map with unix socket streaming | RUN-691

daemon1024:

Remove the logs_ext eBPF hash map used for per-process resource attributes and stream them over the existing unix socket connection instead, using a text protocol (R<pid>:<attrs> / U<pid>) with persistent connection and snapshot+subscribe

```release-note
feat: move logs attribute handling from ebpf maps to socket exchange
```

### DIFF
--- a/common/unixfd/client.go
+++ b/common/unixfd/client.go
@@ -1,6 +1,7 @@
 package unixfd
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"net"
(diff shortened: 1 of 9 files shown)


common/unixfd/client.go
```diff
@@ -104,6 +105,66 @@ func ConnectAndListenMulti(ctx context.Context, socketPath string, requestType s
...
+			return ctx.Err()
+		}
+
+		err := connectLogsSession(ctx, socketPath, onFD, onAttr)
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+
+		// Session ended (server disconnected or error) — retry after backoff.
+		sleepTime := 2 * time.Second
```
aviadhahami: lets move this out of the ctx

daemon1024: I didn't understand it

aviadhahami: it gets allocated each iteration, so just to hoist it

common/unixfd/logsattr.go
RonFed, line 1: 💯

RonFed (approved):

PR #4609, opened 2026-04-14
feat: add db query collection policy to payload collection rule

blumamir:

Fixes: CORE-913

For db queries, there are 2 variants:

- sanitize - meaning that we redact the query to avoid potential PII data)
- full - the query as-is, might contain PII data

This PR allows the user to choose a policy for collecting db queries:

1. "full" - collect the full db query, even if there is an option to reduct PII data. this is best visibility with PII risk
2. "sanitized" - collect only redacted query string. if missing, avoid collecting at all
3. "sanitized-or-full" - use sanitised query string if available, and fallback to "full" otherwise 

```release-note
NONE
```


common/consts/payloadcollection.go
```diff
@@ -0,0 +1,25 @@
+package consts
+
+type DbQueryCollectionPolicy string
+
+const (
+	// DbQueryCollectionPolicySanitized: Only collect sanitized (redacted) DB query payloads, which avoids collecting PII.
+	// If the sanitization is not supported, the payload will not be collected.
+	DbQueryCollectionPolicySanitized DbQueryCollectionPolicy = "sanitized"
```
RonFed: if the query is truncated in this case - do we collect it?

common/consts/payloadcollection.go
```diff
@@ -0,0 +1,25 @@
+package consts
+
+type DbQueryCollectionPolicy string
```
RonFed:
```suggestion
type DbQuerySanitizationPolicy string
```

api/odigos/v1alpha1/instrumentationrules/payloadcollection.go
```diff
@@ -30,6 +32,10 @@ type DbQueryPayloadCollection struct {
 	// If the payload is larger than the MaxPayloadLength, this parameter will determine if the payload should be partially collected up to the allowed length, or not collected at all.
 	// This is useful if you require some decoding of the payload (like json) and having it partially is not useful.
 	DropPartialPayloads *bool `json:"dropPartialPayloads,omitempty"`
+
+	// The policy to use for collecting the DB query payloads.
+	// If not specified, the default policy will be used (collect sanitized payloads if possible, and fall back to full if sanitization isn't supported).
+	CollectionPolicy *consts.DbQueryCollectionPolicy `json:"collectionPolicy,omitempty"`
```
RonFed:
```suggestion
	SanitizationPolicy *consts.DbQuerySanitizationPolicy `json:"sanitizationPolicy,omitempty"`
```

common/consts/payloadcollection.go
```diff
@@ -0,0 +1,25 @@
...
+	// and fall-back to full collection otherwise, so we won't miss any payloads.
+	DbQueryCollectionPolicySanitizedOrFull DbQueryCollectionPolicy = "sanitized-or-full"
+)
+
+// Higher value = higher priority (more restrictive).
+var DbQueryCollectionPolicyPriority = map[DbQueryCollectionPolicy]int{
+	DbQueryCollectionPolicySanitized:       3,
+	DbQueryCollectionPolicySanitizedOrFull: 2,
+	DbQueryCollectionPolicyFull:            1,
+}
```
RonFed: I think a cleaner approach would be to have a function on the type `DbQuerySanitizationPolicy` which will return its priority (a switch case)

common/consts/payloadcollection.go
blumamir: Based on the configuration that the user supplied. we will either truncate or drop

RonFed (approved):

PR #4611, opened 2026-04-14
CORE-887 | fix: deterministic gateway config creation

tamirdavid1:

We want to ensure that any non-deterministic behavior in gateway configuration updates becomes deterministic, in order to prevent unnecessary reloads on gateway.

```release-note
fix periodic config reload in gateway without real config change. this improve the collector memory stability and availability 
```


blumamir (approved):
🎉

PR #4615, opened 2026-04-14
ci(sts): fix for homebrew repos

aviadhahami:
STS didn't mention the odigos-io/% prefix in the scope;
```release-note
NONE
```

emergency-ticket id: EMR-1462


blumamir (approved):
🎉

PR #4616, opened 2026-04-14
fix(odiglet): single-pass mountinfo scan to reduce allocation churn

aviadhahami:
## Summary

- Rewrite `GroupByPodContainer` as a single-pass `/proc` scan instead of using `process.Group()` with a per-PID mountinfo-reading predicate
- Eliminates 80% of allocation churn (1.9GB lifetime) caused by repeated `bufio.Scanner` and `strings.genSplit` allocations in mountinfo parsing

## Problem (RUN-448)

Odiglets crash with OOM on nodes with many instrumented services. Heap profiling shows the dominant allocator is `mountinfo.GetMountsFromReader` called via `isInPodContainersBatchPredicate`:

```
1016 MB  bufio.Scanner.Text
 515 MB  strings.genSplit
 316 MB  mountinfo.GetMountsFromReader
```

The old code used `process.Group(predicate)` where the predicate opened and scanned `/proc/<pid>/mountinfo` for every PID on the node. With 200+ processes, each `GroupByPodContainer` call triggered 200+ file opens and full line-by-line scans.

## Fix

Single-pass design: iterate `/proc` once, read each PID's mountinfo, match against expected containers, build result directly. No `process.Group()` call, no predicate closure, short-circuit on first match.

- Same API, same return type — all 4 call sites unchanged
- PID reuse safe — mountinfo is read immediately after discovering the PID (same window as before)

## Test plan

- [x] `go build ./pkg/process/` — compiles
- [x] `go build ./pkg/kube/runtime_details/` — consumers compile
- [x] `go vet ./pkg/process/` — passes
- [ ] Deploy to a node with many instrumented services, compare heap profile before/after

### DIFF
--- a/.github/chainguard/publish-helm.sts.yaml
+++ b/.github/chainguard/publish-helm.sts.yaml
@@ -1,0 +1,5 @@
+issuer: https://token.actions.githubusercontent.com
+subject_pattern: repo:odigos-io/odigos:ref:refs/(tags/.*|heads/main)
+
+permissions:
+  contents: write
--- a/.github/workflows/publish-modules-rhel.yml
+++ b/.github/workflows/publish-modules-rhel.yml
@@ -12,7 +12,7 @@ jobs:
         runs-on: ${{ matrix.runner }}
         permissions:
           contents: 'read'
-          id-token: 'write'
+          id-token: write
           packages: 'write'
         strategy:
           matrix:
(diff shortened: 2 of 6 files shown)


aviadhahami:
Replaced by clean branch — reopening as new PR.
https://github.com/odigos-io/odigos/pull/4617

PR #4621, opened 2026-04-15
fix: UI OOM

BenElferink:
# Customer Issue Report: Homepage 502 / UI OOM

**Ticket**: [PLAT-862](https://linear.app/odigos/issue/PLAT-862/homepage-returns-502-and-ui-goes-oom)

**Symptom**: The Odigos UI homepage returns 502 errors and the UI pod goes OOM (out of memory).

---

## Root Cause

1. **SSE amplification loop**: IC had 3 duplicate watchers (tripling event processing). The backend `EventBatcher` used debounce that never settled under sustained load, causing unbounded memory growth. The frontend debounce also never settled, triggering repeated full `GET_WORKLOADS` queries.

2. **GET_WORKLOADS does not scale**: A single query deep-copied ALL ICs, ALL pods, ALL manifests cluster-wide via 6+ `List` operations. gqlgen spawned 170K goroutines (10K workloads × 17 resolver fields = ~430MB stack memory). `GetNamespacesWithWorkloads` and `GetDataStreams` also fired on page load with full cluster-wide API calls.

3. **Memory budget**: The informer cache alone consumes ~15Mi per 1K instrumented workloads at idle. The `GET_WORKLOADS` query adds ~2.5KB per workload for model objects. Both scale linearly — at 20K workloads the cache uses ~309Mi idle and the query adds ~32Mi, leaving no headroom in a 512Mi pod for gqlgen serialization.

---

## Fixes

| #   | Fix                                                                                                                           | Impact                                                   |
| --- | ----------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------- |
| 1   | **EventBatcher throttle** — replaced debounce with pure throttle (3s window, `MaxBatchSize: 100`, `MaxDelay: 10s`)            | Bounded batch memory                                     |
| 2   | **Unified watchers** — IC 3→1, Destination 2→1, API watch connections 5→2                                                     | Eliminated duplicate event processing                    |
| 3   | **Frontend debounce** — 5s debounce now settles between backend throttle bursts                                               | At most 1 refetch per 5s                                 |
| 4   | **Targeted manifest fetching** — per-workload `Get` instead of 6+ cluster-wide `List` when IDs are known                      | Deep copies: O(cluster) → O(requested)                   |
| 5   | **Namespace-scoped pod fetching** — list pods only in namespaces with workloads                                               | Avoids pods from irrelevant namespaces                   |
| 6   | **Lazy namespace loading** — lightweight `GET_NAMESPACES` for page load; `workloads` is a lazy resolver field                 | No workload loading on page load                         |
| 7   | **Zero-copy cache reads** — `DefaultUnsafeDisableDeepCopy: true` on cache options                                             | Eliminates deep-copy allocations on all cache reads      |
| 8   | **DataStreams uses cache** — switched from direct API call to cache client                                                    | Eliminates 20-40MB API fetch on page load                |
| 9   | **Config loading dedup** — `LoadConfig` caches result; `SetFilters`/`SetWorkloadIdsDirect` reuse it                           | Eliminates redundant ConfigMap fetches + YAML unmarshals |
| 10  | **Sequential field pre-computation** — `populateWorkloadFields` computes all fields inline; resolvers return early if pre-set | Goroutines: 170K → 10K (list items only)                 |
| 11  | **Deduplicated status computation** — conditions and healthStatus share computed values instead of calculating each twice     | Halved status allocations per workload                   |
| 12  | **Concurrent query serialization** — mutex on `Workloads` resolver prevents concurrent heavy queries                          | Prevents memory doubling from overlapping requests       |
| 13  | **Metrics map race fix** — `metricsByID`/`removeNodeCollector` now hold `sourcesMu`                                           | Fixes `concurrent map read and map write` crash          |

---

## Memory Profile (from instrumented build)

Measured at 20K instrumented workloads, 512Mi pod limit, `GOMEMLIMIT=409Mi`:

| Stage                              | Alloc  | HeapInuse |
| ---------------------------------- | ------ | --------- |
| Idle (cache synced)                | 309Mi  | 323Mi     |
| After SetFilters (IC listing)      | 338Mi  | 352Mi     |
| After populating 10K/20K workloads | 316Mi  | 337Mi     |
| After populating 20K/20K workloads | 341Mi  | 361Mi     |
| gqlgen serialization (estimated)   | ~360Mi | ~390Mi    |

The query itself adds ~32Mi on top of the 309Mi idle cache. The OOM occurs when gqlgen serializes the 20MB JSON response, pushing HeapInuse past the 409Mi GOMEMLIMIT and triggering aggressive GC that can't free enough before the 512Mi hard limit is hit.

---

## Memory Sizing Guide

The UI pod memory should be sized based on the number of instrumented workloads (workloads with InstrumentationConfig resources, not just pods):

| Instrumented workloads | Recommended memory limit |
| ---------------------- | ------------------------ |
| Up to 2K               | 512Mi (default)          |
| 2K – 5K                | 768Mi                    |
| 5K – 10K               | 1Gi                      |
| 10K – 20K              | 1.5Gi                    |
| 20K+                   | 2Gi                      |

**Formula**: `~15Mi per 1K workloads (cache) + ~2.5Mi per 1K workloads (query) + 200Mi (Go runtime + gqlgen) + 50Mi headroom`.

Note: this scales with **instrumented workloads** (unique deployments/daemonsets/statefulsets with Sources), not with pod count. A cluster with 500 workloads × 40 replicas = 20K pods needs only 512Mi, while 20K workloads × 1 replica = 20K pods needs 1.5Gi.

---

## Release Note

```release-note
Enhance Kubernetes cache handling and event watcher functionality for improved performance and integration.
Optimize GET_WORKLOADS query with targeted manifest fetching, namespace-scoped pod fetching, and zero-copy cache reads.
Pre-compute workload fields sequentially to eliminate excessive goroutine spawning at scale.
```

cherry-pick: v1.23, v1.24

### DIFF
--- a/frontend/graph/loaders/fetchers.go
+++ b/frontend/graph/loaders/fetchers.go
@@ -303,12 +303,13 @@ func fetchWorkloadManifests(ctx context.Context, logger logr.Logger, filters *Wo
 			return workloadManifests, nil
 
 		case k8sconsts.WorkloadKindDeploymentConfig:
-			// Only try to get DeploymentConfig if it's available in the cluster
+			// OpenShift DeploymentConfig is not in the standard K8s API, so it's not
+			// in the informer cache. We use the dynamic client to fetch it directly
+			// from the API server (only on OpenShift clusters where this CRD exists).
 			if !kube.IsOpenShiftDeploymentConfigAvailable {
 				return nil, nil
 			}
 
-			// Use dynamic client for DeploymentConfig
 			gvr := schema.GroupVersionResource{
 				Group:    "apps.openshift.io",
 				Version:  "v1",
@@ -854,6 +855,9 @@ func getWorkloadManifest(ctx context.Context, logger logr.Logger, id model.K8sWo
 		}, nil
 
 	case k8sconsts.WorkloadKindDeploymentConfig:
+		// OpenShift DeploymentConfig is not in the standard K8s API, so it's not
+		// in the informer cache. We use the dynamic client to fetch it directly
+		// from the API server (only on OpenShift clusters where this CRD exists).
 		if !kube.IsDeploymentConfigAvailable() {
 			return nil, nil
 		}
--- a/frontend/graph/loaders/fetchers.go
+++ b/frontend/graph/loaders/fetchers.go
@@ -55,7 +55,7 @@ func timedAPICall[T any](logger logr.Logger, operation string, apiCall func() (T
 
 func fetchNamespaces(ctx context.Context, k8sCacheClient client.Client) (*corev1.NamespaceList, error) {
 	namespaces := &corev1.NamespaceList{}
-	err := k8sCacheClient.List(ctx, namespaces)
+	err := k8sCacheClient.List(ctx, namespaces, client.UnsafeDisableDeepCopy)
 	if err != nil {
 		return nil, err
 	}
@@ -74,7 +74,7 @@ func fetchInstrumentationConfigs(ctx context.Context, logger logr.Logger, filter
 		err := k8sCacheClient.Get(ctx, client.ObjectKey{
 			Namespace: filters.NamespaceString,
 			Name:      instrumentationConfigName,
-		}, &instrumentationConfig)
+		}, &instrumentationConfig, client.UnsafeDisableDeepCopy)
 		if err != nil {
 			if apierrors.IsNotFound(err) {
 				// workload cam be not found and it is not an error.
@@ -92,7 +92,7 @@ func fetchInstrumentationConfigs(ctx context.Context, logger logr.Logger, filter
 		}, nil
 	} else {
 		var instrumentationConfigs odigosv1.InstrumentationConfigList
-		err := k8sCacheClient.List(ctx, &instrumentationConfigs, client.InNamespace(filters.NamespaceString))
+		err := k8sCacheClient.List(ctx, &instrumentationConfigs, client.InNamespace(filters.NamespaceString), client.UnsafeDisableDeepCopy)
 		if err != nil {
 			return nil, err
 		}
@@ -124,7 +124,7 @@ func fetchSourcesForWorkload(ctx context.Context, filters *WorkloadFilterSingleW
 		k8sconsts.WorkloadNameLabel:      filters.WorkloadName,
 	}
 	workloadSources := &odigosv1.SourceList{}
-	err := k8sCacheClient.List(ctx, workloadSources, client.InNamespace(filters.Namespace), client.MatchingLabels(workloadLabels))
+	err := k8sCacheClient.List(ctx, workloadSources, client.InNamespace(filters.Namespace), client.MatchingLabels(workloadLabels), client.UnsafeDisableDeepCopy)
 	if err != nil {
 		return nil, err
 	}
@@ -135,7 +135,7 @@ func fetchSourcesForWorkload(ctx context.Context, filters *WorkloadFilterSingleW
 		k8sconsts.WorkloadNameLabel:      filters.Namespace,
 	}
 	namespaceSources := &odigosv1.SourceList{}
-	err = k8sCacheClient.List(ctx, namespaceSources, client.InNamespace(filters.Namespace), client.MatchingLabels(namespaceLabels))
+	err = k8sCacheClient.List(ctx, namespaceSources, client.InNamespace(filters.Namespace), client.MatchingLabels(namespaceLabels), client.UnsafeDisableDeepCopy)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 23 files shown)


frontend/graph/loaders/fetchers.go
```diff
@@ -757,3 +785,141 @@ func fetchInstrumentationInstances(ctx context.Context, logger logr.Logger, filt
...
+			Selector:             obj.Spec.JobTemplate.Spec.Selector,
+			WorkloadHealthStatus: status.CalculateCronJobHealthStatus(obj.Status),
+		}, nil
+
+	case k8sconsts.WorkloadKindDeploymentConfig:
+		if !kube.IsDeploymentConfigAvailable() {
+			return nil, nil
+		}
+		gvr := schema.GroupVersionResource{
+			Group:    "apps.openshift.io",
```
aviadhahami, line 864: what is this for?

frontend/graph/loaders/loader.go
```diff
@@ -295,6 +308,36 @@ func (l *Loaders) loadInstrumentationInstances(ctx context.Context) error {
...
+	var odigosConfigurationConfigMap corev1.ConfigMap
+	err := l.k8sCacheClient.Get(ctx, client.ObjectKey{
+		Namespace: odigosns,
+		Name:      consts.OdigosEffectiveConfigName,
+	}, &odigosConfigurationConfigMap)
+	if err != nil {
+		return err
+	}
+
+	if err := yaml.Unmarshal([]byte(odigosConfigurationConfigMap.Data[consts.OdigosConfigurationFileName]), &l.odigosConfiguration); err != nil {
```
aviadhahami, line 329: `Unmarshal` is maybe costly; do we go through this path a lot?

frontend/graph/workload.resolvers.go
```diff
@@ -719,19 +768,25 @@ func (r *k8sWorkloadTelemetryMetricsResolver) ExpectingTelemetry(ctx context.Con
...
 
 // Workloads is the resolver for the workloads field.
+// Pre-computes all field values sequentially to avoid gqlgen spawning O(N × fields)
+// goroutines for concurrent field resolution. At 10K workloads × 17 resolver fields,
+// that would be 170K goroutines (~430MB of stack memory), exceeding the pod limit.
 func (r *queryResolver) Workloads(ctx context.Context, filter *model.WorkloadFilter) ([]*model.K8sWorkload, error) {
 	l := loaders.For(ctx)
-	err := l.SetFilters(ctx, filter)
-	if err != nil {
+	if err := l.SetFilters(ctx, filter); err != nil {
```
aviadhahami, line 780: According to the PR desc this outputs a `20MB response` for 10k workloads. How does the client behave here? a ~75s loading time will make the user refresh and will trigger a re-fetch;; maybe we should block user interaction during a fetch? or cancel a mid-flight request?

frontend/kube/watchers/batcher.go
```diff
@@ -26,6 +26,7 @@ type EventBatcher struct {
 	mu       sync.Mutex
 	batch    []sse.SSEMessage
 	timer    *time.Timer
+	maxTimer *time.Timer // fires after MaxDelay to guarantee flush under sustained load
```
aviadhahami: could u add a comment differentiating `maxTimer` from `MaxDelay`?

frontend/webapp/hooks/sources/useSourceCRUD.ts
```diff
@@ -19,6 +19,7 @@ import {
   ProgressKeys,
 } from '@odigos/ui-kit/store';
 
+// When SSE targets exceed this count, fall back to a full fetchSources() instead of fetching by individual IDs.
```
aviadhahami, line 22: ? remains?

aviadhahami (commented):
overall lgtm with some comments

frontend/graph/loaders/fetchers.go
BenElferink: OpenShift DeploymentConfig is not in the standard K8s API, so it's not in the informer cache.
We use the dynamic client to fetch it directly from the API server (only on OpenShift clusters where this CRD exists).

frontend/graph/workload.resolvers.go
BenElferink: The actual request takes ~10s, the rest of the time is just "content download" on the browser side, refetch should not be a problem as the server has already responded.

frontend/webapp/hooks/sources/useSourceCRUD.ts
BenElferink: Yes, this is still relevant when more than X sources were modified (currently X=50).

When fetching by ID, you need to make separate API calls for each ID, this cannot run in parallel at scale, so you'd have to queue the requests which adds a long waiting time, and a laggy UI that cannot handle the load.

The best option is to refetch all at once, if the number of modified is above the X (50).

blumamir (approved):

PR #4633, opened 2026-04-15
chore: pr policy to not run for releases PRs

tamirdavid1:

Pre-release PRs are automated and don't require linear tickets or release notes. GitHub Actions' pull_request trigger only filters by target branch, not source branch, so we use job-level if conditions on github.head_ref instead.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/pr-policy.yml
+++ b/.github/workflows/pr-policy.yml
@@ -10,11 +10,13 @@ permissions:
 
 jobs:
   require-linear:
+    if: ${{ !startsWith(github.head_ref, 'pre-releases') }}
     runs-on: ubuntu-latest
     steps: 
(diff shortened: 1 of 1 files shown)


blumamir (approved):
Since it happens in many cases, I wonder if we can make it more general going forward so it will cover all automatic PRs. I remember they used to be excluded by the action implementation in ci-core, but it stopped working for some reason

BenElferink (changes requested):
A change like this should be done via ci-core, I think we just need to add the octo bot:
https://github.com/odigos-io/ci-core/blob/main/require-linear/action.yml
https://github.com/odigos-io/ci-core/blob/main/require-release-note/action.yml

PR #4699, opened 2026-04-23
fix: optimize workload loading for namespaces query

BenElferink:
## Issue

On large clusters (100s of namespaces, 10K+ workloads) the query stalled and could OOM the UI pod.

gqlgen resolved `K8sNamespace.Workloads` concurrently per namespace (each calling `SetFilters`) and then spawned a goroutine per requested workload field → ~30K concurrent goroutines grabbing loader mutexes. Concurrent requests weren't serialized, so a second tab doubled memory.

## Fix

Resolver-level only:

- **`queryResolver.Namespaces`**: if the caller requested `workloads` (detected via `graphql.CollectFieldsCtx`), acquire `heavyWorkloadQueryMu`, call `SetFilters` once, and eagerly pre-populate every `K8sWorkload` in a single sequential loop.
- **`k8sNamespaceResolver.Workloads`**: short-circuits on the pre-populated slice.
- **`populateNamespaceQueryWorkloadFields`**: slim populator for the 3 fields this query actually uses (`markedForInstrumentation`, `dataStreamNames`, `numberOfInstances`). Skips InstrumentationConfig/Pod loads.

`GET_NAMESPACES` (no workloads) keeps the unchanged lightweight path.

---

```release-note
feat: optimize workload loading for namespaces query to enhance performance in large clusters
```

cherry-pick: v1.23, v1.24

emergency-ticket id: EMR-1566

### DIFF
--- a/frontend/graph/workload.resolvers.go
+++ b/frontend/graph/workload.resolvers.go
@@ -80,7 +80,6 @@ func (r *k8sNamespaceResolver) Workloads(ctx context.Context, obj *model.K8sName
 	workloadIds := l.GetWorkloadIdsInNamespace(obj.Name)
 	workloads := make([]*model.K8sWorkload, 0, len(workloadIds))
 	for _, id := range workloadIds {
-		id := id
 		workloads = append(workloads, &model.K8sWorkload{ID: &id})
 	}
 	return workloads, nil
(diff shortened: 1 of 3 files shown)


frontend/graph/workload.resolvers.go
```diff
@@ -74,6 +80,7 @@ func (r *k8sNamespaceResolver) Workloads(ctx context.Context, obj *model.K8sName
 	workloadIds := l.GetWorkloadIdsInNamespace(obj.Name)
 	workloads := make([]*model.K8sWorkload, 0, len(workloadIds))
 	for _, id := range workloadIds {
+		id := id
```
tamirdavid1: remove?

PR #4702, opened 2026-04-23
fix: performance for querying all namespaces with all workloads

BenElferink:
- UI pod still OOMs/timeouts when running `GetNamespacesWithWorkloads` despite earlier fixes.
- Root issue: this query path bypasses optimizations (no mutex + no precomputed fields).
- It triggers full cluster-wide workload + manifest loading, doubling memory under concurrency.
- gqlgen spawns ~60K goroutines (for ~20K workloads), causing ~150–250Mi transient memory spike.
- Additional inefficiency: duplicate `GetSources` calls per workload add lock contention + allocations.

**Fixes:**
- Precompute required workload fields to avoid heavy resolver execution.
- Serialize heavy queries using a mutex + `sync.Once`.
- Reduce redundant `GetSources` calls to one per workload.

**Result:**
- Peak memory drops ~470–570Mi → ~420–440Mi (within limits).
- Eliminates OOM spikes but not baseline memory usage.

---

### 200 namespaces with 20,000 workloads (in total), took only 3 seconds to fetch with only 512Mi allocated to the UI, zero OOMs/restarts:

Screenshot 2026-04-23 at 17 40 01

Screenshot 2026-04-23 at 17 36 54

### Release notes

```release-note
fix: performance for querying all namespaces with all workloads
```

cherry-pick: v1.23, v1.24

emergency-ticket id: EMR-1565

### DIFF
--- a/frontend/graph/loaders/loader.go
+++ b/frontend/graph/loaders/loader.go
@@ -78,6 +78,12 @@ type Loaders struct {
 	instrumentationInstancesFetched             bool
 	instrumentationInstancesByPodContainer      map[PodContainerId][]*v1alpha1.InstrumentationInstance
 	instrumentationInstancesByWorkloadContainer map[WorkloadContainerId][]*v1alpha1.InstrumentationInstance
+
+	// heavyWorkloadsOnce guards the first call to SetFilters(nil) for the
+	// K8sNamespace.Workloads resolver path. Subsequent namespaces in the same
+	// request fast-path without re-acquiring the global heavy-query mutex.
+	heavyWorkloadsOnce sync.Once
+	heavyWorkloadsErr  error
 }
 
 func WithLoaders(ctx context.Context, loaders *Loaders) context.Context {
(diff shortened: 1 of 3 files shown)


edeNFed:
@BenElferink can you give a short explanation what this is doing? I am not sure I entirely follow (I don't know this code well 😀).
- why do we need the sync.Once? 
- is SetFilters a heavy function? (Btw maybe better give it a different name... set filters sounds like we do this.filters = something inside and not doing many complex queries)
- if I am not mistaken sync.Once can introduce overhead as well causing many goroutines to wait for a lock, maybe we should just run it once outside of this flow?

BenElferink:
> * why do we need the sync.Once?

gqlgen runs the `K8sNamespace.Workloads` resolver concurrently - one goroutine per namespace in the response. For the customer (RP) that's ~500 concurrent calls in a single request, all racing into `SetFilters(ctx, nil)`.

The inner mutexes (`sourcesMutex`, `workloadManifestsMutex`) prevent duplicate K8s API calls, but the tail of `SetFilters` reassigns `l.workloadFilter` / `l.workloadIds` / `l.nsToWorkloadIds` without holding a lock - so without `sync.Once` those 500 goroutines race on those fields.

> * is SetFilters a heavy function? (Btw maybe better give it a different name... set filters sounds like we do this.filters = something inside and not doing many complex queries)

Yes, when called with `nil` (cluster-wide).

It does a lot more than its name implies - it triggers `loadSources` (list all `Source` CRDs) + `loadWorkloadManifests` (6+ cluster-wide List calls: Deployments, DaemonSets, StatefulSets, CronJobs, static pods, Rollouts, DeploymentConfigs) and then builds the workload-id maps.

At 20K workloads it's the single most expensive operation in the request.

I agree that the name is misleading. Should we rename to `LoadWorkloadsWithFilter`?

> * if I am not mistaken sync.Once can introduce overhead as well causing many goroutines to wait for a lock, maybe we should just run it once outside of this flow?

The 499 waiting goroutines aren't really overhead - they need the data `SetFilters` loads to do their job, so they'd wait anyway.

Without `sync.Once` they'd just contend on the inner per-resource mutexes one at a time, which is more lock thrashing for the same total wait time.

`sync.Once` itself is essentially free on the fast path (one atomic load) once the work is done.

edeNFed (approved):

PR #4703, opened 2026-04-23
CORE-934 | Migrate datastream config to use odigosconfigextenstion

tamirdavid1:

Refactored the `odigosrouterconnector` to resolve source-to-datastream dynamically via the `odigosconfigk8sextension` instead of embedding it statically in the gateway ConfigMap. The extension's existing workloadKeysIndex cache was extended with a workloadEntry struct that stores both container keys and data stream names, populated atomically during syncWorkloadToDesiredState from InstrumentationConfig labels. 
Removed the Sources field and SourceFilter struct from pipelinegen.DataStreams, and prevents ConfigMap churn and collector restarts when sources change. Dead code from the old static routing path (routingmap.go, connector_test.go, DataStreams config field) was removed.

```release-note
odigosrouterconnector resolves datastream routing dynamically via extension cache, eliminating ConfigMap updates and collector restarts on changes.
```

cherry-pick: v1.23, v1.24


collector/connectors/odigosrouterconnector/connector.go
```diff
@@ -133,75 +144,45 @@ func createLogsConnector(
...
-	if !ok {
-		return nil, ""
-	}
-
-	pipelines, ok := routingIndex[signal]
-	if !ok {
-		return nil, ""
+func (r *routerConnector) resolveDataStreams(attrs pcommon.Map) []string {
+	resource := pcommon.NewResource()
+	attrs.CopyTo(resource.Attributes())
```
edeNFed: Consider making `GetDataStreamsForWorkload` recieve `attrs` directly instead of copying them. If i understand correctly this function is called per trace so better to avoid allocations as much as possible

collector/extension/odigosconfigk8sextension/cache.go
```diff
@@ -139,15 +150,51 @@ func (c *cache) Delete(key string) {
...
 	}
 	c.mu.RUnlock()
 	return out
 }
+
+// SetDataStreams stores the data stream names for the given workload key prefix.
+func (c *cache) SetDataStreams(workloadKey string, streams []string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	entry := c.workloadKeysIndex[workloadKey]
```
edeNFed, line 169: as we now wont be able to see datastreams in the configmap anymore, consider adding some logging everytime a new datastream is created/deleted/updated in cache (shouldnt happen a lot)

edeNFed (approved):

collector/extension/odigosconfigk8sextension/cache.go
tamirdavid1: In This specific place It is created per workload which happens a lot.
also we will have a pipeline + forward connector per datastream still in the gateway cm

collector/connectors/odigosrouterconnector/connector.go
tamirdavid1: Done, Thanks

collector/extension/odigosconfigk8sextension/go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/odigos-io/odigos/collector/extension/odigosconfigk8sextension
 go 1.26.2
 
 require (
+	github.com/odigos-io/odigos/api v1.23.7
```
blumamir: Since it's a module in the monorepo, we should use "replace" directive so the code is aligned and we don't use old versions that creates conflicts.

I think it's really hard so we currently just copy the consts into the relevant collector components.

collector/extension/odigosconfigk8sextension/cache.go
```diff
@@ -139,15 +150,51 @@ func (c *cache) Delete(key string) {
...
+	if entry == nil {
+		entry = &workloadEntry{containerKeys: make(map[string]struct{})}
+		c.workloadKeysIndex[workloadKey] = entry
+	}
+	entry.dataStreams = streams
+}
```
blumamir, line 175: consider merging it with the `Set` function of the cache, so we don't need to duplicate code, and the cache sync is atomic. 

Currently you are locking twice which means if someone access the cache between those 2 updates he will read inconsistent state, and also if the object is created and delete or changes fast, this can cause nasty race conditions

collector/extension/odigosconfigk8sextension/informer.go
```diff
@@ -206,7 +210,34 @@ func (o *OdigosWorkloadConfig) syncWorkloadToDesiredState(workloadKey workloadKe
 		}
 	}
 
-	o.logger.Debug("synced workload to desired state", zap.String("workload", workloadKeyStr), zap.Int("desired", len(desired)), zap.Int("removed", numRemoved))
+	// 3) Update data stream membership for this workload.
+	if dataStreams != nil {
+		o.cache.SetDataStreams(workloadIndexKey, dataStreams)
+	} else {
+		o.cache.DeleteDataStreams(workloadIndexKey)
```
blumamir: Instead of "DeleteDataStream", use Set with empty or nil array

collector/extension/odigosconfigk8sextension/cache.go
```diff
@@ -139,15 +150,51 @@ func (c *cache) Delete(key string) {
...
+
+// DeleteDataStreams removes data stream names for the given workload key prefix.
+// If the entry has no container keys left either, the entire entry is removed.
+func (c *cache) DeleteDataStreams(workloadKey string) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if entry := c.workloadKeysIndex[workloadKey]; entry != nil {
+		entry.dataStreams = nil
+		if len(entry.containerKeys) == 0 {
```
blumamir: What if this entry has container configs, this will delete them along with the workload, right?

collector/extension/odigosconfigk8sextension/cache.go
```diff
@@ -139,15 +150,51 @@ func (c *cache) Delete(key string) {
...
+	entry := c.workloadKeysIndex[workloadKey]
+	if entry == nil {
+		return nil, false
+	}
+	return entry.dataStreams, true
+}
+
+// DeleteDataStreams removes data stream names for the given workload key prefix.
+// If the entry has no container keys left either, the entire entry is removed.
+func (c *cache) DeleteDataStreams(workloadKey string) {
```
blumamir: There is a callback for deleting keys from the cache, which will not be called here

collector/extension/odigosconfigk8sextension/informer.go
```diff
@@ -206,7 +210,34 @@ func (o *OdigosWorkloadConfig) syncWorkloadToDesiredState(workloadKey workloadKe
...
+func extractDataStreamLabels(labels map[string]string) []string {
+	var streams []string
+	for k, v := range labels {
+		if strings.HasPrefix(k, k8sconsts.SourceDataStreamLabelPrefix) && v == "true" {
+			streams = append(streams, strings.TrimPrefix(k, k8sconsts.SourceDataStreamLabelPrefix))
+		}
+	}
+	if len(streams) == 0 {
+		return []string{"default"}
+	}
```
blumamir: I wonder if we should do it here, or simply reflect what we find on the IC and use empty value defaulting where it is used. WDYT?

collector/extension/odigosconfigk8sextension/odigosconfig.go
```diff
@@ -87,3 +87,17 @@ func (o *OdigosWorkloadConfig) UnregisterWorkloadConfigCacheCallback(cb collecto
...
+	key, err := workloadKeyPrefixFromResourceAttributes(res.Attributes())
+	if err != nil {
+		return nil, false
+	}
+	streams, found := o.cache.GetDataStreams(key)
+	if !found {
+		o.logger.Debug("data streams not found for workload", zap.String("key", key))
+		return nil, false
+	}
+	o.logger.Debug("resolved data streams for workload", zap.String("key", key), zap.Strings("dataStreams", streams))
```
blumamir: I think this log have no real value, and can cause a storm of lines. we should be carful of having a log record on every span we process for normal flow

collector/extension/odigosconfigk8sextension/resourceattrs.go
```diff
@@ -78,3 +78,15 @@ func getContainerName(attrs pcommon.Map) string {
 	}
 	return containerName.Str()
 }
+
+// workloadKeyPrefixFromResourceAttributes builds a workload-level cache key prefix
+// from resource attributes. Unlike workloadKeyFromResourceAttributes, it does not
+// require k8s.container.name. Returns the key in "ns/Kind/name/" format.
+func workloadKeyPrefixFromResourceAttributes(attrs pcommon.Map) (string, error) {
```
blumamir: I think the above function should be renamed to `workloadContainerKeyFromResourceAttributes`, so we have 2 sets of keys: workload and container

collector/extension/odigosconfigk8sextension/informer.go
```diff
@@ -206,7 +210,34 @@ func (o *OdigosWorkloadConfig) syncWorkloadToDesiredState(workloadKey workloadKe
...
+		zap.Int("removed", numRemoved),
+		zap.Strings("dataStreams", dataStreams),
+	)
+}
+
+// extractDataStreamLabels reads IC labels and returns the data stream names.
+// If no data-stream labels are present, returns ["default"].
+func extractDataStreamLabels(labels map[string]string) []string {
+	var streams []string
+	for k, v := range labels {
```
blumamir, line 231: if labels is nil this will panic I think

collector/extension/odigosconfigk8sextension/odigosconfig.go
tamirdavid1: This is a debug log, was very useful during development to understand if adding a source to a new datastream works.
I can see future value in it, lmk what do you think

collector/extension/odigosconfigk8sextension/informer.go
tamirdavid1: good idea, done.

collector/extension/odigosconfigk8sextension/informer.go
tamirdavid1: We can potentially keep zombie cache objects without using the DeleteDataStreams

collector/extension/odigosconfigk8sextension/informer.go
tamirdavid1: range on nil is safe in go

blumamir (approved):

collector/extension/odigosconfigk8sextension/cache.go
```diff
@@ -17,6 +17,13 @@ type workloadKey struct {
 	Name      string
 }
 
+// workloadEntry holds per-workload state: the set of container cache keys and the
+type workloadEntry struct {
+	containerKeys map[string]struct{}
+	// data stream names extracted from IC labels.
+	dataStreams []string
+}
```
RonFed:
```suggestion
// workloadEntry holds per-workload state: the set of container cache keys and the data stream names extracted from IC labels.
type workloadEntry struct {
	containerKeys map[string]struct{}
	dataStreams []string
}
```

collector/extension/odigosconfigk8sextension/go.mod
tamirdavid1: sounds good, done.

RonFed (approved):

PR #4723, opened 2026-04-26
feat: tail sampling for highly-relevant and cost-reduction categories

blumamir:

Fixes: CORE-634

Add tail sampling evaluation for highly-relevant and cost-reduction categories. 
This completes the missing categories.

The processor:
- evaluate the rules per span, based on the config of each workload
- make a decision based on the deciding rule
- record attributes on the span for sampling info based on config

```release-note
add tail-sampling rules for highly-relevant and cost-reduction categories
```


common/odigosattributes/sampling.go
```diff
@@ -3,13 +3,26 @@ package odigosattributes
...
+	// The unique identifier of the matching sampling rule for the span. Type: string.
+	// In the same trace, different spans can match different rules,
+	// and this attribute indicate the most restrictive rule that matched a specific span.
+	SamplingSpanMatchingRuleId = "odigos.sampling.span.matching_rule.id"
+	// The human-readable name of the matching sampling rule. Type: string.
+	// It the rule has no name, this attribute will not be set.
+	SamplingSpanMatchingRuleName = "odigos.sampling.span.matching_rule.name"
+	// The keep percentage configured on the matching sampling rule. Type: double.
+	SamplingSpanMatchingRuleKeepPercentage = "odigos.sampling.span.matching_rule.keep_percentage"
+
```
RonFed, line 25: IMO, we can drop the "span" word from all of theses (the variable names and the attributes)
I think it is obvious from the context - unless you think it will be someday 
`odigos.sampling.<something that is not a span>.matching_rule.name`?
making these shorter makes it more clear and easy to read.

collector/processors/odigostailsamplingprocessor/processor.go
```diff
@@ -65,6 +65,30 @@ func (p *tailSamplingProcessor) processTraces(ctx context.Context, td ptrace.Tra
 		}
 	}
 
+	matched, highlyRelevantOperationRule := category.EvaluateHighlyRelevantOperations(td, p.odigosConfigExtension)
```
RonFed: `evaluateNoisyOperations` is a member function of `p` - maybe worth to align to either way for consistency?
having `p.EvaluateHighlyRelevantOperations`?

collector/processors/odigostailsamplingprocessor/processor.go
```diff
@@ -65,6 +65,30 @@ func (p *tailSamplingProcessor) processTraces(ctx context.Context, td ptrace.Tra
...
+		keepTrace := tracePercentage <= percentageAtLeast
+
+		if keepTrace || p.config.DryRun {
+			enrichSpansWithSamplingAttributes(td, consts.SamplingCategoryHighlyRelevant, highlyRelevantOperationRule.Id, highlyRelevantOperationRule.Name, percentageAtLeast, p.config.DryRun, keepTrace, p.config.SpanSamplingAttributes)
+			return td, nil
+		}
+		return ptrace.NewTraces(), nil
+	}
+
+	matched, costReductionRule := category.EvaluateCostReductionOperations(td, p.odigosConfigExtension)
```
RonFed: I still think that having the package name for the categories such as
`costreduction.Evalute`, `highlyrelevant.Evalute` will make this clearer

collector/processors/odigostailsamplingprocessor/matchers/tailsampling.go
```diff
@@ -0,0 +1,66 @@
...
+	if operation.HttpServer != nil {
+		return operationHttpServerMatcher(operation.HttpServer, span)
+	}
+	if operation.KafkaConsumer != nil {
+		return operationKafkaConsumerMatcher(operation.KafkaConsumer, span)
+	}
+	if operation.KafkaProducer != nil {
+		return operationKafkaProducerMatcher(operation.KafkaProducer, span)
+	}
+	// no operation type specified, match any.
```
RonFed, line 22: consider a switch case pattern

collector/processors/odigostailsamplingprocessor/category/costreductionoperations.go
```diff
@@ -0,0 +1,120 @@
...
+
+	"github.com/odigos-io/odigos/collector/processors/odigostailsamplingprocessor/matchers"
+	commonapisampling "github.com/odigos-io/odigos/common/api/sampling"
+	"github.com/odigos-io/odigos/common/collector"
+	"github.com/odigos-io/odigos/common/odigosattributes"
+)
+
+// EvaluateCostReductionOperations matches cost-reduction rules on each span, sets per-span attributes,
+// aggregates trace-level matches, and returns whether a non-disabled deciding rule applies.
+func EvaluateCostReductionOperations(trace ptrace.Traces, configProvider collector.OdigosConfigExtension) (bool, *commonapisampling.CostReductionRule) {
```
RonFed: this is 90% the same as EvaluateHighlyRelevantOperations
can we have a common function for this - something like `tailsampling.Evalute` - and pass the matcher as a parameter

collector/processors/odigostailsamplingprocessor/category/highlyrelevant/highlyrelevanceoperations.go
```diff
@@ -0,0 +1,131 @@
...
+
+// matchHighlyRelevantRulesForSingleSpan returns every highly-relevant rule whose matchers all pass for this span.
+func matchHighlyRelevantRulesForSingleSpan(span ptrace.Span, highlyRelevantOperations []commonapisampling.HighlyRelevantOperation) []*commonapisampling.HighlyRelevantOperation {
+	matchedRules := []*commonapisampling.HighlyRelevantOperation{}
+
+	for _, highlyRelevantOperation := range highlyRelevantOperations {
+		matched := true
+		matched = matched && matchers.TailSamplingOperationMatcher(highlyRelevantOperation.Operation, span)
+		matched = matched && matchers.SpanErrorMatcher(span, highlyRelevantOperation.Error)
+		matched = matched && matchers.SpanDurationMatcher(span, highlyRelevantOperation.DurationAtLeastMs)
```
RonFed, line 64: should we have a short circuit here if one of the matchers returns false - to avoid un-necessery computations?
i.e if `TailSamplingOperationMatcher` is false - there is not need to do `SpanErrorMatcher` and `SpanDurationMatcher`, right?

collector/processors/odigostailsamplingprocessor/category/highlyrelevant/highlyrelevanceoperations.go
```diff
@@ -0,0 +1,131 @@
...
+	for _, matchingRule := range matchingRules {
+		if matchingRule.Disabled {
+			continue
+		}
+		percentage := GetPercentageOrDefault100(matchingRule.PercentageAtLeast)
+
+		// once we hit maximum, no point in keeping iterating.
+		if percentage == 100.0 {
+			return matchingRule
+		}
```
RonFed, line 117: if we have more than one 100% rules - this will be not deterministic - since we are iterating a map - do we want to make it determistic?

collector/processors/odigostailsamplingprocessor/category/costreduction/costreductionoperations.go
```diff
@@ -0,0 +1,120 @@
...
+	// check for lowest percentage enabled sampling rule.
+	var selectedRule *commonapisampling.CostReductionRule
+	for _, r := range matchingRules {
+		if r.Disabled {
+			continue
+		}
+		// shortcut for when we hit the bottom and don't need to keep iterating.
+		if r.PercentageAtMost == 0 {
+			return r
+		}
```
RonFed, line 107: if we have multiple 0% rules - this will be non deterministic - since we are iterating a map - do we want to make this deterministic?

RonFed (approved):

collector/processors/odigostailsamplingprocessor/category/costreductionoperations.go
blumamir: I did extract a lot into utils, but they have many differences in the percentage default / max or min / config types, matchers, etc.

When I tried to make it more generic it turned up ugly and hard to read, but I am open to ideas on specific improvements if you want to suggest them

collector/processors/odigostailsamplingprocessor/category/costreduction/costreductionoperations.go
blumamir: Currently if there are multiple rules with the same percentage, the deciding rule is choose arbitrary. we can relax this in the future and make it deterministic, but for now let's avoid the extra complexity

collector/processors/odigostailsamplingprocessor/category/highlyrelevant/highlyrelevanceoperations.go
blumamir: Isn't it the default behavior of the && operator in go? so if the first argument is false the second one is not evaluated?

collector/processors/odigostailsamplingprocessor/category/highlyrelevant/highlyrelevanceoperations.go
blumamir: same comment as above, current behavior is arbitrary and we can improve in the future. trying to make it simple first

collector/processors/odigostailsamplingprocessor/matchers/tailsampling.go
blumamir: done

collector/processors/odigostailsamplingprocessor/processor.go
blumamir: With noisy operation, we have an extra step of extracting the root span before calling the evaluate function, where as for these categories, we send the entire trace as is. this is where the difference comes from. We can chat about it

common/odigosattributes/sampling.go
blumamir: we also have these attributes for a "trace". the "span" here means it's showing the rule that matched this one specific span (vs the rule that matched the trace that can is likely a different one)

PR #4732, opened 2026-04-27
CORE-843: extract attribute processor

natyharary:

Adds a processor that can extract attributes from spans and put the value under a new attribute.
Useful for extracting nested attributes in url or json strings in attributes.
Uses regex.

Processor config example:
```yaml

kind: Processor
metadata:
  name: extract-study-id-from-studies
  namespace: odigos-system
spec:
  type: odigosextractattribute
  processorName: "extract study.id from /studies/<id>/ and project.id from /projects/<id>/ URL path"
  signals:
    - TRACES
  collectorRoles:
    - CLUSTER_GATEWAY
  orderHint: 1
  processorConfig:
    extractions:
      - source: studies
        data_format: url
        target: extracted_study.id
      - source: projects
        data_format: url
        target: extracted_project.id


```

```release-note
Adds the ExtractAttributeProcessor - used for extracting attributes from other attributes in spans
```


collector/processors/odigosextractattributeprocessor/README.md
```diff
@@ -0,0 +1,96 @@
+# Odigos Extract Attribute Processor
+
+## Overview
+
+The **Odigos Extract Attribute Processor** extracts values embedded inside existing string-valued span
+attributes and surfaces them as new, normalized attributes.
+
+Useful when identifiers (study IDs, tenant IDs, request IDs, etc.) are buried inside SQL statements,
+JSON blobs, or URL paths, and you want them as first-class span attributes.
```
blumamir, line 9: like 👍

collector/processors/odigosextractattributeprocessor/processor.go
```diff
@@ -0,0 +1,130 @@
...
+	return result, found
+}
+
+// buildExtractionRegex returns the pattern that captures the value of key for the given format. The key is anchored
+// on a JSON/SQL/URL boundary so substrings like "myfoo_bar" don't cross-match "foo_bar".
+func buildExtractionRegex(key string, format DataFormat) (*regexp.Regexp, error) {
+	escapedKey := regexp.QuoteMeta(key)
+	switch format {
+	case FormatJSON: // Also works for SQL
+		return regexp.MustCompile(`(?:^|[\s,{("'])["']?` + escapedKey + `["']?\s*[:=]\s*["']?([^"'\s,;)}]+)`), nil
```
blumamir, line 141: nit: consider adding a comment about what this actually checks. 

few things that come to my mind:

```
"key":"value"
"key": "value" (with a space after colon)
"key": true (not a string)
"key": ["value1", "value2"]
"key": "abc\"def"
"key": "abc,def"
```

It's ok not to support everything, but it's hard to tell what is going to work and what not

collector/processors/odigosextractattributeprocessor/processor.go
```diff
@@ -0,0 +1,130 @@
...
+	case FormatURL:
+		return regexp.MustCompile(`(?:^|/)` + escapedKey + `/([^/\s"?&#]+)`), nil
```
blumamir, line 148: same here, could be nice to have a comment about what this actually checks

collector/processors/odigosextractattributeprocessor/processor.go
```diff
@@ -0,0 +1,130 @@
...
+			return true
+		}
+		// Take the first regex match
+		if matches := re.FindStringSubmatch(content); len(matches) > 1 {
+			result = matches[1]
+			found = true
+			return false
+		}
+		return true
+	})
```
blumamir: Running the regexp on all the attributes can be expensive, as regex is not a cheap opeartions and there are many attributes.

I think we should target just those for the payload right now, which are the ones that are actually needed at the moment.

What do you think we create a list of few attributes and just check those? we can also later allow setting it from config option

collector/processors/odigosextractattributeprocessor/processor.go
```diff
@@ -0,0 +1,130 @@
...
+	span.Attributes().Range(func(_ string, value pcommon.Value) bool {
+		if value.Type() != pcommon.ValueTypeStr {
+			return true
+		}
+		content := value.Str()
+		if content == "" {
+			return true
+		}
+		// Take the first regex match
+		if matches := re.FindStringSubmatch(content); len(matches) > 1 {
```
blumamir, line 119: why `> 1`? what is in index `0`?

collector/processors/odigosextractattributeprocessor/processor.go
```diff
@@ -0,0 +1,130 @@
...
+func (p *extractAttributeProcessor) processSpan(span ptrace.Span) {
+	for _, e := range p.extractors {
+		if value, ok := extractFromAttributes(span, e.regex); ok {
+			p.logger.Debug("extraction matched",
+				zap.String("target", e.target),
+				zap.String("value", value),
+				zap.String("regex", e.regex.String()),
+				zap.Stringer("spanId", span.SpanID()),
+			)
+			span.Attributes().PutStr(e.target, value)
```
blumamir, line 94: will this override the attribute if it's already there for any reason? (e.g. is it insert or upsert?)

collector/processors/odigosextractattributeprocessor/processor.go
```diff
@@ -0,0 +1,130 @@
...
+		} else {
+			p.logger.Debug("extraction did not match any attribute",
+				zap.String("target", e.target),
+				zap.String("regex", e.regex.String()),
+				zap.Stringer("spanId", span.SpanID()),
+			)
```
blumamir: If we ever need to enable debug logging in the collector, this is going to create a flood.
do you think it's important to write to logs? most extractions will fail I guess (very few spans in a trace actually contains the payload)

collector/processors/odigosextractattributeprocessor/processor_test.go
```diff
@@ -0,0 +1,263 @@
...
+			name:      "unquoted value with spaced equals",
+			key:       "study_id",
+			input:     `study_id = abc-123`,
+			extracted: "abc-123",
+		},
+		{
+			name:      "numeric value",
+			key:       "study_id",
+			input:     `{"study_id": 123}`,
+			extracted: "123",
```
blumamir, line 92: specifically in this case it could have been nice to have it in the span attribute as a number. but I guess it's not that simple

blumamir (approved):

collector/processors/odigosextractattributeprocessor/processor.go
natyharary: The way that `re.FindStringSubmatch` works is a bit weird:
https://pkg.go.dev/regexp#Regexp.FindStringSubmatch

index 0 is the entire string we want to matchm and index 1 and so forth are the actual matches.
Example from the documentation:

```go

package main

import (
	"fmt"
	"regexp"
)

func main() {
	re := regexp.MustCompile(`a(x*)b(y|z)c`)
	fmt.Printf("%q\n", re.FindStringSubmatch("-axxxbyc-"))
	fmt.Printf("%q\n", re.FindStringSubmatch("-abzc-"))
}

Output:
["axxxbyc" "xxx" "y"]
```

So we want not only to get at least 2 items in the `matches` array (the first one is the entire string and second being the sub string that matched) but also we want to get the item in index 1, which is the actual substring matched.

collector/processors/odigosextractattributeprocessor/processor.go
natyharary: Come to think about it, that's right, most extractions will fail anyway. I'll remove it

collector/processors/odigosextractattributeprocessor/processor.go
natyharary: Good catch. I added it before the actual regex run

collector/processors/odigosextractattributeprocessor/processor.go
natyharary: Good idea, added `RELEVANT_SPAN_ATTRIBUTES` and check for it

collector/processors/odigosextractattributeprocessor/processor_test.go
natyharary: Yeah that would make things difficult as I would need to try to turn it into an int each time and record it as an int if I was successful and a string if not. Could be costly

collector/processors/odigosextractattributeprocessor/processor.go
```diff
@@ -0,0 +1,149 @@
...
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+	"go.opentelemetry.io/collector/processor"
+	"go.uber.org/zap"
+)
+
+var RELEVANT_SPAN_ATTRIBUTES = map[string]struct{}{
+	"db.statement":              {},
+	"messaging.message.payload": {},
+}
```
blumamir, line 20: consider also
```suggestion
var RELEVANT_SPAN_ATTRIBUTES = map[string]struct{}{
	"db.statement":              {},
	"db.query.text": {},
	"messaging.message.payload": {},
	"http.request.payload": {},
	"http.response.payload": {},
}
```

blumamir (approved):

PR #4737, opened 2026-04-28
CORE-942 | feat: adding a recordSpans option for headsampling

tamirdavid1:

Introducing a recordSpans option that records the sampling decision at the agent level, while deferring span dropping to the data collection pipeline instead of discarding them at the agent.
This enables span-based metrics, since spans are still available for metric computation. The tradeoff is increased resource usage in the collector, as spans are not filtered out early.
This PR is the first step of the feature and includes only the CRD changes. A follow-up PR will implement the filtering logic in the data collection pipeline.

```release-note
NONE
```


helm/odigos/values.yaml
```diff
@@ -1504,6 +1504,21 @@ sampling:
...
+  headSampling:
+
+    # @schema
+    # description: |-
+    #   If true, the sampling decision is recorded on spans but the spans are dropped
+    #   in the data collection pipeline rather than at the agent level.
+    #   This is useful when span metrics are needed, since spans remain available for metric computation.
+    #   The tradeoff is higher resource usage in the collector, as spans are not discarded early at the agent.
+    # @schema
+    # recordSpans: false
```
blumamir: Thinking about it again, what this value actualy sets is the strategy for how to acheieve span metrics. 

There can be 3 options eventually:
1. `sampled-spans-only` - most performant, but not acurate - as we only count what we didn't drop in the metrics.
2. `record` - less performant but most accurate
3. `statistical` - very performant, and metrics are "corrected" to account for drops which should give accurate results in large numbers.

What do you say that we move this config option to be part of the `spanMetrics` configuration, and have the default by `sampled-spans-only`, and make it an enum so when we support option 3 in the future it is easy to add in helm?

I also think that logically it is more relevant to span metrics (e.g. why someone would want to change it), vs the implementation (e.g. that it's propgated to head-sampler).

Then users can go to spanMetrics and choose the level of accuracy they need for it in the right context

blumamir (commented):
Looks great, wrote one suggestion that I would love to discuss when you are available

helm/odigos/values.yaml
```diff
@@ -1397,6 +1397,18 @@ metricsSources:
...
+    # description: |-
+    #   Controls how span metrics interact with head-sampling decisions.
+    #   Possible values:
+    #     - "sampled-spans-only" (default): metrics are computed only from sampled spans;
+    #       dropped spans are discarded at the agent level.
+    #     - "record": all spans are forwarded to the collector for metric computation,
+    #       then dropped spans are deleted in the pipeline (higher resource usage).
+    # enum: [sampled-spans-only, record]
+    # @schema
+    # spanMetricsMode: sampled-spans-only
```
blumamir, line 1412: how about?

```suggestion
    # fidelityMode: sampled-spans-only
```

helm/odigos/values.yaml
```diff
@@ -1397,6 +1397,18 @@ metricsSources:
...
 
+    # @schema
+    # description: |-
+    #   Controls how span metrics interact with head-sampling decisions.
+    #   Possible values:
+    #     - "sampled-spans-only" (default): metrics are computed only from sampled spans;
+    #       dropped spans are discarded at the agent level.
+    #     - "record": all spans are forwarded to the collector for metric computation,
+    #       then dropped spans are deleted in the pipeline (higher resource usage).
+    # enum: [sampled-spans-only, record]
```
blumamir: how about?

```
# description: |-
#   Controls the tradeoff between metric accuracy and resource usage.
#   Determines how sampling affects which spans are used to compute metrics.
#   Possible values:
#     - "sampled-spans-only" (default): metrics are computed only from sampled spans.
#       Unsampled spans are dropped early, resulting in lower resource usage but reduced accuracy.
#     - "record": metrics are computed from all spans, regardless of sampling.
#       Unsampled spans are forwarded for metric computation and dropped later in the pipeline,
#       resulting in higher accuracy at the cost of increased resource usage.
# enum: [sampled-spans-only, record]
```

blumamir (approved):

PR #4740, opened 2026-04-28
CORE-960 | feat: initial commit of migrating the custom instrumentaiton from sdkconfig to containerconfig

tamirdavid1:

We are migrating the sdkconfig into the container configuration. This PR introduces the same customInstrumentation into the container config. Once the migration is complete, the instrumentations will rely on the container configuration.
image

```release-note
NONE
```


instrumentor/controllers/agentenabled/dynamicconfig/traces/custominstrumentations.go
```diff
@@ -0,0 +1,48 @@
...
+
+	golangProbes := make([]instrumentationrules.GolangCustomProbe, 0, len(rule1.Golang)+len(rule2.Golang))
+	golangProbes = append(golangProbes, rule1.Golang...)
+	golangProbes = append(golangProbes, rule2.Golang...)
+	merged.Golang = golangProbes
+
+	javaProbes := make([]instrumentationrules.JavaCustomProbe, 0, len(rule1.Java)+len(rule2.Java))
+	javaProbes = append(javaProbes, rule1.Java...)
+	javaProbes = append(javaProbes, rule2.Java...)
+	merged.Java = javaProbes
```
blumamir: you can pass the distro here, and avoid setting java probes if it's a go distro and the other way around

blumamir (approved):

instrumentor/controllers/agentenabled/dynamicconfig/traces/custominstrumentations.go
blumamir: consider switch on the distro language and populate only the relevant section based on the language

tamirdavid1: good point, will do.

PR #4756, opened 2026-04-30
feat: integrate read/edit/delete Rule Drawer V2 for Sources

BenElferink:
```release-note
feat: integrate read/edit/delete Rule Drawer V2 for Sources
```

emergency-ticket id: EMR-1617


frontend/webapp/utils/functions/sources.ts
```diff
@@ -65,19 +56,14 @@ export function mapConditionsToConditionArray(conditions: K8sWorkloadConditions
 // container shape is now structurally identical to `SourceContainer`.
 export function mapWorkloadToSource(w: WorkloadResponse): Source {
   return {
+    ...(w as unknown as Source),
```
alonkeyval: Suggest reverting the spread+cast with an explicit field-by-field mapping so the object is clean and the compiler stays in the loop

frontend/webapp/graphql/queries/workload.ts
```diff
@@ -40,84 +40,318 @@ export const GET_WORKLOADS_BY_IDS = gql`
...
+        }
+        expectingTelemetry {
+          name
+          status
+          reasonEnum
+          message
+        }
       }
       runtimeInfo {
+        completed
```
alonkeyval: do we need the top level runtimeInfo and agentEnabled blocks? I think the drawer use containers[].runtimeInfo and containers[].agentEnabled

frontend/webapp/utils/functions/sources.ts
alonkeyval: and also need to add to the Source type the missing fields like pods and telemetryMetrics

frontend/webapp/graphql/queries/workload.ts
```diff
@@ -40,84 +40,318 @@ export const GET_WORKLOADS_BY_IDS = gql`
...
+          }
         }
-        rollout {
+      }
+      pods {
+        podName
+        nodeName
+        startTime
+        agentInjected
+        startedPostAgentMetaHashChange
```
alonkeyval: some fields are fetched but not used, like odigosInstrumentationDeviceName and startedPostAgentMetaHashChange. we have plan to show them?

alonkeyval:
commented what we discussed internally

frontend/webapp/utils/functions/sources.ts
BenElferink: I've committed the removal of this, but we will have to update the UI Kit with this to support the change: https://github.com/odigos-io/ui-kit/pull/927

frontend/graph/workload.resolvers.go
```diff
@@ -761,6 +762,30 @@ func (r *k8sWorkloadResolver) RollbackOccurred(ctx context.Context, obj *model.K
...
+	if obj == nil || obj.ID == nil {
+		return nil, nil
+	}
+	instrumentationConfigYAML, err := services.K8sManifest(ctx, obj.ID.Namespace, model.K8sResourceKindInstrumentationConfig, workload.CalculateWorkloadRuntimeObjectName(obj.ID.Name, string(obj.ID.Kind)))
+	if err != nil {
+		return nil, fmt.Errorf("failed to get InstrumentationConfig YAML: %w", err)
+	}
+	return &instrumentationConfigYAML, nil
+}
+
```
alonkeyval: I think it better to fetch those yamls on demand rather then with the workload data. we only need them if the user reach to the debug tab. in addition, we use fetchSourceById when we get sse event, so calling the k8s api for multiple sources can be expensive.

frontend/webapp/graphql/queries/workload.ts
```diff
@@ -1,30 +1,42 @@
...
-        reasonEnum
-        message
-      }
+      ${WORKLOAD_FIELDS_SLIM}
+    }
+  }
+`;
+
+export const GET_WORKLOADS_BY_IDS_SLIM = gql`
+  query GetWorkloadsByIds($ids: [K8sWorkloadIdInput!]!) {
```
alonkeyval: we already have query func with this name - GetWorkloadsByIds -> GetWorkloadsByIdsSlim

alonkeyval:
still thinking moving the yamls to it on demand call is the best choice for as before we ship this api change.

1. we still fetch it all again even where we are not in the debug tab - refetch each 3 sec
2. if the the yaml will field to fetch for some reason the whole workload will be undifend

alonkeyval (approved):

PR #4764, opened 2026-05-01
CORE-935: Update RHEL release jobs to use sts, depot, and AWS auth

damemi:
RHEL release is hitting AWS rate limit (see [example run](https://github.com/odigos-io/odigos/actions/runs/25220876629)) so this switches us to our depot runners, adds AWS auth, and also switches to STS to trigger the enterprise RHEL release job

enterprise: https://github.com/odigos-io/odigos-enterprise/pull/2752

```release-note
NONE
```


.github/workflows/publish-modules-rhel.yml
RonFed, line 1: Just as a sanity check worth to compare the time it took for these compared to the non depot runners.

RonFed (approved):

tamirdavid1 (approved):

PR #4810, opened 2026-05-05
CORE-951: support SourcesScopes in InstrumentationRule

natyharary:

1. Add the SourcesScopes field to expand the InstrumentationRule capabilities
2. Add tests for the matching function

```release-note
Add SourcesScopes in InstrumentationRule (Workloads will be deprecated)
```


instrumentor/controllers/utils/instrumentationrules.go
```diff
@@ -3,25 +3,87 @@ package utils
...
+		if len(scopes) == 0 {
+			return false
+		}
+		for i := range scopes {
+			if scope.SourcesScopeMatchesContainer(scopes[i], workload, containerName, language) {
+				return true
+			}
+		}
 		return false
 	}
```
blumamir: consider `AnySourceScopeMatchesContainer`

instrumentor/controllers/utils/instrumentationrules.go
```diff
@@ -3,25 +3,87 @@ package utils
...
 	for _, allowedWorkload := range *rule.Spec.Workloads {
 		if allowedWorkload == workload {
 			return true
 		}
 	}
 	return false
 }
+
+// Resolves whether the rule applies to the workload for at least one container on the InstrumentationConfig
+func IsInstrumentationConfigParticipatingInRule(
```
blumamir: I think it made sense to use for the `Workloads` filter, but for the `SourcesScope` it's less of a good fit. should we just use all the rules and let the reconciler logic take care of it once?

frontend/graph/schema.graphqls
```diff
@@ -150,6 +150,23 @@ input K8sSourceId {
...
+  workloadLanguage: SamplingWorkloadLanguage
+}
+
+input InstrumentationRuleSourcesScopeInput {
+  workloadName: String
+  workloadKind: K8sResourceKind
+  workloadNamespace: String
+  containerName: String
+  workloadLanguage: SamplingWorkloadLanguage
+}
```
blumamir, line 166: It's already exists for sampling and url-templatization. can we reuse?

instrumentor/controllers/utils/instrumentationrules.go
natyharary: Done! Much cleaner

instrumentor/controllers/utils/instrumentationrules.go
natyharary: This is used by 3 controllers (effective config controller, IC controller and instrumentation rule controller).

This happens either when we have updates that should be written to the IC spec, or modifications to the IC itself.


The reconciler logic is what takes care of it, or did I misunderstand you?

frontend/graph/schema.graphqls
natyharary: The sampling rules do not have `containerName` so it isn't 1:1
I can add it to be an optional argument for sampling and then we can reuse it.
I think I am OK with this duplication so we don't force another field on sampling.

WDYT?

instrumentor/controllers/utils/instrumentationrules.go
```diff
@@ -3,23 +3,76 @@ package utils
...
+		containerName := ic.Spec.ContainersOverrides[i].ContainerName
+		runtimeDetails := detailsByContainer[containerName]
+		lang := common.UnknownProgrammingLanguage
+		if runtimeDetails != nil && runtimeDetails.Language != "" {
+			lang = runtimeDetails.Language
+		}
+		if IsWorkloadParticipatingInRule(workload, rule, &containerName, &lang) {
 			return true
 		}
 	}
```
blumamir: Is this filtering needed now that we also filter it again per container when computing it's config?

This code is not trivial both in compute and logic, and I wonder why we still need to keep it.

instrumentor/controllers/agentenabled/sync.go
```diff
@@ -218,6 +217,34 @@ func updateInstrumentationConfigSpec(ctx context.Context, c client.Client, pw k8
...
+		// In we have instrumentation rules and none of them apply to the container, mark the container as AgentEnabled == false
+		if len(rulesForContainer) == 0 && irls != nil && len(*irls) > 0 {
+			containersConfig = append(containersConfig, odigosv1.ContainerAgentConfig{
+				ContainerName:       containerName,
+				AgentEnabled:        false,
+				AgentEnabledReason:  odigosv1.AgentEnabledReasonNoAvailableAgent,
+				AgentEnabledMessage: "container is not included in the source scope of any instrumentation rule that applies to this workload",
+			})
+			continue
+		}
```
blumamir: I don't understand this case. why is it a problem to have container without any rules?

instrumentor/controllers/agentenabled/sync.go
```diff
@@ -218,6 +217,34 @@ func updateInstrumentationConfigSpec(ctx context.Context, c client.Client, pw k8
...
+		rulesForContainer := make([]odigosv1.InstrumentationRule, 0)
+		if irls != nil {
+			for _, rule := range *irls {
+				name := containerName
+				language := containerLanguage
+				if instrutils.IsWorkloadParticipatingInRule(pw, &rule, &name, &language) {
+					rulesForContainer = append(rulesForContainer, rule)
+				}
+			}
+		}
```
blumamir, line 235: shouldn't we check the language after we apply the override?

instrumentor/controllers/agentenabled/distroresolver/distroresolver.go
```diff
@@ -75,7 +77,21 @@ func CalculateDefaultDistroPerLanguage(defaultDistros map[common.ProgrammingLang
...
+
+	// The rules dictate the distro per language. Since we can have multiple rules editing the distro for the same language,
+	// sort the rules since the list order is not deterministic
+	rules := slices.Clone(*instrumentationRules)
+	slices.SortFunc(rules, func(a, b odigosv1.InstrumentationRule) int {
+		if c := cmp.Compare(a.Namespace, b.Namespace); c != 0 {
+			return c
+		}
+		return cmp.Compare(a.Name, b.Name)
+	})
```
blumamir: why sort only for `CalculateDefaultDistroPerLanguage` and not in general for other flows also?

instrumentor/controllers/agentenabled/sync.go
natyharary: I got confused- I thought about how instrumentation rules dictate which and how certain workloads and containers are instrumented.
So in my head -> no IR -> not applied on a container -> error state
(but in the default it actually means that it is instrumented without an IR)

natyharary: Removed it

instrumentor/controllers/agentenabled/distroresolver/distroresolver.go
natyharary: Because languages are the ones that have multiple distros. WDYM by other flows?

instrumentor/controllers/agentenabled/sync.go
natyharary: why? not sure I understand the logic behind double checking it

instrumentor/controllers/utils/instrumentationrules.go
natyharary: I removed checking for relevant instrumentation rules when calling `getRelevantInstrumentationRules()` - it will now not scope down to containers/workloads there, but keep the check here.

instrumentor/controllers/agentenabled/sync.go
natyharary: we are checking the language here: `	runtimeDetailsByContainer := ic.RuntimeDetailsByContainer()`

instrumentor/controllers/agentenabled/distroresolver/distroresolver.go
natyharary: Done

instrumentor/controllers/agentenabled/resources.go
```diff
@@ -16,7 +15,7 @@ import (
 // if err is returned, the reconciliation should be retried.
 // the function can return without error, but some of the resources might be nil.
 // this indicates that they were not found, but it is valid.
-func getRelevantResources(ctx context.Context, c client.Client, pw k8sconsts.PodWorkload) (
+func getRelevantResources(ctx context.Context, c client.Client, pw k8sconsts.PodWorkload, ic *odigosv1.InstrumentationConfig) (
```
blumamir: now do we need the ic here?

instrumentor/controllers/agentenabled/resources.go
```diff
@@ -117,16 +118,6 @@ func getRelevantInstrumentationRules(ctx context.Context, c client.Client, pw k8
 	for i := range irList.Items {
 		ir := &irList.Items[i]
 
-		// ignore disabled rules
-		if ir.Spec.Disabled {
-			continue
-		}
```
blumamir, line 123: why is it removed? what should filter out disabled rules?

instrumentor/controllers/utils/instrumentationrules.go
```diff
@@ -3,23 +3,83 @@ package utils
...
 	"github.com/odigos-io/odigos/api/k8sconsts"
 	odigosv1alpha1 "github.com/odigos-io/odigos/api/odigos/v1alpha1"
+	"github.com/odigos-io/odigos/common"
+	"github.com/odigos-io/odigos/k8sutils/pkg/scope"
 )
 
 // naive implementation, can be optimized.
 // assumption is that the list of workloads is small
-func IsWorkloadParticipatingInRule(workload k8sconsts.PodWorkload, rule *odigosv1alpha1.InstrumentationRule) bool {
+func IsWorkloadParticipatingInRule(
```
blumamir:
```suggestion
func IsContainerParticipatingInRule(
```

instrumentor/controllers/utils/instrumentationrules.go
```diff
@@ -3,23 +3,83 @@ package utils
...
+) bool {
+	if rule.Spec.Disabled {
+		return false
+	}
+	if rule.Spec.SourcesScopes == nil {
+		return true
+	}
+
+	// check if we have SourcesScopes in the instrumentation rule - check those first
+	if rule.Spec.SourcesScopes != nil {
```
blumamir, line 28: we already check `rule.Spec.SourcesScopes == nil` above so this is redundant

instrumentor/controllers/utils/instrumentationrules.go
```diff
@@ -3,23 +3,83 @@ package utils
...
+		scopes := *rule.Spec.SourcesScopes
+		if len(scopes) == 0 {
+			return false
+		}
```
blumamir: shouldn't this be true?

instrumentor/controllers/utils/instrumentationrules.go
```diff
@@ -3,23 +3,83 @@ package utils
...
+		name := ""
+		if containerName != nil {
+			name = *containerName
+		}
+		lang := common.UnknownProgrammingLanguage
+		if language != nil {
+			lang = *language
+		}
```
blumamir: can they be `nil`? should we make them required?

instrumentor/controllers/agentenabled/sync.go
```diff
@@ -218,6 +219,32 @@ func updateInstrumentationConfigSpec(ctx context.Context, c client.Client, pw k8
...
+		// Since we can have multiple rules editing the distro for the same language, sort the rules since the list order is not deterministic
+		rulesForThisContainer := slices.Clone(rulesForContainer)
```
blumamir: why do we need the clone here?

natyharary: We don't. Removed

instrumentor/controllers/agentenabled/resources.go
natyharary: now we dont

instrumentor/controllers/utils/instrumentationrules.go
natyharary: Now they are no longer nullable as we discussed

instrumentor/controllers/utils/instrumentationrules.go
natyharary: True, changed it

instrumentor/controllers/agentenabled/resources.go
natyharary: Brought it back in - this was checked in an earlier piece of the code but now merged back

api/odigos/v1alpha1/instrumentationrule_type.go
```diff
@@ -54,8 +54,8 @@ type InstrumentationRuleSpec struct {
 	// A boolean field allowing to temporarily disable the rule, but keep it around for future use
 	Disabled bool `json:"disabled,omitempty"`
 
-	// An array of workload objects (name, namespace, kind) to which the rule should be applied. If not specified, the rule will be applied to all workloads. empty array will render the rule inactive.
-	Workloads *[]k8sconsts.PodWorkload `json:"workloads,omitempty"`
+	// SourcesScopes lists SourcesScope entries to which the rule should be applied. If not specified, the rule will be applied to all workloads. empty array will render the rule inactive.
+	SourcesScopes *[]k8sconsts.SourcesScope `json:"sourcesScopes,omitempty"`
```
blumamir:
```suggestion
	SourcesScopes []k8sconsts.SourcesScope `json:"sourcesScopes,omitempty"`
```

k8sutils/pkg/scope/scope.go
```diff
@@ -48,10 +22,23 @@ func AnySourceScopeMatchesContainer(
...
+			continue
+		}
+		if scope.ContainerName != "" && scope.ContainerName != containerName {
+			continue
+		}
+		if scope.WorkloadLanguage != "" && scope.WorkloadLanguage != language {
+			continue
+		}
+		return true
 	}
```
blumamir, line 42: out of curiosity, what is the reason for this change?

natyharary: I wanted to combine the two functions - it felt redundant to separate this function into two (that was the original) since it just wraps one function by another in a for loop.
Essentially they do the same - they return early on any mismatch and iterate on all scopes to make sure if we have a scope that matches. If there is a match, it returns True.

image

blumamir (approved):

PR #4827, opened 2026-05-07
CORE-00 | Update Sampling Documentation for Sampling Refactoring

rayray221:
## Summary

Restructure the sampling docs around the rules engine (Noisy Operation, Highly Relevant, Cost Reduction) and retire the legacy per-sampler pages. Adds a configuration reference, end-to-end UI + YAML examples, operational considerations, and FAQs, mirrored across the OSS and Enterprise sidebars.

## Changes

- **New pages**: `introduction`, `head-and-tail-sampling`, `rules/{overview,noisy-operation,highly-relevant,cost-reduction}`, `configuration`, `examples`, `considerations`, `faqs`.
- **Removed**: `errorsampler`, `latencysampler`, `probabilisticsampler`, `servicenamesampler`, `spanattributesampler` (replaced by the rules-based structure above).
- **Sidebar**: OSS and Enterprise navigation updated with a nested `Rules` group; `docs.json` redirects added for the old URLs.
- **Screenshots**: UI walkthroughs for each scenario plus enabled/disabled toggle states under `docs/images/sampling/`.
- **Examples page** covers five canonical scenarios (drop noisy endpoint, keep slow requests, cluster-wide baseline, scope by namespace + language, combine HR + CR) with paired UI steps and CRD YAML, plus a YAML reference accordion for non-obvious schema details.

```release-note
Sampling documentation rewritten around the rules engine, with examples and FAQs.
```


docs/snippets/shared/pipeline/actions/sampling/introduction.mdx
```diff
@@ -0,0 +1,62 @@
+## What is sampling
+
+**Sampling** is how you choose which traces are **kept** in the pipeline and which are **dropped** before they reach backends and UIs. It trims volume so ingest, storage, and search stay manageable, while policies still aim to leave enough signal for debugging and reliability work.
+
+In Odigos, sampling is **per trace**: the entire trace is either retained end to end or not exported, so parent–child structure and timing stay coherent when something is kept. How and when that choice is made—early in the **Data Collector** (**head** sampling) or at the **Odigos Gateway** after spans are assembled (**tail** sampling)—is covered on [Head and tail sampling](./head-and-tail-sampling).
```
blumamir: It actually happens in the agent itself, even before the data collection, right at time of span creation

docs/snippets/shared/pipeline/sampling/introduction.mdx
```diff
@@ -0,0 +1,62 @@
+## What is sampling
+
+**Sampling** is how you choose which traces are **kept** in the pipeline and which are **dropped** before they reach backends and UIs. It trims volume so ingest, storage, and search stay manageable, while policies still aim to leave enough signal for debugging and reliability work.
+
+In Odigos, sampling is **per trace**: the entire trace is either retained end to end or not exported, so parent–child structure and timing stay coherent when something is kept. How and when that choice is made—early in the **Data Collector** (**head** sampling) or at the **Odigos Gateway** after spans are assembled (**tail** sampling)—is covered on [Head and tail sampling](./head-and-tail-sampling).
+
+## Core motivations and definitions
+
+### Primary driver: cost reduction
```
blumamir, line 9: I wonder if we should also mention removing noisy traces for senitation ( health-checks, metrics scraping, infrastructure, etc). since it's also a very common usecase for sampling.

Some people just browse through the traces, and get overwhelmed by tons of health checks and other things that are distraction and make it hard to find what actually matters.

docs/snippets/shared/pipeline/actions/sampling/introduction.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+Traditional framing often starts from dropping data to save money. Odigos shifts the emphasis to **prioritizing what matters**: configuration expresses what should be treated as important (errors, slow paths, specific services or attributes, and so on), and the pipeline retains those traces preferentially. Cost control follows from that prioritization rather than from arbitrary bulk deletion alone.
+
+<Tip>
+  When designing sampling policies, **define “important” first** (failures, SLO risks, critical services), then set fallbacks or rates for everything else. That mirrors how Odigos surfaces configuration and avoids tuning only by a global drop percentage.
+</Tip>
+
+### Sampling level
+
+Sampling in Odigos is **per trace**. The decision to keep or drop is made **at the root span** and applies to every span in the trace. Partial retention would break parent–child links, timing, and attributes that span the whole request, so a kept trace is always **complete** from the root span through its children.
```
blumamir: nit: the root-span decision is only for "Noisy Operation" traces (head-sampling). for tail sampling it's not the case (always any span in the trace)

docs/snippets/shared/pipeline/sampling/introduction.mdx
```diff
@@ -0,0 +1,62 @@
...
+### Sampling level
+
+Sampling in Odigos is **per trace**. The decision to keep or drop is made **at the root span** and applies to every span in the trace. Partial retention would break parent–child links, timing, and attributes that span the whole request, so a kept trace is always **complete** from the root span through its children.
+
+### Key terminology
+
+These terms describe how traces are classified in the sampling model:
+
+<CardGroup cols={3}>
+  <Card title="Important traces" icon="star">
```
blumamir, line 61: should we also include "Noisy Operations" here? so we cover all 3 categories together.

docs/snippets/shared/pipeline/actions/sampling/head-and-tail-sampling.mdx
```diff
@@ -0,0 +1,79 @@
+## Types of sampling implementation
+
+This page contrasts head sampling (decisions in the **Data Collector** when a trace starts) and tail sampling (decisions at the **Odigos Gateway** after spans are brought together). For how Odigos treats whole traces and root-span decisions, see [Introduction to Sampling](./introduction).
+
+### Head sampling (Data Collector)
```
blumamir: nit: it's in the Agent code itself, not Data Collector

docs/snippets/shared/pipeline/actions/sampling/head-and-tail-sampling.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+Head sampling is the most basic and efficient form of sampling. The decision is made at the root span before child spans are processed; the trace in its entirety is either kept or dropped to match that decision.
+
+- **Decision point** — At the root span as the trace starts, inside the Data Collector. The same keep or drop applies to the entire trace (parent-based).
```
blumamir:
```suggestion
- **Decision point** — At the root span as the trace starts, inside the supported instrumentation agents. The same keep or drop applies to the entire trace (parent-based).
```

docs/snippets/shared/pipeline/actions/sampling/head-and-tail-sampling.mdx
```diff
@@ -0,0 +1,79 @@
...
+- **Context** — Limited to root-level fields: service name, entry endpoint, and similar. Typical rule inputs are things like service, root span details, and specific routes (e.g. `foo/about`). You cannot base the decision on errors or latency that appear only on child spans.
```
blumamir:
```suggestion
- **Context** — Limited to root-level fields: service name, entry endpoint, and similar. Typical rule inputs are things like service, root span details, and specific routes (e.g. `foo/about`). You cannot base the decision on errors or latency that appear only on child spans, or at span end time (http response codes, etc)
```

docs/snippets/shared/pipeline/actions/sampling/head-and-tail-sampling.mdx
```diff
@@ -0,0 +1,79 @@
...
+- **Efficiency** — High for the pipeline: for traces that are dropped, only the root span is processed, and Odigos Gateway work stays low on that path. Volume is cut early, before downstream storage, indexing, and query cost.
```
blumamir: If we want to be more accurate, we can say all the spans in a dropped trace are skipped right at span creation time, which uses almost no resources. they are not processed in the agent, not serialized into eBPF maps, not processed at odiglet / data collection, and requires no networking / IO, which is very efficient (when span metrics are enabled this is not the case, but a special case)

docs/snippets/shared/pipeline/sampling/head-and-tail-sampling.mdx
```diff
@@ -0,0 +1,79 @@
...
+**When to use it:**
+
+- **High-Volume Environments:** If your system generates an overwhelming number of traces, head sampling is the most efficient way to drop data early, saving CPU and memory on your application agents and reducing network traffic immediately.
+
+- **Simple Logic:** Use it when you only need to sample based on known "head" attributes, such as the service name, the initial endpoint (e.g., /health vs. /api/orders), or a simple percentage (e.g., "keep 5% of all traffic").
+
+- **Resource Constraints on Collectors:** Since decisions happen at the source, your collectors/gateways don't need to buffer or process data that is already marked to be dropped.
+
+**The Trade-off:** <br/>
+It is a "blind" decision. Because the decision is made at the start, you cannot know if that specific trace will eventually encounter an error or experience high latency. You risk losing the exact "needle in the haystack" traces you might need for debugging.
```
blumamir, line 64: OpenTelemetry describes the "Representativness" principle [here](https://opentelemetry.io/docs/concepts/sampling/#why-sampling) which really helps in understanding the trade-offs and justification for why it's considered good-enough to have statistical sampling with high-volumes (for volume reduction). we can consider saying something about it here

docs/snippets/shared/pipeline/actions/sampling/head-and-tail-sampling.mdx
```diff
@@ -0,0 +1,79 @@
...
+- **Simple Logic:** Use it when you only need to sample based on known "head" attributes, such as the service name, the initial endpoint (e.g., /health vs. /api/orders), or a simple percentage (e.g., "keep 5% of all traffic").
+
+- **Resource Constraints on Collectors:** Since decisions happen at the source, your collectors/gateways don't need to buffer or process data that is already marked to be dropped.
+
+**The Trade-off:** <br/>
+It is a "blind" decision. Because the decision is made at the start, you cannot know if that specific trace will eventually encounter an error or experience high latency. You risk losing the exact "needle in the haystack" traces you might need for debugging.
+
+### Tail-Based Sampling
+
+Tail-based sampling waits until the trace is complete (or a timeout is reached) before deciding whether to keep it. This is typically handled at the collector or gateway level.
```
blumamir:
```suggestion
Tail-based sampling waits until the trace is complete (or a timeout is reached) before deciding whether to keep it. This is handled at the gateway collector level.
```

docs/docs.json
```diff
@@ -211,11 +211,22 @@
                         "group": "Sampling",
```
blumamir: I think it's more appropriate to have it one level above as it's own category, and not under actions, as it is no longer an action, and a big enough topic to be discoverable easily. WDYT?

docs/snippets/shared/pipeline/actions/sampling/rules/overview.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+The pipeline runs the categories in this order. At each stage, every rule in that category is considered; how multiple matches resolve is covered in [Rule evaluation](#rule-evaluation).
+
+## Rule evaluation
+
+### Evaluation inside a category
+
+Within a **single** category, **rule order does not matter**: there is **no** hierarchy among rules.
+
+If a trace **matches more than one rule** in the same category, the outcome uses the **most conservative** interpretation for retention: the rule with the **highest keep percentage** wins among the matches.
```
blumamir: It's the highest for highly-relevance (at-least semantics), and the lowest for cost-reduction and noisy-traces (at-most semantics). this text can be understood differently currently

docs/snippets/shared/pipeline/sampling/rules/noisy-operation.mdx
```diff
@@ -0,0 +1,36 @@
+**Noisy Operation** rules drop or down-sample traffic that you treat as having little or no value—for example Kubernetes liveness and readiness probes, periodic health checks, or scraper traffic. They are the **first** stage in the [rule evaluation pipeline](./overview#order-of-operations-between-categories): a trace that matches a Noisy rule is **not** evaluated against Highly Relevant or Cost Reduction rules.
+
+## How they're evaluated
+
+- **Head sampling at the root span.** The keep/drop decision is made on the entry span, before child spans are processed.
+- **No trace buffering.** Because the decision happens at the start of the trace, the gateway does not need to wait for an [aggregation window](../head-and-tail-sampling#aggregation-window).
+- **Lowest overhead.** Head sampling is the **most efficient** sampling type, with the **least CPU and memory** impact on collectors and gateway.
```
blumamir, line 7: only in supported agents (nodejs, python, java-enterprise and go-enterprise). when not supported, this category is evaluated at tail-sampling time

docs/snippets/shared/pipeline/actions/sampling/rules/noisy-operation.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+<Note>
+  Dropped traces do not continue to other categories. If you need a noisy route to remain visible when something goes wrong (for example a failing health check), replace the Noisy rule with a **Highly Relevant** rule on errors and use a **Cost Reduction** rule to cap the successful traffic.
+</Note>
+
+## Built-in rule: Kubernetes Health Probes
+
+Odigos ships a single built-in Noisy rule, **Kubernetes Health Probes**, which drops traces for the **liveness**, **readiness**, and **startup** probe paths discovered on each workload's Kubernetes manifest.
+
+The rule's **Keep Percentage** controls how much probe traffic is retained. The default is **0%** (drop everything). Raise it temporarily (for example to 100%) while debugging probe failures, then reset to 0%.
```
blumamir: It is sometimes useful to still keep like 2% health-checks just to have couple of examples here and there. can be useful to still have "working" indication for services that have little or no traffic

docs/snippets/shared/pipeline/actions/sampling/rules/noisy-operation.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+<Tip>
+  Only **HTTPGet** probes declared on the manifest are matched. Probes defined as `exec` or `tcpSocket`, or probes routed through different paths (custom configuration, ingress rewrites, service-mesh wrappers), are not auto-handled—add an explicit Noisy rule for those paths if needed.
```
blumamir:
```suggestion
  Only **HTTPGet** probes declared on the manifest are matched. Probes defined as `exec` or `tcpSocket`, or probes routed through different paths (custom configuration, ingress rewrites, service-mesh wrappers), are not auto-handled; add an explicit Noisy rule for those paths if needed.
```

docs/snippets/shared/pipeline/actions/sampling/rules/highly-relevant.mdx
```diff
@@ -0,0 +1,31 @@
+**Highly Relevant** rules mark traces you want to treat as **important**, aligned with **RED**-style signals (request, **error**, **duration**). They are the **second** stage in the [rule evaluation pipeline](./overview#order-of-operations-between-categories), evaluated only on traces that did **not** match any Noisy rule.
+
+## How they're evaluated
+
+- **Tail-aware.** Importance often depends on something that only appears mid-trace—an error on a child span, or a long downstream call—so the evaluator typically needs the **assembled trace**. See [Head vs. tail sampling](../head-and-tail-sampling).
+- **One error span marks the entire trace.** When a Highly Relevant error rule applies, a single error span anywhere in the tree is enough to classify the whole trace as important.
+- **Importance does not mean 100% retention.** Each rule has its own **Keep Percentage** that controls how much matching traffic is retained. You can prioritize errors while still keeping only a fractional sample.
```
blumamir: I would mention that they are kept 100% by default which works for most case, but can be reduced if too much volume is produced from specific highly-relevance operations

docs/snippets/shared/pipeline/sampling/configuration.mdx
```diff
@@ -0,0 +1,87 @@
+You can configure sampling rules in the **UI** or as **YAML** (Kubernetes **CRDs**). The schema matches the same concepts as the UI, so GitOps workflows can express the same behavior as in the product.
+
+Every rule has three parts: a **Source scope** that defines which workloads the rule applies to, an **Operation** that picks the specific operations within that scope, and a **Keep Percentage** that controls how much matching traffic is retained. The Source scope options are identical across all three categories; the Operation and Keep Percentage options differ by category. Each rule can also carry an optional **Name** and **Notes** for easier identification.
```
blumamir, line 3: consider also mentioning that each rule can be enabled/disabled, allowing one to:
- temporary disabled specific sampling for evaluation / incidents / when need to check something
-  not lose the configuration for this rule which might still be useful (disable vs delete)
- disabled rules still record metrics, and allow one to see "what-if" scenarios - e.g. if this disabled rule was enabled, how much traces it would keep/drop and how much traces would it match

docs/snippets/shared/pipeline/actions/sampling/configuration.mdx
```diff
@@ -0,0 +1,87 @@
...
+</Tabs>
+
+## Keep Percentage
+
+Each rule defines what fraction of matching traffic is retained. The available choices have the same shape across all three categories—**Drop All** or **Sample**—but the **meaning** of the Sample percentage differs:
+
+| Category          | Drop All | Sample             |
+| ----------------- | :------: | ------------------ |
+| Noisy Operation   | ✓        | **Keep at most** *N*% |
+| Highly Relevant   | ✓        | **Keep at least** *N*% |
```
blumamir: For highly-relevance it's "Keep All", not Drop, and for Cost-Reduction the percentage is required (no built in option to "Drop-All", although one can write 0% in the custom percentage)

docs/snippets/shared/pipeline/sampling/examples.mdx
```diff
@@ -0,0 +1,363 @@
...
+      namespace: odigos-system
+    spec:
+      name: Drop /api/v1/poll
+      noisyOperations:
+        - name: Drop polling endpoint
+          notes: High-volume poll endpoint with no SLO/debug value.
+          percentageAtMost: 0
+          operation:
+            httpServer:
+              routePrefix: /api/v1/poll
```
blumamir, line 55: For the completeness of the example, you can also include http method here, to scope it only for `GET` for example.

docs/snippets/shared/pipeline/sampling/examples.mdx
```diff
@@ -0,0 +1,363 @@
...
+  </Tab>
+  <Tab title="YAML">
+    ```yaml drop-poll-endpoint.yaml
+    apiVersion: odigos.io/v1alpha1
+    kind: Sampling
+    metadata:
+      name: drop-poll-endpoint
+      namespace: odigos-system
+    spec:
+      name: Drop /api/v1/poll
```
blumamir, line 44: Having rules on the entire cluster means odigos need to evaluate them many times when unnecessary. for specific endpoints, I would include a scope for specific source, to encourage users that might copy-paste from here to populate the scope correctly

docs/snippets/shared/pipeline/sampling/examples.mdx
```diff
@@ -0,0 +1,363 @@
...
+  </Tab>
+  <Tab title="YAML">
+    ```yaml keep-slow-products.yaml
+    apiVersion: odigos.io/v1alpha1
+    kind: Sampling
+    metadata:
+      name: keep-slow-products
+      namespace: odigos-system
+    spec:
+      name: Keep slow product reads
```
blumamir, line 90: also here consider adding a scope for showing best practices

docs/snippets/shared/pipeline/sampling/examples.mdx
```diff
@@ -0,0 +1,363 @@
...
+  </Tab>
+</Tabs>
+
+<Tip>
+  Cost Reduction rules with **no** Operation block apply to every operation in the scope. Combine with Highly Relevant rules to make sure error and latency traces survive the cap—Highly Relevant always wins over Cost Reduction.
+</Tip>
+
+## 4. Scope a rule to namespace and language
+
+**Goal:** Drop scrape traffic to `/internal/metrics` for **Java** workloads in the `prod` namespace only—leave other namespaces and other languages untouched. This example demonstrates the **AND** semantics across scope types: a single source scope item with both `workloadNamespace` and `workloadLanguage` set.
```
blumamir, line 150: might be worth adding that it is recommended to scope rules to only what's relevant, to avoid evaluating them in unrelated cases and wasting resources

docs/snippets/shared/pipeline/sampling/examples.mdx
```diff
@@ -0,0 +1,363 @@
...
+
+## YAML reference notes
+
+A few details that aren't obvious from YAML alone but matter when authoring rules. The full schema—every field, type, and validation rule—is in the [`Sampling` API reference](../../../api-reference/odigos.io.v1alpha1#odigos-io-v1alpha1-Sampling).
+
+<AccordionGroup>
+  <Accordion title="Source scope: language enum and workloadKind casing">
+    The YAML enum for `workloadLanguage` uses lowercase identifiers: `java`, `python`, `go`, `dotnet`, `javascript`, `php`, `ruby`. The UI displays the friendly names (Java, Python, Go, .Net, NodeJS, PHP, Ruby).
+
+    `workloadKind` must match the Kubernetes resource kind exactly (case-sensitive PascalCase): `Deployment`, `StatefulSet`, `DaemonSet`, `Job`, `CronJob`. Lowercase or pluralized values will silently fail to match.
```
blumamir, line 278: We should open a ticket to relax this, I think it's quite easy

docs/snippets/shared/pipeline/sampling/examples.mdx
```diff
@@ -0,0 +1,363 @@
...
+    - **`operation` omitted** — the rule matches **every** operation in the scope.
+    - **Noisy `percentageAtMost` omitted** — defaults to **0%** (drops everything).
+    - **Highly Relevant `percentageAtLeast` omitted** — defaults to **100%** (keeps everything).
+    - **Cost Reduction `percentageAtMost`** — **required**. There is no default; the resource will fail validation if the field is absent.
+  </Accordion>
+
+  <Accordion title="HTTP matchers: route, method, and templated paths">
+    - On HTTP Server matchers, set **either** `route` (exact match) **or** `routePrefix` (prefix match)—not both.
+    - `method` is a free-form string in YAML. The UI restricts it to GET, POST, PATCH, DELETE, and PUT, but YAML accepts any method (e.g., HEAD, OPTIONS). Leave it empty to match any method.
+    - HTTP Client matchers (Noisy only) use `templatedPath` / `templatedPathPrefix`—the framework's parameterized path (e.g., `/users/{id}`), **not** the resolved URL (e.g., `/users/42`). Match against the route as your HTTP library reports it.
```
blumamir, line 294: the route in http server also resolves templated values (`/users/{id}`, or `/users/:id`, or `/users/*`). this is not only http client feature

docs/snippets/shared/pipeline/actions/sampling/considerations.mdx
```diff
@@ -0,0 +1,13 @@
...
+
+For tail sampling, prefer **scaling up** (bigger collector pods) over **scaling out** (more replicas). All spans of a single trace must land on the same pod for aggregation, so fewer, larger pods simplify routing and load balancing. Validate sizing against your own load—there's no one-size-fits-all.
+
+## When any sampler enables tail mode
+
+<Warning>
+  Any sampling rule that requires tail evaluation activates **tail mode** for the matching traffic, even rules that look inactive or minimal—for example a placeholder service name or a broad `/` route. Once tail mode is on, the gateway buffers each trace for the [aggregation window](./head-and-tail-sampling#aggregation-window), which increases memory use, export latency, and the risk of dropped spans. If you don't need tail behavior, remove any unused or test sampling rules from the cluster rather than leaving them in place.
```
blumamir:
```suggestion
  Any sampling rule that requires tail evaluation activates **tail mode** for the entire cluster, even rules that look inactive or minimal—for example a placeholder service name or a broad `/` route. Once tail mode is on, the gateway buffers each trace for the [aggregation window](./head-and-tail-sampling#aggregation-window), which increases memory use, export latency, and the risk of dropped spans. If you don't need tail behavior, remove any unused or test sampling rules from the cluster, or disable tail sampling in odigos configuration
```

docs/snippets/shared/pipeline/actions/sampling/faqs.mdx
```diff
@@ -0,0 +1,101 @@
...
+  </Accordion>
+
+  <Accordion title="Why is dropping health checks called 'sampling' if I'm dropping 100% of them?">
+    "Sampling" is the name of the **pipeline mechanism** that makes a per-trace keep/drop decision—not a description of "always keep some, always drop some." Dropping all health probes is just sampling at 0%. The same engine that drops 100% of probe traffic also retains 25% of `/api/products` traffic; both are sampling decisions, just at different keep percentages.
+  </Accordion>
+
+  <Accordion title="Why does Odigos always keep or drop the entire trace?">
+    Odigos uses **parent-based sampling**: the keep/drop decision is made at the root span and propagated to every child span. Splitting a trace—keeping some spans and dropping others—would break parent-child links, end-to-end timing, and any analysis that depends on the full request path. Trace-level decisions guarantee that a kept trace is **complete** from root through children. See [Sampling level](./introduction#sampling-level).
```
blumamir: this applies to head-sampling only. tail mechanism is different

blumamir: there are actually mechanisms to drop parts of a trace without breaking it to fragments, and we might explore them in the future. currently we use entire-trace sampling as it is most common and straight-forward to use.

docs/snippets/shared/pipeline/sampling/faqs.mdx
```diff
@@ -0,0 +1,101 @@
...
+    Odigos uses **parent-based sampling**: the keep/drop decision is made at the root span and propagated to every child span. Splitting a trace—keeping some spans and dropping others—would break parent-child links, end-to-end timing, and any analysis that depends on the full request path. Trace-level decisions guarantee that a kept trace is **complete** from root through children. See [Sampling level](./introduction#sampling-level).
+  </Accordion>
+</AccordionGroup>
+
+## How rules are evaluated
+
+<AccordionGroup>
+  <Accordion title="If a trace matches rules in more than one category, which category wins?">
+    Categories are evaluated in a fixed sequence: **Noisy → Highly Relevant → Cost Reduction**. A **Noisy** match short-circuits the rest—the trace is dropped (or sampled down) and the other categories are skipped. A **Highly Relevant** match does **not** short-circuit—Cost Reduction rules are still evaluated, but the Highly Relevant Keep Percentage wins. Cost Reduction caps only take effect on traces with no Highly Relevant match. See [Order of operations between categories](./rules/overview#order-of-operations-between-categories).
```
blumamir, line 23: > A **Highly Relevant** match does **not** short-circuit—Cost Reduction rules are still evaluated

Once a highly-relevant rule matched, we use it to make a decision and does not evaluate cost-reduction for this trace. It's the same outcome eventually, just how the feature is implemented (also the metrics for cost-reduction only includes what actually tested which exclude highly-relevant traces)

docs/snippets/shared/pipeline/actions/sampling/faqs.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+<AccordionGroup>
+  <Accordion title="If a trace matches rules in more than one category, which category wins?">
+    Categories are evaluated in a fixed sequence: **Noisy → Highly Relevant → Cost Reduction**. A **Noisy** match short-circuits the rest—the trace is dropped (or sampled down) and the other categories are skipped. A **Highly Relevant** match does **not** short-circuit—Cost Reduction rules are still evaluated, but the Highly Relevant Keep Percentage wins. Cost Reduction caps only take effect on traces with no Highly Relevant match. See [Order of operations between categories](./rules/overview#order-of-operations-between-categories).
+  </Accordion>
+
+  <Accordion title="Within a category, do rules have a priority order?">
+    No. Within a single category, rules have **no hierarchy**—they're all equal. When a trace matches multiple rules in the same category, the engine takes the **most conservative** outcome for retention (the highest Keep Percentage among the matches). This is a deliberate design choice: ordered rule lists (like firewall rules) become unmaintainable when you have dozens of rules and have to mentally simulate the order to predict behavior.
```
blumamir: > the engine takes the **most conservative** outcome for retention (the highest Keep Percentage among the matches)

only in highly-relevance. for the "dropping" categories it takes the minimum value from all matches

docs/snippets/shared/pipeline/actions/sampling/faqs.mdx
```diff
@@ -0,0 +1,101 @@
...
+  <Accordion title="Can I combine multiple scope types (e.g. namespace + language) in a single rule?">
+    Yes. Within a single scope type, multiple selections are combined with **OR** (e.g. namespaces `prod` OR `staging`). Across different scope types, selections are combined with **AND** (e.g. namespace `prod` AND language Java matches **only** Java workloads in `prod`—not every workload in `prod` and not every Java workload across the cluster). See [Source scope](./configuration#source-scope).
+  </Accordion>
+</AccordionGroup>
+
+## Operational impact
+
+<AccordionGroup>
+  <Accordion title="Does adding any sampling rule increase load on the gateway?">
+    It depends on the rule. **Noisy** rules use head sampling and run in the data collector—they reduce gateway load. **Highly Relevant** and **Cost Reduction** rules require tail evaluation, which means the gateway buffers each affected trace for the [aggregation window](./head-and-tail-sampling#aggregation-window). Even a single tail-mode rule activates buffering for matching traffic, raising memory and CPU on the gateway.
```
blumamir:
```suggestion
    It depends on the rule. **Noisy** rules use head sampling and run in the instrumentation agent where supported—they reduce gateway load. **Highly Relevant**, **Cost Reduction** and **Noisy** rules in agents with no head-sampling support, require tail evaluation, which means the gateway buffers each affected trace for the [aggregation window](./head-and-tail-sampling#aggregation-window). Even a single tail-mode rule activates buffering for all traffic, raising memory and CPU on the gateway.
```

docs/snippets/shared/pipeline/actions/sampling/faqs.mdx
```diff
@@ -0,0 +1,101 @@
...
+  </Accordion>
+
+  <Accordion title="What happens to in-flight traces if a gateway crashes?">
+    They are lost. Odigos does not currently use persistent queues or full backpressure between the data collector and the gateway, so a crash drops everything the gateway was buffering. **Graceful** shutdowns (rolling updates, scale-downs) flush their buffers before exit; only ungraceful crashes lose data.
+  </Accordion>
+
+  <Accordion title="My sampler looks empty or inactive—why is it still affecting my pipeline?">
+    The presence of **any** sampling rule that requires tail evaluation activates tail mode for matching traffic—even rules that look minimal, such as a placeholder service name or a broad `/` route. Once tail mode is on, the gateway buffers traces for the aggregation window, which raises memory and latency to export. If you don't need tail behavior, remove unused or test sampling rules from the cluster. See [When any sampler enables tail mode](./considerations#when-any-sampler-enables-tail-mode).
```
blumamir: or disable tail-sampling explicitly in odigos configuration (helm values, or the ui "Settings" page)

docs/snippets/shared/pipeline/actions/sampling/faqs.mdx
```diff
@@ -0,0 +1,101 @@
...
+  </Accordion>
+
+  <Accordion title="Why isn't there a built-in rule for high-latency traces?">
+    Latency thresholds depend on the application: 200 ms is fast for one service and unacceptable for another, so there's no sensible default the product can ship. Errors are different—an error span is universally a signal worth keeping—which is why the error rule is built in but a duration rule is not. Define your own Highly Relevant duration rule with the threshold and Keep Percentage that fit your service.
+  </Accordion>
+
+  <Accordion title="Are there built-in operations beyond HTTP and Kafka (e.g. SQS, gRPC, database calls)?">
+    Today's Operation options are **HTTP Server**, **HTTP Client** (Noisy only), and **Kafka Consumer / Kafka Producer** (Highly Relevant and Cost Reduction). Odigos will add support for additional protocols as customer needs demand it.
```
blumamir:
```suggestion
    Today's Operation options are **HTTP Server**, **HTTP Client** (Noisy only), and **Kafka Consumer / Kafka Producer** (Highly Relevant and Cost Reduction). Odigos will add support for additional protocols as users needs demand it.
```

blumamir (commented):
@rayray221 this is great!!!!! I love it 👑 thank you so much 🙏

I left few nits here and there you can choose to address if you like

blumamir (approved):

docs/snippets/shared/pipeline/actions/sampling/faqs.mdx
rayray221: Replaced

docs/snippets/shared/pipeline/actions/sampling/faqs.mdx
rayray221: Added

docs/snippets/shared/pipeline/actions/sampling/faqs.mdx
rayray221: Replaced

docs/snippets/shared/pipeline/actions/sampling/faqs.mdx
rayray221: Clarified

docs/snippets/shared/pipeline/sampling/faqs.mdx
rayray221: I'm confused then.  If the highly relevant is a percentage, do we not consider the Cost Reduction rules?  How is it the same outcome if the Cost Reduction rules are not evaluated?

docs/snippets/shared/pipeline/actions/sampling/faqs.mdx
rayray221: Leaving for now for clarify until we add support for it

rayray221: Modified

docs/snippets/shared/pipeline/actions/sampling/considerations.mdx
rayray221: Replaced

docs/snippets/shared/pipeline/sampling/examples.mdx
rayray221: Fixed

docs/snippets/shared/pipeline/sampling/examples.mdx
rayray221: https://linear.app/odigos/issue/CORE-1005/relax-naming-for-programing-language-when-creating-sampling-rules

docs/snippets/shared/pipeline/sampling/examples.mdx
rayray221: Added tip to top of page

docs/snippets/shared/pipeline/sampling/examples.mdx
rayray221: Added

docs/snippets/shared/pipeline/sampling/examples.mdx
rayray221: updated

docs/snippets/shared/pipeline/sampling/examples.mdx
rayray221: Updated

docs/snippets/shared/pipeline/actions/sampling/configuration.mdx
rayray221: Fixed

docs/snippets/shared/pipeline/sampling/configuration.mdx
rayray221: Added

docs/snippets/shared/pipeline/actions/sampling/rules/highly-relevant.mdx
rayray221: Added

docs/snippets/shared/pipeline/actions/sampling/rules/noisy-operation.mdx
rayray221: Replaced

docs/snippets/shared/pipeline/actions/sampling/rules/noisy-operation.mdx
rayray221: Added this point

docs/snippets/shared/pipeline/sampling/rules/noisy-operation.mdx
rayray221: Added for clarity

docs/snippets/shared/pipeline/actions/sampling/rules/overview.mdx
rayray221: Modified

docs/docs.json
rayray221: Agreed!!  Moved

docs/snippets/shared/pipeline/actions/sampling/head-and-tail-sampling.mdx
rayray221: Replaced

docs/snippets/shared/pipeline/sampling/head-and-tail-sampling.mdx
rayray221: Added reference and link

docs/snippets/shared/pipeline/actions/sampling/head-and-tail-sampling.mdx
rayray221: Adjusted

docs/snippets/shared/pipeline/actions/sampling/head-and-tail-sampling.mdx
rayray221: Replaced

docs/snippets/shared/pipeline/actions/sampling/head-and-tail-sampling.mdx
rayray221: Replaced

docs/snippets/shared/pipeline/actions/sampling/head-and-tail-sampling.mdx
rayray221: Corrected

docs/snippets/shared/pipeline/sampling/introduction.mdx
rayray221: Added

docs/snippets/shared/pipeline/actions/sampling/introduction.mdx
rayray221: Adjusted

docs/snippets/shared/pipeline/sampling/introduction.mdx
rayray221: Added

docs/snippets/shared/pipeline/actions/sampling/introduction.mdx
rayray221: corrected

docs/snippets/shared/pipeline/sampling/head-and-tail-sampling.mdx
```diff
@@ -0,0 +1,83 @@
...
+- **Accuracy** — Full-trace context lets retention match policy on errors, latency, or any span—but evaluation runs only after the [aggregation window](#aggregation-window), so traces show up later in UIs, and very long requests may be judged on a partial trace.
+- **Best for** — Debugging errors and slow requests end-to-end; full-trace sampling goals; percentage-style or volume cuts after you have full-path context; policies that need services or routes off the entry span.
+
+#### Aggregation window
+
+The pipeline does not wait for a definitive “end of trace” event, because that would require an indefinite wait. Instead, spans are held up to a bounded window to bound memory and latency. The implementation uses a **default 30-second** interval from the first observed span for a trace ID, consistent with OpenTelemetry-style tail processing, then forwards the trace to the sampler with the spans received by then. Evaluation can therefore run on partial traces for very long requests.
```
blumamir, line 27: > The pipeline does not wait for a definitive “end of trace” event, because that would require an indefinite wait.

the issue is not indefinite wait, it's more of the mechanism of how a trace is composed, which inherently carries the outcome that a span for a trace can in practice arrive at any time. this is common when tasks are pushed to queues, and can "sit and wait" in the queue for unbounded time, before being consumed on the other end and continuing a trace.

side-note: This is a common argument for why messaging producer-consumer traces should prefer to use span links instead of parent-child relationship - since it creates those huge traces that can sometimes last days with short activity at start and end, while confusing sampling and other components.

docs/snippets/shared/pipeline/sampling/considerations.mdx
```diff
@@ -0,0 +1,13 @@
+Operational and design considerations to keep in mind when running sampling in production.
+
+## Gateway load, OOM, and scaling
+
+When the gateway accepts a trace for tail evaluation, it allocates memory to hold spans until the [aggregation window](./head-and-tail-sampling#aggregation-window) closes. If the gateway isn't sized correctly, bursts of traffic can cause **OOM** restarts. Symptoms include **dropped spans** or collector crashes that aren't caused by your sampling rules.
```
blumamir, line 5: we can mention that in this case, one can either allocate more resources (maximum visibility with cost), or add head-sampling for specific high-volume traces to reduce volumes early (representativeness traces, with less cpu-memory cluster costs)

docs/snippets/shared/pipeline/sampling/configuration.mdx
```diff
@@ -0,0 +1,93 @@
+You can configure sampling rules in the **UI** or as **YAML** (Kubernetes **CRDs**). The schema matches the same concepts as the UI, so GitOps workflows can express the same behavior as in the product.
```
blumamir, line 1: Now that I read it again, we should probably have a section for the static cluster-deployment configurations via helm / central / ui, and the rules.

there is [this section in helm](https://github.com/odigos-io/odigos/blob/f01e5de59b3ced27965db670b11b52f1a9ee532a/helm/odigos/values.yaml#L1567) with many configuration that we might want to document here? or should we add these settings in a global configuration docs section we will add later on? open to hear your thoughts @rayray221

docs/snippets/shared/pipeline/sampling/rules/overview.mdx
```diff
@@ -0,0 +1,137 @@
+This page describes how **sampling rules** are evaluated in Odigos: the three **rule categories** (Noisy Operation, Highly Relevant, Cost Reduction), the **order between categories**, and how **conflicts inside a category** resolve. Each category has its own page; this overview focuses on how the categories interact.
+
+## Rule categories
+
+<CardGroup cols={3}>
+  <Card title="Noisy Operation" icon="filter" href="./noisy-operation">
+    Traffic with little or no value (for example Kubernetes health probes). Decided early at the **root span** with **head sampling**.
+  </Card>
+  <Card title="Highly Relevant" icon="star" href="./highly-relevant">
+    **Important** traffic aligned with **RED**-style signals (errors, duration). Each rule's **Keep Percentage** controls how much is retained.
```
blumamir, line 10: nit: consider shortening it here and removing the 
```suggestion
    **Important** traffic aligned with **RED**-style signals (errors, duration). 
```

since the default and what most people will want is 100%. the "advanced" options are documented later :)

docs/snippets/shared/pipeline/sampling/introduction.mdx
```diff
@@ -0,0 +1,75 @@
+## What is sampling
```
blumamir, line 1: WDYT after we experiment with this feature a bit, we can later add some "quick start" or `TL;DR` section to quickly on-board users to sampling with common basic settings.

Could be nice to have them with a live setup in a minute or two, and then let them explore and fine tune certain parts based on the full documentation. 

I am sure many users would highly appreciate your thorough work on capturing the complexity of sampling and various trade-offs in these docs. guessing most users would like quick simple steps that will get them started with something minimal (perhaps with dry-run enabled so they can experiment freely without any risk).

docs/snippets/shared/pipeline/sampling/rules/overview.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+<Note>
+  See [Head vs. tail sampling](../head-and-tail-sampling) for how head and tail sampling differ in Odigos (root-span decisions versus aggregation at the gateway).
+</Note>
+
+If you only define **Highly Relevant** rules and **no Cost Reduction** rules, unimportant traffic is not aggressively reduced by those rules alone—policy may still require full-trace context to evaluate importance, which increases gateway work without cutting volume. Use **Highly Relevant** and **Cost Reduction** together when you need both signal and volume control.
+
+## Order of operations between categories
+
+Categories are **not** all evaluated for every trace in parallel. Odigos applies them in this fixed **sequence**:
```
blumamir, line 25: suggestion: while the detailed explanation is awesome, the mechanism is designed to be intuitive, so that people can just classify traces and don't need to worry about the exact sequence. perhaps we can have a `TL;DR` or diagram here that shows how it works (just a quick draft):

```
Noisy -> No                -> Highly relevance -> No       -> Cost reduction -> No -> Keep
-> Yes (drop or sample)      -> Yes (keep or sample)        -> Yes (sample)      
```

docs/snippets/shared/pipeline/sampling/rules/overview.mdx
```diff
@@ -0,0 +1,137 @@
...
+    - **Highly Relevant** — keep **25%** of traces longer than **500 ms**.
+    - **Cost Reduction** — keep at most **50%** of `/api/products` traffic.
+
+    **Outcome:** the trace is **kept at 25%**. Highly Relevant still wins, even though its keep rate is lower than the Cost Reduction cap.
+  </Card>
+</CardGroup>
+
+## Why full traces still matter
+
+For traces that are evaluated on the **tail** path, the evaluator needs the assembled trace (within the aggregation window) so it can apply **Highly Relevant** conditions that depend on any span. That is why tail sampling and aggregation are part of the pipeline for those policies, even when the category order is Noisy → Highly Relevant → Cost Reduction.
```
blumamir, line 137:
```suggestion
For traces that are evaluated on the **tail** path, the evaluator needs the assembled trace (within the aggregation window) so it can apply **Highly Relevant** and **Cost Reduction** conditions that depend on any span. That is why tail sampling and aggregation are part of the pipeline for those policies, even when the category order is Noisy → Highly Relevant → Cost Reduction.
```

docs/snippets/shared/pipeline/sampling/rules/highly-relevant.mdx
```diff
@@ -0,0 +1,31 @@
+**Highly Relevant** rules mark traces you want to treat as **important**, aligned with **RED**-style signals (request, **error**, **duration**). They are the **second** stage in the [rule evaluation pipeline](./overview#order-of-operations-between-categories), evaluated only on traces that did **not** match any Noisy rule.
+
+## How they're evaluated
+
+- **Tail-aware.** Importance often depends on something that only appears mid-trace—an error on a child span, or a long downstream call—so the evaluator typically needs the assembled trace. See [Head vs. tail sampling](../head-and-tail-sampling).
+- **One error span marks the entire trace.** When a Highly Relevant error rule applies, a single error span anywhere in the tree is enough to classify the whole trace as important.
```
blumamir, line 6: to be more accurate - one can define rules for errors and use the source scopes to only apply them to specific sources/endpoints, or have different keep percentage for different sources or endpoints. 

kind of an edge case I know, I guess most people would prefer to just have a global rule for any operation and source, but still possible :)

docs/snippets/shared/pipeline/sampling/faqs.mdx
blumamir: If something matches "highly relevance" then we either keep it or drop it based on the percentage, but there is no need to evaluate the "cost reduction" as it will not be applied in this case.

> A **Highly Relevant** match does **not** short-circuit—Cost Reduction rules are still evaluated, but the Highly Relevant Keep Percentage wins

the accurate mechanism is that highly relevant match **does** short-circuit and cost reduction are not evaluated, which means that the highly-relevance percentage is taken either way.

docs/snippets/shared/pipeline/sampling/examples.mdx
```diff
@@ -0,0 +1,376 @@
...
+    spec:
+      name: Drop /api/v1/poll
+      noisyOperations:
+        - name: Drop polling endpoint
+          notes: High-volume poll endpoint with no SLO/debug value.
+          percentageAtMost: 0
+          sourceScopes:
+            - workloadNamespace: odimall
+              workloadKind: Deployment
+              workloadName: notification-service
```
blumamir, line 52: there was actually a bug in the implementation that carried a change in the schema of this field: https://github.com/odigos-io/odigos/blob/f01e5de59b3ced27965db670b11b52f1a9ee532a/helm/odigos/templates/crds/odigos.io_samplings.yaml#L119

It is now an object with 3 options: `sources`, `languages`, and `namespaces`, each is optional, and if set is a list of values.

This aligns the schema with the ui and the logical behavior

blumamir (approved):
@rayray221 this is super super great, it's written so well and covers the important aspect in clear and through way. you are the best.

While this already amazing as is, I read it again and added few nits or suggestions, but they are all minors we can choose to address or skip.

Thank you again for this great work 👑

PR #4828, opened 2026-05-07
fix: handle readonly mode for APITokens resolver

BenElferink:
```release-note
fix: handle readonly mode for APITokens resolver
```

emergency-ticket id: EMR-1668


frontend/graph/schema.resolvers.go
```diff
@@ -30,6 +30,12 @@ import (
 
 // APITokens is the resolver for the apiTokens field.
 func (r *computePlatformResolver) APITokens(ctx context.Context, obj *model.ComputePlatform) ([]*model.APIToken, error) {
+	// In readonly mode the UI's RBAC role does not grant `get` on secrets,
+	// and readonly users should not be able to see/edit the on-prem token anyway.
+	if services.IsReadonlyMode(ctx) {
+		return make([]*model.APIToken, 0), nil
```
alonkeyval: looks good, small suggest:

```suggestion
return []*model.APIToken{}, nil
```

or 

```
return nil, nil
```

alonkeyval (approved):

PR #4831, opened 2026-05-10
CORE-812 | OTLP profiles export to Destinations (Pyroscope full support)

shivanshuraj1333:

Demo Video: https://www.loom.com/share/e41c5e22e66f4be2aa4b483017890c26

Adds PROFILES as a first-class destination signal and wires the cluster gateway and OTLP exporters (gRPC, HTTP, Pyroscope preset).

Changes inclludes:
- [X] API:
  - PROFILES on ObservabilitySignal
  - CRD/Helm regen
  - catalog YAMLs + destinations/model
  - profiles.supported changes

- [X] Gateway:
  - addProfilesPipeline adds profiles/<p>-<dest> (otlp → exporter)

- [X] Dest config:
  - OTLP gRPC and OTLP HTTP add profile pipelines when the destination exports PROFILES
  - HTTP supports OTLP_HTTP_PROFILES_ENDPOINT
  - New Pyroscope preset + destinations/data/pyroscope.yaml.

- [X] config_builder:
  - After destinations run, scan Service.Pipelines for profiles/* so enabledSignals / gateway ReceiverSignals include PROFILES

- [X] Node collector:
  - New transform in the profiles pipeline maps K8s workload metadata to service.name so backends (e.g. Pyroscope) show sensible service names instead of generic process labels.

- [X] GraphQL:
  - Schema + gqlgen + Go destination services + webapp destination queries/mutations include profiles
  - docs snippets synced 

```release-note
OLTP Profiles support in Odigos Destinations
```


BenElferink (approved):

destinations/data/dynamic.yaml
```diff
@@ -13,14 +13,16 @@ spec:
       supported: true
     logs:
       supported: true
+    profiles:
+      supported: false
```
blumamir: I think dynamic should also support profiles, as it is used a lot for adding debug exporter which is convenient

autoscaler/controllers/common/profiling_otlp.go
```diff
@@ -101,3 +101,21 @@ func ProfilingFilterProcessorConfig() config.GenericMap {
...
 }
+
+// ProfilingServiceNameTransformConfig sets resource attribute service.name from K8s workload
+// metadata enriched by k8sattributes. Without this, Pyroscope derives labels from
+// process.executable.name and shows names like unknown_service:java.
+func ProfilingServiceNameTransformConfig() config.GenericMap {
+	return config.GenericMap{
+		"error_mode": "ignore",
+		"profile_statements": []string{
+			`set(resource.attributes["service.name"], Concat([resource.attributes["k8s.namespace.name"], resource.attributes["k8s.deployment.name"]], "/")) where resource.attributes["k8s.deployment.name"] != nil and resource.attributes["k8s.namespace.name"] != nil`,
```
blumamir: this is not consistent with other signals (which does not include the namespace).
wonder if we should call it just by the workload name and change them together?

blumamir (approved):

PR #4843, opened 2026-05-11
CORE-956 | URL templatization explicit rules not applied when attribute already set

shivanshuraj1333:

When a span arrives with url.template or http.route already set (by a Java framework or a previous heuristic-only processor pass during startup), the processor exits early and user-defined explicit rules are never applied, the span silently keeps the stale value.

Now, we have changed the unconditional early-exit in enhanceSpanWithRules to only skip when no explicit rules are configured, so explicit rules always take priority over any pre-existing attribute value

```release-note
None
```


blumamir:
the original issue is already fixed, current PR is not relevant

PR #4859, opened 2026-05-13
CORE-911: (odiglet) Support retry failed instrumentations

damemi:
This adds a retryInstrumentations channel to the odiglet manager, which can be triggered to call `tryInstrument()` on apps that previously failed. The goal is to allow retrying the same instrumentation on the same PID without restarting the odiglet or app, such as when Go offsets are updated in the cluster.

Go enterprise implementation here: https://github.com/odigos-io/odigos-enterprise/pull/2819

It adds a new map `failedDetailsByDistro` that keys instrumentation attempts by Distro and PID. This allows triggering failed instrumentations only for a specific distro.

When an instrumentation fails, the details are added to the new map. Then, whenever a message is received on the retry channel, the matching Distro (or all) instrumentations will be re-attempted.

The channel is added as a field in InstrumentationManagerOptions, which allows it to be used by enterprise.

`startTrackInstrumentation()` is updated to handle the error cases (adding to the failed map), and the new success cases (ie, a retry where we remove from the failed map). Metrics semantics are kept the same by checking if this is the first time we've tried to instrument this app.

```release-note
NONE
```

no user-facing changes in this PR

cherry-pick:v1.27


instrumentation/manager.go
```diff
@@ -57,6 +57,10 @@ type instrumentationDetails[processGroup ProcessGroup, configGroup ConfigGroup,
 	pd   processDetails
 	cg   configGroup
 	pg   processGroup
+	// distroName is the name of the OTel distribution that was used (or attempted) to instrument
+	// this process. Tracked here so we can target retry of failed instrumentations by distribution
+	// without re-resolving the process details.
+	distroName string
```
RonFed: `pd` has a `Distribution` function we can use
```suggestion
```

instrumentation/manager.go
```diff
@@ -90,6 +94,18 @@ type ManagerOptions[processGroup ProcessGroup, configGroup ConfigGroup, processD
...
+	// failed instrumentations. When a value is received, the manager will iterate through all
+	// processes whose instrumentation has previously failed (or never been created) and try to
+	// instrument them again. The value is a list of OTel distribution names that act as a filter:
+	// only failed instrumentations whose distribution matches one of the supplied names will be
+	// retried. An empty or nil slice retries every failed instrumentation regardless of distribution.
+	//
+	// This is intended for cases where some external state that affects instrumentation has been
+	// updated (for example, custom Go offsets being reloaded after a ConfigMap change) and we want
+	// to give previously-failed processes another chance without restarting the manager.
+	RetryFailedInstrumentations <-chan []string
```
RonFed: why not re-use the current `InstrumentationRequests` channel? If needed we can add a retryFailed field to the request struct

instrumentation/manager.go
```diff
@@ -139,10 +155,21 @@ type manager[processGroup ProcessGroup, configGroup ConfigGroup, processDetails
...
 	detailsByProcessGroup map[processGroup]map[int]*instrumentationDetails[processGroup, configGroup, processDetails]
 
+	// failedDetailsByDistro indexes only the instrumentations that previously failed (inst == nil),
+	// keyed by their OTel distribution name and then by pid. It is a sub-index of detailsByPid and
+	// is kept in sync by startTrackInstrumentation / stopTrackInstrumentation. Having a dedicated
+	// index lets retryFailedInstrumentationsForDistros target exactly the entries that may benefit
+	// from a retry without scanning every successfully instrumented process.
+	//
+	// this map is not concurrent safe, so it should be accessed only from the main event loop.
+	failedDetailsByDistro map[string]map[int]*instrumentationDetails[processGroup, configGroup, processDetails]
```
RonFed: Can we re-use the existing `detailsByPid` map? note that this map tracks also failed instrumentations (when inst is nil (you can see the comment there).

RonFed (changes requested):

cli/cmd/pro.go
```diff
@@ -176,13 +176,7 @@ odigos pro update-offsets --from-file /path/to/local/offsets.json
 			os.Exit(1)
 		}
 
-		fmt.Println("Updated Go Offsets, restarting Odiglet to use the new offsets.")
-		err = kube.RestartOdiglet(ctx, client, ns)
```
RonFed, line 180: should we remove this function?

instrumentation/manager.go
damemi: Yeah I thought that iterating over every detailsByPid when we get a retry request for a distro wouldn't be very efficient if the failed ones are a smaller subset of overall instrumentations. wdyt?

odiglet/odiglet.go
```diff
@@ -193,9 +209,10 @@ func (o *Odiglet) Run(ctx context.Context) {
 		if o.configUpdates != nil {
 			close(o.configUpdates)
 		}
-		if o.instrumentationRequests != nil {
-			close(o.instrumentationRequests)
-		}
+		// We don't close instrumentationRequests here: the manager's runEventLoop returns on
+		// ctx.Done() so it doesn't need a channel close to terminate, and external producers
+		// can obtain the write-end via InstrumentationRequests(). Closing would race with
+		// those producers (including the OSS kube reconcilers themselves if they're mid-send).
```
RonFed: so where are we closing the channel?

odiglet/odiglet.go
```diff
@@ -32,14 +32,30 @@ import (
...
 
+// InstrumentationRequests returns the write-end of the instrumentation manager's request
+// channel. External producers (e.g. the enterprise odiglet's Go offsets file watcher) can use
+// this to send instrumentation, un-instrumentation, or retry-failed requests alongside the OSS
+// odiglet's own kube reconcilers. See instrumentation.Request for the encoding of each request
+// kind.
+//
+// Callers must use a non-blocking send (select with a default branch) and must NOT close the
+// returned channel: the OSS odiglet stops the consumer via ctx.Done(), and an external close()
+// would race with the kube reconcilers that also write to the same channel.
```
RonFed: think we should not make assumptions on how other parts of the code are doing stuff in the comments - these kind of comments tend to become stale and not up to date when the code changes.

```suggestion
// returned channel.
```

instrumentation/manager.go
```diff
@@ -85,9 +95,10 @@ type ManagerOptions[processGroup ProcessGroup, configGroup ConfigGroup, processD
 	// The caller is responsible for closing the channel once no more updates are expected.
 	ConfigUpdates <-chan ConfigUpdate[configGroup]
 
-	// InstrumentationRequests is a channel for receiving explicit instrumentation/un- instrumentation requests.
-	// The sender can request instrumentation for specific processes by providing their process details mapped by pid.
-	// For un-instrumentation requests, the sender provides the process group to un-instrument all processes that match it.
+	// InstrumentationRequests is a channel for receiving explicit instrumentation, un-
+	// instrumentation, or retry-failed requests. See the Request docs for the encoding of each
+	// request type. The same channel carries all three so callers (e.g. enterprise odiglet) can
+	// piggy-back retry signals on the existing pipeline without plumbing a second channel.
```
RonFed:
```suggestion
	// request type.
```

instrumentation/manager.go
```diff
@@ -309,6 +327,67 @@ func (m *manager[ProcessGroup, ConfigGroup, ProcessDetails]) runEventLoop(ctx co
...
+// retryFailedInstrumentationsForDistros re-attempts instrumentation for every entry in
+// detailsByPid whose previous initialize/load attempt failed (inst == nil). When distroFilter is
+// non-empty, only failed entries whose OTel distribution name matches one of the supplied values
+// are retried; an empty (but non-nil) filter retries every failed entry regardless of
+// distribution.
+//
+// We iterate the full detailsByPid map here because there is typically a small number of tracked
+// processes per node and maintaining a parallel "failed" sub-index doesn't pay for itself. The
+// distribution name comes from pd.Distribution(ctx), which is already memoized after the first
+// successful resolution that happened when we tracked the entry.
```
RonFed: since this is an internal function I'd either drop this comment, or split it into the method based on the relevant lines.

instrumentation/manager.go
```diff
@@ -38,15 +38,25 @@ var meter = otel.Meter(otelMeterName)
...
+// over its tracked instrumentations and retry any whose previous initialize/load failed (i.e.
+// inst == nil) AND whose OTel distribution name matches one of the supplied values. A non-nil
+// empty slice retries every failed instrumentation regardless of distribution. When
+// RetryDistros is non-nil the Instrument / ProcessDetailsByPid / ProcessGroup fields are
+// ignored.
 type Request[processGroup ProcessGroup, configGroup ConfigGroup, processDetails ProcessDetails[processGroup, configGroup]] struct {
 	Instrument          bool
 	ProcessDetailsByPid map[int]processDetails
 	ProcessGroup        processGroup
+	RetryDistros        []string
```
RonFed, line 59: kind of a nit:
assuming we'd want other retry filters in the futures (or no filters - just retry everything) - what do you think about passing an interface
```Go
type RetryFilter[ProcessDetails[processGroup, configGroup]] interface {
    ShouldRetry(ProcessDetails[processGroup, configGroup]) bool
}
```

not sure if it would make the code nicer or not, but it would probably remove handling here of the slice/map allocation

instrumentation/manager.go
```diff
@@ -309,6 +327,67 @@ func (m *manager[ProcessGroup, ConfigGroup, ProcessDetails]) runEventLoop(ctx co
...
+			m.handleInstrumentError(err)
+			continue
+		}
+		retriedPIDs = append(retriedPIDs, e.pid)
+	}
+	if len(retriedPIDs) > 0 {
+		// Re-arm the detector for the successfully retried processes so we still get their exit
+		// events for cleanup. TrackProcesses is idempotent for already-tracked PIDs.
+		m.detector.TrackProcesses(retriedPIDs)
+	}
```
RonFed: nit:
this part is duplicated from the main event loop handling for instrumentation requests - the main different is that here we already have the process details cached in the map, and there the caller provides them in the request.
Do you think we can combine these 2 flows?

instrumentation/manager.go
damemi: I really like that, but I think it's more complexity than we need right now. When we go to add more filters or get more specific with the instrumentation errors we should definitely use that approach. but right now it's a lot of extra code just for triggering go offset retries

instrumentation/manager.go
damemi: updated broke this out into a common helper that both sites can use

odiglet/odiglet.go
damemi: the "here" is a little misleading, we don't actually close it anywhere

I was confused by this but it seems like a protection against a "channel closed" panic for multiple writers on different threads, which is a higher risk now that we're also adding the enterprise writer routine. It also doesn't look like anything actually depends on this channel to close for shutdown.

I'm actually curious why are we closing configupdates and instrumentationrequests here currently?

RonFed: that's a good point, since the ebpf manager exit flow is based on context being canceled - I agree we don't need to close these.

instrumentation/manager.go
```diff
@@ -309,6 +309,76 @@ func (m *manager[ProcessGroup, ConfigGroup, ProcessDetails]) runEventLoop(ctx co
...
+// When logExplicitRequests is true, each attempt is logged as an explicit instrumentation request.
+func (m *manager[ProcessGroup, ConfigGroup, ProcessDetails]) instrumentFromDetails(ctx context.Context, byPid map[int]ProcessDetails, logExplicitRequests bool) {
+	var tracked []int
+	for pid, pd := range byPid {
+		// Handle duplicate requests gracefully; this can happen when external systems such as
+		// k8s controllers re-send instrumentation for an already-live process.
+		if m.isInstrumented(pid) {
+			continue
+		}
+		if logExplicitRequests {
```
RonFed: I'd remove this if (and the boolean param) and make the log line something that can fit both flows

RonFed (approved):

odiglet/odiglet.go
```diff
@@ -128,76 +233,11 @@ func (o *Odiglet) Run(ctx context.Context) {
...
-
-	// Start pprof server
-	g.Go(func() error {
-		err := common.StartPprofServer(groupCtx, commonlogger.ToLogr(), int(k8sconsts.DefaultPprofEndpointPort))
-		if err != nil {
-			logger.Error("Failed to start pprof server", "err", err)
-		} else {
-			logger.Info("Pprof server exited")
-		}
-		// if we fail to start the pprof server, don't return an error as it is not critical
```
RonFed, line 142: maybe worth to keep this comment in above `PropagateErr: false,` in the pprof runnable above

odiglet/odiglet.go
```diff
@@ -128,76 +233,11 @@ func (o *Odiglet) Run(ctx context.Context) {
 
 	defer o.criClient.Close()
 
-	// Channel to signal when eBPF manager has exited
```
RonFed, line 131: maybe worth to keep this comment

odiglet/odiglet.go
```diff
@@ -128,76 +233,11 @@ func (o *Odiglet) Run(ctx context.Context) {
...
-		if err != nil {
-			logger.Error("Failed to start opamp server", "err", err)
-		}
-		logger.Info("OpAmp server exited")
-		return err
-	})
-
-	// start kube manager
-	g.Go(func() error {
-		// Create a context that will be cancelled when eBPF manager exits during shutdown
```
RonFed, line 169: worth to keep this comment in the refactored code

RonFed (approved):

PR #4861, opened 2026-05-14
CORE-981 | feat(profiling): VM profile enrichment via unixfd and odigosvmprofileattrsprocessor in collector

shivanshuraj1333:

  Collector-side support for the VM agent's continuous-profiling pipeline.
  
  Introduces a new `odigosvmprofileattrsprocessor` that enriches incoming
  ResourceProfiles with `service.name` (plus VM source kind/name) by looking
  up the sample PID in a cache populated over `common/unixfd`.
  
  The vm-agent streams PID→attrs updates over the existing unixfd channel; the processor
  joins them onto eBPF profile samples in-flight.

Related PR on the VM-agent: https://github.com/odigos-io/vm-agent/pull/639

Demo:
vm-profiling-demo

 
```release-note
Profling support on VM
```


collector/processors/odigoscgroupservicenameprocessor/processor.go
```diff
@@ -0,0 +1,111 @@
...
+
+	p.cache.Store(pid, name)
+	return name, true
+}
+
+// cgroupServiceName reads /proc/<pid>/cgroup and extracts the systemd service unit name.
+// On cgroup v2 the file contains a single line: "0::/system.slice/my-service.service"
+// On cgroup v1 it contains multiple lines: "N:controller:/path"
+// We scan all lines and return the last path segment ending in ".service".
+func cgroupServiceName(pid int64) (string, error) {
```
shivanshuraj1333: currently there's no way to enrich the profiles collected natively from a VM with service names etc

the odigoscgroupservicenameprocessor uses the /proc filesystem to derive the systemd service name and enrich the profiles samples

this can potentially be part of upstream either in the collector as a processor or in the eBPF profiler itself, so that the collected profiles are atleast enriched with systemd service name

TheErikNaz (approved):

collector/processors/odigosvmprofileattrsprocessor/processor.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+		key := strings.TrimSpace(kv[0])
+		val := strings.TrimSpace(kv[1])
+		if key == "" || val == "" {
+			continue
+		}
+		if key == attrServiceName {
+			if existing, ok := resourceAttrs.Get(attrServiceName); ok {
+				name := existing.AsString()
+				if name != "" && !strings.HasPrefix(name, unknownServicePrefix) {
```
daemon1024: why is this a a prefix not an exact match?

collector/processors/odigosvmprofileattrsprocessor/samples_attrs.go
```diff
@@ -0,0 +1,76 @@
...
+
+// propagateServiceNameToSamples sets service.name on every profile sample attribute.
+// Grafana Pyroscope derives the query label service_name from sample attributes
+// (see pkg/ingester/otlp/convert.go serviceNameFromSample), not resource attributes.
+func propagateServiceNameToSamples(dict pprofile.ProfilesDictionary, rp pprofile.ResourceProfiles, serviceName string) {
+	if serviceName == "" {
+		return
+	}
+
+	attrIdx := ensureDictionaryServiceNameAttr(dict, serviceName)
```
daemon1024: should we cache this? it's a nested loop, need to walk through it everytime? 
Or the cache won't benefit here?

collector/processors/odigosvmprofileattrsprocessor/processor_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/odigos-io/odigos/common/unixfd"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/pprofile"
+	"go.uber.org/zap"
+)
+
+func TestProfileAttrCache_RegisterUnregister(t *testing.T) {
+	cache := newProfileAttrCache()
+	cache.applyEvent(unixfd.EncodeLogsAttrRegister(42, "service.name:payment-api,odigos.vm.source.kind:systemd"))
```
daemon1024: We should rename this to `EncodeAttrRegister` instead of Logs if it's being used everywhere as is. WDYT?

collector/processors/odigosvmprofileattrsprocessor/processor.go
```diff
@@ -0,0 +1,145 @@
...
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.opentelemetry.io/collector/pdata/pprofile"
+	"go.uber.org/zap"
+)
+
+const (
+	attrProcessPID       = "process.pid"
+	attrServiceName      = "service.name"
```
daemon1024: Can we use semconv for this?
https://opentelemetry.io/docs/specs/semconv/registry/attributes/process/#process-pid

Similarly for service name as well

daemon1024 (commented):
Some comments! PTAL

collector/processors/odigosvmprofileattrsprocessor/processor.go
shivanshuraj1333: receiver emits unknown_service:<exe> (suffix varies)

shivanshuraj1333: thinking again about it, we can just rely on whatever is coming from the VM agent (for the service name)

collector/processors/odigosvmprofileattrsprocessor/samples_attrs.go
shivanshuraj1333: the number of samples should not be much here, in a hot host maybe max around 100, so the extra loop should not cause much problem, but yes can store it in a map and do the look-up from it, updated the changes

PR #4867, opened 2026-05-15
CORE-1008 | feat(profiling): Dynamic filtering to allow only forwarding profiles for Odigos sources

shivanshuraj1333:

Introducing a new `odigosprofilesprocessor` on the node collector 'sprofiles pipeline (after filter + k8s_attributes). 

It will now forward profiles only for workloads with at least one container row in the `odigos_config_k8s` cache.

There's also metrics for dropped profiles captured via `otelcol_odigos.profiles.processor.dropped.resource_profiles`

```release-note
None
```


autoscaler/controllers/common/profiling_pipeline_names.go
```diff
@@ -9,6 +9,8 @@ const (
 	// Node collector profiles domain — receive on host, forward to cluster gateway.
 	ProfilingNodeFilterProcessor        = "filter/profiles-node"
 	ProfilingNodeK8sAttributesProcessor = "k8s_attributes/profiles-node"
+	// ProfilingNodeProfilesAllowlistProcessor keeps only profiles for workloads present in odigos_config_k8s (InstrumentationConfig).
+	ProfilingNodeProfilesAllowlistProcessor = "odigos_profiles_allowlist"
```
tamirdavid1: consider the name as discussed

autoscaler/controllers/nodecollector/collectorconfig/odigos_config_extension.go
```diff
@@ -18,6 +19,14 @@ func NodeHasURLTemplateProcessor(processors []*odigosv1.Processor) bool {
 	return false
 }
 
+// NodeNeedsOdigosConfigK8sExtension reports whether the node collector should merge the odigos_config_k8s extension domain.
+func NodeNeedsOdigosConfigK8sExtension(processors []*odigosv1.Processor, profiling *odigoscommon.ProfilingConfiguration) bool {
+	if NodeHasURLTemplateProcessor(processors) {
+		return true
+	}
+	return odigoscommon.ProfilingPipelineActive(profiling)
+}
```
tamirdavid1:
```suggestion
// NodeNeedsOdigosConfigK8sExtension reports whether the node collector should merge the odigos_config_k8s extension domain.
func NodeNeedsOdigosConfigK8sExtension(processors []*odigosv1.Processor, profiling *odigoscommon.ProfilingConfiguration) bool {
    // odigos odigos_config_k8s added in two senarios
    // 1. Url templatization 
	if NodeHasURLTemplateProcessor(processors) {
		return true
	}
	// 2. Profiling enabled
	if odigoscommon.ProfilingPipelineActive(profiling) {
        return true	
	}
	return false
}
```

collector/extension/odigosconfigk8sextension/resourceattrs.go
```diff
@@ -90,3 +91,14 @@ func workloadContainerKeyFromResourceAttributes(attrs pcommon.Map) (string, erro
 	}
 	return WorkloadKeyString(ns, kind, name) + "/", nil
 }
+
+// WorkloadKeyStringFromResourceAttributes returns namespace/Kind/workloadName (no trailing slash)
+// from resource attributes after k8s enrichment. Used to align profile resources with
+// InstrumentationConfig-derived workload keys from odigos_config_k8s.
+func WorkloadKeyStringFromResourceAttributes(attrs pcommon.Map) (string, error) {
```
tamirdavid1: consider using directly WorkloadKeyStringFromResourceAttributes and remove trailing slash

collector/odigosotelcol/go.mod
```diff
@@ -497,7 +498,7 @@ require (
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/mwitkow/go-conntrack v0.0.0-20190716064945-2f068394615f // indirect
 	github.com/odigos-io/go-rtml v0.0.1 // indirect
-	github.com/odigos-io/odigos/common v0.0.0-00010101000000-000000000000 // indirect
+	github.com/odigos-io/odigos/common v0.148.0 // indirect
```
tamirdavid1: what is this? seems weird

collector/processors/odigosprofilesallowlistprocessor/processor.go
```diff
@@ -0,0 +1,96 @@
+package odigosprofilesallowlistprocessor
```
tamirdavid1: I think this is reduandant .. all this file i will comment on processProfiles

collector/processors/odigosprofilesallowlistprocessor/processor.go
```diff
@@ -0,0 +1,96 @@
...
+	p.mirror.onDelete(key)
+}
+
+func (p *profilesAllowlistProcessor) processProfiles(_ context.Context, pd pprofile.Profiles) (pprofile.Profiles, error) {
+	if p.provider == nil {
+		pd.ResourceProfiles().RemoveIf(func(pprofile.ResourceProfiles) bool { return true })
+		return pd, nil
+	}
+	rps := pd.ResourceProfiles()
+	rps.RemoveIf(func(rp pprofile.ResourceProfiles) bool {
```
tamirdavid1: I think it can be much much simplified

```
rps.RemoveIf(func(rp pprofile.ResourceProfiles) bool {
    _, found := p.provider.GetFromResource(rp.Resource())
    if !found {
        p.logger.Debug("odigos_profiles_allowlist: drop resource (workload not in InstrumentationConfig cache)")
        return true
    }
    return false
})
```
**This would eliminate:**
1. The entire mirror.go file
2. The callback registration/unregistration in Start/Shutdown
3. The OnSet/OnDeleteKey implementations
4. The dependency on WorkloadKeyStringFromResourceAttributes (and that function itself, which I comment earlier)

tamirdavid1: + consider adding metrics if you think it could help you e.g dropped profiles

shivanshuraj1333: I was doing the old way (using onset/ondelete) to have a cache in the processor but in this case since we can just directly use the values from directly from the extention's cache (unlike url template processor where we were also storing the parsed rules).

Updated the changes accordingly

collector/extension/odigosconfigk8sextension/cache.go
```diff
@@ -89,6 +89,22 @@ func (c *cache) Get(key string) (*commonapi.ContainerCollectorConfig, bool) {
...
+// hasContainersForWorkloadPrefix returns true when the workload key prefix (e.g. "ns/Kind/name/")
+// has at least one full container cache key. Used for profile filtering where resource attributes
+// may not include k8s.container.name.
+//
+// The len(containerKeys) > 0 check is load-bearing, not defensive. The index entry's lifetime is
+// the union of its containerKeys and dataStreams: SetDataStreams creates an entry with an empty
+// containerKeys map, and Delete only removes the entry when both sets are empty. So entry != nil
+// alone does not imply "has at least one InstrumentationConfig container entry" — a workload that
+// only has data-stream labels would falsely return true without the length check.
+func (c *cache) hasContainersForWorkloadPrefix(workloadKeyPrefix string) bool {
```
tamirdavid1, line 101: instead of doing that u can use already exists `getContainerKeysForWorkload `

like this 
if len(c.getContainerKeysForWorkload(workloadIndexKey)) > 0 {
    // has containers
}

collector/extension/odigosconfigk8sextension/odigosconfig.go
```diff
@@ -57,6 +58,14 @@ func (o *OdigosWorkloadConfig) GetFromResource(res pcommon.Resource) (*commonapi
 	return o.cache.Get(key)
 }
 
+func (o *OdigosWorkloadConfig) HasWorkloadInstrumentationConfig(res pcommon.Resource) (string, bool) {
```
tamirdavid1: it is not technically checking for instrumentationConfig so i suggest changing the name for:
`HasCachedWorkloadContainerConfig`

also why it return two values? the name of this func suggest it is a return bool one (semantically)
so better to do that and trimSuffix in the caller with comment

collector/processors/odigosprofilesprocessor/processor.go
```diff
@@ -0,0 +1,111 @@
...
+	meter := tel.MeterProvider.Meter("github.com/odigos-io/odigos/collector/processors/odigosprofilesprocessor")
+	unkeyable, _ := meter.Int64Counter(
+		"odigos.profiles.processor.dropped.resource_profiles.unkeyable",
+		metric.WithDescription("ResourceProfiles dropped because namespace/kind/name could not be derived from resource attributes"),
+	)
+	unmanaged, _ := meter.Int64Counter(
+		"odigos.profiles.processor.dropped.resource_profiles.unmanaged_workload",
+		metric.WithDescription("ResourceProfiles dropped because the workload has no InstrumentationConfig entries in odigos_config_k8s"),
+	)
+	noExt, _ := meter.Int64Counter(
```
tamirdavid1: I think it is not valuable as the validate will fail in this cases. please remove

collector/processors/odigosprofilesprocessor/processor.go
```diff
@@ -0,0 +1,111 @@
...
+	return nil
+}
+
+func (p *odigosProfilesProcessor) Shutdown(context.Context) error {
+	p.provider = nil
+	return nil
+}
+
+func (p *odigosProfilesProcessor) processProfiles(ctx context.Context, pd pprofile.Profiles) (pprofile.Profiles, error) {
+	if p.provider == nil {
```
tamirdavid1: can you add comment here? not clear in which cases it happens

collector/extension/odigosconfigk8sextension/cache.go
shivanshuraj1333: We can but i wasn't sure the implication of it at scale,

So, assume a workload 
```go
func (c *cache) getContainerKeysForWorkload(workloadKey string) []string {
	c.mu.RLock()
	entry := c.workloadKeysIndex[workloadKey]
	if entry == nil || len(entry.containerKeys) == 0 {
		c.mu.RUnlock()
		return nil
	}
	out := make([]string, 0, len(entry.containerKeys))
	for k := range entry.containerKeys {
		out = append(out, k)
	}
	c.mu.RUnlock()
	return out
}
```

here for out we are allocating memory and then copying all the containerKeys into it, in our case we don't need it, we just need to check that there's at least one entry.

collector/extension/odigosconfigk8sextension/odigosconfig.go
shivanshuraj1333: were using the workloadKey for debug logging, but can get rid of it
updated it, thanks!

collector/processors/odigosprofilesprocessor/processor.go
shivanshuraj1333: removing it, we won't hit the code path where provider would be nil

tamirdavid1:
@shivanshuraj1333 can you please update this PR description ? the naming and mirroring is not true anymore right?

collector/processors/odigosprofilesprocessor/metadata.yaml
```diff
@@ -0,0 +1,12 @@
+type: odigosprofilesprocessor
+status:
+  class: processor
+  stability:
+    alpha: [profiles]
+  distributions: [odigos]
+  codeowners:
+    active: [odigos]
+
```
tamirdavid1, line 9: i dont see the metrics here, the same as we are doing here:
https://github.com/odigos-io/odigos/blob/main/collector/receivers/odigosebpfreceiver/metadata.yaml

collector/processors/odigosprofilesprocessor/processor.go
```diff
@@ -0,0 +1,72 @@
...
+type odigosProfilesProcessor struct {
+	logger *zap.Logger
+	cfg    *Config
+
+	provider        collector.OdigosConfigExtension
+	droppedProfiles metric.Int64Counter
+}
+
+func newOdigosProfilesProcessor(logger *zap.Logger, tel component.TelemetrySettings, cfg *Config) *odigosProfilesProcessor {
+	meter := tel.MeterProvider.Meter("github.com/odigos-io/odigos/collector/processors/odigosprofilesprocessor")
```
tamirdavid1: if u adding the metric as i mention above you can use it simple by
https://github.com/odigos-io/odigos/blob/main/collector/receivers/odigosebpfreceiver/logs.go#L123

collector/processors/odigosprofilesprocessor/metadata.yaml
shivanshuraj1333: done

collector/processors/odigosprofilesprocessor/processor.go
shivanshuraj1333: done, using TelemetryBuilder for capturing metrics

tamirdavid1 (approved):

PR #4871, opened 2026-05-15
add url templatization config for the default templatization

blumamir:

Fixes: CORE-959

Internet-exposed services are often hit by bots and scanners with random paths (/wp-admin, /.env, etc.). Many of these return 404, and default URL templatization heuristics would treat each path as a distinct route, inflating cardinality on http.route / url.template and span names.

This PR adds new config section in url templatization action which allows to customize the behavior of the default templatization. e.g. if no rule match, now one can:

- specify to which sources the default templatization is applied. I guess most users would want to enable it for the entire cluster.
- disable specific sources from default templatization - if it creates high cardinality while we still want to use the custom rules if set.
- opt-in to a feature that allows to avoid templatization if the request ended with an error, indicating that it might be garbage request where the path is irrelevant. this mechanism is bypassed if a custom rule is supplied on this endpoint (in this case, all status codes will be templatized)
- configure the default templatization on http error based on status code (either all non-success 2xx request - recomended, or by list of status codes).

Examples (for status codes list with one item [404])

- `/users/1234` 200 -> templated (not http error code).
- `/etc/passwd` 404 -> not templated (404 response code).
- `/etc/passwd` 401 -> templated (false positive, only 404 is tested).
- `/users/1234` 404 -> not templated (false positive for legitimate case)
- `/users/1234` 404 with rule for `/users/{id}` -> will be templated since it match a custom rule.

Example real world paths:

```
GET /Audio/{id}/hls/............\\Windows\\win.ini/stream.mp3/

GET /AccessAnywhere/%2e%2e%5c...%5cwindows%5cwin.ini

GET /<style onload=alertdocument.domain;>../{id}/

GET /@fs/etc/passwd

GET //yOxKvlF7"><script>alert(document.domain)</script>/..CFIDE/administrator/index.cfm
```

```release-note
Add configuration to url-templatization action for control on the default templatization behavior when the request ended with error. can be used to protect publicly accessible services from malicious bots probing irrelevant url paths and overwhelming cardinality and usability.
```


common/api/actions/url_templatization.go
```diff
@@ -1,7 +1,47 @@
...
+type DefaultTemplatizationConfig struct {
+
+	// if set to true, the default templatization will be disabled for the services in the scope.
+	// in case of conflict (if other action set this to false), the default templatization will be disabled.
+	Disabled bool `json:"enabled,omitempty"`
+
+	// config if default templatization should be skipped on error.
+	// use it when sources scope describes a service that is publicly accessible to the internet
+	// to filter garbage requests that can contaminate the url-templatization process.
+	SkipOnError *DefaultTemplatizationSkipOnErrorConfig `json:"skipOnError,omitempty"`
```
RonFed: can we inline this struct into here to avoid extra nesting?

collector/processors/odigosurltemplateprocessor/processor_test.go
```diff
@@ -644,9 +644,9 @@ func TestDefaultDateTemplatization(t *testing.T) {
 				"http.request.method": "GET",
 				"url.path":            "04-12-2025T14:15:16",
 			},
-			expectedSpanName:  "GET /04-12-2025T14:15:16",
+			expectedSpanName:  "GET 04-12-2025T14:15:16",
```
RonFed, line 647: is this expected?

api/odigos/v1alpha1/actions/urltemplatization.go
```diff
@@ -39,6 +40,18 @@ type UrlTemplatizationRulesGroup struct {
 	TemplatizationRules []URLTemplatizationRule `json:"templatizationRules,omitempty"`
 }
 
+// URLTemplatizationDefaultTemplatizationGroup is a group of services for which default templatization will be applied.
+// +kubebuilder:object:generate=true
+// +kubebuilder:deepcopy-gen=true
+type URLTemplatizationDefaultTemplatizationGroup struct {
+	// the scope of services for which this templatization config will be applied.
```
RonFed, line 47: if no scope is provided, this default policy will apply to the whole cluster? might worth to mention this in the comment

common/api/actions/url_templatization.go
RonFed: if in the future we want to allow also skip on redirect - maybe we should consider a different name than `SkipOnError`?

common/api/actions/url_templatization.go
```diff
@@ -1,7 +1,47 @@
...
+// either skipForNonSuccessCodes or statusCodes must be set for it to take effect.
+// if both are set, skipForNonSuccessCodes takes precedence.
+//
+// +kubebuilder:object:generate=true
+// +kubebuilder:deepcopy-gen=true
+type DefaultTemplatizationSkipOnErrorConfig struct {
+
+	// If set to true, default templatization will be skipped for any non-success HTTP status code (i.e., any code not in the 2xx range).
+	// When this is true, the StatusCodes list is ignored.
+	SkipForNonSuccessCodes bool `json:"skipForNonSuccessCodes,omitempty"`
```
RonFed, line 17: the otel definition for http errors is a bit different: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#status
do you think we should clarify this? take the span status into account?

collector/processors/odigosurltemplateprocessor/processor_test.go
blumamir: yes, it was a mistake in the test

RonFed (approved):

common/api/actions/url_templatization.go
blumamir: changed to "SkipHttpStatusCodes"

common/api/actions/url_templatization.go
blumamir: changed to SkipPolicy

blumamir: I want to have the skip policy separated from other configs, like which regex to opt-in etc

PR #4904, opened 2026-05-20
Fix agent directory traversal permissions

### LINKED ISSUE #4769: odiglet creates /var/odigos/java/ with incorrect permissions (missing execute bit), causing non-root Java pods to crash
**Version:** Odigos v1.24.0, v1.24.2
**Mount method:** k8s-host-path (default) and k8s-virtual-device

**Description:**
The `odiglet` DaemonSet extracts instrumentation files to the host node under `/var/odigos/java/`, but creates the directory with mode `644` (`drw-r--r--`) instead of `755` (`drwxr-xr-x`).

Because the execute bit (`x`) is missing, any container running as a non-root user is blocked from traversing the directory. This causes the JVM to fail at startup with the following error:
`Error opening zip file or JAR manifest missing : /var/odigos/java/javaagent.jar`

**Steps to Reproduce:**
1. Install Odigos v1.24.2.
2. Instrument a Java workload that runs as a non-root user (e.g., `USER 1000` in Dockerfile).
3. The Pod enters `CrashLoopBackOff`.

**Verification:**
Running a debug pod on the node confirms the missing execute bit:
```bash
kubectl debug node/<worker> -it --image=busybox -- sh -c "ls -la /host/var/odigos/java/"
# Output: drw-r--r--  2 root  root  4096 Apr 24 16:02 .

5p2O5pe25ouT:
## Summary
- ensure copied instrumentation directories keep traversal execute bits when copied to the host
- use 0755 for agent directory creation instead of os.ModePerm
- add a regression test for directories such as /var/odigos/java being left as 0644

## Why
Fixes #4769. A directory mode like 0644 lets the file exist but prevents non-root instrumented containers from traversing /var/odigos/java to read javaagent.jar.

## Verification
- go test ./pkg/instrumentation/fs -run TestEnsureDirectoryTraversalPermissions -count=1
- go test ./pkg/instrumentation/fs -count=1
- go test ./pkg/instrumentation/...

## Note
I also tried go test ./... in the odiglet module on macOS/arm64, but it is blocked by Linux/eBPF-only packages and missing generated bpf objects, for example bpf_arm64_bpfel.o and syscall.Mount symbols.

### DIFF
--- a/odiglet/pkg/instrumentation/fs/agents.go
+++ b/odiglet/pkg/instrumentation/fs/agents.go
@@ -103,6 +103,10 @@ func CopyAgentsDirectoryToHost() error {
 		logger.Error("Error creating dotnet deprecated directories", "err", err)
 		return err
 	}
+	if err := ensureDirectoryTraversalPermissions(k8sconsts.OdigosAgentsDirectory); err != nil {
+		logger.Error("Error ensuring instrumentation directory permissions", "err", err)
+		return err
+	}
 
 	logger.Info("Odigos agents directory copied to host", "elapsed", time.Since(startTime))
 
--- a/odiglet/pkg/instrumentation/fs/copy.go
+++ b/odiglet/pkg/instrumentation/fs/copy.go
@@ -19,6 +19,8 @@ const (
 
 	// 32 KB buffer for I/O operations
 	bufferSize = 32 * 1024
+
+	directoryTraversalMode os.FileMode = 0755
 )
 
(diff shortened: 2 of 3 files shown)


aviadhahami:
Hi @5p2O5pe25ouT ! Thanks for opening this pr
The issue is resolved starting version 1.32.x. (https://github.com/odigos-io/odigos/pull/5343)

Thanks again!

PR #4907, opened 2026-05-20
RUN-779 | perf: cgroup based pid resolution in odiglet

RonFed:

Adding fast path for PIDs resolution - to be based on cgroups.
The current approach is iterating `/proc` and looking for PIDs where the `mountinfo` file contains a mount of the request (pod,container) - this has a complexity of O(num of procs on the machine) and involve a lot of system calls.
The cgroup based approach is using the cgroup fs (mounted in the odiglet) - and locate the (pod,container) entry in the cgroup tree that the kubelet builds. This approach is using a one-time discovery of the cgroup param, the kubelet root etc' - when requested to resolve the PIDs for a (pod,container) we locate the cgroup directory based on the discovery params, the pod uid, QoS and container name. Reading `cgroup.procs` under the directory resolves the PIDs running for that cgroup.

* currently only supporting cgroupv2 (both syutemd and cgroupfs modes) - v1 adds a mounting complexity that currently not included.
* If the cgroup discovery fails, or we discover cgroups v1, fallback to the current `/proc` approach.

Before: cpu profile measure on pod churm in a cluster (rollouts for 80 replicas deployment)
image
after: same scenario with the cgroup approach:
image

```release-note
perf: reduce odiglet CPU usage when resolving (pod,container) to PIDs
```


tamirdavid1:
Suggestion: add release note about the improvement in odiglet memory due to it.

helm/odigos/templates/odiglet/daemonset.yaml
```diff
@@ -251,6 +255,9 @@ spec:
             # while the parent /var/odigos mount stays read-only for agent binaries.
             - name: odigos-opamp-exchange
               mountPath: /var/odigos/exchange
+            - name: host-cgroup
```
tamirdavid1, line 258: here no need to check
{{- if not .Values.odiglet.noHostPathMounts }} ?

helm/odigos/templates/odiglet/daemonset.yaml
```diff
@@ -220,6 +220,10 @@ spec:
             - name: ODIGOS_PROC_DIR
               value: /hostproc
           {{- end}}
+          {{- if not .Values.odiglet.noHostPathMounts }}
+            - name: ODIGOS_CGROUP_ROOT
```
tamirdavid1, line 224: it is static value, we can refer it in the code no?

RonFed: technically yes, I did it this way - so:
1. if we want to make this configurable we can add a helm value for it
2. if we change it we only need to change the mount here and not also in the code.

helm/odigos/templates/odiglet/daemonset.yaml
RonFed: this block is under `{{- if not .Values.odiglet.noHostPathMounts }}` - i.e "optional" mounts - this is optional since we have the `/proc` fallback.

helm/odigos/templates/odiglet/daemonset.yaml
blumamir: @RonFed do you think we will want to change it? what will cause it to be different?

odiglet/pkg/process/cgroup_linux.go
```diff
@@ -0,0 +1,266 @@
...
+const (
+	defaultCgroupRoot = "/sys/fs/cgroup"
+	cgroupRootEnvVar  = "ODIGOS_CGROUP_ROOT"
+	selfCgroupPath    = "/proc/self/cgroup"
+)
+
+// sysCgroupRoot returns the root path where the host cgroupfs is visible
+// inside this process. Reads ODIGOS_CGROUP_ROOT, falling back to
+// /sys/fs/cgroup for non-Kubernetes deployments where the binary runs
+// directly on a node.
```
blumamir: This code is in odiglet. What do you mean by "non-Kubernetes deployments" in this context?

odiglet/pkg/process/cgroup_linux.go
```diff
@@ -0,0 +1,266 @@
...
+	"sync"
+
+	"github.com/odigos-io/odigos/common/logger"
+
+	"github.com/opencontainers/cgroups"
+	"github.com/opencontainers/cgroups/systemd"
+)
+
+const (
+	defaultCgroupRoot = "/sys/fs/cgroup"
```
blumamir, line 21: Since the root is passed as env var, I think it is over complexity to have a default here, as it makes it harder to follow the logic

odiglet/pkg/process/cgroup_linux.go
```diff
@@ -0,0 +1,266 @@
...
+
+// sysCgroupRoot returns the root path where the host cgroupfs is visible
+// inside this process. Reads ODIGOS_CGROUP_ROOT, falling back to
+// /sys/fs/cgroup for non-Kubernetes deployments where the binary runs
+// directly on a node.
+func sysCgroupRoot() string {
+	if v := os.Getenv(cgroupRootEnvVar); v != "" {
+		return v
+	}
+	return defaultCgroupRoot
```
blumamir, line 35: helm semantics are that if the env is missing, this cgroup path should be skipped. wondering if we should return "found" here and skip the rest in this case?

odiglet/pkg/process/cgroup_linux.go
```diff
@@ -0,0 +1,266 @@
...
+		log := logger.LoggerCompat().With("subsystem", "cgroup")
+
+		// currently we only support the fast cgroup-based PID resolution on cgroup v2.
+		// Supporting v1 adds additional complexity for mounting the cgroup fs.
+		// cgroup v1 uses many per-controller mounts that need HostToContainer propagation,
+		// which only works if the host marked /sys/fs/cgroup as shared — not always true, and forcing it requires a privileged init container that mutates host state.
+		if !isCgroupV2 {
+			log.Info("detected cgroup v1, the fast path for PID resolution will be disabled")
+			return
+		}
```
blumamir, line 66: looks like v2 has unified and hybrid mode, and we only support unified mode.
what is the other mode? should we mention it in the comment?

odiglet/pkg/process/cgroup_linux.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+var (
+	// hostCgroupLayout is detected once at DiscoverCgroupLayout.
+	hostCgroupLayout cgroupLayout
+	isCgroupV2       = cgroups.IsCgroup2UnifiedMode()
+	initOnce         sync.Once
+)
+
+func DiscoverCgroupLayout() {
```
blumamir, line 55: nit: consider writing it in a way that avoid side effects (mutating global vars), so it's easier to read and understand

odiglet/pkg/process/cgroup_linux.go
```diff
@@ -0,0 +1,266 @@
...
+	initOnce.Do(func() {
```
blumamir, line 56: is it expected to happen multiple times? or only for safety?

odiglet/pkg/process/cgroup_linux.go
```diff
@@ -0,0 +1,266 @@
...
+		if !isCgroupV2 {
+			log.Info("detected cgroup v1, the fast path for PID resolution will be disabled")
+			return
+		}
+
+		hostCgroupLayout.root = sysCgroupRoot()
+
+		// based on our self cgroup paths,
+		// try to detect if we're running in a systemd-based cgroup hierarchy, and locate the kubepods root.
+		if parsed, err := cgroups.ParseCgroupFile(selfCgroupPath); err == nil {
```
blumamir, line 72: should we log something if err is not nil?

odiglet/pkg/process/process_linux.go
```diff
@@ -32,21 +74,35 @@ func isInPodContainersBatchPredicate(podContainers []PodContainerUID) func(int)
...
-// GroupByPodContainer groups all the current active processes by (podUID, containerName) using the provided list of PodContainerUIDs to filter relevant processes.
-// Processes that do not belong to any of the provided PodContainerUIDs are ignored.
-func GroupByPodContainer(pcs []PodContainerUID) (map[PodContainerUID]map[int]struct{}, error) {
-	groups, err := process.Group(isInPodContainersBatchPredicate(pcs))
-	if err != nil {
-		return nil, fmt.Errorf("failed to group processes by (pod, container) :%w", err)
-	}
+type PodContainer struct {
+	PodContainerKey
+	QOSClass    string
```
blumamir: Why use string and not the enum?

```
// PodQOSClass defines the supported qos classes of Pods.
// +enum
type PodQOSClass string

const (
	// PodQOSGuaranteed is the Guaranteed qos class.
	PodQOSGuaranteed PodQOSClass = "Guaranteed"
	// PodQOSBurstable is the Burstable qos class.
	PodQOSBurstable PodQOSClass = "Burstable"
	// PodQOSBestEffort is the BestEffort qos class.
	PodQOSBestEffort PodQOSClass = "BestEffort"
)

```

Can remove some complexity when used

odiglet/pkg/process/process_linux.go
```diff
@@ -32,21 +74,35 @@ func isInPodContainersBatchPredicate(podContainers []PodContainerUID) func(int)
...
+// GroupByPodContainer groups all the current active processes by
+// (podUID, containerName) using the provided list of PodContainers to
+// filter relevant processes. Processes that do not belong to any of
+// the provided PodContainers are ignored.
+func GroupByPodContainer(pcs []PodContainer) (map[PodContainerKey]map[int]struct{}, error) {
+	// if we have the cgroup layout available,
+	// use it for grouping as it's more efficient.
+	if hostCgroupLayout.Valid {
+		groups, err := groupByCgroup(pcs)
+		// if we had and error or no groups, fallback to the legacy proc mountinfo parsing
```
blumamir, line 103: why no groups also need to fallback to the old mechanism?

odiglet/pkg/kube/instrumentation_ebpf/instrumentationconfig.go
blumamir: nit: comment need to be updated, as now it's not necessarily expensive

odiglet/pkg/process/process_linux.go
```diff
@@ -3,24 +3,66 @@ package process
...
-	expectedMountByPodContainer := make(map[PodContainerUID][]byte)
-	for _, pc := range podContainers {
-		expectedMount := fmt.Sprintf("%s/containers/%s/", pc.PodUID, pc.ContainerName)
-		expectedMountByPodContainer[pc] = []byte(expectedMount)
+func groupByCgroup(pcs []PodContainer) (map[PodContainerKey]map[int]struct{}, error) {
+	result := make(map[PodContainerKey]map[int]struct{}, len(pcs))
+	for _, pc := range pcs {
+		if pc.ContainerID == "" {
+			continue
+		}
```
blumamir, line 20: can it actually happen?

odiglet/pkg/process/cgroup_linux.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// containerCgroupDir returns the absolute on-host directory holding the
+// container's cgroup.procs file.
+func containerCgroupDir(l cgroupLayout, pc PodContainer) (string, error) {
+	i := strings.Index(pc.ContainerID, "://")
+	if i < 0 {
+		return "", fmt.Errorf("invalid container ref: %q", pc.ContainerID)
+	}
+	id := pc.ContainerID[i+3:]
```
blumamir: what is this `id`? an example here can help

odiglet/pkg/process/cgroup_linux.go
```diff
@@ -0,0 +1,266 @@
...
+			return filepath.Join(segs[:i]...), true
+		}
+	}
+	return "", false
+}
+
+// containerCgroupDir returns the absolute on-host directory holding the
+// container's cgroup.procs file.
+func containerCgroupDir(l cgroupLayout, pc PodContainer) (string, error) {
+	i := strings.Index(pc.ContainerID, "://")
```
blumamir, line 198: why container id must contain `"://"`? what is the format of it?

odiglet/pkg/process/cgroup_linux.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	id := pc.ContainerID[i+3:]
+	pod, err := podCgroupDir(l, pc.QOSClass, pc.PodUID)
+	if err != nil {
+		return "", err
+	}
+	if l.Systemd {
+		return filepath.Join(pod, containerSystemdScope(l.SystemdScopePrefix, id)), nil
+	}
+	return filepath.Join(pod, id), nil
```
blumamir: nit: both `podCgroupDir` and this function have different paths for Systemd or not. consider if can be simplified

odiglet/pkg/process/cgroup_linux.go
RonFed: for safety

odiglet/pkg/process/process_linux.go
RonFed: being extra safe, if we have a bug in the cgroup logic that cause us to not find processes - fallback to the current approach.

odiglet/pkg/process/cgroup_linux.go
RonFed: from k8s api docs:
```
	// ContainerID is the ID of the container in the format '<type>://<container_id>'.
	// Where type is a container runtime identifier, returned from Version call of CRI API
	// (for example "containerd").
```

odiglet/pkg/process/cgroup_linux.go
RonFed: But I don't want the callers of this (odiglet setup code) to need to know what this does - its a one time discover function. If we return something to the callers - it means they need to be aware of it and save it

odiglet/pkg/process/process_linux.go
RonFed: being defensive
if somebody decide to remove pods status from the odiglet cache this can be empty for example

odiglet/pkg/process/cgroup_linux.go
RonFed: added comment and renamed the variable

odiglet/pkg/process/cgroup_linux.go
RonFed: refactored the structure of this function

odiglet/pkg/process/cgroup_linux.go
RonFed: updated the comment

odiglet/pkg/process/cgroup_linux.go
```diff
@@ -0,0 +1,255 @@
...
+func DiscoverCgroupLayout() {
+	initOnce.Do(func() {
+		log := logger.LoggerCompat().With("subsystem", "cgroup")
+
+		// currently we only support the fast cgroup-based PID resolution on cgroup v2.
+		// Supporting v1 adds additional complexity for mounting the cgroup fs.
+		// cgroup v1 uses many per-controller mounts that need HostToContainer propagation,
+		// which only works if the host marked /sys/fs/cgroup as shared — not always true, and forcing it requires a privileged init container that mutates host state.
+		if !isCgroupV2 {
+			log.Info("detected cgroup v1, the fast path for PID resolution will be disabled")
```
blumamir: nit: not a v2 unified mode

odiglet/pkg/process/cgroup_linux.go
blumamir: consider documenting why we still need to try v2 even if env is missing

odiglet/pkg/process/cgroup_linux.go
```diff
@@ -0,0 +1,255 @@
...
+			return
+		}
+
+		hostCgroupLayout.root = sysCgroupRoot()
+
+		// based on our self cgroup paths,
+		// try to detect if we're running in a systemd-based cgroup hierarchy, and locate the kubepods root.
+		if parsed, err := cgroups.ParseCgroupFile(selfCgroupPath); err == nil {
```
blumamir, line 72: consider adding that it will show up only in privileged mode

odiglet/pkg/kube/instrumentation_ebpf/instrumentationconfig.go
```diff
@@ -185,19 +185,32 @@ func (i *InstrumentationConfigReconciler) sendInstrumentationRequest(ctx context
...
 	}
 
-	// build all the relevant (podUID, containerName) combinations for this node
-	pcs := make([]process.PodContainerUID, 0, len(selectedPods)*len(distroByContainer))
+	pcs := make([]process.PodContainer, 0, len(selectedPods)*len(distroByContainer))
 	for _, p := range selectedPods {
+		uid := workload.PodUID(&p)
+		statusByName := make(map[string]string, len(p.Status.ContainerStatuses))
+		for _, cs := range p.Status.ContainerStatuses {
+		statusByName[cs.Name] = cs.ContainerID
```
blumamir: lint

blumamir (approved):

PR #4916, opened 2026-05-21
CORE-1015 | Abstract transport layer opamp

tamirdavid1:

This PR adds a transport abstraction for OpAMP so agents can talk to odiglet over Unix as well as HTTP, without changing default behavior for existing distros.

HTTP OpAMP often hits cross-namespace network policy issues; a node-local Unix socket avoids that class of blocks. Odiglet now runs a Unix OpAMP listener alongside the existing HTTP server; instrumentor picks the transport from distro YAML (opAmpTransport, mount method, JVM version for java unix).

No runtime change yet for workloads until java is configured for unix—community python/nodejs still get HTTP as today. A follow-up will add the Java eBPF OpAMP client (POC); nodejs and python can move later.

For java on unix, the first phase may feel stale: connection comes up before feature config is fully driven over OpAMP. We’ll move features one by one (head sampling, span renamer, URL templating, span metrics) off env/configAsEnvVars and onto the client as that work lands.

```release-note
NONE
```


distros/distro/oteldistribution.go
```diff
@@ -45,9 +46,13 @@ type StaticEnvironmentVariable struct {
 }
 
 type EnvironmentVariables struct {
-	// if this distribution runs an opamp client, add the environment variables that configures the server endpoint (local node)
+	// OpAmpClientEnvironments enables the OpAMP client. When opAmpTransport is empty, defaults to http.
```
RonFed: the comment here is not super clear: do we use this to configure the http endpoint? and the uds location? it it either one depending on `OpAmpTransport`

distros/distro/oteldistribution.go
```diff
@@ -45,9 +46,13 @@ type StaticEnvironmentVariable struct {
...
 	OpAmpClientEnvironments bool `yaml:"opAmpClientEnvironments,omitempty"`
 
+	// opAmpTransport selects unix or http. When unix, injects ODIGOS_OPAMP_UNIX_SOCKET (not HTTP).
+	// When empty, distros with opAmpClientEnvironments default to http.
+	OpAmpTransport commonopamp.OpAmpTransport `yaml:"opAmpTransport,omitempty"`
```
RonFed: what do you think about:
1. naming this OpAmpTransportsSupported
2. having a slice `[]commonopamp.OpAmpTransport`

this means that an agent can support both and let the odiglet decide which one to use based on `ODIGOS_OPAMP_SERVER_HOST`

opampserver/pkg/transport/unix/server.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (s *Server) Kind() commonopamp.OpAmpTransport {
+	return commonopamp.OpAmpTransportUnix
+}
+
+func (s *Server) Start(ctx context.Context, processor opamptypes.Processor) error {
+	logger := commonlogger.LoggerCompat().With("subsystem", "opampserver", "transport", "unix")
+
+	if err := os.MkdirAll(k8sconsts.OdigosOpampExchangeDir, 0o777); err != nil {
```
RonFed: `0o777` seems a bit too much, no? 0666 should be enough I think

opampserver/pkg/transport/unix/server.go
```diff
@@ -0,0 +1,121 @@
...
+			continue
+		}
+		go s.handleConnection(ctx, conn, processor, logger)
+	}
+}
+
+func (s *Server) handleConnection(ctx context.Context, conn net.Conn, processor opamptypes.Processor, logger *commonlogger.OdigosLogger) {
+	defer func() { _ = conn.Close() }()
+
+	for {
```
RonFed: why do we have a loop here? seems like from all the flows we return without iterating again

opampserver/pkg/transport/unix/server.go
```diff
@@ -0,0 +1,121 @@
...
+
+		serverToAgent, status := processor.Process(ctx, &agentToServer, commonopamp.OpAmpTransportUnix)
+		if status == opamptypes.ProcessBadRequest {
+			return
+		}
+		if status == opamptypes.ProcessError || serverToAgent == nil {
+			return
+		}
+
+		respBytes, err := proto.Marshal(serverToAgent)
```
RonFed: this part is the same across the transports - do you think we can extract it to a common handler - which will get the transports as an argument - but the proto marshiling and unmarshling is the same

opampserver/pkg/transport/unix/server.go
RonFed, line 1: I think we can use the exact same http logic just with a different `Server` setup
something like
```Go
    ln, err := net.Listen("unix", socketPath)
    if err != nil {
        panic(err)
    }
    defer ln.Close()

    mux := http.NewServeMux()
    mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
         ...
    })

    server := &http.Server{
        Handler: mux,
    }

    server.Serve(ln)
```

distros/distro/oteldistribution.go
tamirdavid1: changed, lmk if it is more clear

distros/distro/oteldistribution.go
tamirdavid1: You mean odiglet to choose based on the known constraints ? e.g init-container used ? 
honestly I dont think it is making sense in the context it is now (in `EnvironmentVariables` distro param).
but I do understand what you say if agent will support both how we says that.
let me work on something

opampserver/pkg/transport/unix/server.go
tamirdavid1: 0666 on the directory is not enought clients can't reach the socket at all.
changed to 0755 which is the less permissive works

distros/distro/oteldistribution.go
tamirdavid1: changed take a look please, now is much robust and better.

opampserver/pkg/transport/unix/server.go
tamirdavid1: The Java agent (later all agents) keeps one unix socket open for its whole lifetime and pushes many framed protobuf messages over it (init, status, ~30s heartbeats, disconnect), so the server has to read frame after frame on the same connection. Without the loop we'd handle one message and close, forcing the agent to reconnect every heartbeat — which defeats the point of using a persistent unix socket in the first place.

opampserver/pkg/transport/unix/server.go
tamirdavid1: done

opampserver/pkg/transport/unix/server.go
tamirdavid1: not relevant because of the next comment

opampserver/pkg/server/server.go
```diff
@@ -176,13 +105,8 @@ func StartOpAmpServer(ctx context.Context, mgr ctrl.Manager, kubeClientSet *kube
 		for {
 			select {
 			case <-ctx.Done():
-
 				// Close the updateChannel here so the worker goroutine exits
 				close(updateChannel)
-
-				if err := server.Shutdown(ctx); err != nil {
```
RonFed, line 183: why is the shutdown moved to a gorotuine?

opampserver/pkg/server/message_processor.go
RonFed, line 1: why do we need this message processor layer? the original handler did not have it

opampserver/pkg/server/server.go
```diff
@@ -40,130 +41,58 @@ func StartOpAmpServer(ctx context.Context, mgr ctrl.Manager, kubeClientSet *kube
...
+	// Start the worker goroutine to process instrumentation instances updates sequentially
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		ProcessInstrumentationUpdates(ctx, updateChannel, handlers)
+	}()
 
-		// Send the response.
-		w.Header().Set("Content-Type", "application/x-protobuf")
-		_, err = w.Write(bytes)
```
RonFed, line 145: some (or all) of the comments here were removed - I think we should keep them for clarity

opampserver/pkg/server/server.go
```diff
@@ -40,130 +41,58 @@ func StartOpAmpServer(ctx context.Context, mgr ctrl.Manager, kubeClientSet *kube
...
+	// Both listeners speak plain HTTP/1.1 over their respective transports and share
+	// the same MessageProcessor + connection cache. Agents choose via injected env:
+	//   ODIGOS_OPAMP_SERVER_HOST    (HTTP over TCP)
+	//   ODIGOS_OPAMP_UNIX_SOCKET    (HTTP over node-local unix socket)
+	tcpServer := &http.Server{
+		Addr:    fmt.Sprintf("0.0.0.0:%d", commonconsts.OpAMPPort),
+		Handler: transport.NewHandler(ctx, processor, commonopamp.OpAmpTransportHTTP),
+	}
+	unixServer := &http.Server{
+		Handler: transport.NewHandler(ctx, processor, commonopamp.OpAmpTransportUnix),
```
RonFed: why do we need 2 different handlers? if it's only for logging I think its nopt necessery

opampserver/README.md
```diff
@@ -6,11 +6,27 @@ This is an implementation of an OpAmp (almost) compatible server.
...
 - the instrumentation device id injected by odigos is only available on the node itself, and cannot be resolved from a different node in the cluster.
 
+## Transports
+
+OpAMP messages (`AgentToServer` / `ServerToAgent` protobuf) are handled by a shared `MessageProcessor`. Two transports run in parallel:
+
+| Transport | Endpoint | Wire format |
+|-----------|----------|-------------|
+| HTTP | `0.0.0.0:4320` `POST /v1/opamp` | Raw protobuf body |
+| Unix | `/var/odigos/exchange/exchange.sock` | 4-byte big-endian length + protobuf |
```
RonFed, line 21: needs to be updated - it is now also using HTTP

common/opamp/transport.go
```diff
@@ -0,0 +1,55 @@
...
+			return t
+		}
+	}
+	return OpAmpTransportNone
+}
+
+func isTransportUsable(t OpAmpTransport, mountMethod common.MountMethod) bool {
+	switch t {
+	case OpAmpTransportHTTP:
+		return true
```
RonFed, line 47: Does this mean we always try HTTP if the agent has it? or does it depend on the order of the `supported` slice? is HTTP always first in the slice?

distros/distro/oteldistribution.go
```diff
@@ -281,6 +285,12 @@ type OtelDistro struct {
 	// to enable the distribution.
 	EnvironmentVariables EnvironmentVariables `yaml:"environmentVariables,omitempty"`
 
+	// OpAmpTransportsSupported is the ordered list of OpAMP client transports this distribution's
+	// agent can speak. The webhook picks the first entry that is usable on the target node given
+	// runtime constraints (mount method, runtime version, …). Empty defaults to [http].
+	// Only consulted when EnvironmentVariables.OpAmpClientEnvironments is true.
+	OpAmpTransportsSupported []commonopamp.OpAmpTransport `yaml:"opAmpTransportsSupported,omitempty"`
```
RonFed, line 292: so if http is first in the distro yaml - it would always be used? if we want to give priority to unix - it should be placed first on the list?
Do you think the priority should be defined by the distro yaml - or should we have a global priority - where we say "we always prioritize unix over http if its available"?

opampserver/pkg/opamptypes/types.go
```diff
@@ -0,0 +1,22 @@
...
+type ProcessStatus int
+
+const (
+	ProcessOK ProcessStatus = iota
+	ProcessBadRequest
+	ProcessError
+)
+
+// Processor handles OpAMP messages for any transport.
+type Processor interface {
```
RonFed: I think now that we have the same HTTP handler we can avoid this abstraction

distros/distro/oteldistribution.go
tamirdavid1: So yes currently it depends on the order in the slice, if http before unix it will be that

opampserver/README.md
tamirdavid1: that what written there

OpAMP messages (`AgentToServer` / `ServerToAgent` protobuf) are accepted on two transports in parallel; both are served by the same `http.Server` + handler:

RonFed (approved):

PR #4920, opened 2026-05-22
CORE-1052 | fix(profiling): forward profiles for every active Source, not just those with collector-side rules

shivanshuraj1333:

Previously we only checked containerKeys for the workload, and those keys were populated only when the IC carried any collector-side rule (URL templatization / tail sampling). This breaks the cases where IC doesn't contain any collector rules.

Now we check whether the workload's entry itself exists in the cache, which directly reflects whether an InstrumentationConfig exists for the workload, which is a direct reflection of the Source.

```release-note
None
```

cherry-pick: 1.27

### DIFF
--- a/collector/extension/odigosconfigk8sextension/cache.go
+++ b/collector/extension/odigosconfigk8sextension/cache.go
@@ -89,20 +89,23 @@ func (c *cache) Get(key string) (*commonapi.ContainerCollectorConfig, bool) {
 	return val, found
 }
 
-// hasContainersForWorkloadPrefix returns true when the workload key prefix (e.g. "ns/Kind/name/")
-// has at least one full container cache key. Used for profile filtering where resource attributes
-// may not include k8s.container.name.
-//
-// The len(containerKeys) > 0 check is load-bearing, not defensive. The index entry's lifetime is
-// the union of its containerKeys and dataStreams: SetDataStreams creates an entry with an empty
-// containerKeys map, and Delete only removes the entry when both sets are empty. So entry != nil
-// alone does not imply "has at least one InstrumentationConfig container entry" — a workload that
-// only has data-stream labels would falsely return true without the length check.
-func (c *cache) hasContainersForWorkloadPrefix(workloadKeyPrefix string) bool {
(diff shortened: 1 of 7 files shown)


collector/extension/odigosconfigk8sextension/cache.go
```diff
@@ -89,20 +89,23 @@ func (c *cache) Get(key string) (*commonapi.ContainerCollectorConfig, bool) {
...
+// isSourceWorkloadPrefix returns true when the workload key prefix (e.g. "ns/Kind/name/")
+// has an entry in the workloadKeysIndex — i.e. an InstrumentationConfig currently exists for
+// the workload. Used by odigosprofilesprocessor (which only has ns/kind/name on the resource,
+// not k8s.container.name) to forward profiles only for active Sources.
+func (c *cache) isSourceWorkloadPrefix(workloadKeyPrefix string) bool {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
-	entry := c.workloadKeysIndex[workloadKeyPrefix]
-	return entry != nil && len(entry.containerKeys) > 0
+	_, ok := c.workloadKeysIndex[workloadKeyPrefix]
```
shivanshuraj1333, line 99: this was the main bug

collector/extension/odigosconfigk8sextension/informer.go
```diff
@@ -147,6 +147,8 @@ func (o *OdigosWorkloadConfig) handleInstrumentationConfigDelete(obj interface{}
 		return
 	}
 	o.syncWorkloadToDesiredState(workloadKey, nil, nil)
+	workloadIndexKey := WorkloadKeyString(workloadKey.Namespace, workloadKey.Kind, workloadKey.Name) + "/"
+	o.cache.removeWorkloadEntry(workloadIndexKey)
```
shivanshuraj1333, line 151: on delete event, we need to ensure that the entry is also cleaned up from the cache

shivanshuraj1333:
Have tested the changes end to end, this fixes the issue!

tamirdavid1 (approved):

PR #4932, opened 2026-05-24
chore: bump ui-kit with new feats

BenElferink:
```release-note
Features
- add Extract Attribute action to UI
- collectors-pipeline: add span-drop-rates column to table
- debug: implement CLI commands for workload
- edit-source-drawer: adapt drawer UI for VM sources
- improve responsiveness to remove height prop-drilling
- instrumentation-rules: add Node.js to supported languages for CodeAttributes
- instrumnetation-rules: add source scope handling in rule forms

Bug Fixes
- actions: rename Extractions to ExtractAttribute and update related structures
- always show hover effect
- column-list-item: hover background color logic
- typography: text wrapping
```

emergency-ticket id: EMR-1739


frontend/graph/schema.graphqls
```diff
@@ -484,6 +495,18 @@ input ActionFieldsInput {
   piiCategories: [String!]
 
   urlTemplatizationRulesGroups: [UrlTemplatizationRulesGroupInput!]
+
+  extractions: [ExtractionInput!]
```
alonkeyval: extractions should probably match the naming pattern of the other action-specific fields. Maybe use extractAttribute / extractAttributes or another name that clearly maps to the ExtractAttribute action instead of the generic extractions.

BenElferink: I handled this 👍
But we cannot merge until we get this kit PR take care of 1st: https://github.com/odigos-io/ui-kit/pull/983

frontend/webapp/graphql/mutations/instrumentation-rule.ts
```diff
@@ -10,6 +10,12 @@ export const CREATE_INSTRUMENTATION_RULE = gql`
       disabled
       mutable
       profileName
+      sourcesScopes {
```
alonkeyval, line 13: We’re adding sourcesScopes in several GraphQL operations with the same field selection. Could we extract this into a shared fragment? It would reduce duplication and make future schema changes safer. Same idea may apply to extractAttribute fields too and other fields

alonkeyval (approved):

frontend/webapp/graphql/mutations/instrumentation-rule.ts
BenElferink: Created follow up tickets 👍 
https://linear.app/odigos/issue/PLAT-1203/refactor-querymutation-strings-to-re-use-snippets
https://linear.app/odigos/issue/PLAT-1204/refactor-querymutation-strings-to-re-use-snippets-copy

PR #4933, opened 2026-05-24
CORE-1001 | Refactor(frontend): Extract importable server package (enables enterprise UI + MCP)

shivanshuraj1333:
## What

Makes the Odigos OSS frontend **importable** so an out-of-tree binary (the enterprise UI image) can reuse the exact backend setup instead of duplicating it.

Extracts all of `main()` into a small, ordered, public `frontend/server` package:

| Function | Role |
|---|---|
| `ParseFlags()` | flag parsing |
| `Bootstrap(ctx, flags, logger)` | synchronous setup (kube client, cache, OTLP, metrics, profiles); no goroutines |
| `StartBackground(ctx, deps)` | long-running workers |
| `BuildRouter(ctx, deps, RouterOpts)` | the gin engine |
| `ServeAndWait(...)` | serve loop + graceful shutdown |

`main.go` shrinks from ~430 lines to a thin sequence of these calls.

## The extension seam

`RouterOpts.ExtraMounts` lets an out-of-tree wrapper attach handlers **after** all OSS routes but **before** the SPA `NoRoute` fallback (so the catch-all doesn't swallow them). The OSS UI passes none. This is how the enterprise UI mounts `/mcp` (odigos-io/odigos-enterprise#2871) without forking the frontend.

`Bootstrap` and `StartBackground` are deliberately separate so an out-of-tree main can interleave its own steps (e.g. the enterprise license check) between them.

## Non-goals / behavior

Pure refactor — **no behavior change** to the OSS UI: same GraphQL surface, same routes, same remote-CLI endpoints. The webapp bundle + `/workloads` page embed inside the `webapp`/`graph` packages.

## Tests
- `server/router_test.go`: asserts an `ExtraMounts` handler is reachable (registered before `NoRoute`), `/healthz` is wired, and nil mounts are skipped.
- `go build ./... && go vet ./... && go test ./server/...` green. Full UI binary builds + ran E2E on a live cluster.

```release-note
NONE
```


frontend/graph/workloads_embed.go
```diff
@@ -0,0 +1,11 @@
+package graph
```
tamirdavid1, line 1: why needed?

shivanshuraj1333: https://github.com/odigos-io/odigos/blob/main/frontend/main.go#L74

shivanshuraj1333: we just refactored it, go embed makes these static html file that we have and the webapp/out to be present in the binary itself

alonkeyval:
LGTM

alonkeyval (approved):

PR #4958, opened 2026-05-31
CORE-1093 move test connection to a common package

natyharary:

Moves the destination's "test connection" functionality to the `common` package.
There's some glue around this implemenation which uses dependency inversion and duck typing when creating the exporter in the otelcol (which actually does that testing), because:
1. We want to use this package in the VMAgent as well - which is the reason for this refactor in the first place
2. Avoiding importing the opentelemetry collector

```release-note
NONE
```


frontend/graph/schema.resolvers.go
```diff
@@ -706,6 +694,33 @@ func (r *mutationResolver) TestConnectionForDestination(ctx context.Context, des
 	}, nil
 }
 
+func destinationInputToConfigurer(destination model.DestinationInput) *testconnection.Configurer {
```
alonkeyval: schema.resolvers.go is a generated file from gqlgen. Anything you add manually outside the protected sections can be overwritten the next time someone runs make gqlgen.

need to move it to conversion file

natyharary: Got it - I didn't know it is generated. Thanks for the comment! Moved it to conversions.

TheErikNaz (approved):
lgtm but the frontend part should also be looked by @alonkeyval or @BenElferink

frontend/services/test_connection/otlp_test_connection_test.go
RonFed, line 1: can we keep this test? also for otlp http?
I originally added it to help us capture breaking changes in the collector packages

frontend/go.mod
```diff
@@ -8,15 +8,15 @@ require (
...
+	github.com/gin-gonic/gin v1.12.0
 	github.com/glebarez/sqlite v1.11.0
 	github.com/go-logr/logr v1.4.3
-	github.com/grafana/pyroscope v1.21.0
+	github.com/grafana/pyroscope v1.21.1
 	github.com/grafana/pyroscope/api v1.5.0
 	github.com/odigos-io/odigos/api v0.0.0
 	github.com/odigos-io/odigos/common v0.0.0
-	github.com/odigos-io/odigos/config v0.0.0
+	github.com/odigos-io/odigos/config v0.0.0-20260602173917-bd81f1250384
```
RonFed: maybe this is off?

frontend/go.mod
```diff
@@ -387,14 +386,9 @@ require (
...
-	github.com/hashicorp/memberlist => github.com/grafana/memberlist v0.3.1-0.20251126142931-6f9f62ab6f86
 	github.com/odigos-io/odigos/api => ../api
 	github.com/odigos-io/odigos/common => ../common
-	github.com/odigos-io/odigos/config => ../config
 	github.com/odigos-io/odigos/destinations => ../destinations
 	github.com/odigos-io/odigos/k8sutils => ../k8sutils
 	github.com/odigos-io/odigos/odigosauth => ../odigosauth
-	// Pyroscope (and Mimir) expect KnownFields YAML decoding; upstream v3.0.1 lacks DecodeWithOptions.
-	// Matches github.com/grafana/pyroscope go.mod.
-	gopkg.in/yaml.v3 => github.com/colega/go-yaml-yaml v0.0.0-20220720105220-255a8d16d094
```
RonFed, line 399: why can we remove this now?

frontend/go.mod
```diff
@@ -387,14 +386,9 @@ require (
 )
 
 replace (
-	github.com/hashicorp/memberlist => github.com/grafana/memberlist v0.3.1-0.20251126142931-6f9f62ab6f86
```
RonFed, line 390: why can we remove this now?

frontend/go.mod
```diff
@@ -341,37 +339,38 @@ require (
...
-	go.opentelemetry.io/collector/extension/extensionauth v1.54.0 // indirect
-	go.opentelemetry.io/collector/extension/xextension v0.148.0 // indirect
-	go.opentelemetry.io/collector/internal/sharedcomponent v0.148.0 // indirect
-	go.opentelemetry.io/collector/internal/telemetry v0.148.0 // indirect
+	go.opentelemetry.io/collector/exporter/exporterhelper/xexporterhelper v0.130.0 // indirect
+	go.opentelemetry.io/collector/exporter/xexporter v0.130.0 // indirect
+	go.opentelemetry.io/collector/extension/extensionauth v1.36.0 // indirect
+	go.opentelemetry.io/collector/extension/xextension v0.130.0 // indirect
+	go.opentelemetry.io/collector/internal/sharedcomponent v0.130.0 // indirect
+	go.opentelemetry.io/collector/internal/telemetry v0.130.0 // indirect
```
RonFed: don't think these should change, when we bump the collector version we do it across all of our components - so common and frontend should be aligned with the collector version we use in our collector.

frontend/services/testconnectionotel/exporttester.go
```diff
@@ -0,0 +1,134 @@
...
+	"go.opentelemetry.io/collector/exporter"
+	"go.opentelemetry.io/collector/exporter/exportertest"
+	"go.opentelemetry.io/collector/pdata/plog"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+
+	"google.golang.org/grpc/status"
+)
+
+var testers = []testconnection.ExporterConnectionTester{NewOTLPTester(), NewOTLPHTTPTester()}
```
RonFed: nit (some formatting):
```Go
var (
	connectionTesters = []ExporterConnectionTester{
		NewOTLPTester(),     // "otlp_grpc/" prefix
		NewOTLPHTTPTester(), // "otlp_http/" prefix
	}
)
```

frontend/services/testconnection/exporttester.go
RonFed, line 1: I'm fine with the package name being `testconnection` or `test_connection`
`testconnectionotel` sounds too verbose to me

common/config/testconnection/testconnection.go
```diff
@@ -0,0 +1,160 @@
...
+	Message    string
+}
+
+// An interface to prevent importing the collector - dependency inverts and delegates the implementation to the caller modules (OSS/VMAgent)
+type ExporterConnectionTester interface {
+	// Prefix is the exporter "type" segment (before the "/") that this tester knows how to test,
+	// e.g. "otlp" matches generated exporter ids like "otlp/my-destination".
+	Prefix() string
+	// builds the exporter from the destination's rawConfig, connects, and exports an empty payload.
+	TestExport(ctx context.Context, rawConfig map[string]any, signals []common.ObservabilitySignal) ExportAttempt
```
RonFed: we didn't have the per signal notion in the original implementation.
I suggest not adding it in this refactor PR, but we can discuss it - there are few question about per signal check - I think the signals to check should ideally be provided by the client (UI) since it is aware of which signals the user wants to send to the destination - this set of signals is not necessarily the same as the set of the destination supported signals.

RonFed (changes requested):
main concern is the go.mod changes that I'm not sure are all correct

frontend/go.mod
natyharary: Reverted

frontend/go.mod
natyharary: Reverted, go mod tidy got cheeky

frontend/go.mod
natyharary: Reverted

frontend/go.mod
natyharary: Reverted

common/config/testconnection/testconnection.go
natyharary: Put it into stash instead :) waiting for that one day when we'll want to do this

natyharary: Maybe this can be a good first issue for new employees?

frontend/services/testconnection/exporttester.go
natyharary: Cool, went with `testconnection`

RonFed (approved):

PR #4961, opened 2026-06-01
fix(ignored-namespaces): enhance workload filtering with bypass logic

BenElferink:
This update introduces a new mechanism for handling workloads in ignored namespaces by implementing a bypass set for workloads that have an associated InstrumentationConfig. The changes ensure that instrumented workloads remain visible and mutable in the UI, even if their namespaces are configured to be ignored.

Key modifications include:
- Addition of `BypassWorkloads` to the `WorkloadFilter` struct.
- Implementation of `ShouldIgnoreWorkload` method to determine if a workload should be excluded based on its namespace and bypass status.
- Refactoring of various fetching functions to utilize the new filtering logic.

This enhancement improves the user experience by allowing visibility into instrumented workloads that would otherwise be hidden due to namespace filtering.

```release-note
fix(ignored-namespaces): enhance workload filtering with bypass logic
```

emergency-ticket id: EMR-1790


frontend/graph/loaders/types.go
```diff
@@ -29,4 +30,26 @@ type WorkloadFilter struct {
...
+func (f *WorkloadFilter) ShouldIgnoreWorkload(id model.K8sWorkloadID) bool {
+	if f == nil {
+		return false
+	}
+	if _, ignored := f.IgnoredNamespaces[id.Namespace]; !ignored {
+		return false
+	}
+	if _, bypass := f.BypassWorkloads[id]; bypass {
+		return false
+	}
```
RonFed, line 53: shouldn't we check this before `IgnoredNamespaces`? since we want to give this priority over ignored namespaces?

frontend/graph/loaders/fetchers.go
```diff
@@ -749,16 +782,39 @@ func fetchInstrumentationInstances(ctx context.Context, logger logr.Logger, filt
...
+		// so the UI can render pod/process health for the user-instrumented component.
+		// Falls back to the instance's own namespace when the workload label is missing.
+		if _, namespaceIgnored := filters.IgnoredNamespaces[ii.Namespace]; namespaceIgnored {
+			if workloadId.Name == "" {
+				continue
+			}
+			if _, bypass := filters.BypassWorkloads[workloadId]; !bypass {
+				continue
+			}
+		}
```
RonFed: why not to use the `ShouldIgnoreWorkload` here?

frontend/graph/loaders/types.go
BenElferink: Flipping the order wouldn't change behavior, but the current order is slightly better:

- Both orderings are logically equivalent: the function returns `true` only when the namespace is ignored AND the workload is not in the bypass set. `(¬A ∨ B) ⇒ false` is the same regardless of which side you check first.
- The "priority" concept doesn't really apply here - `BypassWorkloads` is **only meaningful** when the namespace is ignored. Its sole purpose is to override the ignored-namespace exclusion. If the namespace isn't ignored, there's nothing to bypass, so checking the bypass set first is wasted work.
- The current order short-circuits on the common case: most workloads in a cluster are in non-ignored namespaces, so the first check returns `false` immediately and we never touch the (potentially larger) `BypassWorkloads` map.

So no change needed - the priority is preserved by the second check, not by ordering.

RonFed (approved):

PR #4964, opened 2026-06-01
CORE-1071 | feat(profiling): make VM profile-attrs PID→service cache survive collector reloads

shivanshuraj1333:

The `odigosvmprofileattrsprocessor` PID→service.name cache (and its unixfd client to the VM agent) lived on the processor instance. A collector config reload (SIGHUP) does a full Shutdown()+rebuild, destroying the processor and its cache — so the rebuilt processor enriched profiles from an empty cache during warm-up and silently dropped them (`dropped_resource_profiles`) on every reload.

### Fix
Move the cache + client to a package-level singleton (`sync.Once`, `context.Background()` goroutine) that the processor only references. A reload rebuilds the processor but never touches the singleton, so the cache stays warm — zero drops across reloads. New `shared_cache.go`; `start()` binds the singleton, `shutdown()` no longer tears down the client; a cancellable context keeps the package goleak test clean.

### Testing
- Unit: `TestSharedCache_SurvivesProcessorRebuild`.
- Live on a VM: `systemctl reload odigos-otelcol` under traffic kept profiles flowing
  with no cold-cache drop spike; same collector PID (true reload).

```release-note
Fix VM profiling dropping profiles on every collector config reload by making the PID→service.name enrichment cache process-global
```

### DIFF
--- a/collector/processors/odigosvmprofileattrsprocessor/processor.go
+++ b/collector/processors/odigosvmprofileattrsprocessor/processor.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"strings"
-	"sync"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/consumer"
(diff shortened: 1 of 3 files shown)


natyharary (approved):
lgtm but I want @TheErikNaz 's 👀  as well here

TheErikNaz (approved):

PR #4984, opened 2026-06-04
feat: configurable ebpf map size env var

daemon1024:
Fixes RUN-768

```release-note
add tshirt sizing for relevant map sizes
```


RonFed (commented):
should we add the new value to values.yaml with a comment describing it?

PR #4990, opened 2026-06-05
CORE-1023: Use onPremToken to pull enterprise images

damemi:
This sets it up so that enterprise images are pulled from our private enterprise registry, using the onPremToken JWT as auth.

* If onPremToken is set, create a pull secret using it
* Use the pull secret for enterprise images
* Only use token as pull secret when pulling from default enterprise registry (ie, not a custom registry prefix)

cherry-pick:v1.30

```release-note
enterprise: Enterprise images now pull from enterprise-registry.odigos.io and require valid onPremToken + auto-generated pull secret (for default Odigos enterprise registry)
```


cli/cmd/resources/centralodigos/centralui.go
```diff
@@ -29,7 +29,7 @@ func (m *centralUIResourceManager) Name() string { return k8sconsts.CentralUIApp
 
 func (m *centralUIResourceManager) InstallFromScratch(ctx context.Context) error {
 	return m.client.ApplyResources(ctx, 1, []kube.Object{
-		NewCentralUIDeployment(m.ns, k8sconsts.OdigosImagePrefix, m.managerOpts.ImageReferences.CentralUIImage, m.odigosVersion, m.managerOpts.ImagePullSecrets),
+		NewCentralUIDeployment(m.ns, k8sconsts.OdigosEnterpriseImagePrefix, m.managerOpts.ImageReferences.CentralUIImage, m.odigosVersion, m.managerOpts.ImagePullSecrets),
```
tamirdavid1: as i understood from @alonkeyval this is not needed (everything in resources) as it about to be deprecated

cli/cmd/pro-dep.go
```diff
@@ -248,6 +256,33 @@ func createOdigosCentralSecretDep(ctx context.Context, client *kube.Client, ns,
 	return nil
 }
 
+// ensureCentralEnterpriseRegistryPullSecretsDep creates/updates the enterprise registry pull secret when
```
tamirdavid1: the same, if the cli approach will be deprecated i dont think we need it

helm/odigos-central/templates/_helpers.tpl
```diff
@@ -1,5 +1,9 @@
-{{- define "utils.imageName" -}}
-{{- $defaultRegistry := "registry.odigos.io" -}}
+{{- define "utils.imagePrefix" -}}
+{{- $enterprise := .Enterprise | default true -}}
+{{- $defaultRegistry := "enterprise-registry.odigos.io" -}}
```
tamirdavid1: the logic for overwrite registry the users uses will still work?
i mean `imagePrefix`

helm/odigos-central/templates/_helpers.tpl
```diff
@@ -27,12 +83,16 @@ true
   {{- end -}}
 {{- end -}}
 
-{{/* Render imagePullSecrets in K8s shape from a list of strings */}}
+{{/* Render imagePullSecrets. Set Enterprise=true on workloads that pull enterprise images. */}}
 {{- define "odigos.renderPullSecrets" -}}
-{{- if .Values.imagePullSecrets }}
+{{- $enterpriseSecret := include "odigos.enterpriseRegistryPullSecretToMount" . -}}
+{{- if or .Values.imagePullSecrets $enterpriseSecret }}
 imagePullSecrets:
```
tamirdavid1, line 84: just checking, we wont regression the current option for users to set their own imagePullSecrets right?

helm/odigos-central/values.yaml
```diff
@@ -7,6 +7,13 @@ imagePrefix: ''
 # -- Image pull secrets for private registries (k8s secret names)
 imagePullSecrets: []
 
+# On-prem token for Odigos Enterprise (chart-managed odigos-central secret when set)
+# onPremToken: ''
+
+# Set to true when odigos-central is managed externally (GitOps). The chart reads the token from
+# the existing odigos-central secret to create the enterprise registry pull secret.
+# externalOnpremTokenSecret: false
```
tamirdavid1, line 15: why this all in comment? 
what will be the experience for existing central users? they will need to enable it? if so maybe worth talking with @rayray221 on this?

helm/odigos/values.schema.json
```diff
@@ -524,7 +524,7 @@
     },
     "externalOnpremTokenSecret": {
       "default": "false",
-      "description": "Set to true if you are providing the odigos-pro secret externally (not managed by this chart).\nWhen true, the chart will not create the secret but will still enable Pro features if the secret exists.\nThis is useful for GitOps workflows where secrets are managed separately.",
+      "description": "Set to true if you are providing the odigos-pro secret externally (not managed by this chart).\nWhen true, the chart will not create the odigos-pro secret but will still enable Pro features if the secret exists.\nThe chart reads the token from the existing odigos-pro secret to create the enterprise registry pull secret.\nThis is useful for GitOps workflows where secrets are managed separately.",
```
tamirdavid1: so if users set external onprem token the registry pull secret will be fetched from the customer externally managed one?

k8sutils/pkg/pro/common.go
```diff
@@ -68,3 +117,41 @@ func odigletRolloutTrigger(ctx context.Context, client kubernetes.Interface, nam
 type TokenPayload struct {
 	OnpremToken string `json:"token"`
 }
+
+// EnterpriseImagePrefix returns the registry used for enterprise images, mirroring helm utils.imagePrefix with Enterprise=true.
+func EnterpriseImagePrefix(config common.OdigosConfiguration) string {
```
tamirdavid1: pass a pointer? the linter suggested

tamirdavid1 (commented):
overall lgtm, wrote few questions.
the most "critical" part for me is maybe we can avoid all logic related to installing central using the old approach it will save a lot of code that might be deleted very soon

helm/odigos-central/templates/_helpers.tpl
damemi: correct this just appends the enterprise secret to any existing imagepullsecrets (if the enterprise secret is set. if a customer is using their own registry prefix we don't create it, which will likely be the case if they're using their own pull secrets)

helm/odigos-central/values.yaml
damemi: these were already optional values, just not documented in the helm value. adding the comments doesn't change how they're used, it just documents the behavior

used in [`centralbackend/secret.yaml`](https://github.com/odigos-io/odigos/blob/b2cc97700d3c6767b6a74a065d9aca27605ebeb5/helm/odigos-central/templates/centralbackend/secret.yaml#L6) and [`helpers.tpl`](https://github.com/odigos-io/odigos/blob/b2cc97700d3c6767b6a74a065d9aca27605ebeb5/helm/odigos-central/templates/_helpers.tpl#L24)

helm/odigos/values.schema.json
damemi: the token will be fetched from the users external secret, but we will still generate the docker auth pull secret in cluster. So it doesn't need any change from users, just provide your enterprise token and we handle the auth side

helm/odigos/templates/_helpers.tpl
```diff
@@ -12,6 +12,54 @@
...
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "odigos.enterpriseRegistryPullSecretName" -}}
+odigos-enterprise-registry
+{{- end -}}
+
+{{- define "odigos.hasEnterpriseRegistryPullSecret" -}}
+{{- if and (include "odigos.onPremToken" .) (include "odigos.usesOdigosRegistry" .) -}}
```
blumamir, line 33: We should be honoring externalOnpremTokenSecret here, or at least document support level for it.
don't think anyone is actually using this configuration at the moment, but if someone will try we probably want it to work

helm/odigos/templates/_helpers.tpl
```diff
@@ -12,6 +12,54 @@
...
+true
+{{- end -}}
+{{- end -}}
+
+{{- define "odigos.usesOdigosRegistry" -}}
+{{- eq (include "utils.imagePrefix" (dict "Values" .Values)) "registry.odigos.io" -}}
```
blumamir, line 39: do we need to also address the redhat registry here?
can we simplify this to only check if `Values.imagePrefix` is empty? (e.g. not overwritten by custom registry)

helm/odigos/templates/_helpers.tpl
```diff
@@ -12,6 +12,54 @@
...
+{{- if .Values.onPremToken -}}
+{{- .Values.onPremToken -}}
+{{- else -}}
+{{- $sec := lookup "v1" "Secret" .Release.Namespace "odigos-pro" -}}
+{{- if $sec -}}
+{{- if index $sec.data "odigos-onprem-token" -}}
+{{- index $sec.data "odigos-onprem-token" | b64dec -}}
+{{- else if index $sec.data "ODIGOS_ONPREM_TOKEN" -}}
+{{- index $sec.data "ODIGOS_ONPREM_TOKEN" | b64dec -}}
+{{- end -}}
```
blumamir, line 23: why do we check both `odigos-onprem-token` and `ODIGOS_ONPREM_TOKEN`?

are they both being set?

k8sutils/pkg/pro/common.go
```diff
@@ -68,3 +117,33 @@ func odigletRolloutTrigger(ctx context.Context, client kubernetes.Interface, nam
...
+		return true, nil
+	}
+
+	var config common.OdigosConfiguration
+	if err := yaml.Unmarshal([]byte(configYAML), &config); err != nil {
+		return false, fmt.Errorf("failed to parse odigos configuration: %w", err)
+	}
+
+	return UsesOdigosRegistry(&config), nil
+}
```
blumamir, line 149: consider reusing the functions that already read and parse the config from the configmap:

[this](https://github.com/odigos-io/odigos/blob/cdcd3feabea0a6fbcd65d4d9079ae64918c12c98/k8sutils/pkg/utils/config_util.go#L24) or [this](https://github.com/odigos-io/odigos/blob/main/frontend/services/odigos_config.go)

helm/odigos-central/templates/_helpers.tpl
```diff
@@ -18,6 +18,56 @@ odigos-enterprise-{{- .Component -}}
...
+
+{{- define "odigos.validateEnterpriseRegistryPullSecrets" -}}
+{{- if and (include "odigos.secretExists" .) (include "odigos.usesOdigosRegistry" .) (not (include "odigos.onPremToken" .)) -}}
+{{- fail "Odigos images pull from registry.odigos.io but no on-prem token is available to the chart. Set onPremToken or ensure the odigos-central secret exists in the release namespace before install/upgrade." -}}
+{{- end -}}
+{{- end -}}
+
+{{- define "odigos.enterpriseRegistryDockerConfigJson" -}}
+{{- $token := include "odigos.onPremToken" . -}}
+{{- $auth := printf "odigos:%s" $token | b64enc -}}
```
blumamir, line 67: what is the reason to add the `odigos:` prefix to the token?
do we need to base64 the token? it's already based64 as a JWT, right?

helm/odigos/templates/enterprise-registry-pull-secret.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if include "odigos.hasEnterpriseRegistryPullSecret" . }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "odigos.enterpriseRegistryPullSecretName" . }}
+  namespace: {{ .Release.Namespace }}
```
blumamir, line 6: this secret will live in the odigos namespace.
will it work if someone using the `init-container` mount method? since the pod that try to pull the image live in the application's namespace, right?

blumamir (approved):
overall LGTM.

added few questions / concerns

helm/odigos/templates/_helpers.tpl
damemi: this does honor externalOnpremTokenSecret ... `odigos.onPremToken` checks the `odigos-pro` secret for the onpremtoken. ExternalTokenSecret just means the user is creating the `odigos-pro` themselves

helm/odigos/templates/_helpers.tpl
damemi: no the red hat registry is different right now. we can't easily put those images behind onprem auth for a couple of reasons:

* red hat serves the images through a proxy, so our custom auth handler breaks when it tries to serve GCP artifact registry images (because gcp returns a relative 302 redirect)
* we can't use per-customer auth because red hat's proxy requires any dockerconfig/auth settings to be submitted *with your images* and served to *everyone that pulls them*

It's probably not going to be possible for us to put enterprise auth around red hat images

helm/odigos-central/templates/_helpers.tpl
damemi: this is the docker `user:pass` format

helm/odigos/templates/enterprise-registry-pull-secret.yaml
damemi: https://github.com/odigos-io/odigos/pull/3913 solved that.. the init container is pre pulled in the odiglet daemonset

helm/odigos/templates/_helpers.tpl
damemi: no central uses it and this was a bad copy paste

k8sutils/pkg/pro/common.go
damemi: those helpers take a different type of k8s client (controller runtime)

PR #4991, opened 2026-06-05
CORE-1030: Support GKE Autopilot

damemi:
This adds support for GKE Autopilot based on our approval as an Autopilot partner https://docs.cloud.google.com/kubernetes-engine/docs/how-to/run-autopilot-partner-workloads and https://docs.cloud.google.com/kubernetes-engine/docs/resources/autopilot-partners

It adds a `gke.autopilot` flag that installs an AllowlistSynchronizer (GKE CRD) that will automatically install our Autopilot allowlist to support Odiglet's elevated permissions.

It also disables the odiglet's "node label" init phase, since we can't modify nodes on GKE Autopilot. This means hostpath mount method won't necessarily always work (if an injected pod gets scheduled on a different Autopilot node than the odiglet -- though not sure this is even possible with how Autopilot scales nodes).

Keeping a downstream copy of our git-on-borg (googlesource git repo) here: https://github.com/odigos-io/gke-ap-allowlist, which we could automate to submit new allowlists upstream. We will need a new allowlist if the odiglet daemonset definition changes

```release-note
Add GKE Autopilot which disables odiglet node labeling and gke.autopilot flag to install AllowlistSynchronizer
```

cherry-pick:v1.28

### DIFF
--- a/docs/snippets/shared/setup/installation-options.mdx
+++ b/docs/snippets/shared/setup/installation-options.mdx
@@ -166,3 +166,123 @@ you will need to modify the YAML of the oc route with `spec.tls.termination: edg
 Check the OpenShift documentation or run `oc expose -h` for more
 information on configuring OpenShift routes according to your
 cluster's requirements.
+
+## GKE Installation
+
+Installing on Google Kubernetes Engine (GKE) requires additional settings
+that can be configured based on your cluster type:
+
+<Tabs>
+  <Tab title="Odigos CLI">
+       For GKE Standard clusters:
+
+       ```shell
+       odigos install --set gke.enabled=true
+       ```
+
+       For GKE Autopilot clusters:
+
+       ```shell
+       odigos install --set gke.enabled=true --set gke.autopilot=true
+       ```
+  </Tab>
+  <Tab title="Helm Chart">
+       For GKE Standard clusters, set `gke.enabled=true` in `values.yaml`:
+
+       ```yaml
+       gke:
+         enabled: true
+       ```
+
+       Or pass it as a flag during installation:
+
+       ```shell
+       helm install odigos odigos/odigos --set gke.enabled=true
+       ```
+
+       For GKE Autopilot clusters, also set `gke.autopilot=true`:
+
+       ```yaml
+       gke:
+         enabled: true
+         autopilot: true
+       ```
+
+       Or pass both flags during installation:
+
+       ```shell
+       helm install odigos odigos/odigos --set gke.enabled=true --set gke.autopilot=true
+       ```
+  </Tab>
+</Tabs>
+
+GKE installation with `gke.enabled=true` does the following:
+
+* Configures a `ResourceQuota` for `system-node-critical` pods, which is required for odiglet on GKE
+
+### GKE Autopilot
+
+GKE Autopilot restricts privileged workloads by default. Odiglet requires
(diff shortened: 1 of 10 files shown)


docs/snippets/shared/setup/installation-options.mdx
```diff
@@ -166,3 +166,103 @@ you will need to modify the YAML of the oc route with `spec.tls.termination: edg
...
+GKE Autopilot restricts privileged workloads by default. Odiglet requires
+elevated permissions for eBPF-based instrumentation, so it must be allowlisted
+before it can run on Autopilot clusters.
+
+Allowlist synchronization requires GKE version `1.32.2-gke.1652000` or later.
+
+When `gke.autopilot=true`, the Helm chart:
+
+* Creates an `AllowlistSynchronizer` that syncs the Odigos odiglet allowlist from the [Google-managed partner repository](https://cloud.google.com/kubernetes-engine/docs/how-to/run-autopilot-partner-workloads)
+* Skips node label setup in the odiglet init phase, where Autopilot nodes cannot be modified. Autopilot is auto-detected by the `gk3-` node name prefix; set `gke.autopilot=true` to force this behavior
```
tamirdavid1, line 234: maybe worth adding what we discuss about limitations?

helm/odigos/templates/gke/allowlist-synchronizer.yaml
```diff
@@ -0,0 +1,16 @@
...
+# https://cloud.google.com/kubernetes-engine/docs/how-to/run-autopilot-partner-workloads
+apiVersion: auto.gke.io/v1
+kind: AllowlistSynchronizer
+metadata:
+  name: odigos-allowlist-synchronizer
+  labels:
+    odigos.io/system-object: "true"
+  annotations:
+    helm.sh/hook: pre-install,pre-upgrade
+    helm.sh/hook-weight: "-5"
```
tamirdavid1, line 12: what is this hook-weight?

helm/odigos/templates/odiglet/daemonset.yaml
```diff
@@ -113,6 +113,10 @@ spec:
                 resourceFieldRef:
                   containerName: init
                   resource: limits.cpu
+            {{- if .Values.gke.autopilot }}
```
tamirdavid1, line 116: wdyt about passing it to the odigos-configuration and read it instead of setting an env var?

damemi: I see your point, but right now it's just a small flag used in the odiglet init, where we don't currently wire up the odigos configmap. I think this is less invasive and matches how we get the odigos tier during odiglet init from an env var

helm/odigos/templates/gke/allowlist-synchronizer.yaml
damemi: hook-weight is a finer grained ordering in case you have multiple pre-install hooks for example, lower numbers go first. The allowlistsynchronizer should be a very early hook object. I don't think we specifically need hook weight but it seems good to have

docs/snippets/shared/setup/installation-options.mdx
damemi: yup updated

tamirdavid1 (approved):

PR #4994, opened 2026-06-06
Agent log level

blumamir:

Fixes: CORE-1136

## Summary

Adds an InstrumentationRule action to control agent diagnostics (agentDiagnostics). Log levels must be injected as environment variables at pod admission so they take effect during agent initialization (startup, config loading, instrumentation setup). The resolved config is also written to InstrumentationConfig container config, so levels can be updated dynamically when rules change without requiring manual pod edits.

## What it does

New spec.agentDiagnostics on InstrumentationRule with optional odigosLogLevel and openTelemetryComponentsLogLevel (error / warn / info / debug).
Instrumentor merges matching rules (most verbose wins) and stores the result on ContainerAgentConfig.agentDiagnostics.
Pods webhook injects ODIGOS_LOG_LEVEL and OTEL_LOG_LEVEL when the distro supports it.
Node.js community distro is enabled first via ownLogs capability flags.
Example

```
spec:
  agentDiagnostics:
    odigosLogLevel: debug
    openTelemetryComponentsLogLevel: info
```

```release-note
InstrumentationRule: agent diagnostics — Control Odigos agent and OpenTelemetry SDK log verbosity via agentDiagnostics on InstrumentationRules;
```


shivanshuraj1333:
I wonder if this should or needed to be extended for other env variables, like cache TTL etc, we saw requirement for it in demos otherwise flamgraphs takes time to render

distros/yamls/nodejs-community.yaml
```diff
@@ -32,3 +32,6 @@ spec:
       supported: true
     traceVerbosity:
       supportDisablingOdigosAgentLibraries: true
+  ownLogs:
```
RonFed:
```suggestion
  ownDiagnostics:
```

instrumentor/controllers/utils/predicates/instrumentation_rule.go
```diff
@@ -15,7 +15,8 @@ func isRuleRelevantForAgentInjection(spec *odigosv1alpha1.InstrumentationRuleSpe
 		spec.TraceVerbosity != nil ||
 		spec.CustomInstrumentations != nil ||
 		spec.CodeAttributes != nil ||
-		spec.EbpfLogCapture != nil
+		spec.EbpfLogCapture != nil ||
+		spec.AgentDiagnostics != nil
```
RonFed, line 19: by the shape of this function, it looks like we need to refactor this code (not in this PR just a thought for the future)

RonFed (approved):

blumamir:
> I wonder if this should or needed to be extended for other env variables, like cache TTL etc, we saw requirement for it in demos otherwise flamgraphs takes time to render

not sure I understand how it affects flamgraphs render time, but anyhow it's not related to the current scope which is to write logs :)

instrumentor/controllers/utils/predicates/instrumentation_rule.go
blumamir: agree

distros/yamls/nodejs-community.yaml
blumamir: fixed

PR #5009, opened 2026-06-09
CORE-00 | Enhance grpc metric value

tamirdavid1:

Change the name of the metric for rejecting traces in grpc reciever following rtml.
The metric exists both in data collection and gateway, so change the name to be more generic.
also added a logging in the gateway the same as we have in data-collection

```release-note
NONE
```

### DIFF
--- a/collector/config/configgrpc/configgrpc.go
+++ b/collector/config/configgrpc/configgrpc.go
@@ -619,6 +619,7 @@ func (sc *ServerConfig) getGrpcServerOptions(
 				settings.Logger.Info("OTLP gRPC receiver detected memory limit reached, dropping a batch")
 				return ctx, errMemoryLimitReachedDataDropped
 			} else { // gateway collector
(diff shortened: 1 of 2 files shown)


collector/config/configgrpc/configgrpc.go
```diff
@@ -619,6 +619,7 @@ func (sc *ServerConfig) getGrpcServerOptions(
 				settings.Logger.Info("OTLP gRPC receiver detected memory limit reached, dropping a batch")
 				return ctx, errMemoryLimitReachedDataDropped
 			} else { // gateway collector
+				settings.Logger.Info("OTLP gRPC receiver detected memory limit reached")
```
RonFed: maybe consider putting it in debug, might be noisy in the gateway

RonFed:
```suggestion
				settings.Logger.Info("OTLP gRPC receiver detected memory limit reached, returning a retrible error to sender")
```

blumamir (approved):

PR #5013, opened 2026-06-09
CORE-1092: Fix namespace source instrumentation race

damemi:

This addresses a race where an InstrumentationConfig delete event tricked the instrumentor into thinking we were uninstrumenting an app.

In a user's environment, they are using ArgoCD to delete/replace apps with a command like:

```
argocd app sync connect-app \
    --grpc-web \
    --resource apps:Deployment:default/analytics-api-service \
    --force --replace --async

##  ArgoCD output explicitly showed:
  deployment.apps "analytics-api-service" deleted
  deployment.apps/analytics-api-service replaced
```

After this command, they are noticing that apps which should be covered by an existing Namespace Source get stuck uninstrumented, with no `agents-meta-hash`. I was able to reproduce this only with a local Argo setup. When I tried to do it otherwise, without argo (just by deleting and recreating the app), I did see a pod temporarily go through with no `agents-meta-hash` but it was quickly recovered and rolled back out.

However, I think in the Argo case the situation is more like this:

1. Deployment is instrumented with NS source + valid IC
2. Deployment gets deleted and immediately replaced by Argo
3. New deployment pod gets admitted before old IC gets garbage collected (k8s delete event fires)
4. Old IC delete event finally fires, and our instrumentor controller handles it like an uninstrument event:
  a. `c.Get()`[ returns nil, the IC is deleted](https://github.com/odigos-io/odigos/blob/bfc5da9bb14bb84f0d45fb1577b0e375151fae7a/instrumentor/controllers/agentenabled/sync.go#L96-L98)
  b. `rollout.Do()`[ treats a nil IC as an uninstrument event](https://github.com/odigos-io/odigos/blob/bfc5da9bb14bb84f0d45fb1577b0e375151fae7a/instrumentor/controllers/agentenabled/rollout/rollout.go#L42-L43)
  c. [this assumption about rollbacks is wrong](https://github.com/odigos-io/odigos/blob/bfc5da9bb14bb84f0d45fb1577b0e375151fae7a/instrumentor/controllers/agentenabled/rollout/rollout.go#L87-L88) -- an IC can be nil and a pod can have empty agents hash in this race scenario

So essentially, the InstrumentationConfig delete event comes through after the new Pod has been admitted and our current logic says "Oh, the IC was deleted, we must be uninstrumenting". This makes sense because when we delete a Source (or switch it to disabled) the sourceinstrumentation controller does delete the IC, which is the signal to our internal design to uninstrument.

The fault in that logic is that while the IC is the internal source of truth for instrumentation, the overall source of truth is supposed to be Source objects. A missing IC alone should not be enough for us to consider the app uninstrumented. In this scenario, with the GC race, we have a situation where the IC gets deleted even though the Source says "this should still be instrumented". So, an IC alone is not always enough for us to deduce the current state and we should remember to fall back on our real source of truth.

Also, this updates our IC creation logic to look for UPDATEs on instrumented workloads. Previously we just looked at CREATE, and if a workload was attempted to roll out during a race like this with an IC missing, there was nothing to trigger IC creation again.

```mermaid
sequenceDiagram
    participant Argo
    participant Deploy as Deployment
    participant IC as InstrumentationConfig
    participant SI as sourceinstrumentation
    participant Pod as New Pod
    participant WH as Pod Webhook

    Argo->>Deploy: delete old deployment
    Deploy-->>IC: owner ref GC → IC deleted
    Argo->>Deploy: create new deployment
    Deploy->>Pod: ReplicaSet spins up pods
    Pod->>WH: admit pod
    WH-->>Pod: no IC → skip injection (no hash)

    Note over SI: Without predicate: nothing re-runs syncWorkload
    Note over SI: With predicate: deployment UPDATE → syncWorkload → IC created

```

This also fixes errors in the instrumentor logs like:

```
InstrumentationConfig.odigos.io "deployment-analytics-api-service" is invalid:
[status.conditions[1].lastTransitionTime: Required value,
 status.conditions[2].lastTransitionTime: Required value]
```

https://github.com/odigos-io/odigos/pull/4441 removed the lastTransitionTime from them

cherry-pick:v1.30

```release-note
fix: namespace source instrumentation race which prevented redeployed apps from being properly instrumented
```


RonFed:
@damemi is this fixing the race or taking the approach of making the ic creation faster to reduce the chance for a race?

damemi:
@RonFed you're right, this was originally just trying to make the race less likely. I took a deeper look into it and change what we're looking at here. Updated my PR description with what I found, ptal

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -97,6 +98,17 @@ func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.Instrumentation
 			return RolloutResult{}, nil
 		}
 
+		stillInstrumented, instrumentedErr := workloadStillMarkedForInstrumentation(ctx, c, pw)
```
RonFed, line 108: before this we do 
```
hasAgents, agentsErr := workloadHasOdigosAgents(ctx, c, workloadObj)
```

isn't the point of this check to bypass the above? why are we doing it after it?

k8sutils/pkg/predicate/missing_instrumentation_config.go
RonFed, line 1: since this is specific to a single package - maybe we shouldn't place it in `k8sutils`?

k8sutils/pkg/predicate/missing_instrumentation_config.go
```diff
@@ -0,0 +1,70 @@
...
+		return false
+	}
+
+	sources, err := odigosv1.GetSources(context.Background(), p.Client, pw)
+	enabled, _, err := sourceutils.IsObjectInstrumentedBySource(context.Background(), sources, err)
+	if err != nil || !enabled {
+		return false
+	}
+
+	return true
```
RonFed: if we are doing cache calls and checking other objects (sources) maybe we should just inline this check into the reconciler body?

instrumentor/controllers/sourceinstrumentation/manager.go
```diff
@@ -47,14 +47,15 @@ func SetupWithManager(mgr ctrl.Manager, k8sVersion *version.Version) error {
 		return err
 	}
 
-	// Workload and Namespace reconcilers exist to catch the case where one of these entities is created
-	// after the Source that instruments it (because Sources can exist independently of entities).
-	// For that reason, we only watch for Create events on these controllers.
+	// Workload reconcilers catch workloads created after their Source, and updates while the
+	// workload's InstrumentationConfig is missing (for example after GitOps replace).
```
RonFed: the GitOps replace example deletes and creates the deployment according to the logs in the PR description - so we should have got a create event for the deployment with the existing predicate, no?

RonFed (changes requested):
This PR includes fixes in both the rollout logic and the source controller - I think the change in the rollout code in not behaving as expected (see comment there) - so maybe only the source controller is needed to change?
Conceptually, I also prefer to fix this in the source controller level - as its closer to the source of the problem (🥁 ) - and keep the rollout logic (which is complex) a bit simpler.
Maybe we shouldn't have delete the IC in the first place in the described scenario? since the namespace is instrumented? even thought we are using the owner reference and that deletion is not in "our control" as long as we keep this ownership.

instrumentor/controllers/sourceinstrumentation/manager.go
damemi: Yes but what I think happens is that Deployment create event is coming *before* the IC delete event (the flow chart doesn't show that well)

So it's like:

```
Deployment Delete ----------------->
Deployment Create --->Pod admit
                                 ---->IC Delete
```

That's the race, that the 2nd create squeezes in before the garbage collection has actually cascaded to the IC delete

damemi: At that point, there's nothing you can do to try to recover except delete and recreate the deployment again

k8sutils/pkg/predicate/missing_instrumentation_config.go
damemi: this isn't the only predicate in the predicates package that's currently used in one spot, I think it's good to keep them organized together so they are easy to find if someone does need it

instrumentor/controllers/agentenabled/rollout/rollout.go
damemi: the hasAgents check actually isn't a problem, though the logic there has a bug too because it doesn't consider terminating pods.

In this scenario imagine you have 1 pod that is already successfully instrumented with an `agents-hash`. You delete and recreate that deployment, now that successful pod is terminating while a new pod lands without a hash (because still needs runtime detection). (also this looks like a rolling deployment but is actually 2 different deployments due to delete+recreate)

The hasAgents check just goes off the workload selector of the deployment (which is the same when you delete+recreate even though it's 2 different deployments, but the bug would still be the same if it were a rolling deployment... that's just a different flaw I'm noticing now as I type this). So hasAgents sees the terminating pod with an agent and falls through past the `!hasAgents {rollout}` check

Now say there isn't a terminating pod that already has a hash to save you. The new pod lands without an agents hash, no IC, and we end up passing the `!hasAgents {rollout}` check to trigger a rollout. That actually helps the scenario by triggering a re-sync of the workload, except the sourceinstrumentation controller doesn't try to create IC on workload updates (only creates). That's the bug

There is a lot of edge-based logic and assumptions in the rollout code, this is one example of that because it's acting directly on what it sees as a delete. But there can actually be more reasons for an IC to be deleted that aren't a rollout. The attempt to do any rollout with `!hasAgents` is actually good reconcile logic, even though it was designed specifically for handling uninstrumentation it's really trying to kick off an event that brings the state back into balance. But the thing that could fix that state on that event wasn't listening to it.

k8sutils/pkg/predicate/missing_instrumentation_config.go
damemi: Yeah that's a good question, let me see if there is any benefit to filtering at the predicate level vs in the reconciler (feel like we've looked into it before but it was a long time ago)

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -438,10 +450,21 @@ func workloadHasOdigosAgents(ctx context.Context, c client.Client, obj client.Ob
 		return false, fmt.Errorf("workloadHasOdigosAgents: listing pods failed: %w", err)
 	}
 
-	// any non-empty list means the workload still runs instrumented pods.
 	return len(pods.Items) > 0, nil
 }
 
+func workloadStillMarkedForInstrumentation(ctx context.Context, c client.Client, pw k8sconsts.PodWorkload) (bool, error) {
+	sources, err := odigosv1alpha1.GetSources(ctx, c, pw)
```
blumamir, line 464: This is reading from the cache, which is exposed to race conditions as well.

k8sutils/pkg/predicate/missing_instrumentation_config.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	pw, err := workload.PodWorkloadFromObject(obj)
+	if err != nil {
+		return false
+	}
+
+	icName := workload.CalculateWorkloadRuntimeObjectName(pw.Name, pw.Kind)
+	ic := &odigosv1.InstrumentationConfig{}
+	err = p.Client.Get(context.Background(), client.ObjectKey{Namespace: pw.Namespace, Name: icName}, ic)
```
blumamir: Calling the cache from the predicate feels very very wrong to me.

This should be a simple call with no side effects and no race with the state of the cache

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -97,6 +98,17 @@ func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.Instrumentation
...
+		stillInstrumented, instrumentedErr := workloadStillMarkedForInstrumentation(ctx, c, pw)
+		if instrumentedErr != nil {
+			logger.Error(instrumentedErr, "failed to check if workload is still marked for instrumentation")
+			return RolloutResult{}, instrumentedErr
+		}
+		if stillInstrumented {
+			logger.Info("skipping uninstrumentation rollout - workload is still covered by an active source",
+				"workload", pw.Name, "namespace", pw.Namespace)
+			return RolloutResult{}, nil
+		}
```
blumamir, line 117: What if this source get's deleted after? what will trigger the rollback?

k8sutils/pkg/predicate/missing_instrumentation_config.go
```diff
@@ -0,0 +1,70 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	cr_predicate "sigs.k8s.io/controller-runtime/pkg/predicate"
+
+	odigosv1 "github.com/odigos-io/odigos/api/odigos/v1alpha1"
+	sourceutils "github.com/odigos-io/odigos/k8sutils/pkg/source"
+	"github.com/odigos-io/odigos/k8sutils/pkg/workload"
+)
+
+// MissingInstrumentationConfigPredicate allows update events when the workload's InstrumentationConfig
+// does not exist and the workload is still covered by an active Source.
```
blumamir: what if IC exists and terminating?
What will the predicate do and what will the reconciler do in this case?

From brief look at the code, it will try to delete it, but I can't figure out what will re-create it

blumamir (commented):
Suggesting an alternative which I think is simpler, and more robust for race conditions and edge cases, and also more idomatic to k8s:

how about we add reconciler in source controllers for instrumentation config deletion and call syncWorkload in this case to re-create it?

instrumentor/controllers/agentenabled/rollout/rollout.go
damemi: The race here is due to GC cascade between workloads+instrumentationconfigs due to their ownerreference releationship, Sources are meant to be long lived and that's not what we're racing against

k8sutils/pkg/predicate/missing_instrumentation_config.go
damemi: we don't *need* this predicate at all (or similar to what Ron said we could filter earlier) but something to gate on updates from workloads we actually care about

instrumentor/controllers/agentenabled/rollout/rollout.go
damemi: that will be another event and the entire flow will run back through and trigger the rollback

instrumentor/controllers/sourceinstrumentation/manager.go
```diff
@@ -119,7 +120,7 @@ func SetupWithManager(mgr ctrl.Manager, k8sVersion *version.Version) error {
 		ControllerManagedBy(mgr).
 		Named("sourceinstrumentation-instrumentationconfig").
 		For(&v1alpha1.InstrumentationConfig{}).
-		WithEventFilter(&odigospredicate.CreationPredicate{}).
+		WithEventFilter(predicate.Or(&odigospredicate.CreationPredicate{}, &odigospredicate.DeletionPredicate{})).
```
blumamir: you can use the [ExistencePredicate](https://github.com/odigos-io/odigos/blob/main/k8sutils/pkg/predicate/existence.go) :)

instrumentor/controllers/agentenabled/rollout/rollout.go
```diff
@@ -97,6 +98,20 @@ func Do(ctx context.Context, c client.Client, ic *odigosv1alpha1.Instrumentation
 			return RolloutResult{}, nil
 		}
 
+		// Just because an IC is nil, it doesn't mean the workload is not instrumented.
+		// The workload may be instrumented by a source (and the IC may just temporarily be missing)
+		// So we need to check if the workload is still marked for instrumentation.
```
blumamir, line 103: nit: i think it can be useful for future us to write how we observed this situation happening (with argo rollout sync event?) . that way, if we ever refactor or make changes, we have the full context (which is not a trivial one).

blumamir (approved):
LGTM thank you mike

instrumentor/controllers/agentenabled/rollout/rollout.go
damemi: Sure, I'll add some more detail here... but `git blame` will always point back to this PR and the linked linear :)

instrumentor/controllers/sourceinstrumentation/manager.go
damemi: thank you

PR #5022, opened 2026-06-10
CORE-1150: scrape the odiglets metrics via the data collection new

natyharary:
Does what it says on the tin. Sends odiglet metrics through the data collection to the gateway.

```release-note
Stream odiglet instrumentation metrics unto the victoria metrics via the cluster collector
```

### DIFF
--- a/autoscaler/controllers/nodecollector/collectorconfig/ownmetrics-victoriametrics.go
+++ b/autoscaler/controllers/nodecollector/collectorconfig/ownmetrics-victoriametrics.go
@@ -13,7 +13,12 @@ const (
 	odigletMetricsExporterName = "otlp_http/odiglet-metrics-out"
 	odigletMetricsPipelineName = "metrics/odiglet-metrics"
 
-	// Adds "odiglet" as a pod name so we can filter out the odiglet's own metrics in the UI
+	odigletJobName               = "odiglet"
+	odigletMetricsScrapeInterval = "10s"
+
+	// Stamps the data collection pod's own name as k8s.pod.name on the scraped odiglet metrics.
+	// The scrape carries no per-pod identity, so without this the series from every node's data would aggregate to be the same.
+	// The UI queries these metrics grouped and filtered by k8s.pod.name, matching against the data collection pod names.
 	odigletMetricsPodNameProcessorName = "resource/odiglet-pod-name"
 
 	// keep only the eBPF instrumentation counters (java, python, nodejs) exposed by the odiglet.
@@ -35,8 +40,8 @@ func odigletMetricsReceiverConfig(odigosNamespace string) config.GenericMap {
 			"config": config.GenericMap{
 				"scrape_configs": []config.GenericMap{
 					{
-						"job_name":           "odiglet",
-						"scrape_interval":    "10s",
+						"job_name":           odigletJobName,
+						"scrape_interval":    odigletMetricsScrapeInterval,
 						"enable_compression": false,
 						"static_configs": []config.GenericMap{
 							{
--- a/autoscaler/controllers/nodecollector/collectorconfig/ownmetrics-victoriametrics.go
+++ b/autoscaler/controllers/nodecollector/collectorconfig/ownmetrics-victoriametrics.go
@@ -1,0 +1,112 @@
+package collectorconfig
+
+import (
+	"fmt"
+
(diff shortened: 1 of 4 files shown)


autoscaler/controllers/nodecollector/collectorconfig/ownmetrics-victoriametrics.go
```diff
@@ -0,0 +1,112 @@
...
+		k8sconsts.OdigletMetricsServerPort,
+	)
+
+	return config.GenericMap{
+		odigletMetricsReceiverName: config.GenericMap{
+			"config": config.GenericMap{
+				"scrape_configs": []config.GenericMap{
+					{
+						"job_name":           "odiglet",
+						"scrape_interval":    "10s",
```
tamirdavid1: const?

autoscaler/controllers/nodecollector/collectorconfig/ownmetrics-victoriametrics.go
```diff
@@ -0,0 +1,112 @@
...
+						"enable_compression": false,
```
tamirdavid1: If the default is false we can drop it

autoscaler/controllers/nodecollector/collectorconfig/ownmetrics-victoriametrics.go
```diff
@@ -0,0 +1,112 @@
...
+	odigletMetricsPipelineName = "metrics/odiglet-metrics"
+
+	// Adds "odiglet" as a pod name so we can filter out the odiglet's own metrics in the UI
+	odigletMetricsPodNameProcessorName = "resource/odiglet-pod-name"
+
+	// keep only the eBPF instrumentation counters (java, python, nodejs) exposed by the odiglet.
+	ebpfInstrumentationMetricsRegexPattern = "odigos_(java|python|nodejs)_ebpf_instrumentation_.*"
+
+	// the cluster collector's own-metrics OTLP http receiver listens on this port.
+	clusterCollectorOwnMetricsOtlpHttpPort = 44318
```
tamirdavid1, line 28: it can be configured?

autoscaler/controllers/nodecollector/collectorconfig/ownmetrics-victoriametrics.go
```diff
@@ -0,0 +1,112 @@
...
+			},
+		},
+	}
+}
+
+func odigletMetricsProcessorConfig() config.GenericMap {
+	return config.GenericMap{
+		odigletMetricsPodNameProcessorName: config.GenericMap{
+			"attributes": []config.GenericMap{{
+				"key":    string(semconv.K8SPodNameKey),
```
tamirdavid1, line 89: why needed? we dont have it as part of the k8s attribute processor?

autoscaler/controllers/nodecollector/collectorconfig/ownmetrics-victoriametrics.go
natyharary: will move it

autoscaler/controllers/nodecollector/collectorconfig/ownmetrics-victoriametrics.go
natyharary: This is hardcoded in many places, so I guess the answer is no

autoscaler/controllers/nodecollector/collectorconfig/ownmetrics-victoriametrics.go
natyharary: This is meant for satisfying the UI's need for pod name, since we aggregate the metrics per pod name:
https://github.com/natyharary/odigos/blob/core-1150-scrape-the-odiglets-metrics-via-the-data-collection-new/frontend/services/metrics/utils.go#L22-L27

If we don't supply it, we won't be able to differentiate between different node collectors and it will make it useless in the UI.

Regarding the k8sattribute processor - it has some K8s API calls which are a bit wasteful here, since {POD_NAME} is already there.

autoscaler/controllers/nodecollector/collectorconfig/ownmetrics-victoriametrics.go
natyharary: it isnt

tamirdavid1 (approved):

PR #5028, opened 2026-06-10
CORE-1158 | feat: support applying actions to the profiles signal

shivanshuraj1333:

**Problem**

While looking at profiles samples in a customer env, we couldn't tell whether a sample came from an Odigos-instrumented source or another agent. 

**What changed**
Profiles now go through the **same pipeline-building path as traces / metrics /logs**

- **Same router topology.** using the `odigosrouterconnector` 

This PR includes the collector change (`odigosrouterconnector` profiles support) alongside the config generation, so the generated config and the collector that runs it.

```
GATEWAY — profiles built identically to traces / metrics / logs (minus `batch`):

  profiles/in        otlp ─► resource/odigos-version ─► [action processors] ─► odigosrouterconnector/profiles
                                                         (run ONCE at root)                 │ routes per data-stream
  profiles/<stream>  odigosrouterconnector/profiles ─► forward/profiles/<dest> …            ▼
  profiles/<dest>    forward/profiles/<dest> ─► otlphttp/<dest>     fan-out ─► Pyroscope-1, Pyroscope-2, Odigos-UI
```

```release-note
Support applying AddClusterInfo, RenameAttribute, and DeleteAttribute actions to the profiles signal.
```

### DIFF
--- a/common/config/processor.go
+++ b/common/config/processor.go
@@ -10,6 +10,7 @@ type CrdProcessorResults struct {
 	TracesProcessorsPostSpanMetrics []string
 	MetricsProcessors               []string
 	LogsProcessors                  []string
+	ProfilesProcessors              []string
 	Errs                            map[string]error
 }
 
@@ -21,6 +22,7 @@ func CrdProcessorToConfig(processors []ProcessorConfigurer) CrdProcessorResults 
 		TracesProcessorsPostSpanMetrics: []string{},
 		MetricsProcessors:               []string{},
 		LogsProcessors:                  []string{},
+		ProfilesProcessors:              []string{},
 		Errs:                            make(map[string]error),
 	}
 
@@ -54,6 +56,9 @@ func CrdProcessorToConfig(processors []ProcessorConfigurer) CrdProcessorResults 
 		}
 		if isLoggingEnabled(processor) {
 			results.LogsProcessors = append(results.LogsProcessors, processorKey)
+		}
+		if isProfilingEnabled(processor) {
+			results.ProfilesProcessors = append(results.ProfilesProcessors, processorKey)
 		}
 	}
 	if len(results.Errs) != 0 {
--- a/common/config/root.go
+++ b/common/config/root.go
@@ -127,22 +127,15 @@ func isProfilingEnabled(dest SignalSpecific) bool {
 	return isSignalExists(dest, common.ProfilesObservabilitySignal)
 }
 
-// addProfilesPipeline registers a direct profiles pipeline that consumes from
-// the gateway "otlp" receiver and exports via the supplied exporter.
-//
-// Profile pipelines deliberately bypass the data-stream router connector that
-// traces/metrics/logs flow through. The router connector and per-data-stream
-// pipelines do not yet support the profiles signal, so the OTel collector fans
-// the gateway OTLP receiver out to this pipeline directly. Callers must not
-// include the registered name in the destinationPipelineNames slice they
-// return from ModifyConfig; doing so would cause the gateway builder to wire
-// a forward connector that has no upstream producer.
-func addProfilesPipeline(currentConfig *Config, prefix, destID, exporterName string) {
+// addProfilesPipeline registers a "profiles/<prefix>-<id>" destination pipeline (exporter only) and
+// returns its name. Like traces/metrics/logs destination pipelines, its receiver (the forward
+// connector) is attached by the gateway builder, so it flows through the odigosrouterconnector.
+func addProfilesPipeline(currentConfig *Config, prefix, destID, exporterName string) string {
 	name := fmt.Sprintf("profiles/%s-%s", prefix, destID)
 	currentConfig.Service.Pipelines[name] = Pipeline{
-		Receivers: []string{"otlp"},
 		Exporters: []string{exporterName},
 	}
+	return name
 }
 
 func addProtocol(s string) string {
--- a/common/pipelinegen/config_builder.go
+++ b/common/pipelinegen/config_builder.go
@@ -142,8 +142,11 @@ func CalculateGatewayConfig(
 			pipeline := currentConfig.Service.Pipelines[pipelineName]
 			// add the forward connector as a receiver to the pipeline
 			pipeline.Receivers = append(pipeline.Receivers, connectorName)
-			// every destination pipeline should have a generic batch processor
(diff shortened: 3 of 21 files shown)


common/pipelinegen/config_builder.go
```diff
@@ -222,6 +222,18 @@ func CalculateGatewayConfig(
...
+	// processors directly onto every profiles pipeline (registered above by addProfilesPipeline and
+	// the applySelfTelemetry UI callback).
+	if len(processorsResults.ProfilesProcessors) > 0 {
+		for name, pipe := range currentConfig.Service.Pipelines {
+			if name == "profiles" || strings.HasPrefix(name, "profiles/") {
+				pipe.Processors = append(slices.Clone(processorsResults.ProfilesProcessors), pipe.Processors...)
+				currentConfig.Service.Pipelines[name] = pipe
+			}
+		}
+	}
```
blumamir: if there is more than one destination, we would now run the same processors independently on each pipeline, right?
It's ok for now to move fast, but eventually we should build a proper pipeline that process the profiles once and then fan out the data into all destinations.

common/pipelinegen/config_builder.go
```diff
@@ -222,6 +222,18 @@ func CalculateGatewayConfig(
 		}
 	}
 
+	// Profiles bypass the data-stream router, so there is no root pipeline: prepend the profiles
+	// processors directly onto every profiles pipeline (registered above by addProfilesPipeline and
+	// the applySelfTelemetry UI callback).
+	if len(processorsResults.ProfilesProcessors) > 0 {
+		for name, pipe := range currentConfig.Service.Pipelines {
+			if name == "profiles" || strings.HasPrefix(name, "profiles/") {
```
blumamir: nit: did you consider to pass the processor as parameter to the `addProfilingGatewayPipeline` function and to `applySelfTelemetry`, so it is created together with the pipeline and not added after? that makes it easier to follow the flow of code IMO.

common/config/processor.go
```diff
@@ -55,6 +64,11 @@ func CrdProcessorToConfig(processors []ProcessorConfigurer) CrdProcessorResults
 		if isLoggingEnabled(processor) {
 			results.LogsProcessors = append(results.LogsProcessors, processorKey)
 		}
+		if isProfilingEnabled(processor) {
+			if _, ok := profilesCapableProcessorTypes[processor.GetType()]; ok {
+				results.ProfilesProcessors = append(results.ProfilesProcessors, processorKey)
+			}
+		}
```
blumamir, line 62: why do we limit profiles only to `profilesCapableProcessorTypes`?
I guess someone can use it just like other signals to add any supported processor with any config they like if there is a need. 
wonder what different in profiles from others

blumamir (commented):
LGTM, one thing I think we should consider is why limit the processors for just those 2 options and not allow anything user might want to put there

common/pipelinegen/config_builder.go
shivanshuraj1333: Yeah I initially avoided it since the connector support is not there in the upstream, but fan out is something we can do, updated the PR with it now the profiles pipeline processes once and fans out to all destination exporters.

common/pipelinegen/config_builder.go
shivanshuraj1333: thanks, reorganized in the fan-out change

common/config/processor.go
shivanshuraj1333: it's by design, profiles is only supported by a subset of processors (experimental signal), so an unsupported pick would crash the collector at startup, and I don't want the user to configure things like T/M/L till it's stable in upstream.

the current implementation shall allow them to do most of the things that they may want to do

common/pipelinegen/config_builder.go
```diff
@@ -246,6 +253,28 @@ func CalculateGatewayConfig(
...
+	}
+	slices.Sort(exporters)
+	exporters = slices.Compact(exporters)
+	currentConfig.Service.Pipelines["profiles"] = config.Pipeline{
+		Receivers:  []string{"otlp"},
+		Processors: processors,
+		Exporters:  exporters,
+	}
+}
+
```
blumamir: I am highly against this. going to make maintenance and following the code way harder in the future.

Please create the pipeline in the way they should run, instead of creating them in the wrong way and them modifying the config like with calls like

```
delete(currentConfig.Service.Pipelines, name)
```

Also, what it the reason to make profiles different from other signals for building the pipelines?

common/config/processor.go
blumamir: We have the same issue for all other signals, nothing prevents a user from hooking up a processor that does not exists / not stable / not supporting the used signal.

Your suggestion carries a price - if we ever want to add a processor to profiles which is not in this list, it will not work. Also, the inconsistency between the signals will degrade odigos user experience as it's more things to be aware of.

I think we should keep it consistent.

common/pipelinegen/config_builder.go
shivanshuraj1333: so I was hesitant to use the `odigosrouterconnector` since connectors are alpha for profiles, but I validated and tested it and I think we are already alpha features from upstream.


Now profiling is following the T/M/L route only

common/config/root.go
```diff
@@ -137,12 +137,15 @@ func isProfilingEnabled(dest SignalSpecific) bool {
 // include the registered name in the destinationPipelineNames slice they
 // return from ModifyConfig; doing so would cause the gateway builder to wire
 // a forward connector that has no upstream producer.
-func addProfilesPipeline(currentConfig *Config, prefix, destID, exporterName string) {
+// addProfilesPipeline registers a "profiles/<prefix>-<id>" destination pipeline (exporter only) and
+// returns its name. Like traces/metrics/logs destination pipelines, its receiver (the forward
+// connector) is attached by the gateway builder, so it flows through the odigosrouterconnector.
+func addProfilesPipeline(currentConfig *Config, prefix, destID, exporterName string) string {
```
blumamir, line 133: the comment above is not incorrect:

> registers a direct profiles pipeline that consumes from the gateway "otlp" receiver

and others

common/pipelinegen/pipeline_builder.go
blumamir:
```suggestion
// BuildDataStreamPipelines constructs data stream pipelines for logs, metrics, traces, profiles
```

blumamir (approved):
👍

PR #5047, opened 2026-06-12
CORE-00 | Add Istio service mesh tracing integration guide

rayray221:
#### Summary

Adds documentation for sending Istio (Envoy) telemetry into Odigos so mesh spans and Odigos application spans share the same traces.

-New shared snippet docs/snippets/shared/instrumentations/advanced/istio.mdx, surfaced in both OSS and Enterprise docs via stub pages.
-Walks through configuring an Istio extensionProvider + Telemetry resource to export OTLP/HTTP to the node collector (odigos-data-collection-local-traffic:4318).
-Documents how Istio identifies the telemetry backend (port naming / appProtocol for protocol negotiation) and the validation steps to confirm spans are flowing.
-Calls out the current limitation that Envoy/mesh spans land in the default data stream (they lack Odigos workload attributes), so destinations must be on that stream to receive them.
-Includes an ASCII waterfall example showing how Istio client/server sidecar spans bracket each service's application work across a two-hop frontend → checkout → payments trace, illustrating on-the-wire vs. proxy overhead.

#### Navigation

-Registers the new pages under the Advanced instrumentation groups in docs/docs.json.

#### Notes

-Docs-only change; no code or runtime behavior affected.

```release-note
Add Istio service mesh tracing integration guide, so Envoy mesh spans join Odigos application traces.
```


docs/snippets/shared/instrumentations/advanced/istio.mdx
```diff
@@ -0,0 +1,151 @@
...
+<Tab title="Helm">
+Add the `meshConfig` block to the values for the `istiod` chart (for example, `istiod-values.yaml`):
+
+```yaml istiod-values.yaml
+meshConfig:
+  enableTracing: true
+  extensionProviders:
+    - name: odigos-otel
+      opentelemetry:
+        service: odigos-data-collection-local-traffic.odigos-system.svc.cluster.local
```
RonFed, line 48: Note that the local traffic service is a feature that works on k8s versions 1.26 or newer so might want to specify that?

I think we might be able to point to the data collection with local host - but that only works if it runs with hostNetwork: true

docs/snippets/shared/instrumentations/advanced/istio.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+<Note>
+This page covers how to point an **existing** Istio mesh at Odigos. It is not a guide to installing or operating Istio. For mesh setup, see the [Istio installation guide](https://istio.io/latest/docs/setup/install/helm/).
+</Note>
+
+## Prerequisites
+
+<Steps>
+<Step title="A running Istio mesh">
+Istio's control plane (`istiod`) must be installed in your cluster. Follow the [Istio installation guide](https://istio.io/latest/docs/setup/install/helm/). You do not need to enroll any workloads in the mesh yet.
```
tamirdavid1:
```suggestion
Istio's control plane (`istiod`) must be installed in your cluster. Follow the [Istio installation guide](https://istio.io/latest/docs/setup/install/helm/).
```

tamirdavid1: just feel a bit redundant, for your call

tamirdavid1 (approved):

docs/snippets/shared/instrumentations/advanced/istio.mdx
rayray221: Added notice for K8s minimum version

PR #5058, opened 2026-06-14
chore: autoscaler to support sending to abnormal service

tamirdavid1:

Autoscaler to add all relevant needed gateway configuration changes for the abnormal service

```release-note
NONE
```

emergency-ticket id: EMR-1879


api/k8sconsts/abnormal.go
```diff
@@ -0,0 +1,6 @@
+package k8sconsts
+
+const (
+	OdigosAbnormalServiceName = "odigos-security"
```
blumamir: how about `odigos-insignts`? or `odigos-trace-insights`?

helm/odigos/templates/odigos-configuration-cm.yaml
```diff
@@ -61,6 +61,10 @@ data:
       clientSecret: secretRef:odigos-oidc
       {{- end }}
     {{- end }}
+    {{- if .Values.security.enabled }}
+    security:
+      enabled: true
+    {{- end }}
```
blumamir, line 67: take from helm value...

autoscaler/controllers/clustercollector/abnormal_exporter.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/odigos-io/odigos/common"
+	"github.com/odigos-io/odigos/common/config"
+	pipelinegen "github.com/odigos-io/odigos/common/pipelinegen"
+)
+
+// addAbnormalGatewayExporter appends an OTLP gRPC exporter to the gateway's
+// root traces pipeline so every processed span fans out to the in-cluster
+// sidecar alongside the destination router. Noop when disabled or when no
+// root traces pipeline exists.
+func addAbnormalGatewayExporter(c *config.Config, odigosNs string, abnormal *common.AbnormalConfiguration) error {
```
blumamir: `AbnormalGatewayExporter` reads so strange in the code 😅

common/pipelinegen/config_builder.go
```diff
@@ -28,6 +28,12 @@ type GatewayConfigOptions struct {
 	TraceAggregationWaitDuration *string
 	SamplingDryRun               bool
 	SamplingSpanAttributes       *sampling.SpanSamplingAttributesConfiguration
+
+	// AbnormalEnabled is true when the optional side-channel exporter is
+	// wired onto the root traces pipeline. It shares the upstream
+	// groupbytrace processor with sampling v2 so a trace is assembled at
+	// most once on the gateway, regardless of which consumers asked for it.
+	AbnormalEnabled bool
```
blumamir: consider passing `odigoscommon.AbnormalConfiguration` here, that way if we ever what to reuse this feature for another platform, we can share the logic to add this pipeline generically

common/pipelinegen/config_builder.go
```diff
@@ -85,14 +91,25 @@ func CalculateGatewayConfig(
...
+	// groupbytrace is shared across consumers that need fully assembled traces
+	// on the root traces pipeline:
+	//   - sampling v2 (tail sampler can't decide on individual spans)
+	//   - the optional side-channel exporter tapped onto the root pipeline [odigos-abnormal]
+	// Tail sampling itself stays gated strictly on SamplingEnabled so that
+	// turning on the side-channel never starts dropping spans.
+	if gatewayOptions.OdigosConfigExtensionName != nil {
+		samplingOn := gatewayOptions.SamplingEnabled != nil && *gatewayOptions.SamplingEnabled
+		if samplingOn || gatewayOptions.AbnormalEnabled {
+			currentConfig.Processors[consts.GroupByTraceProcessorV2] = config.GenericMap{
```
blumamir, line 104: consider renaming this processor, which is now called `groupbytrace/samplingv2`. since it's not only for sampling now, we can call it just `groupbytrace`

common/pipelinegen/config_builder.go
```diff
@@ -85,14 +91,25 @@ func CalculateGatewayConfig(
...
+				"wait_duration": resolveTraceAggregationWaitDuration(gatewayOptions.TraceAggregationWaitDuration),
+			}
+			prefix := []string{consts.GroupByTraceProcessorV2}
```
blumamir: I think prefix is not the right name here, right? or am I missing something

common/pipelinegen/config_builder.go
```diff
@@ -85,14 +91,25 @@ func CalculateGatewayConfig(
...
+			if samplingOn {
+				currentConfig.Processors[consts.OdigosTailSamplingProcessorName] = getTailSamplingProcessorConfig(gatewayOptions)
+				prefix = append(prefix, consts.OdigosTailSamplingProcessorName)
```
blumamir: Do we want to send only sampled traces to abnormal? (how it currently wired)

common/pipelinegen/config_builder.go
```diff
@@ -85,14 +91,25 @@ func CalculateGatewayConfig(
...
+			}
+			processorsResults.TracesProcessors = append(prefix, processorsResults.TracesProcessors...)
```
blumamir: I suggest we do these after the group by (to reduce the things we store in memory) but before tail-sampling, since it might need url-templatization etc. 
currently they are placed after the url-templatization

common/pipelinegen/config_builder.go
```diff
@@ -468,7 +485,21 @@ func insertClusterMetricsResources(currentConfig *config.Config, odigosNs string
 	currentConfig.Service.Pipelines[rootPipelineName] = pipeline
 }
 
-func getTailSamplingProcessors(gatewayOptions *GatewayConfigOptions) ([]string, map[string]config.GenericMap) {
+// defaultTraceAggregationWaitDuration is used when groupbytrace is installed
+// without sampling v2 being enabled (e.g. only the side-channel exporter is
+// the consumer that needs assembly). When sampling v2 is on, the scheduler
+// validates and defaults this value, so this fallback is only hit on the
+// side-channel-only path.
+const defaultTraceAggregationWaitDuration = "10s"
```
blumamir: I think the default is 30s and it's written in helm, comments, docs etc. 
wondering what the motivation for this

blumamir (approved):
Approved with few comments.
Notice this now enables the feature by default instead of reading from values which is not what we want

PR #5093, opened 2026-06-18
CORE-00 | Docs: Add NGINX ingress OpenTelemetry tracing guide

rayray221:
## Summary
- Add an advanced-instrumentation page documenting how to configure an existing Ingress-NGINX Controller to export OpenTelemetry traces to Odigos via the node collector's OTLP/gRPC endpoint.
- Shared snippet (`snippets/shared/instrumentations/advanced/nginx.mdx`) imported by OSS and Enterprise stub pages, mirroring the Istio page structure.
- Register the new page under the "Advanced" instrumentation group for both editions in `docs.json`.

## Details
Covers prerequisites, controller configuration (Helm + ConfigMap), per-route enablement, validation (with real controller/collector error examples), an example trace timeline, tuning, and troubleshooting. Sampling guidance points to Odigos tail sampling rather than controller-side sampling.

```release-note
Added documentation for tracing NGINX ingress traffic with OpenTelemetry through Odigos.
```

### DIFF
--- a/docs/snippets/shared/instrumentations/advanced/nginx.mdx
+++ b/docs/snippets/shared/instrumentations/advanced/nginx.mdx
@@ -1,0 +1,191 @@
+Odigos can collect the distributed traces emitted by the [Ingress-NGINX Controller](https://kubernetes.github.io/ingress-nginx/) and route them through the same pipeline as your application traces. Once enabled, the controller reports a span for every request that enters the cluster through it — route, upstream, response code, and latency — over OTLP to the Odigos collector, where they are correlated with the spans Odigos already generates for your applications. The result is a single trace that begins at the edge and follows the request all the way through your services.
+
+<Note>
+This page covers how to point an **existing** Ingress-NGINX deployment at Odigos. It is not a guide to installing or operating the controller. For controller setup, see the [Ingress-NGINX installation guide](https://kubernetes.github.io/ingress-nginx/deploy/).
+</Note>
+
+<Note>
+Tracing happens at the **Ingress-NGINX Controller**, not in your application's nginx pods. Stock `nginx` images lack the OpenTelemetry module, but the controller bundles it and turns a few ConfigMap keys into the right OTel directives. Put the workloads you want traced behind an Ingress and enable OTel on the controller.
+</Note>
+
+## Prerequisites
+
+<Steps>
+<Step title="Kubernetes 1.26 or newer">
+The `odigos-data-collection-local-traffic` service depends on `internalTrafficPolicy: Local`, which became GA in 1.26.
+</Step>
+
+<Step title="A running Ingress-NGINX Controller">
+The Ingress-NGINX Controller must be installed in your cluster, with the OpenTelemetry module available. Follow the [Ingress-NGINX installation guide](https://kubernetes.github.io/ingress-nginx/deploy/).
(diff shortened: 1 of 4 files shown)


docs/snippets/shared/instrumentations/advanced/nginx.mdx
```diff
@@ -0,0 +1,191 @@
...
+  name: ingress-nginx-controller
+  namespace: ingress-nginx
+data:
+  enable-opentelemetry: "true"
+  otlp-collector-host: "odigos-data-collection-local-traffic.odigos-system.svc.cluster.local"
+  otlp-collector-port: "4317"
+  otel-service-name: "nginx-ingress"
+  opentelemetry-trust-incoming-span: "true"
+  opentelemetry-operation-name: "HTTP $request_method $uri"
+```
```
RonFed, line 71: we need both the helm command and the config map updates to be performed? isn't one causing the other? not sure just making sure

docs/snippets/shared/instrumentations/advanced/nginx.mdx
```diff
@@ -0,0 +1,191 @@
...
+</Tab>
+</Tabs>
+
+<Warning>
+**OTLP/gRPC only.** The Ingress-NGINX OTel module speaks OTLP gRPC on port **4317**. There is no HTTP/protobuf (4318) option — pointing it at 4318 will silently fail to export. The `odigos-data-collection-local-traffic` service accepts OTLP/gRPC on 4317.
+</Warning>
+
+<Tip>
+By default the controller samples every request through the ingress, so no sampler settings are needed. If you need to reduce volume, do it with [Odigos tail sampling](../../pipeline/sampling/head-and-tail-sampling) rather than at the controller — that way the decision sees the full trace (errors, latency, and downstream spans) instead of being made blindly per request at the edge.
```
RonFed, line 86: 💯

RonFed (commented):
Considering ingress-nginx is deprecated, should we consider:
1. mention this in our docs?
2. provide an equivalent solution to the alternative https://gateway-api.sigs.k8s.io/guides/getting-started/introduction/ ?

Another idea is to have some sort of integration of this with our UI - so maybe we can identify that nginx is instrumented and pointed to our pipeline and add it as an "external" source in the UI?

RonFed (approved):

docs/snippets/shared/instrumentations/advanced/nginx.mdx
rayray221: Correct, it's one or the other.  I updated the comment before to make it more explicit.

rayray221:
> Considering ingress-nginx is deprecated, should we consider:
> 
> 1. mention this in our docs?
> 2. provide an equivalent solution to the alternative https://gateway-api.sigs.k8s.io/guides/getting-started/introduction/ ?
> 
> Another idea is to have some sort of integration of this with our UI - so maybe we can identify that nginx is instrumented and pointed to our pipeline and add it as an "external" source in the UI?

I added a deprecation notice with a reference to Gateway API.  Challenge with adding an equivalent solution for gateway is that there are multiple implementations and I'm not sure which would make the most sense to provide an instrumentation guide for.  I love the last idea of adding this to the UI; we should do the same with Istio.

PR #5095, opened 2026-06-20
Instrumentor debug

blumamir:

Fixes: CORE-1272

Debugging instrumentor in vsc used to work, but it's not anymore. This PR fixes it so we can run the instrumentor locally and debug it with breakpoints and all that jazz.

Changes:

- k8sutils - check if we are running in the cluster or not, and adjust the kubeconfig setting accordingly (so components can also start when not `InClusterConfig`)
- add required env vars to instrumentor launch settings
- for debug - use metrics and health ports which are not common (8080/8081) and less likely to conflict
- add relevant files to gitignore

```release-note
NONE
```


.vscode/settings.json
```diff
@@ -7,9 +7,5 @@
         "Otel"
     ],
 
-    "go.buildTags": "embed_manifests",
-    "go.toolsEnvVars": {
-        "GOOS": "linux",
-        "GOARCH": "arm64"
```
RonFed, line 13: The Linux and arch fields help vscode to detect and reference Linux only files

blumamir: reverted

RonFed (approved):

PR #5106, opened 2026-06-23
CORE-00: add encoding to mock destination, configurable

natyharary:

Add encoding to mock destination, configurable.

```release-note
NONE
```

### DIFF
--- a/collector/exporters/mockdestinationexporter/config.go
+++ b/collector/exporters/mockdestinationexporter/config.go
@@ -14,10 +14,10 @@ import (
 type EncodingType string
 
 const (
-	// EncodingNone skips serialization entirely (the default, cheapest behavior).
-	EncodingNone EncodingType = "none"
-	// EncodingProto serializes telemetry to OTLP protobuf bytes, like most real destinations.
+	// EncodingProto serializes telemetry to OTLP protobuf bytes, like most real destinations. Default.
 	EncodingProto EncodingType = "proto"
+	// EncodingNone skips serialization entirely
+	EncodingNone EncodingType = "none"
 	// EncodingJSON serializes telemetry to OTLP JSON bytes.
 	EncodingJSON EncodingType = "json"
 )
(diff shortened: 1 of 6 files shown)


collector/exporters/mockdestinationexporter/config.go
```diff
@@ -10,6 +10,18 @@ import (
...
 )
 
+// EncodingType controls how the exporter serializes telemetry before "sending" it.
+type EncodingType string
+
+const (
+	// EncodingNone skips serialization entirely (the default, cheapest behavior).
+	EncodingNone EncodingType = "none"
+	// EncodingProto serializes telemetry to OTLP protobuf bytes, like most real destinations.
+	EncodingProto EncodingType = "proto"
```
blumamir, line 18: I suggest to make this the default, to make it more realistic

collector/exporters/mockdestinationexporter/exporter.go
```diff
@@ -41,17 +41,44 @@ func (e *MockDestinationExporter) Capabilities() consumer.Capabilities {
...
 }
 
 func (e *MockDestinationExporter) ConsumeLogs(ctx context.Context, logs plog.Logs) error {
+	switch e.config.Encoding {
+	case EncodingProto:
+		e.discardEncoded((&plog.ProtoMarshaler{}).MarshalLogs(logs))
+	case EncodingJSON:
+		e.discardEncoded((&plog.JSONMarshaler{}).MarshalLogs(logs))
+	}
 	return e.mockExport(ctx)
```
blumamir: discard the encoded bytes only after `mockExport` returns, so it matches what "real" exporters are doing

blumamir (approved):
approved but added suggestions to make it more accurate and better mock what real exporters are doing

collector/exporters/mockdestinationexporter/config.go
natyharary: Done

collector/exporters/mockdestinationexporter/exporter.go
natyharary: Also done

PR #5129, opened 2026-06-25
feat: add browser instrumentation

BenElferink:
## Summary

Adds **browser (web) instrumentation** to Odigos via an opt-in `browser-community` distribution and the `odigos-browser-proxy` sidecar.

Agent: https://github.com/odigos-io/opentelemetry-browser (`browser-community` **v0.2.0**)

Related PRs:
- Enterprise: https://github.com/odigos-io/odigos-enterprise/pull/3029
- UI Kit: https://github.com/odigos-io/ui-kit/pull/1035
- VM Agent: https://github.com/odigos-io/vm-agent/pull/789

```release-note
feat: add browser instrumentation
```

## Architecture

Because the OpenTelemetry Web SDK runs in the **end-user's browser**, Odigos cannot attach to an in-pod process. Instead it injects a sidecar in front of the HTML-serving container:

```mermaid
flowchart TD
    User["End-user browser"] -->|"GET / (HTML)"| SC["odigos-browser-proxy sidecar"]
    SC -->|"forward"| App["web server container<br/>(nginx/serve/etc)"]
    App -->|"HTML response"| SC
    SC -->|"inject script tag + recompress"| User
    User -->|"GET /__odigos/agent.js"| SC
    User -->|"POST /__odigos/v1/traces (OTLP)"| SC
    SC -->|"forward + CORS"| NC["node-local collector :4318"]
```

1. Init container installs an iptables redirect (app port → sidecar)
2. Sidecar injects `window.__ODIGOS__` + `<script src="/__odigos/agent.js">` into `text/html`
3. Sidecar serves the SDK bundle and relays OTLP/HTTP to the node-local collector (same-origin, no public collector / CORS config required)

Enable via Source `containerOverrides` → `otelDistroName: browser-community`.

## What's in this PR

- `browser-proxy/` — sidecar binary (HTML inject, gzip recompress, agent serve, OTLP relay)
- Instrumentor webhook injection + distro/runtime detection support
- Helm: `ODIGOS_BROWSER_PROXY_IMAGE` on instrumentor
- Release: `browser-proxy` added to publish / full-build / vuln-scan / ECR
- Docs: `docs/oss/instrumentations/browser/native.mdx`
- E2E: folded into existing `source` scenario (nginx `browser-frontend` + sidecar assert) — **not** a new matrix entry

## Review feedback addressed

| Feedback | Resolution |
|---|---|
| Separate e2e across 3 k8s versions | Folded into `tests/e2e/source` |
| `browser-proxy` missing from release | Added to publish-modules, full-build, vuln-scan, ECR |
| Browser vs Node mutually exclusive? | Yes for v1 (one distro/container); documented — split HTML vs API containers for both |
| `NODE_IP` / hostNetwork | Uses `LocalTrafficOTLPHttpDataCollectionEndpoint` (data-collection local-traffic on ≥1.26); clarified comments |
| Istio / mesh iptables collision | Unsupported with sidecar meshes; skip inject when `istio-proxy` / `linkerd-proxy` present |
| Sidecar crash / no probes | Liveness + readiness on `/__odigos/healthz` |
| HTML served uncompressed after gzip strip | Recompress gzip responses after injection |

Security topics (same-origin agent trust, unauthenticated OTLP path, CSP) are tracked for a follow-up design doc.

## Limitations (v1)

- Opt-in only (not auto-detected)
- One distro per container — `browser-community` replaces in-pod language agents on that container
- Not supported alongside Istio/Linkerd sidecars
- Init container needs `NET_ADMIN`

## Test plan

- [ ] `go test ./browser-proxy/...`
- [ ] `tests/e2e/source` — browser-frontend sidecar + init injection assert
- [ ] Manual: Source override → `browser-community`, confirm script injection + browser traces at destination
- [ ] Confirm `odigos-browser-proxy` image is published on release


tests/e2e/source/01-browser-frontend.yaml
RonFed, line 1: maybe instead of adding a new e2e (while will run across 3 k8s versions), we can add it to one of the existing ones? by extending simple demo, having one source for the browser instrumentation?

helm/odigos/templates/instrumentor/deployment.yaml
```diff
@@ -108,6 +108,8 @@ spec:
             {{- else }}
             value: {{ template "utils.imageName" (dict "Values" .Values "Component" "agents" "Tag" $imageTag) }}
             {{- end }}
+          - name: ODIGOS_BROWSER_PROXY_IMAGE
+            value: {{ template "utils.imageName" (dict "Values" .Values "Component" "browser-proxy" "Tag" $imageTag) }}
```
RonFed, line 112: do we need to add this proxy image to be part of our release? didn't see it was added

instrumentor/controllers/agentenabled/distroresolver/distroresolver.go
```diff
@@ -25,6 +25,14 @@ func resolveDistroByOverride(overwriteDistroName string, distroGetter *distros.G
...
 	}
 
+	// Browser (web) distros are delivered by a sidecar in front of whatever web server runs in the
+	// pod. The frontend nature of a workload cannot be reliably auto-detected from the in-pod
+	// process, so browser instrumentation is opt-in via this override and must work regardless of
+	// the detected in-pod language (nginx, javascript/node static server, unknown, ...).
+	if distro.BrowserSidecar != nil {
+		return distro, nil
+	}
+
```
RonFed, line 35: if a use pod is written in node js and it serves requests from a browser, we would have to choose either browser distro or js distro for the backend? we can't have both at the same time?

instrumentor/controllers/agentenabled/browser_proxy.go
```diff
@@ -0,0 +1,144 @@
...
+	agentDirResolved := strings.ReplaceAll(distroMetadata.BrowserSidecar.AgentDirectory, distro.AgentPlaceholderDirectory, k8sconsts.OdigosAgentsDirectory)
+
+	falsePtr := false
+	truePtr := true
+	runAsProxy := k8sconsts.BrowserProxyRunAsUser
+	runAsRoot := int64(0)
+
+	// The browser SDK exports to the sidecar over the application's own origin; the sidecar then
+	// forwards to the node-local collector reachable at $(NODE_IP):4318.
+	otlpEndpoint := service.LocalTrafficOTLPHttpDataCollectionEndpoint("$(NODE_IP)")
```
RonFed: I think using the node ip would only work if the user pod has `hostNetwork: true` - we can't add this to users pods since it has security implications, but maybe we can use the odiglet local service instead

instrumentor/controllers/agentenabled/browser_proxy.go
```diff
@@ -0,0 +1,144 @@
...
+		podswebhook.MountDirectory(&sidecar, distroMetadata.BrowserSidecar.AgentDirectory)
+		dirsToCopy[distroMetadata.BrowserSidecar.AgentDirectory] = struct{}{}
+		volumeMounted = true
+	case common.K8sVirtualDeviceMountMethod:
+		podswebhook.InjectDeviceToContainer(&sidecar, k8sconsts.OdigosGenericDeviceName)
+	}
+
+	// The init container installs the iptables redirect (inbound app port -> sidecar) before the
+	// application starts. It needs root + CAP_NET_ADMIN; the sidecar's own traffic (UID
+	// BrowserProxyRunAsUser) is excluded from the redirect so it can reach the app on loopback.
```
RonFed, line 138: it is very common in k8s to have tools like istio that do stuff like that, will this collide with them?

RonFed (commented):
wrote some general questions, didn't go into the proxy itself yet
I think we should wait for @blumamir to review this as well

tests/e2e/source/01-browser-frontend.yaml
BenElferink: Good call — folded this into the existing `source` e2e instead of a separate matrix scenario:

- Added a small nginx `browser-frontend` workload alongside simple-demo
- Source uses `containerOverrides` → `browser-community`
- Asserts sidecar + init injection in that same test
- Removed `tests/e2e/browser-instrumentation/`

So it rides the existing `source` scenario across the 3 k8s versions without adding a new matrix entry.

helm/odigos/templates/instrumentor/deployment.yaml
BenElferink: Yep, it was missing from the release path. Added `browser-proxy` to:

- `.github/workflows/publish-modules.yml` (publishes `odigos-browser-proxy`)
- `.github/workflows/full-build.yaml`
- `.github/workflows/vulnerability-scan.yaml`
- `Makefile` `publish-to-ecr`

Skipped RHEL for now — `browser-proxy/Dockerfile` is alpine/`iptables`-based and has no `rhel` target yet.

instrumentor/controllers/agentenabled/distroresolver/distroresolver.go
BenElferink: Correct for v1: one distro per container. Selecting `browser-community` replaces the in-pod language agent on that container (the webhook injects the sidecar and skips env/agent mounts).

To get both Node backend + browser HTML instrumentation, split them into separate containers (or workloads) and only override the HTML-serving one. Documented this under Requirements & notes in `docs/oss/instrumentations/browser/native.mdx`.

Supporting dual instrumentation on the same container is a follow-up (needs a separate browser flag alongside the language distro).

instrumentor/controllers/agentenabled/browser_proxy.go
BenElferink: Good catch on the misleading comment — fixed.

The code already uses `LocalTrafficOTLPHttpDataCollectionEndpoint("$(NODE_IP)")`, same helper as in-pod agents:
- k8s ≥ 1.26 → `http://odigos-data-collection-local-traffic.<ns>:4318` (`InternalTrafficPolicy=Local`) — no `hostNetwork` needed
- older clusters → falls back to `http://$(NODE_IP):4318`

`odiglet-local` is OpAMP (:4320), not OTLP — the right local service here is the data-collection one, which the helper already selects.

instrumentor/controllers/agentenabled/browser_proxy.go
BenElferink: Yes — same iptables REDIRECT pattern as Istio/Linkerd, so they would collide on the same ports.

For v1:
- Documented as unsupported with sidecar meshes
- Skip browser-proxy injection when `istio-proxy` / `linkerd-proxy` is already present on the pod (log + no-op)

Longer-term alternatives (CNI / ambient / explicit opt-in) can be a follow-up if we need mesh coexistence.

PR #5172, opened 2026-07-02
fix(frontend): surface auto-rollback status when injection is disabled (PLAT-1264)

BenElferink:
## Summary

When an instrumented workload enters `CrashLoopBackOff` and Odigos auto-rolls it back, the rollback condition and the **recover from rollback** button in the Edit Source drawer did not render — even though the workload was rolled back and needs recovery.

Root cause is the **ordering of checks** in `CalculateAutoRollbackStatus` (`frontend/graph/status/rollback.go`):

- Performing a rollback inherently disables agent injection — the rollout controller sets `Spec.AgentInjectionEnabled = false` alongside `Status.RollbackOccurred = true` (see `rollout_rollback_test.go`).
- The `!ic.Spec.AgentInjectionEnabled` check ran **before** the `ic.Status.RollbackOccurred` check, so a rolled-back source returned reason `AgentNotEnabled` with progress `Irrelevant`.
- The UI maps `Irrelevant` → `null` (`mapDesiredStatusToConditionStatus`), so `statuses-card.tsx` rendered nothing — no condition, no recover button.

## Fix

- Move the `Status.RollbackOccurred` check to the top of `CalculateAutoRollbackStatus` (right after the `ic == nil` guard) so it takes precedence over the config-disabled and injection-not-enabled checks. `RollbackOccurred` is a terminal state that requires user recovery and must surface regardless of the (necessarily disabled) injection state or a later config toggle.
- Add unit tests (`frontend/graph/status/rollback_test.go`) covering the ordering and the other status paths.

## Test plan

- [x] `go build ./graph/...`
- [x] `go vet ./graph/status/`
- [x] `go test ./graph/status/ -run TestCalculateAutoRollbackStatus` (8 cases pass)
- [ ] Manual: crashloop an instrumented workload so Odigos auto-rolls it back (`rollbackOccurred: true`, `agentInjectionEnabled: false`); confirm `autoRollback.autoRollbackStatus` returns reason `RollbackOccurred` / status `Notice` and the recover-from-rollback button renders in the Edit Source drawer overview.

Fixes PLAT-1264.

```release-note
fix(frontend): surface auto-rollback status when injection is disabled
```

cherry-pick: v1.28, v1.29, v1.30, v1.31


frontend/graph/status/rollback_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+func reasonOf(t *testing.T, s *model.DesiredConditionStatus) string {
+	t.Helper()
+	if s == nil || s.ReasonEnum == nil {
+		return ""
+	}
+	return *s.ReasonEnum
+}
+
+func TestCalculateAutoRollbackStatus_NilConfig_ReturnsNil(t *testing.T) {
```
natyharary: redundant - a nil config is a naive case, we shouldn't even calculate auto rollback status if we have a nil config

frontend/graph/status/rollback_test.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+}
+
+func TestCalculateAutoRollbackStatus_ConfigDisabled_NoRollback(t *testing.T) {
+	ic := &odigosv1alpha1.InstrumentationConfig{}
+	ic.Spec.AgentInjectionEnabled = true
+
+	got := CalculateAutoRollbackStatus(ic, &computed.AutoRollbackConfig{Enabled: false})
+
+	if reasonOf(t, got) != string(AutoRollbackReasonDisabled) {
```
natyharary: I'm not sure what this checks - it checks if we got the wrong reason? We could alternatively just check it we get the right reason.

either way, I think it is redundant since we just want to test that we see the recover button if the state is right

frontend/graph/status/rollback_test.go
```diff
@@ -0,0 +1,133 @@
...
+	ic.Spec.AgentInjectionEnabled = false
+
+	got := CalculateAutoRollbackStatus(ic, enabledConfig())
+
+	if reasonOf(t, got) != string(AutoRollbackReasonAgentNotEnabled) {
+		t.Fatalf("expected reason %q, got %q", AutoRollbackReasonAgentNotEnabled, reasonOf(t, got))
+	}
+}
+
+func TestCalculateAutoRollbackStatus_WaitingForRollout(t *testing.T) {
```
natyharary: same as above

frontend/graph/status/rollback_test.go
```diff
@@ -0,0 +1,133 @@
...
+	ic.Status.InstrumentationTime = nil
+
+	got := CalculateAutoRollbackStatus(ic, enabledConfig())
+
+	if reasonOf(t, got) != string(AutoRollbackReasonWaitingForRollout) {
+		t.Fatalf("expected reason %q, got %q", AutoRollbackReasonWaitingForRollout, reasonOf(t, got))
+	}
+}
+
+func TestCalculateAutoRollbackStatus_Evaluating_WithinStabilityWindow(t *testing.T) {
```
natyharary: I'm not sure how urgent is to check that evaluation within the stability window for this PR - we don't do any work related to that. Up to you if this test should remain or not

frontend/graph/status/rollback_test.go
```diff
@@ -0,0 +1,133 @@
...
+	ic.Status.InstrumentationTime = &now
+
+	got := CalculateAutoRollbackStatus(ic, enabledConfig())
+
+	if reasonOf(t, got) != string(AutoRollbackReasonEvaluating) {
+		t.Fatalf("expected reason %q, got %q", AutoRollbackReasonEvaluating, reasonOf(t, got))
+	}
+}
+
+func TestCalculateAutoRollbackStatus_Stable_AfterStabilityWindow(t *testing.T) {
```
natyharary: same as above

natyharary (changes requested):

frontend/graph/status/rollback_test.go
BenElferink: Removed. Agreed — nil IC is a naive guard case and not related to the recover-button fix.

frontend/graph/status/rollback_test.go
BenElferink: Removed. Trimmed the suite down to the two `RollbackOccurred` cases that assert we get reason `RollbackOccurred` + status `Notice` (what the UI needs to show the recover button).

frontend/graph/status/rollback_test.go
BenElferink: Removed as well — same rationale, out of scope for this PR.

frontend/graph/status/rollback_test.go
BenElferink: Dropped it. Not related to the ordering fix / recover button, so keeping the suite focused.

frontend/graph/status/rollback_test.go
BenElferink: Dropped as well for the same reason.

natyharary (approved):

PR #5184, opened 2026-07-05
feat: support query params in head sampling

blumamir:

Fixes: CORE-1149

Head sampling for http server currently supports route and method. this PR also add supports for query params,
we observed in a user environment, that the kubelet health probes also contained query params, which the sampler currently ignores.

The usecase is graphql endpoint, where there is one path `/graphql` and it can be called for health-check or for other queries depending on the `query` url param.

```release-note
Support query params for noisy operations sampling
```


common/api/sampling/common.go
```diff
@@ -0,0 +1,12 @@
...
+
+// QueryParamMatcher is a matcher for a query parameter in a url.
+// +kubebuilder:object:generate=true
+type QueryParamMatcher struct {
+
+	// name of the query parameter
+	Name string `json:"name"`
+
+	// value of the query parameter, that should be matched exactly
+	ValueExact string `json:"valueExact"`
```
RonFed:
```suggestion
	ValueExact string `json:"valueExact",omitempty`
```

RonFed (approved):

PR #5199, opened 2026-07-08
refactor: update cli-install make target arguments and add default Ja…

kushalShukla-web:
This PR updates the `cli-install` target in the `Makefile` to align with the new Helm-based implementation of the `odigos install` command. 

Previously, `make cli-install` relied on deprecated flags (`--version`, `--cluster-name`, `--central-backend-url`, and `--nowait`). These have been updated to use their standard Helm-equivalent flags:
- `--version` -> `--chart-version`
- `--cluster-name` -> `--set clusterName=`
- `--central-backend-url` -> `--set centralProxy.centralBackendURL=`

This resolves the `Error: unknown flag: --version` issue and allows contributors to successfully run `make cli-install` from local source code again.

```release-note
Fix `make cli-install` to use the updated Helm-based installation flags for Odigos CLI.

### CODE REVIEW

File: Makefile [hunk 1]
```diff
@@ -360,10 +360,9 @@ check-clean-work-tree:
 cli-install:
 	@echo "Installing odigos from source. version: $(ODIGOS_CLI_VERSION)"
 	cd ./cli ; go run -tags=embed_manifests . install \
-		--version $(ODIGOS_CLI_VERSION) \
-		--nowait \
-		$(if $(CLUSTER_NAME),--cluster-name $(CLUSTER_NAME)) \
-		$(if $(CENTRAL_BACKEND_URL),--central-backend-url $(CENTRAL_BACKEND_URL)) \
+		$(if $(ODIGOS_CLI_VERSION),--chart-version $(ODIGOS_CLI_VERSION)) \
```
damemi, line 363: does `ODIGOS_CLI_VERSION` make sense to use for this target?

tbh I'm not sure if this target is actually very useful anymore or not. The main pain that I usually have is if I make changes to the CLI (especially around the install command), trying to run that built CLI doesn't work because the helm chart expects a real version, and this gets defaulted to something like `v0.0.0` (don't remember exactly). So I can't actually test the install because those tags don't exist

@kushalShukla-web have you tested this target out?

PR #5207, opened 2026-07-11
CORE-1315: feat: implement rustc commit hash detection for runtime versioning 

kushalShukla-web:
This PR introduces the ability to detect the `rustc` commit hash from compiled Rust binaries. This improves our process discovery by allowing more accurate runtime versioning for Rust applications.

```release-note
Implement `rustc` commit hash detection for accurate Rust runtime versioning

### CODE REVIEW

**Comment by kushalShukla-web** [comment]
@damemi, could you please take a look at this PR when you have a chance? Thanks!

File: procdiscovery/pkg/inspectors/rust/rust.go [hunk 1]
```diff
@@ -10,6 +11,30 @@ import (
...
+// The standard library is compiled into virtually every Rust binary (it is
+// statically linked by default), and its panic/debug machinery embeds these
+// paths as string literals, so the hash survives even in release builds
+// without debug symbols.
+//
+// Unlike Go (which embeds a full semantic version, see debug/buildinfo),
+// rustc does not embed its own semver into the binary it produces, so the
+// commit hash is the most reliable runtime identifier available without
+// making an external call to resolve hash -> release version.
+var rustcCommitHashRe = regexp.MustCompile(`/rustc/([0-9a-f]{40})/`)
```
damemi: I'm not a rust developer so I'm curious, how useful is it for a rust user to see the commit hash vs an actual semantic version?

an external lookup is probably not useful, but depending on the rust release cadence maybe a hardcoded mapping for hash->version could be embedded (kind of similar to the Go offsets embedding) and automatically updated on rust releases

damemi (commented):
thanks for the PR @kushalShukla-web ! Our rust support is currently limited but if we can get the runtime version efficiently I think it would be a good improvement

procdiscovery/pkg/inspectors/rust/rust.go
kushalShukla-web: I agree that a semantic version would be much more useful than a commit hash. I started with the commit hash since it's directly available from the runtime, but I'll explore an embedded hash-to-version mapping similar to the Go offsets approach.

kushalShukla-web:
> thanks for the PR @kushalShukla-web ! Our rust support is currently limited but if we can get the runtime version efficiently I think it would be a good improvement

Added the hardcoded map! For future Rust releases, we'll likely also need a script to detect new release commits/tags and automatically update the mapping

damemi:
> > thanks for the PR @kushalShukla-web ! Our rust support is currently limited but if we can get the runtime version efficiently I think it would be a good improvement
> 
> Added the hardcoded map! For future Rust releases, we'll likely also need a script to detect new release commits/tags and automatically update the mapping

is this possible to do now? It could even run as a nightly check, but we would need some deterministic way to regularly update the version mapping that doesn't rely on manual code changes. For an example, I'd check out how we updated Go offsets: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/.github/workflows/offsets.yml

In that case, we generated the offsets as a JSON file using a custom tool/script, then embedded that with `go:embed`. Something like that might be easier

kushalShukla-web:
> > > thanks for the PR @kushalShukla-web ! Our rust support is currently limited but if we can get the runtime version efficiently I think it would be a good improvement
> > 
> > 
> > Added the hardcoded map! For future Rust releases, we'll likely also need a script to detect new release commits/tags and automatically update the mapping
> 
> is this possible to do now? It could even run as a nightly check, but we would need some deterministic way to regularly update the version mapping that doesn't rely on manual code changes. For an example, I'd check out how we updated Go offsets: https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/.github/workflows/offsets.yml
> 
> In that case, we generated the offsets as a JSON file using a custom tool/script, then embedded that with `go:embed`. Something like that might be easier

Hi @damemi i have updated code as per like you said.  i have tested rust_version.json file generation code and its working fine.

damemi:
Thanks @kushalShukla-web I like it, one kind of organizational note though the generator script is probably better in the [`scripts`](https://github.com/odigos-io/odigos/tree/main/scripts) directory than `procdiscovery`, the latter is meant to be a consumed package by other parts of the code.

also thinking about it now the rust versions file might be better as YAML, that might make it easier for your script to just append to the end of the list (instead of having to reformat the json)

procdiscovery/pkg/inspectors/rust/rust.go
```diff
@@ -49,6 +74,36 @@ func (n *RustInspector) DeepScan(pcx *process.ProcessContext) (common.Programmin
...
+
+	for _, section := range file.Sections {
+		// String literals live in allocated, non-executable data sections
+		// (e.g. .rodata, .data.rel.ro). Skip .text/.bss/etc to avoid
+		// unnecessary work scanning code or zero-initialized memory.
+		if section.Flags&elf.SHF_ALLOC == 0 || section.Flags&elf.SHF_EXECINSTR != 0 {
+			continue
+		}
+
+		data, err := section.Data()
```
RonFed: the `Data()` function, allocates a byte slice of the entire section, and can lead to big memory spikes
Do we have a way to make this a more focused search and less memory internsive?

RonFed (commented):
Thanks for the contribution.
This looks great!, left one comment about resource concern

kushalShukla-web:
I have made the requested changes based on the comments. @damemi @RonFed, could you please take another look?

RonFed (approved):

damemi (approved):
thanks @kushalShukla-web!

damemi:
@kushalShukla-web could you run `make go-mod-tidy` and push the updated go mod? https://github.com/odigos-io/odigos/actions/runs/29743513905/job/88357146864?pr=5207#step:5:38

kushalShukla-web:
Done

PR #5221, opened 2026-07-12
feat: add instrumentation config condition for pods injection

blumamir:

Fixes: CORE-1274

Currently we have the "rollout" condition, with the following issues:

1. it only tracks if auto rollout was applied. if it was not successful (for example, pods started without manifest injection), there is no condition to track it.
2. it is not updated with changes to the cluster. for example if rollout succeeded and then some time after pod started without manifest injection, it will not show up anywhere.
3. conditions are meant to be stateless and reconcilable. the "rollout" condition tracks "event" thus it is not a great fit for condition, which causes some edge cases and bugs.

```release-note
add condition to track pods injection and rollout status
```


status/generate/main.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/odigos-io/odigos/status"
+	"gopkg.in/yaml.v3"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const fileHeader = `// Code generated by "make -C status generate". DO NOT EDIT.
+
+`
+
+const fileTemplate = `package {{ .Package }}
```
RonFed: can be moved to a .tpl file

status/generate/main.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+{{- end }}
+)
+`
+
+type reasonData struct {
+	Name                    string
+	Message                 string
+	TechnicalDescription    string
+	K8sConditionStatusConst string
```
RonFed:
```suggestion
	K8sConditionStatus <correct type>
```

instrumentor/controllers/podsinjectionstatus/effectiveconfig_controller.go
```diff
@@ -0,0 +1,45 @@
...
+	"errors"
+
+	odigosv1 "github.com/odigos-io/odigos/api/odigos/v1alpha1"
+	"github.com/odigos-io/odigos/k8sutils/pkg/utils"
+	"github.com/odigos-io/odigos/k8sutils/pkg/workload"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type EffectiveConfigReconciler struct {
```
RonFed, line 16: write what changes in effective config are relevant

RonFed (approved):

PR #5231, opened 2026-07-13
feat: add verification logic for Cpp custom instrumentation probes

kushalShukla-web:
This PR adds verification logic for C++ custom instrumentation probes. It ensures that when users configure `Cpp` custom instrumentation rules, the configuration is properly validated using the `Verify()` method, preventing invalid rules from being processed by the controller. It also includes the necessary unit tests to cover this validation logic.

```release-note
Add validation for C++ custom instrumentation probes in instrumentation rules.


kushalShukla-web:
@damemi @RonFed This PR adds Verify() for CppCustomProbe. While doing this, I realized Cpp isn't wired up in the GraphQL schema or frontend/services/instrumentationrule.go yet (only Golang/Java/Php are) — so there's no way to actually create a Cpp probe from the UI/API right now.

Should I extend this PR to add that wiring too ?

kushalShukla-web:
Added CPP in InstrumentationRule — so that cpp custom instrumentation probes defined on an InstrumentationRule actually reach the agent config instead of being silently dropped.

PR #5238, opened 2026-07-14
CORE-1246 | feat(profiling): shared profile cache + gated ingest + configureProfilingCache

shivanshuraj1333:

Today the vm-agent and the k8s frontend each carry their own profiling byte-cache. This introduces one shared `common/profilecache` package, adopts it on the k8s side, and standardizes both platforms on the same cache design and API.

- **shared cache** (`common/profilecache`): per-source slots, LRU + idle-TTL eviction, byte-bounded buffers. Memory is bounded by `maxSlots × slotMaxBytes` — the VM-only global byte cap is removed so both platforms behave identically.
- **gated ingest**: the frontend only buffers *enabled* sources (`AddProfileData` no-ops until a slot is opened); the vm-agent now matches this.
- **`configureProfilingCache` mutation**: persists the cache limits (`profiling.ui`) and applies them live (prune → resize) with no UI-pod restart. The scheduler now propagates `profiling.ui` through the effective-config merge and the frontend watcher reconfigures the live store.
- per-source control resolvers unchanged (`enable/disable/clearSourceProfiling`, `profilingSlots`).

Companion PRs:
- agents-api — per-source control + flamegraph: https://github.com/odigos-io/agents-api/pull/36
- agents-api — `configureProfilingCache`: https://github.com/odigos-io/agents-api/pull/39
- vm-agent — adopt shared cache + gated ingest + `configureProfilingCache`: https://github.com/odigos-io/vm-agent/pull/841

```release-note
NONE
```


frontend/services/profiles/store.go
```diff
@@ -1,218 +1,15 @@
...
-		}
-	}
-}
-
-// StopCleanup stops the TTL cleanup goroutine
-func (s *ProfileStore) StopCleanup() {
-	if s.stopCleanup != nil {
-		s.stopCleanup()
-	}
+	return profilecache.NewStore(maxSlots, ttlSeconds, slotMaxBytes, cleanupInterval)
```
tamirdavid1: Why the NewProfileStore is not a common func? then the frontend/agent can run it from the common/profilecache `New` method.

common/profilecache/store.go
```diff
@@ -0,0 +1,310 @@
+// Package profilecache is the shared in-memory cache of raw OTLP CPU-profile
+// chunks used by both the Odigos frontend (buffering profiles for the flamegraph
+// UI) and the vm-agent (buffering profiles for the diagnostic bundle export).
+// Each source gets one byte-bounded buffer; sourceKey is an opaque string chosen
+// by the caller (namespace/kind/name in k8s, service.name on a VM).
```
tamirdavid1:
```suggestion
// Package profilecache is the shared in-memory cache of raw OTLP CPU-profile
// chunks.
// Each source gets one byte-bounded buffer; sourceKey is an opaque string chosen
```

common/profilecache/buffer.go
```diff
@@ -0,0 +1,127 @@
...
+func (b *BoundedBuffer) Resize(maxBytes int) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	b.maxBytes = maxBytes
+	b.trimLocked()
+}
+
+// trimLocked evicts oldest chunks until the buffer fits maxBytes, zeroing evicted
+// entries and compacting the backing slice so it does not grow without bound.
+func (b *BoundedBuffer) trimLocked() {
```
liranm-odigos, line 68: No mutex here? or you assume that the caller will always call this method with running mutex itself?

common/profilecache/store.go
```diff
@@ -0,0 +1,286 @@
...
+	RemoveSlot(sourceKey string)
+	ClearSlotBuffer(sourceKey string) bool
+	GetProfileData(sourceKey string) [][]byte
+	MaxSlots() int
+	ActiveSlots() (activeKeys []string, keysWithData []string)
+	MemoryStats() MemoryStats
+	Reconfigure(maxSlots, slotMaxBytes, ttlSeconds int)
+}
+
+func NewStore(maxSlots, ttlSeconds, slotMaxBytes int, cleanupInterval time.Duration) *Store {
```
liranm-odigos: nit; Consider adding a struct like StoreConfig instead of accepting many params

common/profilecache/store.go
```diff
@@ -0,0 +1,286 @@
...
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				s.SweepNow()
+			}
+		}
+	}()
```
liranm-odigos: No blocking?
RunCleanup can run in the background?

common/profilecache/buffer.go
shivanshuraj1333: It's a locked helper, both callers (AddAt, Resize) already hold b.mu. Added a doc line making that contract explicit.

liranm-odigos (approved):

PR #5246, opened 2026-07-15
feat(graphql): add instrumentations field to containers in workload queries

BenElferink:
```release-note
feat(graphql): add instrumentations field to containers in workload queries
```


alonkeyval (approved):

alonkeyval:
need also uikit upgrade

PR #5264, opened 2026-07-16
RUN-146: Add PHP support for Custom Instrumentation rules

BenElferink:
Need to bump ui-kit with: https://github.com/odigos-io/ui-kit/pull/1110

---

Screenshot 2026-07-22 at 14 35 58

Screenshot 2026-07-22 at 14 36 32

---

Adds PHP as a supported language for the Custom Instrumentation rule.

Inspired by the config-driven hook approach in https://github.com/newfold-labs/otel-php-instrumentation, this wires PHP probes through the Odigos control plane. The PHP agent registrar lives in a companion PR on `odigos-io/opentelemetry-php`.

**Changes:**
- Add `PhpCustomProbe` (`className` optional, `functionName` required) to `CustomInstrumentations`
- Regenerate CRDs / deepcopy; expose PHP in GraphQL, UI rule catalog, and docs
- Merge PHP probes into container agent config
- Mark `php-community` distro as supporting custom instrumentation with `configAsEnvVars: true`
- Inject `ODIGOS_AGENT_CUSTOM_INSTRUMENTATIONS` JSON into PHP pods via the instrumentor webhook

**Usage example:**
```yaml
kind: InstrumentationRule
metadata:
  name: php-custom-instrumentation
  namespace: odigos-system
spec:
  ruleName: "Instrument OrderService::processOrder"
  customInstrumentations:
    php:
      - className: "App\\Service\\OrderService"
        functionName: "processOrder"
```

PHP rule changes require a workload restart (env-var delivery).

**Companion PRs:**
- `odigos-io/opentelemetry-php` — agent-side hook registrar
- `odigos-io/odigos-enterprise` — MCP tool support for PHP probes

Rebased onto latest `main` (resolved conflicts from deprecated `sdkConfigs` removal).

#### Changelog entry:
```release-note
Add PHP support for Custom Instrumentation rules (class/function probes via ODIGOS_AGENT_CUSTOM_INSTRUMENTATIONS).
```


common/api/instrumentationrules/custom_instrumentation.go
```diff
@@ -111,3 +124,26 @@ func (c *CppCustomProbe) Verify() error {
...
 	return nil
 }
+
+// PhpCustomProbe contains the details for a custom probe for PHP applications.
+// className is optional; when empty, functionName refers to a global function.
+// +kubebuilder:object:generate=true
+// +kubebuilder:deepcopy-gen=true
+type PhpCustomProbe struct {
+	ClassName    string `json:"className,omitempty" yaml:"className,omitempty"`
+	FunctionName string `json:"functionName" yaml:"functionName"`
```
RonFed, line 128: I'm not familiar with PHP so a few questions:
1. Can an app have 2 classes with the same name in different packages?
2. does it have overloading of functions? i.e can a class have few methods with the same name, but with different arguments?

instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -416,6 +416,17 @@ func (p *PodsWebhook) injectOdigosToContainer(containerConfig *odigosv1.Containe
...
 
+	// Custom Instrumentation configuration (native agents that receive config via env vars)
+	customInstrumentationsEnabled := containerConfig.Traces != nil && containerConfig.Traces.CustomInstrumentations != nil
+	supportsCustomInstrumentations := distroMetadata.Traces != nil && distroMetadata.Traces.CustomInstrumentations != nil && distroMetadata.Traces.CustomInstrumentations.Supported
+	if customInstrumentationsEnabled && supportsCustomInstrumentations && distroMetadata.ConfigAsEnvVars {
+		customInstrumentationsConfigJson, err := json.Marshal(containerConfig.Traces.CustomInstrumentations)
+		if err != nil {
+			return false, nil, fmt.Errorf("failed to marshal custom instrumentations config: %w", err)
+		}
+		existingEnvNames = podswebhook.InjectConstEnvVarToPodContainer(existingEnvNames, podContainerSpec, "ODIGOS_AGENT_CUSTOM_INSTRUMENTATIONS", string(customInstrumentationsConfigJson))
```
RonFed: can we have a dedicated const for `ODIGOS_AGENT_CUSTOM_INSTRUMENTATIONS`? I also think we should make this env var PHP specific to avoid possible confusions, so something like `ODIGOS_PHP_AGENT_CUSTOM_INSTRUMENTATIONS`

common/api/instrumentationrules/custom_instrumentation.go
BenElferink: Good questions!

1. **Same class name in different namespaces:** Yes. PHP uses namespaces, so two classes can share a short name in different namespaces (e.g. `App\Service\OrderService` vs `Vendor\Other\OrderService`). That’s why `className` is the fully qualified class name (FQCN), including the namespace — matching how PHP resolves classes at runtime.

2. **Method overloading:** No. PHP does not support traditional overloading (multiple methods with the same name but different argument lists). A class can only declare one method with a given name, so `className` + `functionName` uniquely identifies the target without needing signature/arity fields.

instrumentor/controllers/agentenabled/pods_webhook.go
BenElferink: Good call - done in 41b5e38f4:

- Renamed to the PHP-specific `ODIGOS_PHP_AGENT_CUSTOM_INSTRUMENTATIONS`
- Extracted as `k8sconsts.OdigosPhpAgentCustomInstrumentationsEnvVar`
- Updated the OSS docs accordingly
- Updated the PHP agent (`odigos-io/opentelemetry-php`) to read the new env var name

frontend/services/instrumentationrule.go
```diff
@@ -434,6 +434,20 @@ func getCustomInstrumentationsInput(input model.InstrumentationRuleInput) *instr
 		}
 	}
 
+	if input.CustomInstrumentations.Php != nil {
+		customInstrumentations.Php = make([]instrumentationrules.PhpCustomProbe, 0, len(input.CustomInstrumentations.Php))
+		for _, probe := range input.CustomInstrumentations.Php {
+			apiProbe := instrumentationrules.PhpCustomProbe{}
```
aviadhahami, line 440: but what if we have only one of the values class/func?

distros/yamls/php-community.yaml
```diff
@@ -33,3 +33,8 @@ spec:
       - '{{ODIGOS_AGENTS_DIR}}/php'
     device: 'instrumentation.odigos.io/generic'
     k8sAttrsViaEnvVars: true
+  traces:
+    customInstrumentations:
+      supported: true
```
aviadhahami, line 38: why this key?

common/api/instrumentationrules/custom_instrumentation.go
```diff
@@ -31,6 +32,18 @@ func (ci *CustomInstrumentations) Verify() error {
 			return fmt.Errorf("invalid configuration for java custom instrumentation: %w", err)
 		}
 	}
+	// Validate C++ probes
+	for _, p := range ci.Cpp {
```
aviadhahami: what are those?

frontend/services/instrumentationrule.go
BenElferink: Good catch.

- **Only `functionName`** → valid (global PHP function; `className` is optional)
- **Only `className`** → invalid (`functionName` is required by `PhpCustomProbe.Verify()`)

Previously the GraphQL convert path didn’t call `Verify()`, so a class-only probe could be persisted. Fixed in 130288a19 — create/update now runs `CustomInstrumentations.Verify()` and returns an error for incomplete probes.

distros/yamls/php-community.yaml
BenElferink: This is the distro capability flag. It tells the control plane that `php-community` supports custom instrumentation probes (`distro.Traces.CustomInstrumentations.Supported`).

Without it, `DistroSupportsCustomInstrumentations()` / the pods webhook would skip PHP for this feature, so probes wouldn’t be merged into agent config or injected via `ODIGOS_PHP_AGENT_CUSTOM_INSTRUMENTATIONS`.

common/api/instrumentationrules/custom_instrumentation.go
BenElferink: Those are the existing `CppCustomProbe` entries (VM Agent C++ custom instrumentation, added in #4865). `CppCustomProbe.Verify()` already existed, but `CustomInstrumentations.Verify()` never iterated C++ probes.

While adding the PHP validation loop here, I also filled that gap so C++ probes get validated the same way as Go/Java/PHP. Happy to split that into a separate PR if you’d rather keep this one PHP-only.

aviadhahami (approved):

PR #5270, opened 2026-07-18
Pods manifest injection texts

blumamir:

Fixes: CORE-1291

experimenting with a new concept in how we manage our statues in k8s.
This is so we can show more useful texts and relevant state in ui source drawer.

Challenge: the statuses are now used to

1. write conditions on objects in k8s
2. show things in the ui
3. (future effort): generate docs to explain what they mean

I want to consolidate it all into one yaml format, which we can use to write statuses uniformly, and make any changes like texts and new cases, just in one place.

It also allows:

- setting parameters and templatizing the messages (currently with workload kind, in the future with relevant times, languages, distro names, etc as we see fit)
- repeating multiple states for the same condition (like enabled and disabled) so each one get's it's own text, but all the other common things does not have to be repeated
- declaring action items which ui can consume (and later we can show in docs or popups or some other aggregated way)

This PR improves previous work for the first status which is PodsManifestInjection, which is a bundle of the rollout and the check if pod manifest is injected in all pods (e.g. if not, then what state the rollout is in?)

It manages much more corner cases for this flow, and also replaces some bugs or unhandled paths in the rollout mechanism, by a different one that branches more cases.

Graphql endpoints are updated with more relevant rollout related status to show in UI.

```release-note
Improve text in UI for showing rollout status and if all pods are applied with an agent or not
```

cherry-pick: v1.32


status/instrumentationconfig/podsmanifestinjection.yaml
```diff
@@ -2,134 +2,286 @@ apiVersion: internal.odigos.io/v1beta1
...
+      It may be an Odigos automatic rollout, a manual rollout triggered from Odigos, or an unrelated rollout that happens to be running.
+      Until the rollout finishes, some pods may still be out of date; afterward, all running pods should match.
+    states:
+    - state: "enabled"
+      title: "Rollout: In Progress"
+      message: "Instrumentation Enabled; Agent will be applied to all pods once {{ .WorkloadKind }} rollout is completed"
+      summary: "After the rollout completes, new pods should start with the instrumentation agent applied."
+    - state: "disabled"
+      title: "Rollout: In Progress"
+      message: "Instrumentation Disabled; Some pods have agent applied, new pods after {{ .WorkloadKind }} rollout will start without it"
```
RonFed: align messages in 3 states (complete instead of start rollout)

RonFed (approved):

RonFed (approved):

PR #5302, opened 2026-07-20
CORE-1307 | fix: cap the native symbolizer's transient symbol-table decode

shivanshuraj1333:

The native symbolize processor caps its **retained** symbol cache (256 MiB, LRU), but had no bound on the **transient** parse. Go's `elf.File.Symbols()` materialises the entire `.symtab` + string table before we filter to `STT_FUNC`, so on a large unstripped binary (Oracle/C++) that transient which is what spikes RSS. 

#### Where this fits (CORE-1307)

```
node-collector profiles pipeline — memory guards

  profiles in
     |
     v
  [ memory_limiter ]   (1) pipeline gate: reject under memory pressure    #5304
     |
     v
  filter > k8sattributes > odigosprofiles
     |
     v
  [ symbolize ]        (2) transient .symtab decode cap   max_symtab_bytes  #5302  <-- THIS PR
     |                 (3) retained symbol LRU cache       max_symbol_bytes  #5307
     v                      one knob -> maxMemoryMiB: symbol=B, symtab=B/5
  service-name > export
```

#### Changelog entry:

```release-note
odigossymbolizeprocessor: cap the transient ELF symbol-table decode.
```
---

### DIFF
--- a/collector/processors/odigossymbolizeprocessor/config.go
+++ b/collector/processors/odigossymbolizeprocessor/config.go
@@ -17,6 +17,9 @@ type Config struct {
 	// MaxSymbolBytes caps the total bytes of parsed symbols held across all cached
 	// binaries (LRU eviction by memory — the real guard at scale). 0 = default.
 	MaxSymbolBytes int64 `mapstructure:"max_symbol_bytes"`
+	// MaxSymtabBytes caps the transient decode: a symbol table larger than this on
+	// disk is skipped without decoding (the peak guard). 0 = default.
+	MaxSymtabBytes int64 `mapstructure:"max_symtab_bytes"`
 	// MaxMapsCache caps cached per-pid /proc maps (LRU). 0 = default.
 	MaxMapsCache int `mapstructure:"max_maps_cache"`
 	// MapsTTLSeconds is how long a cached /proc/<pid>/maps is reused. 0 = default.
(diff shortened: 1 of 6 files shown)


damemi (approved):
@shivanshuraj1333 what's the effective outcome of skipping over tables that exceed the limit? Does it mean we essentially can't profile/visualize that app? If so that should probably be logged or materialized somewhere so users don't just wonder what's happening

shivanshuraj1333:
Right now there isn't a way to properly fix this, we can, but we're fundamentally limited by how much memory we can afford to spend on symbolization on a busy node.

If a binary's symbol table doesn't fit the configured budget, we genuinely can't symbolize it, and that's a real tradeoff we're choosing to accept rather than risk an OOM.


On separate note, to solve such challenges, exactly why we started this whole design (memory_limiter + the configurable budget + this transient cap), to make that tradeoff explicit and tunable instead of an uncontrolled memory spike.

More context on the thinking here: https://keyval.slack.com/archives/C05F343GJTG/p1783425554785909

---

Though your point is stilll valid and what we can fix is the silence around it, pushed a change so `readFunctionSymbols` reports whether a table was skipped for size (vs. genuinely absent), and `parseAndCache` logs a `Warn` (not `Debug`, so it's visible by default) naming the path and the configured `max_symtab_bytes` whenever that happens, so at least users aren't left wondering why an app isn't symbolizing.

PR #5307, opened 2026-07-20
CORE-1307 | feat: make the native symbolizer memory budget configurable

shivanshuraj1333:
#### What this PR does:

Bounds the node-collector symbolize processor's memory. Adds `profiling.symbolization.maxMemoryMiB` as a single budget; the processor derives both its internal guards from it, so they can't be set out of relation to each other:

- retained symbol-cache LRU byte budget — gets the full amount
- transient `.symtab` decode gate (per binary) — gets `budget / symtabDecodeCostFactor`

`symtabDecodeCostFactor` used to be an unverified `5`. Re-measured it directly (`runtime.MemStats` around the actual decode call, against real ELF binaries from 132 KiB to 8.7 MiB of symtab): real ratio is 2.65x-3.3x. Set to `4`. Zero-config default (256 MiB) now derives its transient cap from the same formula instead of an unrelated hardcoded 512 MiB.

Also: a binary skipped for exceeding the decode gate no longer logs a `Warn` per binary (storm risk on a busy node) — moved to `Debug`, with one aggregated `Warn` per sweep.

A second `odigossymbolizeprocessor` instance is still reachable via the Processor CRD's raw `processorConfig`, with its own independent `max_memory_bytes`. The pipeline now warns when that's present, naming the instance.

#### Where this fits (CORE-1307)

```
node-collector profiles pipeline

  profiles in -> memory_limiter (process-wide gate, #5304)
             -> filter -> k8sattributes -> odigosprofiles
             -> symbolize (this PR: one knob, measured ratio)
             -> service-name -> export
```

#### Changelog:

```release-note
Add profiling.symbolization.maxMemoryMiB to bound the node-collector symbolizer's memory.
```


autoscaler/controllers/common/profiling_otlp.go
```diff
@@ -118,10 +118,20 @@ func OdigosProfilesProcessorConfig() config.GenericMap {
...
 // (interpreted runtimes, Go) pass through. Defaults to bounded caches and async parsing.
-func OdigosSymbolizeProcessorConfig() config.GenericMap {
-	return config.GenericMap{}
+func OdigosSymbolizeProcessorConfig(symbolization *odigoscommon.ProfilingSymbolizationConfiguration) config.GenericMap {
+	cfg := config.GenericMap{}
+	// One budget bounds both caps; unset → the processor defaults apply.
+	if symbolization != nil && symbolization.MaxMemoryMiB != nil && *symbolization.MaxMemoryMiB > 0 {
+		budget := int64(*symbolization.MaxMemoryMiB) << 20
+		cfg["max_symbol_bytes"] = budget
+		cfg["max_symtab_bytes"] = budget / symtabDecodeCostFactor
```
damemi: what if `max_symtab_bytes` is explicitly set to a different value?

damemi (approved):
this seems more usable than the `max_symtab_bytes` config added in #5302, and in my comment it's not clear why have both or what happens if both are manually set. Maybe consider just reducing to this one

PR #5308, opened 2026-07-20
feat: add 2 new actions for db sql processing

blumamir:

Fixes: RUN-909

Add the following 2 actions:
- dbQueryTemplatization (similar to url templatization). remove high-cardinality or possible PII data from sql query texts
- inferDbAttributes

```release-note
Add 2 new actions to control db query processing behavior
```


api/odigos/v1alpha1/actions/inferdbattributes.go
```diff
@@ -0,0 +1,36 @@
...
+type InferDbAttributesConfig struct {
+	// the scope of services for which this config will be applied.
+	// if empty, the provided config will be applied to all sources.
+	Scopes *k8sconsts.SourcesScopes `json:"scopes,omitempty"`
+
+	actionsapi.InferDbAttributesConfig `json:",inline"`
+}
+
+func (InferDbAttributesConfig) ProcessorType() string {
+	return "odigossqlquery"
```
RonFed: nit: can use a const (guess autoscaler will use it too)

common/api/actions/db_query_templatization.go
```diff
@@ -0,0 +1,23 @@
+package actions
+
+// DbQueryTemplatizationConfig is the per-container collector config for
```
RonFed: since this is in the common package maybe "per-container" is not accurate here?

common/api/actions/db_query_templatization.go
```diff
@@ -0,0 +1,23 @@
...
+	// - number literals
+	// - string literals
+	// what is not templatized:
+	// - identifiers (e.g. table names, column names, etc.)
+	// - positional parameters (e.g. "SELECT * FROM users WHERE id = $1")
+	// - bind parameters (e.g. "SELECT * FROM users WHERE id = :name")
+	// - boolean literals
+	// - NULL
+	// - keywords, operators, punctuation, whitespace, comments, etc.
+	TemplatizeLiterals bool `json:"templatizeLiterals,omitempty"`
```
RonFed, line 19: if we would want to expand this in the future, this flag might conflict with possible future configs?
for example only template literals after `password=`

common/api/actions/db_query_templatization.go
blumamir: updated the comment

RonFed (approved):

PR #5310, opened 2026-07-20
refactor: pii masking processor

blumamir:

Fixes: RUN-909

write our own pii masking processor, to use instead of the redaction processor from contrib which is not a best fit for us:

- only handles attributes which are strings (bypass the http header attributes which are strings array)
- regex mask everything it match, which is not what people usually want (the want to match something like "password=1234" and replace just the value.
- it has tons of configuration options and features which are overkill to the simple case we use.
- we can add code to masking categories where regex cannot be used

```release-note
NONE
```


autoscaler/controllers/actions/piimasking_sync.go
```diff
@@ -0,0 +1,146 @@
...
+
+// SyncPiiMaskingProcessor creates, patches, or deletes the shared PII-masking Processor from Actions.
+// Categories from all non-disabled PiiMasking actions are unioned into a single processor.
+// When no enabled actions remain, the shared Processor is deleted.
+// Legacy per-action Processor CRs (named after the Action) are also removed.
+func SyncPiiMaskingProcessor(ctx context.Context, c client.Client) error {
+	logger := commonlogger.FromContext(ctx).WithName("pii-masking")
+	ns := env.GetCurrentNamespace()
+
+	if err := deleteLegacyPiiMaskingProcessors(ctx, c, ns); err != nil {
```
natyharary, line 96: Nice

autoscaler/controllers/actions/piimasking_sync.go
```diff
@@ -0,0 +1,146 @@
...
+		TypeMeta: metav1.TypeMeta{APIVersion: "odigos.io/v1alpha1", Kind: "Processor"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      consts.PiiMaskingProcessorName,
+			Namespace: namespace,
+		},
+		Spec: odigosv1.ProcessorSpec{
+			Type:            cfg.ProcessorType(),
+			ProcessorName:   "PII Masking",
+			Disabled:        false,
+			Signals:         []common.ObservabilitySignal{common.TracesObservabilitySignal},
```
natyharary, line 80: we probably should support other signals as well. logs are classic for pii masking for example

natyharary: follow up: I see that we do support multiple signals:
https://github.com/odigos-io/odigos/pull/5310/changes#diff-234c8c0646b313dc0db812e83bbf27322c7cfa0c2921cda68e4a6e69d8144cefR246-R258

will this not always create a ProcessorSpec with traces only?

collector/processors/odigospiimaskingprocessor/factory.go
```diff
@@ -0,0 +1,48 @@
...
+		createDefaultConfig,
+		processor.WithTraces(createTracesProcessor, metadata.TracesStability),
+	)
+}
+
+func createDefaultConfig() component.Config {
+	return &Config{}
+}
+
+func createTracesProcessor(
```
natyharary, line 31: same here - do we plan to also support logs?

natyharary (commented):
looks good - having some thoughts about multiple signals

collector/processors/odigospiimaskingprocessor/mask.go
```diff
@@ -0,0 +1,59 @@
...
+	},
+	actions.UuidMasking: {
+		maskedValue: "***UUID***",
+		patterns: []*regexp.Regexp{
+			regexp.MustCompile(`[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}`),
+		},
+	},
+}
+
+func maskCategory(category actions.PiiCategory, value string) (string, bool) {
```
tamirdavid1, line 43: are we giving the user the ability to create it's own regex?

collector/processors/odigospiimaskingprocessor/metadata.yaml
```diff
@@ -0,0 +1,8 @@
+type: odigospiimasking
```
tamirdavid1, line 1: idea - add metrics for each category so we can check how much we actually masked?

common/consts/consts.go
```diff
@@ -153,6 +153,8 @@ const (
 	OdigosTraceStateProcessorName         = "odigostracestate"
 	URLTemplatizationProcessorName        = "odigos-url-templatization"
 	OdigosURLTemplateProcessorType        = "odigosurltemplate"
+	PiiMaskingProcessorName               = "odigospiimasking"
```
tamirdavid1:
```suggestion
	PiiMaskingProcessorName               = "odigos-pii-masking"
```

tamirdavid1 (approved):

collector/processors/odigospiimaskingprocessor/mask.go
blumamir: in a followup PR, this one is just for the migration (to keep them small and scoped)

collector/processors/odigospiimaskingprocessor/metadata.yaml
blumamir: sounds great. let's have a dedicated PR for that to keep this one scoped and easy to review

autoscaler/controllers/actions/piimasking_sync.go
blumamir: We should add logs support. should be it's own PR as followup

collector/processors/odigospiimaskingprocessor/factory.go
blumamir: yes, as a dedicated task

natyharary (approved):

PR #5312, opened 2026-07-20
Sql query config extension

blumamir:

Fixes: RUN-909

make it so that the 2 new db actions
- db query templatization
- infer db attributes

are causing the relevant processor to be injected into the config map of the collectors.

the config itself is passed via the config extension

```release-note
NONE
```


autoscaler/controllers/actions/sql_query_sync.go
```diff
@@ -0,0 +1,144 @@
...
+		return nil
+	}
+
+	if mode == SQLQuerySyncCreateIfMissing {
+		existing := &odigosv1.Processor{}
+		key := client.ObjectKey{
+			Namespace: ns,
+			Name:      consts.SQLQueryProcessorName,
+		}
+		if err := c.Get(ctx, key, existing); err == nil {
```
tamirdavid1, line 120: we should not check for not exists specifically?

autoscaler/controllers/nodecollector/collectorconfig/odigos_config_extension.go
```diff
@@ -19,16 +19,26 @@ func NodeHasURLTemplateProcessor(processors []*odigosv1.Processor) bool {
 	return false
 }
 
+// NodeHasSQLQueryProcessor reports whether processors (node-collector role only) include SQL query processing.
+func NodeHasSQLQueryProcessor(processors []*odigosv1.Processor) bool {
```
tamirdavid1, line 23: maybe we can generalise to any processor? so we can pass the sqlQueryProcessorType in this ccase and other in different cases?

collector/processors/odigossqlqueryprocessor/factory.go
```diff
@@ -47,5 +44,7 @@ func createTracesProcessor(
 		nextConsumer,
 		proc.processTraces,
 		processorhelper.WithCapabilities(consumerCapabilities),
+		processorhelper.WithStart(proc.Start),
+		processorhelper.WithShutdown(proc.Shutdown),
```
tamirdavid1, line 48: what is this ?

collector/processors/odigossqlqueryprocessor/go.mod
```diff
@@ -4,10 +4,12 @@ go 1.26.2
 
 require (
 	github.com/DataDog/go-sqllexer v0.2.4
+	github.com/odigos-io/odigos/common v0.0.0-00010101000000-000000000000
```
tamirdavid1: maybe we can do v0.0.0 like all other replaces?

common/consts/consts.go
```diff
@@ -153,6 +153,7 @@ const (
 	OdigosTraceStateProcessorName         = "odigostracestate"
 	URLTemplatizationProcessorName        = "odigos-url-templatization"
 	OdigosURLTemplateProcessorType        = "odigosurltemplate"
+	SQLQueryProcessorName                 = "odigos-sql-query"
```
tamirdavid1, line 156:
```suggestion
	SQLQueryProcessorName                 = "odigos-sql-query-templatization"
```

tamirdavid1: I remember you told me we want to use similar terms

tamirdavid1 (approved):

autoscaler/controllers/actions/sql_query_sync.go
blumamir: could be improvement to handle the err correctly here. need to align it also in url templatization.

let's keep it the same as existing pattern (url-templatization) and improve this flow in both of them in followup pr

autoscaler/controllers/nodecollector/collectorconfig/odigos_config_extension.go
blumamir: we can! let's do in dedicated PR to keep it focused

collector/processors/odigossqlqueryprocessor/factory.go
blumamir: removing the provider on shutdown (probably not required, but good to have)

collector/processors/odigossqlqueryprocessor/go.mod
blumamir: updated

common/consts/consts.go
blumamir: it's a general processor that does both templatization and infer attributes in one sql query parse. this is intentional

PR #5326, opened 2026-07-21
CORE-1110 | fix: shared log-downgrade for benign profiler errors (common logger + collector)

shivanshuraj1333:
## What this PR does

When the eBPF profiler recognizes an interpreter (Perl, PHP, .NET, V8, BEAM, Go-labels) whose runtime version is outside the supported range — e.g. Perl 5.26 on RHEL 8 — it logs `Failed to load … unsupported Perl 5.26.3 (need >= 5.28 and < 5.43)` at **ERROR**. The condition is expected and fully fail-safe (native stack frames are still captured; only interpreter frames are missing), yet it surfaces as an error in the collector logs (CORE-1110).

This fixes it **in Odigos, with no dependency on any eBPF-profiler change**, and does it as a **shared, standardized mechanism** rather than a one-off:

- **`common/logger`**: adds `NewDowngradeCore` + `DowngradeRule` + `DefaultDowngradeRules()`. A zap core wrapper that lowers the level of records matching a benign, fail-safe pattern (default: `Failed to load …` + `unsupported`/`not supported`/`need >=` → WARN). `Init` applies it, so every component that logs through the common logger (odiglet, instrumentor, scheduler, autoscaler, …) shares one behavior and one extensible rule set. Real errors are untouched — the default rule set is intentionally narrow.
- **collector** (`ebpfprofilerwrapper`): the profiler runs as an OTel receiver whose logs go to the collector's own zap core (not the common logger), so a thin wrapper factory applies **the same `common/logger` rules** to the profiler receiver's logger. One source of truth.

No behavioral change to profiling — log-severity only. Native profiling for these processes already worked and is unchanged.

## Standardization (k8s + VM)

```
common/logger.NewDowngradeCore(rules)   ← single mechanism + rule set
   ├─ common/logger.Init          → odiglet, instrumentor, scheduler, autoscaler (k8s)
   ├─ collector ebpfprofilerwrapper → node collector + VM collector (odigosotelcol)
   └─ vm-agent pkg/logger          → VM agent   (follow-up, same one-line wrap)
```

The profiler bridges its `slog` logger to the collector's zap `Core` (`factory_linux.go`: `log.SetLogger(zapslog.NewHandler(rs.Logger.Core()))`) — Odigos owns that Core, so wrapping it is the correct, upstream-safe seam.

## Validation

- `common/logger` unit test: Perl/.NET unsupported → WARN; real failures (`permission denied`, connection errors) → ERROR; below-error levels untouched. Builds + tests pass on host.
- `ebpfprofilerwrapper` cross-compiles for linux/amd64 against the real profiler + `xreceiver` API; `go vet` clean.
- Full collector link validated in CI (the profiler receiver is Linux-only).

> `components.go` is builder-generated; the edit matches what `make genodigoscol` produces from the updated `builder-config.yaml`.

## Changelog entry

Unsupported interpreter runtime versions (e.g. Perl 5.26) now log a warning instead of an error, via a shared log-downgrade layer in the common logger and the collector.

```release-note
Log expected, fail-safe conditions (e.g. unsupported interpreter runtime versions) as warnings instead of errors, via a shared downgrade layer in the common logger and collector.
```

Refs: CORE-1110


shivanshuraj1333:
fixing this in upstram https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1683

PR #5329, opened 2026-07-21
feat: expose data throughput for node and cluster collector

natyharary:

Expose data throughput for node and cluster collector by scraping otelcol_*_metric_data_size from VictoriaMetrics which were generated by odigostrafficmetrics.

Important note:
The `odigostrafficmetrics` processor is located differently in the pipelines for node and cluster collectors -
In the node collector, this is for the throughput that goes through the node collector, so head sampled data is taken into consideration, while tail sampled does not.
In the cluster collector, this is the throughput that is sent to the destinations.
**this means that if we send to multiple destinations per signal, we will multiply the throughput**

This is used to help customers select the right resource sizing for their cluster.

image

Related ui-kit PR:
https://github.com/odigos-io/ui-kit/pull/1111

```release-note
Show data collectors (node, clusters) data throughput when activating MetricsStore (VictoriaMetrics) in the collectors pipeline page
```

### DIFF
--- a/frontend/graph/collectors.graphqls
+++ b/frontend/graph/collectors.graphqls
@@ -53,9 +53,9 @@ enum ManifestFormat {
 type GatewayDeploymentInfo {
   status: WorkloadRolloutStatus!
   hpa: HorizontalPodAutoscalerInfo
-  throughputTracesBytesPerSec: Float
-  throughputMetricsBytesPerSec: Float
-  throughputLogsBytesPerSec: Float
+  throughputTracesBytesPerSec: Int
+  throughputMetricsBytesPerSec: Int
+  throughputLogsBytesPerSec: Int
   resources: Resources
   imageVersion: String
   lastRolloutAt: String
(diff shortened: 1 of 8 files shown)


frontend/graph/collectors.graphqls
```diff
@@ -53,6 +53,9 @@ enum ManifestFormat {
 type GatewayDeploymentInfo {
   status: WorkloadRolloutStatus!
   hpa: HorizontalPodAutoscalerInfo
+  throughputTracesBytesPerSec: Float
+  throughputMetricsBytesPerSec: Float
+  throughputLogsBytesPerSec: Float
```
BenElferink:
```suggestion
  throughputTracesBytesPerSec: Int
  throughputMetricsBytesPerSec: Int
  throughputLogsBytesPerSec: Int
```

frontend/graph/collectors.graphqls
```diff
@@ -64,6 +67,9 @@ type GatewayDeploymentInfo {
 type CollectorDaemonSetInfo {
   status: WorkloadRolloutStatus!
   nodes: NodesSummary!
+  throughputTracesBytesPerSec: Float
+  throughputMetricsBytesPerSec: Float
+  throughputLogsBytesPerSec: Float
```
BenElferink:
```suggestion
  throughputTracesBytesPerSec: Int
  throughputMetricsBytesPerSec: Int
  throughputLogsBytesPerSec: Int
```

BenElferink (approved):

frontend/graph/collectors.graphqls
natyharary: Makes sense since those are bytes anyway. changed

PR #5334, opened 2026-07-22
feat: pii masking support regex

blumamir:

Fixes: RUN-909

add options similar to extract attributes config - to mask values from attributes based on custom regex

```release-note
pii masking action - support masking by custom regex and for specific formats (json, sql, resource path)
```


autoscaler/controllers/actions/piimasking_controller.go
```diff
@@ -18,17 +18,12 @@ package actions
...
 	"github.com/odigos-io/odigos/common"
-	actionsapi "github.com/odigos-io/odigos/common/api/actions"
 )
 
 var piiMaskingSupportedSignals = map[common.ObservabilitySignal]struct{}{
 	common.TracesObservabilitySignal: {},
 }
 
-type piiMaskingProcessorConfig struct {
-	PiiCategories []actionsapi.PiiCategory `json:"pii_categories"`
```
tamirdavid1, line 29: remind me what this removed? because it is moved to the common/api/actions ?

tamirdavid1 (approved):

autoscaler/controllers/actions/piimasking_controller.go
blumamir: it was duplicate of the struct in common, so I cleaned it up

PR #5336, opened 2026-07-22
CORE-1331: fix CI flake from webhook cert secret mount delay

BenElferink:
## Summary
- Fix autoscaler/instrumentor readiness flake caused by slow kubelet Secret volume sync during webhook cert bootstrap
- Instead of restarting pods (`RestartOnSecretRefresh`), sync TLS material from the Secret into a writable local `CertDir` via `SecretDiskSync`
- Point webhook server + cert-controller at that local dir, and remove unused Secret volume mounts

Fixes [CORE-1331](https://linear.app/odigos/issue/CORE-1331/ci-flake-autoscaler-not-ready-due-to-webhook-cert-secret-volume-sync)

Related: [cert-controller#35](https://github.com/open-policy-agent/cert-controller/issues/35)

## Test plan
- [ ] Fresh install: autoscaler/instrumentor become Ready with `restartCount: 0`
- [ ] Confirm cert files appear under `/tmp/odigos-webhook-certs` shortly after Secret is written
- [ ] Confirm `/readyz` passes without waiting on kubelet secret projection
- [ ] Watch CI `central-install` / kubernetes-tests for the previous autoscaler Ready flake

```release-note
NONE
```


shivanshuraj1333 (dismissed):

RonFed (changes requested):
not sure we should re-start the autoscaler to handle this.
The slowness is only happening in initial install (which is what we test all the time) - users' common use-case is upgrade, which doesn't have this slowness issue.
Having always restart count of 1, might raise some questions, and cause users to think something is wrong

shivanshuraj1333:
Need to unblock the CI, lets put some fix in this PR which help us unblock the CI, also I agree with Ron about the autoscaler restart

BenElferink:
@RonFed @shivanshuraj1333 Agreed on avoiding autoscaler restart.

Dropped `RestartOnSecretRefresh`. Instead we now sync webhook TLS material from the Secret into a writable local `CertDir` (`SecretDiskSync`), so readiness no longer waits on kubelet secret volume projection and `restartCount` stays 0.

Also removed the unused Secret volume mounts from autoscaler/instrumentor deployments.

shivanshuraj1333:
LGTM, waiting for Ron's approval

PR #5341, opened 2026-07-22
feat: add scopes to pii masking 

blumamir:

Fixes: RUN-909

Add scopes to pii masking and reconcile it in instrumentation config to pass to processor as config extension

```release-note
add option to set scopes on pii masking action (which services participate in each pii masking action)
```


collector/processors/odigospiimaskingprocessor/cache.go
RonFed, line 1: this seems like common boilerplate code that each component that use the cache has? maybe we can think of a generic common implementation for components to use in the future?

RonFed (approved):

collector/processors/odigospiimaskingprocessor/cache.go
blumamir: good idea, let's have a dedicated PR to migrate them to common cache

PR #5347, opened 2026-07-23
sqlquery: handle cassandra as SQL best effort

blumamir:

Fixes: RUN-909

handle cassandra as SQL db for the db infere attributes and db templatization.
they are CQL but will be tokenize as SQL which will work most of the times (but not always).

in future PR we will improve cassandra sql processor handling

```release-note
process cassandra as sql for infer db attributes and db query templatization
```

### DIFF
--- a/collector/processors/odigossqlqueryprocessor/processor.go
+++ b/collector/processors/odigossqlqueryprocessor/processor.go
@@ -18,7 +18,11 @@ import (
 	"github.com/odigos-io/odigos/common/collector"
 )
 
-const dbStatementKey = "db.statement"
+const (
+	dbStatementKey = "db.statement"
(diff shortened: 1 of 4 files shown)


collector/processors/odigossqlqueryprocessor/processor.go
```diff
@@ -18,7 +18,11 @@ import (
 	"github.com/odigos-io/odigos/common/collector"
 )
 
-const dbStatementKey = "db.statement"
+const (
+	dbStatementKey = "db.statement"
+	dbOperationKey = "db.operation"
+	dbSQLTableKey  = "db.sql.table"
```
RonFed: we can use older semconv packages for these

RonFed (approved):

PR #5359, opened 2026-07-24
CORE-1307 | fix: aggregate symtab-size-skip warnings instead of logging per binary

shivanshuraj1333:

Follow-up to #5302. The Warn added there logged once per binary that tripped
`max_symtab_bytes` — but that's deduped only by the symbol cache's lifetime, not
by content, so a busy node with many oversized native libraries (or the same
library across many restarting containers, each resolving to its own path) could
turn into a per-binary, per-restart warning storm. That's worst exactly where
operators dial `maxMemoryMiB` down to protect a constrained node — the case that
skips the most binaries also logs the most.

Per-binary detail moves to `Debug`; a counter now feeds one `Warn` per sweep
interval (the existing 30s sweeper) naming how many binaries were skipped since
the last report — a bounded, constant-rate signal regardless of churn.

#### Changelog entry:

```release-note
NONE
```

### DIFF
--- a/collector/processors/odigossymbolizeprocessor/internal/symbolize/memory_bound_test.go
+++ b/collector/processors/odigossymbolizeprocessor/internal/symbolize/memory_bound_test.go
@@ -12,8 +12,13 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


shivanshuraj1333:
will cover the changes in already approved PR

PR #5372, opened 2026-07-26
fix: block CRUD on Instrumentation Rules for OSS installations

BenElferink:
```release-note
fix: block CRUD on Instrumentation Rules for OSS installations
```


frontend/graph/instrumentationrules.resolvers.go
```diff
@@ -18,21 +19,44 @@ func (r *computePlatformResolver) InstrumentationRules(ctx context.Context, obj
...
 // CreateInstrumentationRule is the resolver for the createInstrumentationRule field.
 func (r *mutationResolver) CreateInstrumentationRule(ctx context.Context, instrumentationRule model.InstrumentationRuleInput) (*model.InstrumentationRule, error) {
+	isOnprem, err := IsOnprem(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if !isOnprem {
+		return nil, fmt.Errorf("instrumentation rules are only supported for onprem installations")
+	}
+
```
alonkeyval: this can be wrap as func: requireOnPrem and reuse it in all 3 places

frontend/graph/utils.go
```diff
@@ -5,12 +5,29 @@ import (
...
 	"github.com/odigos-io/odigos/frontend/graph/status"
+	"github.com/odigos-io/odigos/frontend/kube"
+	"github.com/odigos-io/odigos/k8sutils/pkg/env"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+func IsOnprem(ctx context.Context) (bool, error) {
+	odigosNs := env.GetCurrentNamespace()
+	configMap, err := kube.DefaultClient.CoreV1().ConfigMaps(odigosNs).Get(ctx, k8sconsts.OdigosDeploymentConfigMapName, metav1.GetOptions{})
```
alonkeyval: can we take this from the cache?

alonkeyval (approved):
approved with comments

PR #5374, opened 2026-07-26
CORE-1360: add option to strip PostgreSQL :: casts in query templatization

blumamir:

Fixes: CORE-1360

Adds a `removePostgresCastOperator` option to the DbQueryTemplatization action. When enabled, PostgreSQL `::type` casts are stripped from query text (e.g. `SELECT ... WHERE id = ?::uuid` → `SELECT ... WHERE id = ?`), so queries that only differ by cast type collapse to the same template and do not inflate cardinality.

```release-note
Add DbQueryTemplatization option `removePostgresCastOperator` to strip PostgreSQL `::type` casts from query text.
```


BenElferink (approved):

collector/processors/odigossqlqueryprocessor/processor.go
```diff
@@ -25,6 +25,11 @@ import (
 // it checks for standalone UUIDs, not part of a longer identifier/literal.
 var uuidRegex = regexp.MustCompile(`(^|[^0-9A-Fa-f])([0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12})($|[^0-9A-Fa-f])`)
 
+// postgresCastRegex matches PostgreSQL `::type` casts, including optional
+// schema qualification and array brackets (e.g. ::uuid, ::public.my_type, ::int[]).
+// Leading whitespace is included so normalized forms like "? :: jsonb" collapse cleanly.
+var postgresCastRegex = regexp.MustCompile(`(?i)\s*::\s*(?:[A-Za-z_][\w$]*\.)?[A-Za-z_][\w$]*(?:\s*\[\s*\])*`)
```
tamirdavid1, line 31: I think we might very soon see yourselves add more use cases for it,
maybe we can send it as map of 
name:
   enabled:
   regex:

and then it will be easy to extend it?

tamirdavid1 (approved):

collector/processors/odigossqlqueryprocessor/processor.go
blumamir: sounds good, lets wait for one or few more examples and then generalize it?

PR #5379, opened 2026-07-26
CORE-1328 | chore: add hpa and fix service ip resolution

tamirdavid1:

Insights now scales horizontally out of the box. We added an always-on CPU-based HorizontalPodAutoscaler for the odigos-insights service, **with min/max replicas defaulting to the gateway's sizing (and overridable via Helm)** so capacity tracks load automatically. To make that scaling actually effective, we fixed a gRPC connection-pinning issue where the gateway pinned all telemetry to a single insights pod: the gateway now targets a new headless Service via the dns:/// resolver with round_robin, fanning traffic evenly across every replica. This means as trace volume grows, insights adds pods and the load is distributed across them instead of overwhelming one. Requires metrics-server in the cluster for the HPA to read CPU utilization.

```release-note
NONE
```

### DIFF
--- a/api/k8sconsts/insights.go
+++ b/api/k8sconsts/insights.go
@@ -3,4 +3,11 @@ package k8sconsts
 const (
 	OdigosInsightsServiceName = "odigos-insights"
 	OdigosInsightsOTLPPort    = 4317
+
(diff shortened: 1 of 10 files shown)


shivanshuraj1333 (approved):
LGTM

liranm-odigos (approved):

PR #5390, opened 2026-07-27
CORE-1155: add grpc appprotocol to collector ports

damemi:
Problem When Istio's OpenTelemetry tracer is pointed at odigos-data-collection-local-traffic:4317 over gRPC, all exports fail (rq_error = rq_total, rq_success: 0). OTLP/HTTP on 4318 works. Net effect: no Istio/Envoy spans reach Odigos when gRPC is used.

Root cause The service defines port 4317 as Name: "otlp" with no appProtocol. Istio only marks an upstream cluster as HTTP/2 (required for gRPC) when the port's appProtocol is grpc or the port name is grpc/grpc-*. With neither, Istio falls back to use_downstream_protocol_config; the tracer's internal gRPC client has no downstream to mirror and degrades to HTTP/1.1, which the collector's gRPC receiver rejects. The collector isn't in the mesh, so there's no ALPN to negotiate h2c.

Fix Add AppProtocol: ptr.To("grpc") to the 4317 port in autoscaler/controllers/nodecollector/service.go. (Same pattern applies to the gateway service in k8sutils/pkg/service/service.go if gRPC to the gateway is also wanted.)

This also updates the data collection service handler to patch (previously just created), which the gateway service handler already does

```release-note
fix(collector): support Istio otel tracer on data collection local traffic service
```

### DIFF
--- a/autoscaler/controllers/nodecollector/service.go
+++ b/autoscaler/controllers/nodecollector/service.go
@@ -12,7 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


autoscaler/controllers/nodecollector/service.go
```diff
@@ -71,6 +77,15 @@ func (b *nodeCollectorBaseReconciler) syncService(ctx context.Context, dc *odigo
 		return err
 	}
 
-	err := b.Client.Create(ctx, dcService)
-	return client.IgnoreAlreadyExists(err)
+	// Use server-side apply instead of create/update: the autoscaler cache only caches
+	// Services labeled as the cluster gateway, so this node collector Service is not in the
+	// cache and a cached Get would always miss. SSA writes directly to the API server and
+	// reconciles the desired spec on both create and update (e.g. adding the grpc appProtocol
+	// to an existing Service).
```
blumamir: should we add it to the cache instead?

damemi: yeah I wasn't sure which one would be a messier change, but now that I think about it that's probably better to just do

blumamir (approved):

PR #5394, opened 2026-07-28
CORE-1306: add destination exporter batching

natyharary:

Adds exporter-side batching for destinations, in order to deprecate the batch processor.
It uses the exporter helper package:
https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md

When modifying CRD, the following args can be applied: (these are the default values)
```yaml
...
sendingQueue:
    unit: requests          # queue sizer
    size: 1000              # queue_size
    batch:
      unit: items           # batch sizer
      size: 8192            # min_size
      maxSize: 0            # no max / no splitting
      flushTimeout: "200ms"
...
```

these map to these values in sending_queue:
```yaml
...
sending_queue:
  enabled: true
  sizer: requests
  queue_size: 1000
  batch:
    sizer: items
    min_size: 8192
    max_size: 0
    flush_timeout: 200ms
...
```

Some destinations do not support the exporter helper package so they will default and get the batch processor applied to their 

```release-note
Adds batching at the exporter for supporting destinations
```


api/odigos/v1alpha1/destination_types.go
```diff
@@ -36,6 +36,40 @@ type DestinationMetricsSettings struct {
 	CollectAgentsTelemetry  *bool `json:"agentsTelemetryEnabled,omitempty"`
 }
 
+type SendingQueueBatch struct {
+	// Unit is how batch size is measured (items or bytes). Maps to exporterhelper sending_queue.batch.sizer.
+	// +optional
+	// +kubebuilder:validation:Enum=items;bytes
+	Unit string `json:"unit,omitempty"`
```
blumamir, line 43: Should we use an enum here? 

Can validate that only known values are used, and also that document what are the available options as a type

api/odigos/v1alpha1/destination_types.go
```diff
@@ -36,6 +36,40 @@ type DestinationMetricsSettings struct {
...
+	// Size is the minimum batch size before flush. Maps to exporterhelper sending_queue.batch.min_size.
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	Size int `json:"size,omitempty"`
+	// MaxSize is the maximum batch size (enables splitting). Maps to exporterhelper sending_queue.batch.max_size.
+	// 0 means no maximum (OTel default). Nil means unset (omit from config).
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	MaxSize *int `json:"maxSize,omitempty"`
```
blumamir, line 52: I think normal users will have hard time understanding how these fields play together. instead of sending them to open urls with docs, let's document it here cleanly

api/odigos/v1alpha1/destination_types.go
```diff
@@ -36,6 +36,40 @@ type DestinationMetricsSettings struct {
...
+type SendingQueueBatch struct {
+	// Unit is how batch size is measured (items or bytes). Maps to exporterhelper sending_queue.batch.sizer.
+	// +optional
+	// +kubebuilder:validation:Enum=items;bytes
+	Unit string `json:"unit,omitempty"`
+	// Size is the minimum batch size before flush. Maps to exporterhelper sending_queue.batch.min_size.
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	Size int `json:"size,omitempty"`
+	// MaxSize is the maximum batch size (enables splitting). Maps to exporterhelper sending_queue.batch.max_size.
```
blumamir, line 48: > Maps to exporterhelper sending_queue.batch.max_size.

you have the context in you head, but this comment can be read by odigos users which will have no idea what `exporterhelper` is.

api/odigos/v1alpha1/destination_types.go
```diff
@@ -36,6 +36,40 @@ type DestinationMetricsSettings struct {
...
+	// 0 means no maximum (OTel default). Nil means unset (omit from config).
```
blumamir, line 49: If 0 and nil have the same outcome, then should we state them separately here? does it give any value to the user?

api/odigos/v1alpha1/destination_types.go
```diff
@@ -36,6 +36,40 @@ type DestinationMetricsSettings struct {
...
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	MaxSize *int `json:"maxSize,omitempty"`
+	// FlushTimeout is how long to wait before flushing a batch. Maps to exporterhelper sending_queue.batch.flush_timeout.
+	// Example: "200ms", "1s".
+	// +optional
+	FlushTimeout string `json:"flushTimeout,omitempty"`
```
blumamir, line 56: there is also the minimum batch size above, which can conflict with this setting. what if the time passed and we did not get the minimum size?

blumamir, line 56: use `metav1.Duration` instead of string

api/odigos/v1alpha1/destination_types.go
```diff
@@ -36,6 +36,40 @@ type DestinationMetricsSettings struct {
 	CollectAgentsTelemetry  *bool `json:"agentsTelemetryEnabled,omitempty"`
 }
 
+type SendingQueueBatch struct {
```
blumamir, line 39: wonder if it's a "SendingQueue" batch or just a batch?
It's true that the batch is stored in the sending queue, but it's a batch also for exporting.

api/odigos/v1alpha1/destination_types.go
```diff
@@ -54,6 +88,10 @@ type DestinationSpec struct {
 	// If not specified, defaults to "all".
 	// +optional
 	SourceSelector *SourceSelector `json:"sourceSelector,omitempty"`
+
+	// SendingQueue configures exporterhelper sending_queue for this destination's exporters.
```
blumamir, line 92: same here - this text is exposed to the user, which likely will read `exporterhelper sending_queue` and get confused

api/odigos/v1alpha1/destination_types.go
```diff
@@ -36,6 +36,40 @@ type DestinationMetricsSettings struct {
...
+	// +kubebuilder:validation:Minimum=0
+	MaxSize *int `json:"maxSize,omitempty"`
+	// FlushTimeout is how long to wait before flushing a batch. Maps to exporterhelper sending_queue.batch.flush_timeout.
+	// Example: "200ms", "1s".
+	// +optional
+	FlushTimeout string `json:"flushTimeout,omitempty"`
+}
+
+type SendingQueue struct {
+	// Unit is how queue size is measured (items, bytes, requests). Maps to exporterhelper sending_queue.sizer.
```
blumamir, line 60: consider if we want to use values in bytes here, since the handy measurement is usually Mb

api/odigos/v1alpha1/destination_types.go
```diff
@@ -54,6 +88,10 @@ type DestinationSpec struct {
 	// If not specified, defaults to "all".
 	// +optional
 	SourceSelector *SourceSelector `json:"sourceSelector,omitempty"`
+
+	// SendingQueue configures exporterhelper sending_queue for this destination's exporters.
+	// +optional
+	SendingQueue *SendingQueue `json:"sendingQueue,omitempty"`
```
blumamir, line 94: is it supported for all exporters? or only otlp?

api/odigos/v1alpha1/destination_types.go
```diff
@@ -36,6 +36,40 @@ type DestinationMetricsSettings struct {
...
+	// +optional
+	// +kubebuilder:validation:Enum=items;bytes;requests
+	Unit string `json:"unit,omitempty"`
+	// Size is the maximum queue size. Maps to exporterhelper sending_queue.queue_size.
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	Size int `json:"size,omitempty"`
+	// Batch configures exporterhelper sending_queue.batch for this destination's exporters.
+	// +optional
+	Batch *SendingQueueBatch `json:"batch,omitempty"`
```
blumamir, line 70: what happen if it's unset? can we end up with in-consistent settings?

blumamir, line 70: should this be part of the sending queue? why not have it directly on the destination itself (so users can set it directly without requiring another concept)

common/config/sending_queue.go
```diff
@@ -0,0 +1,169 @@
...
+// applySendingQueue stamps sending_queue onto supported exporters.
+// Returns true when sending_queue was applied to every exporter on every pipeline;
+// false if any exporter could not receive it.
+func applySendingQueue(dest ExporterConfigurer, cfg *Config, pipelineNames []string) bool {
+	q := dest.GetSendingQueueConfig()
+	if q == nil {
+		return false
+	}
+	queue := BuildSendingQueue(*q)
+	applied := len(pipelineNames) > 0
```
blumamir, line 106: if we end up with 2 exporters, where one is applied and one is not, then the final value here will be `false`, right? even though we did applied it to one exporter.

not sure if it's a real scenario or not, but looks like the code is currently logically incorrect

common/config/sending_queue.go
```diff
@@ -0,0 +1,169 @@
...
+			exporterConfig, ok := cfg.Exporters[exporterName]
+			if !ok {
+				applied = false
+				continue
+			}
+			switch m := exporterConfig.(type) {
+			case GenericMap:
+				m["sending_queue"] = queue
+			case map[string]interface{}:
+				m["sending_queue"] = queue
```
blumamir, line 132: nit: I think you can do
```suggestion
			case GenericMap, map[string]interface{}:
				m["sending_queue"] = queue
```

common/config/sending_queue.go
```diff
@@ -0,0 +1,169 @@
...
+// TODO(2026-07-27): revisit when collector exporters gain sending_queue support.
+var unsupportedSendingQueuePrefixes = []string{
+	"nop/",
+	"debug/",
+	"awsemf/",
+	"awsxray/",
+	"azureblobstorage/",
+	"googlecloudstorage/",
+	"prometheusremotewrite/",
+}
```
blumamir, line 158: I wonder if we can write it on the destination manifest, instead of having this in the code

PR #5398, opened 2026-07-28
CORE-1324 | feat: remove profiling processors and eBPF receiver from the OSS collector

shivanshuraj1333:

Removes the eBPF profiler receiver and the profiling processors from the OSS collector. They move to odigos-enterprise, which builds its own distribution from them.

Deletes `collector/receivers/odigosebpfreceiver` and `collector/processors/{odigosprofiles,odigossymbolize,odigosvmprofileattrs}processor`, regenerates the OCB manifest and `odigosotelcol`, and drops the matching release-tagging entries. `go.opentelemetry.io/ebpf-profiler` stays as an indirect dependency — `elasticsearchexporter` still pulls it in.

Both prerequisites are already on `main`: the tier gate landed in #5447 and image selection in #5446.

```release-note
The OSS collector image no longer includes the eBPF profiling receiver and processors. They are now part of the Odigos Enterprise collector.
```


blumamir:
Will review after rebased when related PRs are merged

blumamir (approved):

PR #5400, opened 2026-07-28
CORE-1324 | feat: remove the odiglet eBPF map exchange from OSS

shivanshuraj1333:

The odiglet always creates the shared eBPF maps and starts the unix-socket server that hands out their file descriptors. The only reader is `odigosebpfreceiver`, which is enterprise-only, so on OSS this allocates maps and runs a server nothing ever connects to.

Removes the wiring outright, along with `common/ebpf` and `common/unixfd`, which had no other OSS consumer. Both ends of the exchange now live in odigos-io/odigos-enterprise#3310.

`InstrumentationManagerOptions` keeps its `TracesMap` / `MetricsMap` / `MetricsAttributesMap` / `LogsMap` fields. They are pass-throughs to `instrumentation.ManagerOptions`, which the eBPF SDKs read as external readers, and the enterprise odiglet sets them through `odiglet.New()`.

Needs #5398 in first.

```release-note
NONE
```


RonFed:
if this flag will be set only in enterprise, I think we shouldn't make this as a CLI flag to the binary, rather a configuration on the odiglet struct/manager that will be passed only by the enterprise `main` - this way we avoid changes in `helm` for this

existing mechanisms we already have for this kind of extensions are:
`InstrumentationManagerOptions`
`AddRunnable()`

RonFed (approved):

PR #5410, opened 2026-07-29
CORE-1228: (OBI) Support custom service name and resource attributes

damemi:
This updates the OBI distro + network metrics implementations to support custom service name and resource attributes at instrumentation time. These attributes are shared across signals via the obi internals (so for example you can't have a different service name for traces and metrics, which wouldn't be useful anyway).

These only apply at CreateInstrumentation, so these attributes can't be changed without uninstrumenting and reinstrumenting (ie, disable, edit service name, reenable). From what I could tell the other distros don't support mid-instrumentation changes either so this is consistent. (The OBI dynamic sdk does support changing these, but our pipeline doesn't have a clear way to do that)

cherry-pick:v1.33

```release-note
feat: support service.name and resource attributes for OBI traces and network metrics.
```


shivanshuraj1333 (approved):
lgtm

PR #5420, opened 2026-07-30
CORE-1399 | fix: resolve unresolvable Go module graph across 12 packages

shivanshuraj1333:
`go list -m all` fails on 12 modules because several dependencies declare `replace` directives in their own go.mod that Go does not propagate to consumers.

A module with an unresolvable transitive dependency breaks any tooling that loads its full dependency graph

Adds the missing `replace` to each affected module, pointing to the real resolvable version or local path the upstream module already intends.

```release-note
NONE
```

### DIFF
--- a/autoscaler/go.mod
+++ b/autoscaler/go.mod
@@ -104,5 +104,9 @@ replace (
 	github.com/odigos-io/odigos/api => ../api
 	github.com/odigos-io/odigos/common => ../common
 	github.com/odigos-io/odigos/k8sutils => ../k8sutils
+	// k8sutils requires this via its own local-only replace, which doesn't propagate to us.
+	github.com/odigos-io/odigos/odigosauth => ../odigosauth
 	github.com/odigos-io/odigos/status => ../status
+	// argo-rollouts@v1.9.1 requires this via its own internal-only replace, which doesn't propagate to us.
+	k8s.io/kubelet => k8s.io/kubelet v0.34.1
 )
--- a/cli/go.mod
+++ b/cli/go.mod
@@ -163,10 +163,14 @@ require (
 )
 
 replace (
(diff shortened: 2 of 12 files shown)


autoscaler/go.mod
```diff
@@ -104,5 +104,12 @@ replace (
 	github.com/odigos-io/odigos/api => ../api
 	github.com/odigos-io/odigos/common => ../common
 	github.com/odigos-io/odigos/k8sutils => ../k8sutils
+	// k8sutils requires this via its own local-only replace, which doesn't propagate to us.
+	// TODO: drop if this module's k8sutils dependency is ever removed (see deviceplugin, CORE-1401).
+	github.com/odigos-io/odigos/odigosauth => ../odigosauth
 	github.com/odigos-io/odigos/status => ../status
+	// argo-rollouts requires these two via its own internal-only replace, currently pinned to v0.29.3 there too.
+	// TODO: bump to match if/when argo-rollouts is upgraded (latest v1.9.1 pins v0.34.1 instead).
+	k8s.io/cloud-provider => k8s.io/cloud-provider v0.29.3
```
blumamir: remove now that #5429 is merged

autoscaler/go.mod
```diff
@@ -104,5 +104,12 @@ replace (
...
+	k8s.io/kubelet => k8s.io/kubelet v0.29.3
```
blumamir: where is this one coming from?

blumamir (commented):
final cleanups before merging

autoscaler/go.mod
shivanshuraj1333: after bumping argo also, still `argo-rollouts@v1.9.1` transitively requires [k8s.io/kubelet](http://k8s.io/kubelet)

blumamir (approved):
awesome, thank you @shivanshuraj1333

PR #5427, opened 2026-07-31
ci: verify Go module graphs resolve in CI

shivanshuraj1333:
`go build`, `go test`, and `go mod tidy` only need to resolve packages actually compiled, so they pass even when a module's full dependency graph is unresolvable (see CORE-1399), nothing in CI catches that class of break before it ships.

Adds a `verify-go-mod-graph` job that runs `go list -m all` across every module, the one command that does exercise the full graph, so a broken module graph fails CI directly instead of surfacing later.

Fixes CORE-1400

```release-note
NONE
```

### DIFF
--- a/.github/workflows/go-mod-tidy.yml
+++ b/.github/workflows/go-mod-tidy.yml
@@ -16,3 +16,17 @@ jobs:
         run: make go-mod-tidy
       - name: Check clean repository
         run: make check-clean-work-tree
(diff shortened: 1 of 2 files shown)


blumamir (approved):
awesome 🤩

let's merge after https://github.com/odigos-io/odigos/pull/5420 gets in

PR #5438, opened 2026-08-02
feat: per source/container allow to set AllowConcurrentAgents

blumamir:

Fixes: CORE-1417

Suuport setting `AllowConcurrentAgents` for each container.

Currently it's a global setting for the entire cluster, but we want more fine grained control. Currently adding support in backend, users can opt-in via the Source manifest.

```
kind: Source
metadata:
  name: source-fgdzk
  namespace: default
spec:
  containerOverrides:
  - allowConcurrentAgents: true
    containerName: app
  disableInstrumentation: false
  workload:
    kind: Deployment
    name: app
    namespace: default

```

```release-note
Add settings in the source+container level to run/block instrumentation when other agents are detected
```


tamirdavid1 (approved):

tamirdavid1:
LIKE

PR #5446, opened 2026-08-03
CORE-1324 | feat: select odigos-collector vs odigos-enterprise-collector by tier in helm + operator

shivanshuraj1333:

The collector is being split into an OSS image (`odigos-collector`) and an enterprise image (`odigos-enterprise-collector`, built in odigos-enterprise). Nothing picks between them today.

Selects the image by tier in the three places that name it: the autoscaler's `ODIGOS_COLLECTOR_IMAGE`, the odiglet data-collection container, and the operator's `RELATED_IMAGE_*` wiring. Uses the chart's existing `odigos.secretExists` if/else idiom rather than introducing a second one.

Must land before #5398 — reversed, enterprise installs point at an image that no longer contains the components their config asks for.

Needs odigos-io/odigos-enterprise#3356 in first. Enterprise e2e installs this chart at enterprise tier, so from the moment this merges it renders `odigos-enterprise-collector:e2e-test`, and nothing loads that tag into kind until 3356.

#### Upgrade note for customers pulling from a mirrored registry

From this change on, an enterprise install asks for a new image name. Customers who mirror our images into their own registry rather than pulling ours directly must add `odigos-enterprise-collector` — and `odigos-enterprise-collector-rhel-certified` on OpenShift — to their mirror **before** upgrading, or the collector will fail to pull. A mirror carrying only `odigos-collector` is no longer enough at enterprise tier.

```release-note
Enterprise installations now run the dedicated odigos-enterprise-collector image.
```


blumamir (approved):

shivanshuraj1333:
[3rd one in the row, get back and merge]

PR #5447, opened 2026-08-03
CORE-1324 | feat: inject ODIGOS_ONPREM_TOKEN into the cluster collector on enterprise tier

shivanshuraj1333:

The enterprise collector's auth extension (odigos-enterprise#3354) verifies the Odigos on-prem token at startup and refuses to run without it. Nothing supplies that token today.

Injects `ODIGOS_ONPREM_TOKEN` from the `odigos-pro` secret into the cluster collector and the odiglet data-collection container, on enterprise tier only. Community has neither the extension nor the secret.

Needs odigos-io/odigos-enterprise#3354 in first. This writes `odigos_enterprise_auth` into the enterprise-tier collector config, and the collector rejects unknown component types at config load even when no pipeline references them — so without the extension in the image, the enterprise collector crashloops.

```release-note
NONE
```


autoscaler/controllers/nodecollector/collectorconfig/common.go
```diff
@@ -205,18 +210,46 @@ func init() {
...
+
+func CommonConfig(tier common.OdigosTier) config.Config {
+	if tier == common.CommunityOdigosTier {
+		return config.Config{
+			Extensions: commonExtensions,
+			Service:    commonService,
+		}
+	}
+
+	extensions := cloneGenericMap(commonExtensions)
```
blumamir: nit: instead of cloning, let's refactor so commonExtensions is a function that returns a map and not static variable.

autoscaler/controllers/nodecollector/collectorconfig/common.go
```diff
@@ -205,18 +210,46 @@ func init() {
...
+	return out
+}
+
+func CommonConfig(tier common.OdigosTier) config.Config {
+	if tier == common.CommunityOdigosTier {
+		return config.Config{
+			Extensions: commonExtensions,
+			Service:    commonService,
+		}
+	}
```
blumamir: nit: consider wrapping this into functions, that receives the tier as parameter and return the relevant extensions list etc for this tier.
then the logic is contained in a function and you can immediately tell in one place which extensions are used and what are the parameters it depends on.

It will also make the current function more streamlined to read, instead of branching like this. the code below (most of the function body) is for enterprise, which is not clear from how the code is structured

autoscaler/controllers/nodecollector/collectorconfig/common.go
```diff
@@ -205,18 +210,46 @@ func init() {
 	}
 }
 
-func CommonApplicationTelemetryConfig(nodeCG *odigosv1.CollectorsGroup, onGKE bool, odigosNamespace string, detectors []string) config.Config {
+func CommonApplicationTelemetryConfig(nodeCG *odigosv1.CollectorsGroup, onGKE bool, odigosNamespace string, detectors []string, tier common.OdigosTier) config.Config {
+	receivers := cloneGenericMap(commonReceivers)
+	if tier != common.CommunityOdigosTier {
+		receivers[odigosEbpfReceiverName] = config.GenericMap{}
+	}
```
blumamir: same here - wrap this in a function that gets the tier as parameter and returns the right receivers map based on the input

autoscaler/controllers/nodecollector/collectorconfig/common.go
```diff
@@ -188,7 +193,7 @@ func init() {
 				},
 			},
 		},
-		odigosEbpfReceiverName: config.GenericMap{},
+		// odigosEbpfReceiverName is added per tier in CommonApplicationTelemetryConfig.
```
blumamir: I think this is no longer relevant here after the refactor.

autoscaler/controllers/nodecollector/collectorconfig/metrics.go
```diff
@@ -11,10 +12,12 @@ const (
 	odigosMetricsPipelineName = "metrics"
 )
 
-func metricsReceivers(metricsConfigSettings *odigosv1.CollectorsGroupMetricsCollectionSettings) (config.GenericMap, []string) {
+func metricsReceivers(metricsConfigSettings *odigosv1.CollectorsGroupMetricsCollectionSettings, tier common.OdigosTier) (config.GenericMap, []string) {
 	receivers := config.GenericMap{}
-	// For now, we are adding odigosebpfreceiver in any case if metrics enabled.
-	pipelineReceiverNames := []string{odigosEbpfReceiverName}
+	pipelineReceiverNames := []string{}
+	if tier != common.CommunityOdigosTier {
```
blumamir: nit: make it a positive check. e.g. for enterprise add this pipeline. makes it easier to read and understand and more explicit

autoscaler/controllers/nodecollector/collectorconfig/traces.go
```diff
@@ -144,9 +145,16 @@ func TracesConfig(nodeCG *odigosv1.CollectorsGroup, opts TracesConfigOptions) co
 		tracesMainPipelineExporterNames = append(opts.AdditionalTraceExporters, odigosTracesExportingForwardConnectorName)
 	}
 
+	tracesPipelineReceivers := []string{OTLPInReceiverName}
+	// odigosebpfreceiver only exists in the enterprise collector image - see the comment on
+	// odigosEbpfReceiverName in common.go.
+	if opts.Tier != common.CommunityOdigosTier {
+		tracesPipelineReceivers = append(tracesPipelineReceivers, odigosEbpfReceiverName)
+	}
```
blumamir: same here, lets wrap this in a function, so the receivers logic is contained and this function only handles pipeline logic

autoscaler/controllers/clustercollector/configmap.go
```diff
@@ -218,8 +218,13 @@ func syncConfigMap(enabledDests *odigosv1.DestinationList, allProcessors *odigos
 			if err := addSelfTelemetryPipeline(c, gateway.Spec.CollectorOwnMetricsPort, destinationPipelineNames, signalsRootPipelines); err != nil {
 				return err
 			}
-			if err := addProfilingGatewayPipeline(c, env.GetCurrentNamespace(), profilingCfg); err != nil {
-				return err
+			// The gateway profiles pipeline receives over plain otlp so it would not crash an
+			// OSS collector, but community tier should have no profiling wiring at all.
+			if tier != odigoscommon.CommunityOdigosTier {
```
blumamir: same here nit: positive explicit check == OnPrem

autoscaler/controllers/clustercollector/deployment.go
```diff
@@ -151,6 +151,21 @@ func getDesiredDeployment(ctx context.Context, c client.Client, enabledDests *od
 	}
 
 	extraEnvVars := []corev1.EnvVar{}
+	// The odigos_enterprise_auth extension verifies this token at startup and refuses to run
+	// without it. Community tier has neither the extension nor the odigos-pro secret.
+	if tier != common.CommunityOdigosTier {
```
blumamir: only for OnPrem, instead of not for OSS

blumamir (approved):
LGTM, however added few comments.

Approving so this is not blocked after the comments are addressed

PR #5450, opened 2026-08-03
CORE-1422 | chore: remove RuntimeUpdateState

blumamir:
## Summary

Cleanup from the migration of using workload manifest injection to pods webhook.

Removes the temporary `RuntimeUpdateState` / `ProcessingState` field that tracked whether the new runtime detection path had run during that migration. Also drops the related logic in odiglet runtime inspection and the instrumentor webhook env injector, and updates CRDs, generated code, docs, and e2e asserts.

Linear: https://linear.app/odigos/issue/CORE-1422/remove-runtimeupdatestate-migration-leftovers-after-webhook-injection

Removes leftover migration scaffolding (`RuntimeUpdateState`) now that env injection happens via the pods webhook instead of workload manifest injection.

```release-note
NONE
```


instrumentor/internal/webhook_env_injector/webhook_env_injector.go
```diff
@@ -206,24 +206,10 @@ func applyOdigosEnvDefaults(container *corev1.Container, appendEnvVars []distroT
...
-		if runtimeDetails.CriErrorMessage != nil {
-			criErrorMessage = *runtimeDetails.CriErrorMessage
-		}
-		logger.Info("CRI error message present, skipping environment variable injection", "container", containerName, "message", criErrorMessage)
+	if runtimeDetails.CriErrorMessage != nil {
+		logger.Info("CRI error message present, skipping environment variable injection", "container", containerName, "message", *runtimeDetails.CriErrorMessage)
 		return false
 	}
-
-	// All conditions are satisfied
```
blumamir, line 226: reviewer note: the `runtimeDetails.RuntimeUpdateState` was used to skip injection in 2 cases:

1. `RuntimeUpdateState == nil` - should happen only in these case:
- runtime override - in this case the env vars are irrelevant and not handled by existing code anyway.
- pre-migration runtime detection - not relevant after more than 1.5 years from releasing this feature.
2. `ProcessingStateFailed` - this can only affect if cri had error, which is something we saw just once in a cluster - so not expected to affect anyone

tamirdavid1 (approved):

PR #5458, opened 2026-08-04
CORE-1441 | feat: add Recommendations

blumamir:

Add new "Recommentation" concept:
- manifests per recommendation in `github.com/odigos-io/odigos/recommendations`
- CRD in api package (to store recommendation info)
- reconciler in autoscaler
- graphql schema with query and mutation

the reconciler will sync each recommendation CR based on the manifest, set:
- `Applied` (boolean) - indicates if what the recommendation suggest is effectively applied in the cluster right now
- `ConditionsMet` (boolean) - indicates if the condition for the recommendation are met (for example - auto go offsets is relevant only if there are a `golang-enterprise` agents in the cluster).

Linear: https://linear.app/odigos/issue/CORE-1441/add-recommendation-crd-and-autoscaler-controllers

```release-note
NONE
```


api/k8sconsts/recommendation.go
```diff
@@ -0,0 +1,7 @@
+package k8sconsts
+
+const (
+	// RecommendationDismissedLabel marks a Recommendation CR as dismissed by the user.
+	// Presence with value "true" means the recommendation is dismissed.
+	RecommendationDismissedLabel = "odigos.io/recommendation-dismissed"
```
RonFed, line 6: why is this a label and not a spec field? similar to how we disable Actions

helm/odigos/templates/autoscaler/deployment.yaml
```diff
@@ -64,10 +64,7 @@ spec:
           - name: ODIGOS_LOG_LEVEL
             value: {{ .Values.autoscaler.logLevel | default .Values.logLevel | default "info" | quote }}
           - name: ODIGOS_TIER
-            valueFrom:
-              configMapKeyRef:
-                key: ODIGOS_TIER
-                name: odigos-deployment
+            value: '{{- if include "odigos.secretExists" . }}onprem{{- else }}community{{- end }}'
```
RonFed, line 67: why is this changed?

autoscaler/controllers/recommendations/applied_when.go
```diff
@@ -0,0 +1,113 @@
...
+	"sigs.k8s.io/yaml"
+
+	odigosv1 "github.com/odigos-io/odigos/api/odigos/v1alpha1"
+	"github.com/odigos-io/odigos/common/consts"
+	"github.com/odigos-io/odigos/k8sutils/pkg/env"
+	k8sutils "github.com/odigos-io/odigos/k8sutils/pkg/utils"
+	recommendationcatalog "github.com/odigos-io/odigos/recommendations"
+)
+
+func getEffectiveConfigData(ctx context.Context, c client.Client) (any, error) {
```
RonFed, line 23: we probably already have a helper for getting the effective config, prefer using it instead of duplicating

api/k8sconsts/recommendation.go
blumamir: because this one is reconciled and not managed by the user, so he cannot write things into the spec (which is managed by a reconciler)

helm/odigos/templates/autoscaler/deployment.yaml
blumamir: when user upgrades OSS -> enterprise, nothing changes in the manifest currently so autoscaler is not restarted and keeps running as OSS

autoscaler/controllers/recommendations/applied_when.go
blumamir: here it is getting it untyped (e.g. `any` which is backed by maps). the typed object does not work with JMESPath

RonFed (approved):

PR #5465, opened 2026-08-05
PLAT-1380 | feat: add Insights HTTP client

alonkeyval:
Adds the frontend transport layer required to expose the Insights REST API through GraphQL while preserving the service contract and retryable engine warm-up state.

Introduces a thin HTTP client covering the Insights query and mutation surface, with dedicated wire types, collection-envelope decoding, URL-escaped path parameters, and typed API errors. A distinct engine-starting error allows resolvers and the UI to identify retryable 503 responses without relying on error strings.

Unit tests cover endpoint escaping, query encoding, response decoding, request payloads, typed error mapping, and engine warm-up responses.

```release-note
NONE
```


alonkeyval:
Superseded by #5475 in native stack #5480.

PR #5466, opened 2026-08-05
PLAT-1381 | feat: add Insights REST ↔ GraphQL model conversions

alonkeyval:
Add REST ↔ gqlgen conversion helpers so Insights resolvers can map client wire types without embedding transport or schema quirks.

Implements `frontend/services/insights/conversions.go` for the Insights GraphQL layer: response/request round-trips between the HTTP client models and gqlgen models under `frontend/graph/model`. Transaction and policy IDs are converted between REST `int64` and GraphQL `ID` strings; free-form JSON (baseline data, anomaly evidence, weight maps, catalog tags) stays as encoded strings per schema convention; timestamps remain RFC3339 strings. The LearningCondition camelCase REST quirk is isolated here via custom JSON marshal/unmarshal so the client and resolvers stay unaware of it.

Depends on / includes [PLAT-1380](https://github.com/odigos-io/odigos/pull/5465) (Insights HTTP client).

```release-note
NONE
```


alonkeyval:
Superseded by #5476 in native stack #5480.

PR #5469, opened 2026-08-05
PLAT-1382 | feat: fill Insights GraphQL resolver stubs

alonkeyval:
The insights resolvers were gqlgen stubs that panicked, so the schema landed in PLAT-1376 could not actually serve anything. This fills all 26 of them with thin delegation to `frontend/services/insights`.

```release-note
NONE
```

PR #5471, opened 2026-08-05
PLAT-1384 | feat: expose insightsEnabled on GraphQL Config

alonkeyval:
The UI needs a cheap flag before calling nested `insights { ... }` queries, so it can hide the feature without probing APIs that return `ErrNotEnabled`.

Adds `insightsEnabled: Boolean!` to the GraphQL `Config` type and populates it in `buildConfigResponse` from the effective config via `config.InsightsEnabled()` — the same source of truth as `services/insights.IsEnabled`.

- Schema field on `Config` + gqlgen regeneration
- Value is `true` only when effective config has insights explicitly enabled; otherwise `false`
- Unit tests cover enabled / disabled / unset

Stacked on #5469 — review that one first; this branch contains its commits.

```release-note
NONE
```


alonkeyval:
Superseded by #5479 in native stack #5480.

PR #5481, opened 2026-08-06
RUN-969 | support `runAsNonRoot` on the odiglet pod

RonFed:

Support an opt-in mode to run the odiglet daremonset with `runAsNonRoot=true`.

### Helm changes
* Add `odiglet.runAsNonRoot` to the helm values.
* Add `SETUID` and `SETGID` to the capabilities required for the odiglet container when running in non-privileged mode. This capability is required in order to perform `jattach` and instrument java processes without restarting them.

### Odiglet and data-collection changes
* odiglet base image is updated to contain `libcap` and allow us to `setcap` on the binary.
* When running as non root, the executed files need to have capabilities se on them to determine the permitted capabilities set for a running process - added `setcap` both to the collector and the odiglet Dockerfiles.
* common util function to allow processes to set their effective capabilities to be equal to their permitted capabilities. This util is being used both in the data-collection collector via an extension and in the odiglet main function.

```release-note
feat: support `runAsNonRoot` on the odiglet pod
```


api/k8sconsts/collectors.go
```diff
@@ -10,3 +10,5 @@ const (
 const OdigosCollectorConfigMapProviderScheme = "k8scm"
 
 const OdigosConfigK8sExtensionType = "odigos_config_k8s"
+
+const OdigosCapabilitiesExtensionType = "odigos_capabilities"
```
blumamir: is this a k8s only extension?

collector/extension/odigoscapabilitiesextension/factory.go
blumamir, line 1: should we add a `metadata.yaml` file?

collector/extension/odigoscapabilitiesextension/extension.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/odigos-io/odigos/common/capabilities"
+)
+
+// capabilitiesExtension raises the process effective capability set to match the
+// permitted set on startup, before pipeline components (e.g. the eBPF receiver)
+// start.
+//
+// This is necessary when the collector is run as a non-root user with elevated capabilities (e.g. via `setcap`).
+// since we are using the same image, and same binary for both the node collector and the cluster collector,
+// the node collector needs eBPF related capabilities, but the cluster collector does not.
```
blumamir, line 19: consider a `README.md`

collector/Dockerfile
```diff
@@ -13,7 +13,14 @@ ARG TARGETARCH
...
 RUN --mount=type=cache,target=/go/pkg/mod \
     --mount=type=cache,target=/root/.cache/go-build \
-    GOOS=linux GOARCH=$TARGETARCH make build-odigoscol
+    CGO_ENABLED=0 GOOS=linux GOARCH=$TARGETARCH make build-odigoscol
+
+# Add file capabilities so the binary can use eBPF privileges when run as non-root.
+# Caps match odiglet.dataCollection.capabilities.
+# only setting the capabilities as permitted (p) so that the binary can be shared between the node collector (needs eBPF) and the cluster collector (does not need eBPF).
+# this means that the node collector needs to set its effective capabilities to match the permitted ones, which is done by the odigoscapabilitiesextension.
+RUN apt-get update && apt-get install -y --no-install-recommends libcap2-bin
```
blumamir: consider moving this up for better caching

blumamir: is it safe to get the latest version and not pinned?

helm/odigos/values.yaml
```diff
@@ -1073,12 +1077,20 @@ odiglet:
...
   # @schema
   unPrivileged: false
 
+  # @schema
+  # description: |-
+  #   runAsNonRoot can be set to true to run the containers in the odiglet daemonset as non-root.
+  #   when true, each container uses the non-root USER declared in its image (e.g. 1000 for odiglet).
+  #   by default (false), all containers are forced to run as UID/GID 0, matching historical behavior.
+  # @schema
+  runAsNonRoot: false
```
blumamir, line 1097: should we also document the cons? like... why someone would not want to enable it?

Otherwise it looks just like a feature that everyone would want, right?

odiglet/Dockerfile
```diff
@@ -216,4 +223,7 @@ COPY --from=builder /tmp/grpc_health_probe /root/grpc_health_probe
 COPY --from=rsync-base /usr/bin/rsync /usr/bin/rsync
 WORKDIR /instrumentations/
 COPY --from=agents-builder /instrumentations/ .
+# make sure we run as non root user
+# this is required when this container is added to a pod that has runAsNonRoot: true
+USER 1000:1000
```
blumamir: why use different numbers for the user id?

blumamir (approved):

api/k8sconsts/collectors.go
RonFed: right, moved to common

collector/extension/odigoscapabilitiesextension/extension.go
RonFed: good idea, added

odiglet/Dockerfile
RonFed: aligned the UIDs

PR #5503, opened 2026-08-10
CORE-1451 | test: add Node.js gateway tail sampling e2e

blumamir:
Adds an end-to-end test for gateway tail sampling with the nodejs-community agent. Covers dry-run sampling span attributes for error/duration traffic and VictoriaMetrics sampling counter deltas, and wires the suite into CI.

Related: https://linear.app/odigos/issue/CORE-1451/add-nodejs-gateway-tail-sampling-e2e-test

```release-note
NONE
```


.github/workflows/e2e.yaml
```diff
@@ -67,12 +67,25 @@ jobs:
           TAG=e2e-test make build-cli-image
           docker save registry.odigos.io/odigos-cli:e2e-test -o cli-image.tar
 
+      # Victoria Metrics is published, not built in this repo. Materialize a single-platform
```
natyharary, line 70: might slow down the CI but I don't see any other tway to do this

tests/common/victoriametrics_sampling_metrics_runner.sh
```diff
@@ -0,0 +1,302 @@
+#!/bin/bash
+#
+# Record and assert VictoriaMetrics sampling counter deltas (Prometheus API).
```
natyharary, line 3: where is this used? when I searched victoriametrics_sampling_metrics_runner there's no call to this shell script

natyharary (commented):
Looks good, minor comments

tests/common/victoriametrics_sampling_metrics_runner.sh
blumamir: image

.github/workflows/e2e.yaml
blumamir: We can remove once kind resolve this bug upstream

alonkeyval (approved):

PR #5532, opened 2026-08-12
CORE-1472: Use dockerhub for default OSS registry

damemi:
We are moving toward serving registry.odigos.io as an enterprise service. This updates the helm chart to use the `docker.io/keyval` images by default for OSS, unless an enterprise token is present, then it uses registry.odigos.io.

This just updates the imageName and imagePrefix helpers to be able to check if the odigos enterprise token exists. This is why `.Release` is now passed to the imageName util (so it can look up in `odigos.onPremToken`) and calls to `imagePrefix` now take more than just `.Values`

```release-note
OSS: Switch to docker.io/keyval as default container image registry
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ TAG ?= $(shell odigos version --cluster 2>/dev/null || odigos version --cli 2>/d
 ODIGOS_CLI_VERSION ?= $(shell odigos version --cli)
 CLUSTER_NAME ?= local-dev-cluster
 CENTRAL_BACKEND_URL ?=
-ORG ?= registry.odigos.io
+ORG ?= docker.io/keyval
 # Override ORG for staging pushes
 ifeq ($(STAGING_ORG),true)
     ORG = us-central1-docker.pkg.dev/odigos-cloud/staging-components
--- a/docs/snippets/enterprise/setup/docker-registry.mdx
+++ b/docs/snippets/enterprise/setup/docker-registry.mdx
@@ -14,7 +14,7 @@ Replace `<YOUR-DESIRED-VERSION>` with the version of the Odigos images you want 
 ---
 
 ### Step 2: Pull the Images
-Start by pulling the required images from the official Odigos registry onto your local machine or CI environment:
+Start by pulling the required images from the enterprise registry (`registry.odigos.io`) onto your local machine or CI environment:
 
 ```bash
 docker pull registry.odigos.io/odigos-scheduler:$VERSION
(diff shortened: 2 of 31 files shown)


tests/e2e/cli-upgrade/chainsaw-test.yaml
```diff
@@ -125,7 +125,7 @@ spec:
     - name: Upgrade to HEAD version with the current compiled cli
       try:
         - script:
-            content: ../../../cli/odigos install --ns odigos-test --set image.tag=e2e-test --set imagePrefix=registry.odigos.io
+            content: ../../../cli/odigos install --ns odigos-test --set image.tag=e2e-test
```
tamirdavid1: what about the first install in the cli upgrade? we also need to use the dockerhub?

tests/common/assert/odigos-upgraded.yaml
```diff
@@ -14,7 +14,7 @@ metadata:
 spec:
   containers:
     - name: manager
-      image: registry.odigos.io/odigos-autoscaler:e2e-test
+      image: docker.io/keyval/odigos-autoscaler:e2e-test
```
tamirdavid1, line 17: I think we need to make sure e2e not uses this common assertion file, as it will fail there

tamirdavid1:
Post merging it, we will need to enforce also the proxy to force the auth?

tamirdavid1:
we can share in slack channel a dashboard where we can track this migration numbers?

Makefile
```diff
@@ -2,7 +2,7 @@ TAG ?= $(shell odigos version --cluster 2>/dev/null || odigos version --cli 2>/d
 ODIGOS_CLI_VERSION ?= $(shell odigos version --cli)
 CLUSTER_NAME ?= local-dev-cluster
 CENTRAL_BACKEND_URL ?=
-ORG ?= registry.odigos.io
+ORG ?= docker.io/keyval
```
tamirdavid1, line 5: I didnt check that, but i guess this will handle make deploy etc to use dockerhub?
how it will affect enterprise local tests?
I mean i use to do `make deploy` in oss for oss changes (e.g autoscaler) and then `make deploy` in enterprise repo for enterprise changes (e.g instrumentor)
now we should put a different image prefix this case?
i'm asking because i want to make the development experience easier

docs/snippets/enterprise/setup/docker-registry.mdx
tamirdavid1, line 25: victoria metrics missing here by propose (I see it in the enterprise),
not related to this pr but on the way maybe should added?

tests/e2e/cli-upgrade/chainsaw-test.yaml
damemi: it doesn't make a difference functionally, we use ghcr because we thought that might help with costs and latency since it's pulling from github. but as far as the upgrade itself, no it doesn't matter

tests/common/assert/odigos-upgraded.yaml
damemi: this assert should work the same right?

Makefile
damemi: Yeah I was looking at the enterprise side and I think everything should be good there, since enterprise is going to keep using registry.odigos.io. But if you're trying to run make deploy for an OSS component while testing enterprise, you'll need to specify the different prefix. I should be able to update that here to do it automatically

docs/snippets/enterprise/setup/docker-registry.mdx
damemi: I haven't been following the victoria metrics work but if so yeah

RonFed (approved):

PR #5535, opened 2026-08-13
CORE-941 | feat: match templated HTTP routes in sampling processor

blumamir:
Sampling HTTP `route` / `routePrefix` matchers compared paths as literal strings, so templated rules (`*` / `{name}`) did not match Express `http.route` or concrete URLs.

This shares segment-based path matching in `common/urltemplate`, uses it from the tail sampling processor (and URL templatization), and extends the Node.js tail-sampling e2e with exact and prefix templated HTTP server routes.

Linear: https://linear.app/odigos/issue/CORE-941/add-option-for-templated-path-in-the-sampling-processor

```release-note
Sampling HTTP route and routePrefix rules now match templated path segments (`*` and `{name}`), not only literal strings.
```


collector/processors/odigostailsamplingprocessor/matchers/attrcompare.go
```diff
@@ -10,26 +14,17 @@ func compareHttpMethod(spanMethod string, ruleMethod string) bool {
...
-func compareHttpRoute(spanRoute string, ruleRouteExact string, ruleRoutePrefix string) bool {
-	if ruleRouteExact != "" {
-		return ruleRouteExact == spanRoute
-	}
-	if ruleRoutePrefix != "" {
-		return strings.HasPrefix(spanRoute, ruleRoutePrefix)
+// Matching is segment-based (same as url templatization): static segments must equal;
+// "*" and "{name}" match any single segment. Exact requires the same segment count;
+// prefix matches when the path starts with the rule segments.
+func compareHttpRoute(spanRoute string, ruleRouteExact urltemplate.PathRule, ruleRoutePrefix urltemplate.PathRule) bool {
```
tamirdavid1: nit: if it is return a bool maybe we should call func
isHttpRouteEqual or something like this?

collector/processors/odigostailsamplingprocessor/matchers/headsampling.go
```diff
@@ -27,16 +28,16 @@ func NewHeadSamplingOperationMatcher(operation *commonapisampling.HeadSamplingOp
 }
 
 type headSamplingHttpServerMatcher struct {
-	method      string
-	route       string
-	routePrefix string
+	method              string
+	routeSegments       urltemplate.PathRule
+	routePrefixSegments urltemplate.PathRule
```
tamirdavid1: can we add a comment about what these means?

common/urltemplate/match.go
```diff
@@ -0,0 +1,30 @@
+package urltemplate
+
+// Match reports whether pathSegments matches this rule exactly (same number of segments).
+// Static segments must equal; wildcard and templated segments match any single segment.
+func (r PathRule) Match(pathSegments []string) bool {
```
tamirdavid1:
```suggestion
func (r PathRule) IsMatch(pathSegments []string) bool {
```

tamirdavid1 (approved):

tamirdavid1 (approved):

PR #5583, opened 2026-08-17
chore(ci): source the Go version from a .go-version file

aviadhahami:
The Go toolchain version was hardcoded in **18 `setup-go` steps across 12 workflows**, in three quoting styles (`"1.26.6"`, `'1.26.6'`, `"~1.26.6"`). Bumping it meant editing every one, and a missed occurrence is silent.

## Change

Adds `.go-version` at the repo root and points every `setup-go` step at it via `go-version-file`. The version now lives in one place.

Left alone: `go-version: stable` in `update-rust-versions.yml`, and the existing `go-version-file: cli/go.mod` in `publish-modules-rhel.yml`.

## Verification

18 lines changed, 18 added — no formatting churn. All 40 workflows still parse.

```release-note
NONE
```


aviadhahami:
Closing — the team would rather not introduce a `.go-version` file. The 18 setup-go pins stay as they are; automating them directly is the alternative.

PR #5584, opened 2026-08-17
chore(ci): add a nightly workflow to keep .go-version current

aviadhahami:
⚠️ Stacked on #5583 (`.go-version`) — merge that first.

Reads the latest stable Go patch release from `go.dev/dl/?mode=json` and opens a PR when `.go-version` is behind.

## Behaviour

- **Stays on the current minor line** — never bumps across a minor version on its own
- No-ops when already current
- Warns and exits cleanly if no stable release is found on that line

## Verification

Ran the resolve logic against the live feed:

```
current=1.26.6  minor-line=1.26
latest on 1.26.x = 1.26.6
→ no PR (already current)
```

Correctly ignores `go1.25.13` on the other stable line.

## Scope

Only viable because `.go-version` is now the single source for every `setup-go` step — otherwise this would mean rewriting 18 pins across 12 workflows.

The `golang:` base images are dependabot's job (#5580). `odiglet-base` still needs a republish via the Publish Odiglet Base Builder workflow; both are called out in the PR body this workflow generates.

```release-note
NONE
```

### DIFF
--- a/.github/workflows/update-go-version.yml
+++ b/.github/workflows/update-go-version.yml
@@ -1,0 +1,77 @@
+name: Update Go Version
+
+# Keeps the Go toolchain patch version current. The version lives in .go-version
+# (consumed by every setup-go step) and in the golang: base images, which
+# dependabot handles separately.
+on:
+  schedule:
+    - cron: "0 2 * * *"
+  workflow_dispatch:
+
+permissions: read-all
+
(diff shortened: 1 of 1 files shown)


aviadhahami:
Closing along with #5583 — this depended on `.go-version`, which isn't the direction we're taking.

PR #5588, opened 2026-08-18
ECS agent docs

TheErikNaz:

ECS agent documentation 

```release-note
None
```

### DIFF
--- a/docs/vmagent/ecs/configuration.mdx
+++ b/docs/vmagent/ecs/configuration.mdx
@@ -39,8 +39,7 @@ is managed from Odigos Central or with
 
 ### Platform naming per mode
 
-What the agent reports itself as depends on the topology, so each mode reads
-naturally in the Central UI:
+The agent's name depends on the topology so each mode displays naturally in the Central UI:
 
 | Mode | Platform name | Why |
 |---|---|---|
--- a/docs/vmagent/ecs/configuration.mdx
+++ b/docs/vmagent/ecs/configuration.mdx
@@ -1,0 +1,183 @@
+---
+title: "ECS Configuration Reference"
+sidebarTitle: "Configuration Reference"
+description: "Every environment variable of the Odigos ECS Agent and ECS Proxy, how platform names are derived, where configuration is persisted, and why each elevated task permission is required."
+---
+
+import EnterpriseVMAgent from '/snippets/enterprise/vmagent/enterprise-vm-agent.mdx'
+
+<EnterpriseVMAgent />
+
+On ECS the **task definition is the configuration surface**. On a fresh
+container instance it is the only input the agent has, so the Central endpoint
+and any other defaults travel inside it — nothing depends on service discovery
+or a pre-provisioned file.
+
+Everything beyond bootstrap — [sources](/vmagent/setup/configuration/add-sources),
+[destinations](/vmagent/setup/configuration/add-destinations),
+[actions](/vmagent/setup/configuration/actions/overview), and
+[instrumentation rules](/vmagent/setup/configuration/instrumentation-rules/overview) —
+is managed from Odigos Central or with
+[`odictl`](/vmagent/setup/configuration/odictl), exactly as on any VM Agent.
+
+## Agent environment variables
+
+| Variable | Required | Effect |
+|---|---|---|
+| `ODIGOS_DEFAULT_CONFIG` | Recommended | A full config document (YAML **or** JSON, same shape as `config.yaml`) that seeds the agent on a fresh node. This is where you set the Central endpoint: `{"central":{"enabled":true,"endpoint":"central.example.com:443"}}`. In Agent + Proxy mode, point it at the proxy instead. |
+| `ODIGOS_ONPREM_TOKEN` | Yes, unless `license.jwt` is pre-provisioned | Written to `/etc/odigos-ecs-agent/license.jwt` at startup. Deliver it from AWS Secrets Manager via the task definition's `secrets` block, not as a plain `environment` value. |
+| `ODIGOS_CENTRAL_ENDPOINT` | No | Shorthand for just the endpoint (`host:port` or URL) when you don't need a full document; equivalent to setting `central.enabled` + `central.endpoint`. `ODIGOS_DEFAULT_CONFIG` wins if both are set. |
+| `ODIGOS_ECS_BEHIND_PROXY` | Agent + Proxy mode | `true` tells the agent it connects to an `odigos-ecs-proxy` rather than to Central. Set it in [Agent + Proxy](/vmagent/ecs/installation-with-proxy) mode; omit it in [Agent only](/vmagent/ecs/installation-agent-only) mode. An endpoint alone cannot distinguish the two, and the reported platform name differs per mode. |
+| `ODIGOS_AGENT_NAME` | No | Platform name reported upstream, overriding the per-mode default below. |
+| `ODIGOS_ECS_CLUSTER_NAME` | No | The ECS cluster name. Normally unnecessary — the agent reads the cluster from the ECS task metadata endpoint. Set it when metadata is unavailable, or to report a different name. |
+
+<Note>
+  `:443` in an endpoint implies `wss`. The connection is **outbound-only**, so a
+  public Central endpoint or one in a different AWS account works with no inbound
+  rules on the cluster.
+</Note>
+
+### Platform naming per mode
+
+What the agent reports itself as depends on the topology, so each mode reads
+naturally in the Central UI:
+
+| Mode | Platform name | Why |
+|---|---|---|
+| **Agent only** (no proxy) | the **ECS cluster name**, auto-detected | Each node registers with Central directly; naming the platform after the cluster is what makes a single-node cluster appear as the cluster rather than as an EC2 hostname. |
+| **Agent + Proxy** | the **node hostname** | The proxy is what reports the cluster name upstream; it identifies its agents per node, and its coverage check matches agent names against the instances' private DNS. |
+
+`ODIGOS_AGENT_NAME` overrides both. In Agent-only mode, if the cluster cannot be
+determined (task metadata unavailable), the agent falls back to the hostname and
+logs a warning.
+
+Both modes register with platform type **`aws.ecs`**, so a cluster presents
+consistently in Central either way.
+
+### Configuration precedence and persistence
+
+`ODIGOS_DEFAULT_CONFIG` is a **default**, not an override. Precedence, lowest to
+highest:
+
+<Steps>
+  <Step title="Built-in ECS distribution defaults">The baseline shipped in the image.</Step>
+  <Step title="Discrete env vars">`ODIGOS_CENTRAL_ENDPOINT`, `ODIGOS_AGENT_NAME`, and so on.</Step>
+  <Step title="ODIGOS_DEFAULT_CONFIG">The inline document from the task definition.</Step>
+  <Step title="The persisted config.yaml">Whatever Central or `odictl` has written. **This wins on every boot after the first.**</Step>
+</Steps>
+
+The config tree persists on the container **instance** at
+`/etc/odigos-ecs-agent` (a host bind mount), so the agent keeps its deployment
+identity and configuration across task restarts and image upgrades.
+
+<Warning>
+  Because the persisted `config.yaml` takes precedence, **editing
+  `ODIGOS_DEFAULT_CONFIG` does not change a value that has already been
+  persisted**. To change a seeded value after first boot, update it through
+  Central or `odictl` — or clear the persisted config on the instance so the new
+  default is applied again.
+</Warning>
+
+## Proxy environment variables
+
+| Variable | Required | Effect |
+|---|---|---|
+| `ODIGOS_CLUSTER_NAME` | Yes | The ECS cluster this proxy represents. Becomes Central's platform name and derives the proxy ID (`ecs-cluster/<name>`). |
+| `ODIGOS_CENTRAL_ENDPOINT` | No | Central `host:port`. Empty runs the proxy downstream-only (agents still aggregate against it). |
+| `ODIGOS_CENTRAL_SECURE_WS` | No | `true` forces `wss` upstream (also implied by `:443`). |
+| `ODIGOS_PLATFORM_TYPE` | No | Platform type reported upstream. Defaults to `aws.ecs`, matching the per-node agent. |
+| `ODIGOS_PROXY_ID` | No | Overrides the derived Central connection id. |
+| `ODIGOS_PROXY_LISTEN` | No | Agent listen address. Default `:4321`. |
+| `ODIGOS_PROXY_JOURNAL` | No | Config-journal path. Default `/var/lib/odigos-ecs-proxy/journal.jsonl`. Point it at an EFS mount if the journal must survive task replacement. |
+| `ODIGOS_COVERAGE_CHECK` | No | `true` enables the expected-vs-connected check: the proxy lists the cluster's container instances through the ECS/EC2 APIs and reports ACTIVE instances that have no connected agent. Requires `ecs:ListContainerInstances`, `ecs:DescribeContainerInstances`, and `ec2:DescribeInstances` on the task role. |
+| `ODIGOS_COVERAGE_INTERVAL` | No | Coverage check period (Go duration). Default `60s`. |
+
+The proxy is unprivileged — no host PID, no Docker socket, no host mounts.
+
+### What the proxy does with requests
+
+| Request | Routing |
+|---|---|
+| Config queries (`sources`, `destinations`, `actions`, `instrumentationRules`, `config`, …) | Any one healthy agent — the config is identical on every node. |
+| Live-state queries (`workloadReports`, counters) | Fan out to all agents; lists merged. |
+| Mixed queries (config + live fields in one operation) | Fan out; live lists concatenate, config fields keep a single agent's value so they are never multiplied by node count. |
+| Config mutations (`createSource`, `updateDestination`, …) | Broadcast to every agent **and journaled**. |
+| Live mutations (`restartSource`, …) | Broadcast, not journaled. |
+| Profiling jobs, diagnose bundles | One agent. |
+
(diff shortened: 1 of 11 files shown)


docs/vmagent/ecs/overview.mdx
```diff
@@ -0,0 +1,202 @@
...
+    **Several container instances that should behave as one cluster.**
+
+    The agent daemon still runs on every instance, but the agents dial a single
+    `odigos-ecs-proxy` task instead of Central. The proxy is the cluster's only
+    upstream connection, so Central shows **one** platform with aggregated
+    workloads, cluster-wide configuration, config replay for nodes that join
+    later, and an optional coverage check.
+
+    <Steps>
+      <Step title="Store the license token and create the IAM roles">Same two roles as the single-node flow, plus optional coverage permissions for the proxy.</Step>
```
rayray221, line 126: You should restate the roles from single-node instead of pointing them back to the single node tab.  Ideally they do not have to flip back and forth.

docs/vmagent/ecs/configuration.mdx
```diff
@@ -0,0 +1,183 @@
...
+
+The image bakes the Java instrumentation files into a staging directory. On
+startup the agent rsyncs them onto the container instance's `/var/odigos` (the
+task definition bind mount), and seeds them into a workload's container namespace
+when it attaches. Go, C++, and the OBI fallback are compiled into the agent
+binary and need no on-disk files.
+
+This is why `/var/odigos` must be a writable host mount.
+
+## Required permissions
```
rayray221: This section should be under requirements

docs/vmagent/ecs/installation-agent-only.mdx
```diff
@@ -0,0 +1,306 @@
...
+            "awslogs-stream-prefix": "odigos-ecs-agent"
+          }
+        }
+      }
+    ]
+  }
+  ```
+
+  Every setting that looks unusual is required for a specific capability — see
+  [required permissions](/vmagent/ecs/configuration#required-permissions) before
```
rayray221: Nit, I do not think we should comment about the permissions seaming to be unusual.  What about:

```suggestion
All settings are required for specific capabilities. For details, see [required permissions](https://www.google.com/search?q=/vmagent/ecs/configuration%23required-permissions)
```

docs/vmagent/ecs/overview.mdx
rayray221, line 1: You reference `odictl` later in the documentation, it should be mentioned in the overview as a means of configuration.

docs/vmagent/ecs/installation-agent-only.mdx
```diff
@@ -0,0 +1,306 @@
...
+  [Agent + Proxy](/vmagent/ecs/installation-with-proxy).
+</Note>
+
+## Moving to Agent + Proxy later
+
+Nothing is thrown away. Deploy the proxy, then re-point the agents' endpoint at
+it and roll the daemon service — the agents speak the same protocol to the proxy
+that they speak to Central. See
+[Install: Agent + Proxy](/vmagent/ecs/installation-with-proxy), and remember to
+add `ODIGOS_ECS_BEHIND_PROXY=true` to the agent task definition when you do.
```
rayray221: Nit, I wonder if there is a better place for this.  Seems like it's own page but it is also very little content.

docs/vmagent/ecs/installation-with-proxy.mdx
```diff
@@ -0,0 +1,543 @@
...
+## Deployment order
+
+Deploy the **proxy first**, then the agents. That way the agents connect on
+their very first boot, with the correct proxy address baked into their first task
+definition revision — no retry-loop noise and no redeploy.
+
+<Steps>
+  <Step title="Choose the proxy's stable address">Pinned instance, or an internal NLB.</Step>
+  <Step title="Deploy the proxy">One task per cluster, listening on TCP 4321.</Step>
+  <Step title="Deploy the agent daemon">Pointed at the proxy, with `ODIGOS_ECS_BEHIND_PROXY=true`.</Step>
```
rayray221: I think pointed is the wrong word here since the variable is only configuring to use a proxy not the proxy address itself.  Maybe say:

```suggestion
  <Step title="Deploy the agent daemon">Enable the proxy, with `ODIGOS_ECS_BEHIND_PROXY=true`.</Step>
```

docs/vmagent/ecs/installation-with-proxy.mdx
```diff
@@ -0,0 +1,543 @@
...
+</Info>
+
+## Deployment order
+
+Deploy the **proxy first**, then the agents. That way the agents connect on
+their very first boot, with the correct proxy address baked into their first task
+definition revision — no retry-loop noise and no redeploy.
+
+<Steps>
+  <Step title="Choose the proxy's stable address">Pinned instance, or an internal NLB.</Step>
```
rayray221: Nit, I would push for the NLB:

```suggestion
  <Step title="Choose the proxy's stable address">Internal NLB (preferred), or a pinned instance.</Step>
```

docs/vmagent/ecs/installation-with-proxy.mdx
```diff
@@ -0,0 +1,543 @@
...
+</Accordion>
+
+## Step 3: Give the proxy a stable address
+
+The agents need one fixed `host:4321` reachable from the container instances.
+That is the proxy's **only** networking requirement — no Cloud Map, no service
+discovery. Pick one of two options.
+
+<Tabs>
+  <Tab title="Option A: Pinned instance (no extra AWS resources)">
```
rayray221: Nit, I would put internal NLB as Option A and the pinned instance an Option B since the NLB is really the preferred, production method.

docs/vmagent/ecs/installation-with-proxy.mdx
```diff
@@ -0,0 +1,543 @@
...
+      `list-container-instances` returns the literal string `None`, `EC2_ID` ends
+      up empty — and an **empty** `--instance-ids` makes `describe-instances`
+      query every instance in the region, silently handing you the private IP of
+      an unrelated EC2 instance instead of failing.
+    </Warning>
+
+    **Trade-off:** if that instance is ever replaced, the proxy's address changes
+    and you must re-point both the placement constraint and the agents. See
+    [the pinned-IP caveat](#caveat-the-pinned-proxy-address).
+  </Tab>
```
rayray221: Nit, I would put this at the top of the Step 3.  IMO this method should only be used for test/dev to begin with.

docs/vmagent/ecs/installation-with-proxy.mdx
```diff
@@ -0,0 +1,543 @@
...
+  <Accordion title="Availability Zone Rebalancing disabled">
+    AZ rebalancing defaults to `ENABLED` on new services and explicitly rejects
+    `maximumPercent <= 100`, failing with
+    `Availability Zone Rebalancing does not support maximumPercent <= 100%`.
+
+    ```bash
+    --availability-zone-rebalancing DISABLED
+    ```
+  </Accordion>
+</AccordionGroup>
```
rayray221: If I understand this correctly, these steps are only needed for Option A, pinned instance.  Why not put this in the tab as well?

docs/vmagent/ecs/installation-with-proxy.mdx
```diff
@@ -0,0 +1,543 @@
...
+aws ecs register-task-definition --cli-input-json file:///tmp/agent-td.json >/dev/null && echo "agent task definition registered"
+```
+
+<Warning>
+  **`ODIGOS_ECS_BEHIND_PROXY=true` matters.** The agent cannot tell a proxy
+  address from a Central address — both are just `host:port` — and the two modes
+  report different platform names. Behind the proxy, each agent identifies itself
+  by **node hostname** (the proxy is what reports the cluster name upstream, and
+  its coverage check matches agent names against the instances' private DNS).
+  Leave it `false` here and every agent registers under the cluster name instead,
```
rayray221: Nit

```suggestion
  If you leave it `false` here and every agent registers under the cluster name instead,
```

docs/vmagent/ecs/installation-with-proxy.mdx
```diff
@@ -0,0 +1,543 @@
...
+   new revision, and `update-service --force-new-deployment`.
+
+If that fragility is unacceptable, use **Option B** (internal NLB) — it survives
+instance and task replacement at the cost of one extra AWS resource.
+
+## Current limits
+
+- **Journal durability** — the config journal lives on the proxy task's local
+  storage (`/var/lib/odigos-ecs-proxy/journal.jsonl`). Point
+  `ODIGOS_PROXY_JOURNAL` at an EFS mount if it must survive task replacement.
```
rayray221: We should have more content around setting ODIGOS_PROXY_JOURNAL to an EFS mount.  Perhaps an example?

docs/vmagent/ecs/installation-with-proxy.mdx
```diff
@@ -0,0 +1,543 @@
...
+  Steady state is unaffected either way: agents persist their configuration
+  locally, and the journal only closes the gap for nodes that were absent when a
+  change was made.
+- **Per-node workload entries** — workload lists are concatenated across nodes,
+  so a service with tasks on several nodes currently appears once per node.
+- **Single-node targeting** — profiling jobs and diagnose bundles are served by
+  one agent; targeting a specific node is not yet supported.
```
rayray221, line 575: I'm not sure I understand this.  Does this mean that you have to go to the individual node for profiling and diagnose?

docs/vmagent/ecs/configuration.mdx
```diff
@@ -0,0 +1,183 @@
...
+<Note>
+  `:443` in an endpoint implies `wss`. The connection is **outbound-only**, so a
+  public Central endpoint or one in a different AWS account works with no inbound
+  rules on the cluster.
+</Note>
+
+### Platform naming per mode
+
+What the agent reports itself as depends on the topology, so each mode reads
+naturally in the Central UI:
```
rayray221: Nit,

```suggestion
The agent's name depends on the topology so each mode displays naturally in the Central UI:
```

docs/vmagent/ecs/configuration.mdx
```diff
@@ -0,0 +1,183 @@
...
+  persisted**. To change a seeded value after first boot, update it through
+  Central or `odictl` — or clear the persisted config on the instance so the new
+  default is applied again.
+</Warning>
+
+## Proxy environment variables
+
+| Variable | Required | Effect |
+|---|---|---|
+| `ODIGOS_CLUSTER_NAME` | Yes | The ECS cluster this proxy represents. Becomes Central's platform name and derives the proxy ID (`ecs-cluster/<name>`). |
```
rayray221, line 84: Is this required if it is derived?

docs/vmagent/ecs/configuration.mdx
```diff
@@ -0,0 +1,183 @@
...
+
+| Setting | Reason |
+|---|---|
+| `pidMode: host` | Classifying and attaching to other tasks' processes via `/proc/<pid>`. |
+| `privileged: true` | eBPF program loading, perf events, and writing instrumentation files into target containers via `/proc/<pid>/root`. |
+| `networkMode: host` | The collector and OpAMP server listen on the instance's network namespace, reachable by instrumented workloads on localhost or the instance IP. |
+| `/var/run/docker.sock` (ro) | Container metadata: mapping PIDs to ECS tasks, service names, task ARNs. |
+| `/var/lib/docker` (ro) | Resolving executable paths through overlay2 for language detection. |
+| `/var/odigos` (rw) | Instrumentation files synced to the instance on startup; the source seeded into workload containers when attaching Java. |
+| `/etc/odigos-ecs-agent` (rw) | Agent configuration and identity, persisted across task restarts and upgrades. |
```
rayray221: Nit,
ro and rw should be explicitly written out to read-only and read-write

docs/vmagent/ecs/maintenance.mdx
```diff
@@ -0,0 +1,183 @@
...
+    <Warning>
+      With a **pinned** proxy (host networking), the service must already have
+      `minimumHealthyPercent=0,maximumPercent=100` and
+      `--availability-zone-rebalancing DISABLED`, or the rollout gets stuck: ECS
+      tries to place the new task before stopping the old one, and both want port
+      `4321` on the same instance. If the deployment is already stuck, apply the
+      deployment configuration and stop the old task once to unblock it.
+    </Warning>
+
+    Upgrade the proxy and agents to the same version.
```
rayray221, line 63: Does it matter which is upgraded first?  Is there a best practice?

docs/vmagent/ecs/maintenance.mdx
```diff
@@ -0,0 +1,183 @@
...
+
+<Note>
+  Changing an environment variable follows the exact same flow: new revision,
+  then `update-service --force-new-deployment`. Remember that values already
+  persisted in `config.yaml` on the instance
+  [take precedence](/vmagent/ecs/configuration#configuration-precedence-and-persistence)
+  over `ODIGOS_DEFAULT_CONFIG`.
+</Note>
+
+## Switch from single node to multi node
```
rayray221: Similar to my other comment, I think this should have its own page under the Installation section.  On this page and the other reference on the agent only install could have links to the dedicated page.

docs/vmagent/ecs/configuration.mdx
```diff
@@ -0,0 +1,183 @@
...
+there is no `systemd` inside the container:
+
+- `odigos-otelcol` runs as a **supervised child process** of the agent, so its
+  lifecycle follows the agent task.
+- Agent and collector **share the task's stdout**, captured by the `awslogs`
+  driver into the `/odigos/ecs-agent` log group.
+- With host networking, the collector and OpAMP server listen on the instance's
+  network namespace, reachable by instrumented tasks on localhost or the instance
+  IP.
+- Collector logs are JSON-formatted on ECS (the `awslogs` driver is the primary
```
rayray221: You mention in the troubleshooting that this goes to cloudwatch, you should mention it here too.

docs/vmagent/ecs/installation-agent-only.mdx
```diff
@@ -0,0 +1,306 @@
...
+  <Card title="Actions" icon="wand-magic-sparkles" href="/vmagent/setup/configuration/actions/overview">
+    Transform, filter, and enrich telemetry.
+  </Card>
+  <Card title="Configuration Reference" icon="sliders" href="/vmagent/ecs/configuration">
+    ECS-specific environment variables and permissions.
+  </Card>
+</CardGroup>
+
+<Note>
+  In this mode `odictl` targets **one node's** agent — there is no cluster-wide
```
rayray221: Nit,

```suggestion
  In this mode `odictl` targets **single node's** agent — there is no cluster-wide
```

docs/vmagent/ecs/troubleshooting.mdx
```diff
@@ -0,0 +1,161 @@
...
+aws ecs execute-command --cluster "$CLUSTER" --task "$TASK" \
+  --container odigos-ecs-agent --interactive --command odictl
+```
+
+A shell works the same way (`--command /bin/bash`), which is useful for
+`uname -r`, checking `/sys/kernel/btf/vmlinux`, or reading
+`/etc/odigos-ecs-agent/config.yaml`.
+
+<Note>
+  `odictl` always targets **one node's** agent — there is no cluster-wide
```
rayray221: Nit,

```suggestion
  `odictl` always targets **single node's** agent — there is no cluster-wide
```

docs/vmagent/ecs/troubleshooting.mdx
```diff
@@ -0,0 +1,161 @@
...
+
+aws ecs execute-command --cluster "$CLUSTER" --task "$PTASK" \
+  --container odigos-ecs-proxy --interactive --command "/bin/bash"
+```
+
+```bash
+exec 3<>/dev/tcp/127.0.0.1/4321
+printf 'GET /healthz HTTP/1.0\r\n\r\n' >&3
+cat <&3
+```
```
rayray221, line 67: Nit,
Would be good to include example output.

docs/vmagent/ecs/troubleshooting.mdx
```diff
@@ -0,0 +1,161 @@
...
+
+    Fix: move the instances to the **Amazon Linux 2023** ECS-optimized AMI (kernel
+    6.1), or at minimum the AL2 **kernel-5.10** ECS AMI. See
+    [System Requirements](/vmagent/ecs/requirements#container-instance-kernel-and-ami).
+  </Accordion>
+
+  <Accordion title="The daemon task is not running on every instance">
+    ```bash
+    aws ecs describe-services --cluster "$CLUSTER" --services odigos-ecs-agent \
+      --query 'services[0].{running:runningCount,desired:desiredCount,events:events[:5].message}'
```
rayray221: Nit,

```suggestion
Run the following:
    ```bash
    aws ecs describe-services --cluster "$CLUSTER" --services odigos-ecs-agent \
      --query 'services[0].{running:runningCount,desired:desiredCount,events:events[:5].message}'
```

docs/vmagent/ecs/configuration.mdx
TheErikNaz: It is not derived, What is derived is ODIGOS_PROXY_ID, from that value

docs/vmagent/ecs/installation-agent-only.mdx
TheErikNaz: I see your point, but it is really a transition from single node to multi node, 
I think leaving it here is okay

docs/vmagent/ecs/installation-with-proxy.mdx
TheErikNaz: Exactly
Diagnose can be easy to implement, I'm not sure profiling is worth the trouble until it isn't a requirement

docs/vmagent/ecs/maintenance.mdx
TheErikNaz: Shouldn't matter, can be also rolled out as a one shot upgrade

rayray221 (approved):

PR #5661, opened 2026-08-24
Added docs for AWS connector

TheErikNaz:

AWS connector docs

```release-note
AWS connector docs
```

### DIFF
--- a/docs/cloud-connectors/add-connectors/add-aws-connector.mdx
+++ b/docs/cloud-connectors/add-connectors/add-aws-connector.mdx
@@ -16,9 +16,21 @@ Add an AWS Cloud Connector to manage AWS accounts. The connector runs in the Cen
 | Type | Domain | Discovery | Instrumentation |
 | --- | --- | --- | --- |
 | `aws.lambda` | Workload | Yes | Yes (OpenTelemetry Lambda layer) |
-| `aws.fargate-task` | Workload | Yes | Coming soon |
+| `aws.fargate-task` | Workload | Yes | Yes (agent-staging sidecar in the task definition) |
 | `aws.ecs-cluster` | Compute platform | Yes | Coming soon |
 | `aws.fargate-cluster` | Compute platform | Yes | Coming soon |
+
+<Info>
+  `aws.ecs-cluster` and `aws.fargate-cluster` are discovered for coverage today;
+  Odigos does not yet install a separate agent onto them. `aws.fargate-task` — an
+  individual Fargate **service** — is the ECS resource Odigos instruments directly.
+</Info>
+
+Both `aws.lambda` and `aws.fargate-task` can be instrumented **Odigos-managed** (the connector applies and
+maintains the change) or **self-managed** via your own CloudFormation, CDK, or Terraform — see
+[Odigos-managed vs. self-managed instrumentation](/cloud-connectors/overview#odigos-managed-vs-self-managed-instrumentation).
+For the self-managed path, artifact preloading, and per-language CDK/CloudFormation snippets, see the
+[AWS Connector](/cloud-connectors/aws/overview) section.
 
 ## Add the connector
 
@@ -47,20 +59,29 @@ Add an AWS Cloud Connector to manage AWS accounts. The connector runs in the Cen
         | --- | --- |
         | Authentication | `sts:GetCallerIdentity` |
         | Region enumeration | `ec2:DescribeRegions` |
-        | Lambda discovery | `lambda:ListFunctions` |
-        | ECS / Fargate discovery | `ecs:ListClusters`, `ecs:DescribeClusters`, `ecs:ListServices`, `ecs:DescribeServices`, `ecs:ListTasks`, `ecs:DescribeTasks` |
+        | Lambda discovery | `lambda:ListFunctions`, `lambda:ListTags` |
+        | ECS / Fargate discovery | `ecs:ListClusters`, `ecs:DescribeClusters`, `ecs:ListServices`, `ecs:DescribeServices`, `ecs:ListTasks`, `ecs:DescribeTasks`, `ecs:DescribeTaskDefinition` |
+
+        <Info>
+          `ecs:DescribeTaskDefinition` is a discovery permission, not only an
+          instrumentation one — it's what lets Odigos see whether a self-managed
+          Fargate service already carries your own instrumentation, so grant it
+          even if you only ever plan to self-manage instrumentation.
+        </Info>
       </Tab>
       <Tab title="Discovery + instrumentation">
         The connector can discover resources and instrument supported workloads. Prefer this for a full setup in one step.
 
-        Includes all discovery actions, plus Lambda instrumentation:
+        Includes all discovery actions, plus Odigos-managed instrumentation for Lambda and Fargate:
 
         | Purpose | Actions |
         | --- | --- |
         | Read and update function config | `lambda:GetFunctionConfiguration`, `lambda:UpdateFunctionConfiguration` |
         | Manage OTel layers | `lambda:GetLayerVersion`, `lambda:ListLayers`, `lambda:PublishLayerVersion`, `lambda:DeleteLayerVersion` |
+        | Manage Fargate task definitions | `ecs:RegisterTaskDefinition`, `ecs:UpdateService` |
+        | Pass task/execution roles when registering a task definition | `iam:PassRole` (scope with condition `iam:PassedToService=ecs-tasks.amazonaws.com`) |
 
-        Lambda instrumentation also requires an [OTLP destination](/cloud-connectors/instrument-workloads#add-an-otlp-destination) after the connector is online.
+        Odigos-managed instrumentation also requires an [OTLP destination](/cloud-connectors/instrument-workloads#add-an-otlp-destination) after the connector is online. If you plan to self-manage instrumentation instead (your own CloudFormation/CDK/Terraform), you only need the discovery actions above — see [AWS Connector](/cloud-connectors/aws/overview).
       </Tab>
       <Tab title="Custom">
         Pick which resource types may be discovered and which may be instrumented. The UI updates the **policy preview** to match your selections so the IAM policy stays aligned with what verification will validate.
@@ -209,10 +230,19 @@ Add an AWS Cloud Connector to manage AWS accounts. The connector runs in the Cen
 
 ## Next steps
 
-<Card
-  title="Instrument workloads"
-  icon="microchip"
-  href="/cloud-connectors/instrument-workloads#add-an-otlp-destination"
->
-  Add an OTLP destination, create Sources for discovered Lambdas, and track instrumentation.
-</Card>
+<CardGroup cols={2}>
+  <Card
+    title="Instrument workloads"
+    icon="microchip"
+    href="/cloud-connectors/instrument-workloads#add-an-otlp-destination"
+  >
+    Add an OTLP destination, create Sources for discovered Lambdas and Fargate services, and track instrumentation.
+  </Card>
+  <Card
+    title="AWS Connector"
+    icon="aws"
+    href="/cloud-connectors/aws/overview"
+  >
+    Self-managed instrumentation, artifact preloading, and CDK/CloudFormation for every language.
+  </Card>
+</CardGroup>
--- a/docs/cloud-connectors/add-connectors/add-aws-connector.mdx
+++ b/docs/cloud-connectors/add-connectors/add-aws-connector.mdx
@@ -26,9 +26,12 @@ Add an AWS Cloud Connector to manage AWS accounts. The connector runs in the Cen
   individual Fargate **service** — is the ECS resource Odigos instruments directly.
 </Info>
 
-Both `aws.lambda` and `aws.fargate-task` can be instrumented **Odigos-managed** (the connector applies and
-maintains the change) or **self-managed** via your own CloudFormation, CDK, or Terraform — see
-[Odigos-managed vs. self-managed instrumentation](/cloud-connectors/overview#odigos-managed-vs-self-managed-instrumentation).
+aws.lambda and aws.fargate-task support two instrumentation approaches:
+- Odigos-managed: The connector automatically applies and maintains the required changes.
+
+- Self-managed: You maintain full control by provisioning through your own CloudFormation, CDK, or Terraform pipelines.
+
+(For a deeper comparison, see [Odigos-managed vs. self-managed instrumentation](/cloud-connectors/overview#odigos-managed-vs-self-managed-instrumentation).)
 For the self-managed path, artifact preloading, and per-language CDK/CloudFormation snippets, see the
 [AWS Connector](/cloud-connectors/aws/overview) section.
 
--- a/docs/cloud-connectors/add-connectors/add-aws-connector.mdx
+++ b/docs/cloud-connectors/add-connectors/add-aws-connector.mdx
@@ -84,7 +84,7 @@ For the self-managed path, artifact preloading, and per-language CDK/CloudFormat
         | Manage Fargate task definitions | `ecs:RegisterTaskDefinition`, `ecs:UpdateService` |
         | Pass task/execution roles when registering a task definition | `iam:PassRole` (scope with condition `iam:PassedToService=ecs-tasks.amazonaws.com`) |
 
-        Odigos-managed instrumentation also requires an [OTLP destination](/cloud-connectors/instrument-workloads#add-an-otlp-destination) after the connector is online. If you plan to self-manage instrumentation instead (your own CloudFormation/CDK/Terraform), you only need the discovery actions above — see [AWS Connector](/cloud-connectors/aws/overview).
+        Odigos-managed instrumentation also requires an [OTLP destination](/cloud-connectors/instrument-workloads#add-an-otlp-destination) after the connector is online.
       </Tab>
       <Tab title="Custom">
         Pick which resource types may be discovered and which may be instrumented. The UI updates the **policy preview** to match your selections so the IAM policy stays aligned with what verification will validate.
--- a/docs/cloud-connectors/aws/cdk/fargate.mdx
+++ b/docs/cloud-connectors/aws/cdk/fargate.mdx
@@ -1,0 +1,307 @@
+---
+title: 'AWS CDK Reference for Fargate'
+description: 'Full reference for the OdigosFargateInstrumentation construct in @odigos/aws-cdk / odigos-aws-cdk: how it works, every option, and the edge cases to know about.'
+sidebarTitle: 'Fargate'
+---
+
+This page is the complete reference for the Odigos CDK construct introduced in
+[Instrument Fargate](/cloud-connectors/aws/workloads/fargate/instrument#aws-cdk). Read that page first
+for the quick version; come here when you need an option this page's basic example doesn't cover.
+
+<Info>
+  This is the **Fargate** construct. For Lambda, see [CDK → Lambda](/cloud-connectors/aws/cdk/lambda) —
+  the two share the same package but work differently under the hood (see the comparison below).
+</Info>
+
+## How Fargate differs from Lambda here
+
+| | Lambda (`OdigosLambdaInstrumentation`) | Fargate (`OdigosFargateInstrumentation`) |
+| --- | --- | --- |
+| Mechanism | A layer, plus an exec wrapper or loader variable | Agents staged into a shared task volume by a run-to-completion container |
+| Languages | Python, Node.js, Java, Ruby | Java, Python, Node.js, .NET, PHP, Ruby |
+| Where telemetry goes | A collector in the same Lambda sandbox, which forwards after the response returns | Your destination, directly from the application process |
+| Language detection | From the function's declared runtime | None — you state the language explicitly |
+
+Both write `ODIGOS_IAC` and never the connector's own ownership marker (`ODIGOS_MANAGED`): instrumented by
+Odigos, owned by you.
+
+## Install and basic usage
+
+```bash
+npm install @odigos/aws-cdk        # TypeScript / JavaScript
+pip install odigos-aws-cdk         # Python
+```
+
+<CodeGroup>
+
+```typescript TypeScript
+import { OdigosFargateInstrumentation, OdigosFargateLanguage } from '@odigos/aws-cdk';
+
+const odigos = new OdigosFargateInstrumentation(this, 'Odigos', {
+  destination: { httpEndpoint: 'https://otlp.example.com:4318' },
+});
+
+odigos.instrument({
+  container: apiContainer,          // the ecs.ContainerDefinition to instrument
+  language: OdigosFargateLanguage.JAVA,
+  service: apiService,              // reads OTEL_SERVICE_NAME off the ECS service
+});
+```
+
+```python Python
+from odigos_aws_cdk import OdigosFargateInstrumentation, OdigosFargateLanguage
+
+odigos = OdigosFargateInstrumentation(self, "Odigos",
+    destination={"http_endpoint": "https://otlp.example.com:4318"},
+)
+odigos.instrument(
+    container=api_container,
+    language=OdigosFargateLanguage.JAVA,
+    service=api_service,
+)
+```
+
+</CodeGroup>
+
+## What it does to the task definition
+
+Four moves, and they only work as one mechanism:
+
+1. A bare-name task volume, `odigos-agent` — on Fargate this is an ephemeral-storage bind mount scoped to
+   the task (there's no tmpfs).
+2. A **non-essential** `odigos-agent` container that copies the language agent out of the agents bundle into
+   that volume, then exits successfully.
+3. That volume mounted read-only at `/var/odigos` in your application container, which `DependsOn` the
+   staging container with condition `SUCCESS` — legal only against a non-essential container, which is why
+   (2) is non-essential.
+4. The language's activation environment, the OTLP export environment, and the `ODIGOS_IAC` marker, all set
+   on your application container.
+
+<Warning>
+  Marker presence alone is a false green here. A container can carry a complete,
+  correct activation environment while the staging container is missing or
+  misconfigured — the application then starts, finds nothing at `/var/odigos`, and
+  exports nothing while still reporting as instrumented. That's why all four moves
+  are one construct call rather than four options you could apply separately.
+</Warning>
+
+## Where the telemetry goes
+
+Straight to your destination, from inside the application process. **This is the substantive difference
+from Lambda**, where the SDK hands batches to a local collector and the export happens after the response
+has already returned. A Fargate task has no such collector, so a slow or unreachable destination is felt
+directly by the application's own exporter.
+
+## Naming the service
+
+`OTEL_SERVICE_NAME` **must equal the workload's name in Odigos — for a Fargate workload, that's the ECS
+service name**, not the task-definition family and not the container name. A different value is reported as
+instrumentation drift, not treated as a naming preference — so there's no default. Pass `service` to read
+the name off an `ecs.FargateService` construct, or `serviceName` when the service is defined elsewhere.
+
+<Warning>
+  The service's physical name has to be one you set explicitly. A name
+  CloudFormation generates is a `Ref` to the service, and the task definition that
+  service runs can't reference it back — CloudFormation rejects that cycle at
+  deploy time — so a generated name **throws at synth**, pointing you at
+  `serviceName`.
+</Warning>
+
+## Languages, and stating them
+
+<ParamField path="language" type="OdigosFargateLanguage" required>
+  One of `JAVA`, `PYTHON`, `NODEJS`, `DOTNET`, `PHP`, `RUBY`.
+</ParamField>
+
+| Language | Activated through | Needs `runtimeVersion`? |
+| --- | --- | --- |
+| `JAVA` | `JAVA_TOOL_OPTIONS` | No |
+| `PYTHON` | `PYTHONPATH` | No |
+| `NODEJS` | `NODE_OPTIONS` | No |
+| `DOTNET` | `CORECLR_*` / `DOTNET_*` (glibc images only) | No |
+| `PHP` | `PHP_INI_SCAN_DIR` | **Yes** — major.minor, e.g. `"8.3"` |
+| `RUBY` | `RUBYOPT` | **Yes** — major.minor, e.g. `"3.3"` |
+
+There's no aspect and no auto-detection for Fargate, and unlike Lambda that's not a matter of taste: a
+Lambda function declares its runtime, but an ECS container just declares an image. A language guessed from
+an image would silently mis-instrument whatever it guessed wrong — a Java agent staged into a Python
+container leaves the container uninstrumented, and `JAVA_TOOL_OPTIONS` set on a JVM-less image is simply
+ignored, so nothing fails loudly enough to notice.
+
+<Warning>
+  PHP and Ruby need `runtimeVersion` because their agents are laid down per exact
+  runtime version inside the bundle, and the version you pass selects the path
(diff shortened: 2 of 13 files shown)


docs/cloud-connectors/aws/cdk/overview.mdx
```diff
@@ -0,0 +1,217 @@
...
+description: 'What the Odigos AWS CDK package is, when to use it, and the concepts shared by the Lambda and Fargate constructs.'
+sidebarTitle: 'Overview'
+---
+
+The Odigos AWS CDK package adds OpenTelemetry instrumentation to your AWS workloads from inside your own
+[AWS CDK](https://aws.amazon.com/cdk/) application — so instrumentation ships through the same pipeline as
+the rest of your infrastructure, reviewed in a pull request and applied by `cdk deploy`.
+
+<Info>
+  **New to this?** AWS CDK is a tool for describing cloud infrastructure in a
```
tamirdavid1, line 12: maybe move that above The `Odigos AWS CDK...` ?

docs/cloud-connectors/aws/cdk/overview.mdx
```diff
@@ -0,0 +1,217 @@
...
+
+## When you'd use this
+
+Odigos can instrument your workloads two ways, and the CDK package is for the second one:
+
+| | Odigos-managed | Self-managed, with this package |
+| --- | --- | --- |
+| Who changes the workload | Odigos, automatically | Your CDK app, on `cdk deploy` |
+| Where instrumentation is defined | In Odigos | In your own source code |
+| What Odigos does | Applies and maintains it | Watches and reports only |
```
tamirdavid1:
```suggestion
| What Odigos does | Applies and maintains it | Watches and reports |
```

docs/cloud-connectors/aws/cdk/overview.mdx
```diff
@@ -0,0 +1,217 @@
...
+
+<ParamField path="destination.metrics" type="boolean" default="false">
+  Metrics — numeric measurements over time. **Off unless you turn it on.**
+</ParamField>
+
+<ParamField path="destination.logs" type="boolean" default="false">
+  Logs. **Off unless you turn it on.**
+</ParamField>
+
+```typescript
```
tamirdavid1: is this example? maybe worth adding example title?

tamirdavid1: Image

docs/cloud-connectors/aws/cdk/overview.mdx
```diff
@@ -0,0 +1,217 @@
...
+destination: {
+  endpoint: 'otlp.example.com:4317',
+  metrics: true,   // off unless you ask
+  logs: true,      // off unless you ask
+}
+```
+
+Metrics and logs are off by default because sending a signal your destination doesn't accept is not free:
+every batch is rejected and then retried on a backoff, and on Lambda that retrying happens inside the time
+you're billed for. Traces are what nearly every destination accepts; the other two depend on your specific
```
tamirdavid1: I think that 
```
Traces are what nearly every destination accepts; the other two depend on your specific destination, so they’re a deliberate choice.
```
is not really true, most of the destination supports logs, I understand why we are setting traces on by default but is it because this reason?

docs/cloud-connectors/aws/cdk/lambda.mdx
```diff
@@ -0,0 +1,239 @@
+---
+title: 'CDK: Lambda'
```
tamirdavid1, line 2: General comment (also for Fargate), lets add some screenshots?
The UI with your additions really shows everything in this doc right?

docs/cloud-connectors/aws/cdk/fargate.mdx
```diff
@@ -0,0 +1,262 @@
...
+});
+```
+
+You create the construct once, then call `instrument()` for each container you want instrumented. Unlike
+Lambda, there's no aspect that sweeps a whole stack — you name each container.
+
+## How it works
+
+Fargate has no equivalent of a Lambda layer, so the agent files have to be put into the task at startup.
+The construct sets up four things that only work together:
```
tamirdavid1:
```suggestion
The construct sets up four things that work together:
```

docs/cloud-connectors/aws/cdk/fargate.mdx
```diff
@@ -0,0 +1,262 @@
...
+  All four are one mechanism, which is why they're one construct call rather than
+  four options. A container can have a perfectly correct set of environment
+  variables while the copy step is missing — the application then starts, finds
+  nothing at `/var/odigos`, and sends no telemetry while still *reporting* as
+  instrumented.
+</Warning>
+
+## Where telemetry goes
+
+**Straight from your application to your destination.** This is the substantive difference from Lambda,
```
tamirdavid1: i dont think we need to mention anything about lambda in this doc, if a user only uses fargate and open this doc there are multiple lambda references here

docs/cloud-connectors/aws/cdk/fargate.mdx
```diff
@@ -0,0 +1,262 @@
...
+
+Practically, that means the destination endpoint has to be reachable from the task's network.
+
+## Stating the language
+
+<ParamField path="language" type="OdigosFargateLanguage" required>
+  One of `JAVA`, `PYTHON`, `NODEJS`, `DOTNET`, `PHP`, `RUBY`.
+</ParamField>
+
+| Language | Loaded through | Needs `runtimeVersion`? |
```
tamirdavid1: the `needs runtimeVersion`, you mention No for python and Java, the sdk probably supports a range of versions

docs/cloud-connectors/aws/cdk/fargate.mdx
```diff
@@ -0,0 +1,262 @@
...
+}
+```
+
+Every agent in the bundle prefers HTTP. Java, Python, Node.js, and .NET will fall back to gRPC if that's all
+you provide. Giving both lets each language use what it prefers.
+
+<Warning>
+  **PHP and Ruby can only speak HTTP** — their community SDKs support nothing else.
+  Instrumenting a PHP or Ruby container against a gRPC-only destination stops the
+  build, rather than deploying a task that silently sends nothing.
```
tamirdavid1:
```suggestion
  **PHP and Ruby can only speak HTTP** —
  Instrumenting a PHP or Ruby container against a gRPC-only destination stops the
  build, rather than deploying a task that silently sends nothing.
```

docs/cloud-connectors/aws/cdk/fargate.mdx
```diff
@@ -0,0 +1,262 @@
...
+  failing later during deployment.
+</Info>
+
+## The agents image
+
+The copy step pulls its files from an image — by default the one published by this package's release,
+built for both Intel and ARM/Graviton.
+
+<Warning>
+  **This image is pulled by your task's execution role every time a task starts.** A
```
tamirdavid1, line 208: there is an option to manually pulling it to some private registry? if not maybe worth mention that we're not supporting fargate tasks in without it being able to reach the public registry?

docs/cloud-connectors/aws/workloads/lambda/instrument.mdx
```diff
@@ -0,0 +1,249 @@
+---
+title: 'Instrument Lambda with CDK or CloudFormation'
+description: 'Self-managed AWS Lambda instrumentation for Java, Python, Node.js, and Ruby using the Odigos CDK construct or hand-written CloudFormation.'
+sidebarTitle: 'Instrument Lambda'
+---
+
+AWS Lambda functions can be instrumented two ways: **Odigos-managed**, where the connector attaches
```
tamirdavid1, line 7: lets add some screenshots?

docs/cloud-connectors/add-connectors/add-aws-connector.mdx
```diff
@@ -16,10 +16,22 @@ Add an AWS Cloud Connector to manage AWS accounts. The connector runs in the Cen
...
 
+<Info>
+  `aws.ecs-cluster` and `aws.fargate-cluster` are discovered for coverage today;
+  Odigos does not yet install a separate agent onto them. `aws.fargate-task` — an
+  individual Fargate **service** — is the ECS resource Odigos instruments directly.
+</Info>
+
+Both `aws.lambda` and `aws.fargate-task` can be instrumented **Odigos-managed** (the connector applies and
+maintains the change) or **self-managed** via your own CloudFormation, CDK, or Terraform — see
+[Odigos-managed vs. self-managed instrumentation](/cloud-connectors/overview#odigos-managed-vs-self-managed-instrumentation).
```
rayray221: This come across a little unclear.  Maybe bullet it?


```suggestion
aws.lambda and aws.fargate-task support two instrumentation approaches:
- Odigos-managed: The connector automatically applies and maintains the required changes.

- Self-managed: You maintain full control by provisioning through your own CloudFormation, CDK, or Terraform pipelines.

(For a deeper comparison, see [Odigos-managed vs. self-managed instrumentation](/cloud-connectors/overview#odigos-managed-vs-self-managed-instrumentation).)
```

docs/cloud-connectors/aws/cdk/overview.mdx
TheErikNaz: wdym? I don't really follow

docs/cloud-connectors/aws/cdk/overview.mdx
TheErikNaz: It is not an example, this is required to define, what signals to collect.
The connector validates it against the already created Destinations and reports a drift if there is a signal mis-match

docs/cloud-connectors/aws/cdk/lambda.mdx
TheErikNaz: There are multiple changes that are going to come, I think we should hold with the screenshots until we finish polishing the connector imp

docs/cloud-connectors/aws/cdk/fargate.mdx
TheErikNaz: yes there is 
And it is defined in the docs

docs/cloud-connectors/aws/workloads/lambda/instrument.mdx
TheErikNaz: Same as I said, lets finish the feature and the UI and only then do the screenshots

docs/cloud-connectors/add-connectors/add-aws-connector.mdx
```diff
@@ -47,20 +59,29 @@ Add an AWS Cloud Connector to manage AWS accounts. The connector runs in the Cen
...
 
         | Purpose | Actions |
         | --- | --- |
         | Read and update function config | `lambda:GetFunctionConfiguration`, `lambda:UpdateFunctionConfiguration` |
         | Manage OTel layers | `lambda:GetLayerVersion`, `lambda:ListLayers`, `lambda:PublishLayerVersion`, `lambda:DeleteLayerVersion` |
+        | Manage Fargate task definitions | `ecs:RegisterTaskDefinition`, `ecs:UpdateService` |
+        | Pass task/execution roles when registering a task definition | `iam:PassRole` (scope with condition `iam:PassedToService=ecs-tasks.amazonaws.com`) |
 
-        Lambda instrumentation also requires an [OTLP destination](/cloud-connectors/instrument-workloads#add-an-otlp-destination) after the connector is online.
+        Odigos-managed instrumentation also requires an [OTLP destination](/cloud-connectors/instrument-workloads#add-an-otlp-destination) after the connector is online. If you plan to self-manage instrumentation instead (your own CloudFormation/CDK/Terraform), you only need the discovery actions above — see [AWS Connector](/cloud-connectors/aws/overview).
```
rayray221: Putting the line:
If you plan to self-manage instrumentation instead (your own CloudFormation/CDK/Terraform), you only need the discovery actions above — see [AWS Connector](/cloud-connectors/aws/overview)

After the line regarding the OTLP destination seems strange.  I'm not sure why this line needs to be here at all.

docs/cloud-connectors/overview.mdx
```diff
@@ -43,15 +43,46 @@ Connectors organize resources into two domains:
 
 You choose which resource types to discover and instrument when you add a connector (per-type capabilities).
 
+## Odigos-managed vs. self-managed instrumentation
+
+Every workload type supports two ways of getting instrumented, chosen per type when you add or edit a connector:
```
rayray221: nit

```suggestion
When adding or editing a connector, choose an instrumentation method for each workload type:
```

docs/cloud-connectors/overview.mdx
```diff
@@ -43,15 +43,46 @@ Connectors organize resources into two domains:
...
+
+| Discovery | Instrumentation | Mode | Who applies the change |
+| --- | --- | --- | --- |
+| On | On | **Odigos-managed (automatic)** | Odigos attaches instrumentation for you (a Lambda layer + environment, or a sidecar injected into an ECS task definition) and keeps it in place on every reconcile cycle. |
+| On | Off | **Self-managed (IaC)** | You apply the change yourself — CloudFormation, CDK, Terraform, or by hand. Odigos only discovers and reports status; it never modifies the resource. |
+| Off | — | Not discovered | The type is not surfaced at all. |
```
rayray221, line 68: nit:
Do we even need this?  It just stating that nothing has been done.

docs/cloud-connectors/overview.mdx
```diff
@@ -43,15 +43,46 @@ Connectors organize resources into two domains:
...
+
+There is no separate "mode" toggle — it's just whether you turned on the instrumentation capability for that type. This is what makes self-managed workloads safe by default: until you grant instrumentation access, Odigos cannot touch the resource no matter what else is enabled.
```
rayray221: Nit, I would word this something like:

```suggestion
Rather than using a dedicated mode toggle, Odigos controls access through per-type instrumentation settings. This ensures self-managed workloads are safe by default—Odigos cannot touch or configure a resource unless its instrumentation capability is active.
```

docs/cloud-connectors/overview.mdx
rayray221, line 1: nit, Prerequisites are usually close to the top since a customer may not be able to fulfill them thus making the rest of the document irrelevant.

docs/cloud-connectors/aws/overview.mdx
```diff
@@ -0,0 +1,131 @@
...
+* **Fargate** has no node-local collector — each instrumented task exports **directly** to the destination
+  over its own network path. The destination endpoint must be reachable from the task's VPC (public IP or
+  NAT for a registry pull, plus whatever egress the destination itself needs).
+
+## Next steps
+
+<CardGroup cols={2}>
+  <Card title="Instrument workloads" icon="microchip" href="/cloud-connectors/instrument-workloads">
+    Add an OTLP destination and create Sources for Odigos-managed instrumentation.
+  </Card>
```
rayray221: nit, Seems disconnected.  I wonder if we should remove the generic Instrument workloads page and point to the service specific one?

docs/cloud-connectors/aws/overview.mdx
```diff
@@ -0,0 +1,131 @@
...
+
+  <Card title="Odigos AWS CDK" icon="code" href="/cloud-connectors/aws/cdk/overview">
+    Instrument from your own CDK app — both Lambda and Fargate.
+  </Card>
```
rayray221: nit, why is this the next logical step?  What if they aren't using cdk?  Shouldn't they be linked to lambda/fargate instrumentation sections?

docs/cloud-connectors/aws/workloads/lambda/preload.mdx
```diff
@@ -0,0 +1,79 @@
+---
+title: 'Preload the Lambda Layer'
+description: 'Preload the Lambda OpenTelemetry layer into your own AWS account before instrumenting with IaC.'
+sidebarTitle: 'Preload'
+---
+
+If you use **Odigos-managed** Lambda instrumentation, you can skip this page — the connector publishes the
+layer version into your account itself (`lambda:PublishLayerVersion`) the first time it instruments a
+function. Nothing to preload.
```
rayray221: nit, this should probably be Mintlify Note or Tip.

docs/cloud-connectors/aws/workloads/lambda/preload.mdx
```diff
@@ -0,0 +1,79 @@
...
+
+If you **self-manage** instrumentation, you have two choices for where the layer your CloudFormation, CDK,
+or Terraform attaches actually lives:
+
+<Tabs>
+  <Tab title="Reference the public Odigos layer">
+    Simplest option: attach the public layer ARN Odigos publishes in every region
+    (`arn:aws:lambda:<region>:<odigos-account>:layer:odigos-otel-<language>-<arch>:<n>`) directly in your
+    function resource. No preloading, no extra deploy — see
+    [Instrument Lambda](/cloud-connectors/aws/workloads/lambda/instrument). Skip this option if your organization's
```
rayray221: nit, this should be more prescriptive. 
```suggestion
    function resource. No preloading, no extra deploy — for next steps see
    [Instrument Lambda](/cloud-connectors/aws/workloads/lambda/instrument). Skip this option if your organization's
```

docs/cloud-connectors/aws/workloads/lambda/preload.mdx
```diff
@@ -0,0 +1,79 @@
...
+    policy forbids attaching a Lambda layer owned by another AWS account.
```
rayray221: If the companies policy forbids it, shouldn't this point them to the Preloading tab?

docs/cloud-connectors/aws/workloads/lambda/preload.mdx
```diff
@@ -0,0 +1,79 @@
...
+  </Tab>
+
+  <Tab title="Preload into your own account (SAR)">
+    Deploy the Odigos layer as a Serverless Application Repository (SAR) application first, so the
+    `AWS::Lambda::LayerVersion` it creates is owned by **your own** CloudFormation stack. Then reference that
+    stack's `LayerVersionArn` output everywhere you'd otherwise use the public ARN.
```
rayray221: I wonder if we can point them to AWS documentation for preloading instead of putting in our own write up.  What you have are not very detailed and it is not really an Odigos configuration but an AWS one.

docs/cloud-connectors/aws/workloads/fargate/preload.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+    Nothing to configure — this is what Odigos-managed instrumentation uses, and it's also the default a
+    self-managed CloudFormation task definition references unless you override it.
+  </Tab>
+
+  <Tab title="Your own private registry">
+    Mirror the image into a registry your task's execution role can already reach — typically a private ECR
+    repository in the same account.
+
+    <Steps>
```
rayray221, line 30: nit, I think you should specify "For example" since these are not definitive steps for them to perform.

docs/cloud-connectors/aws/workloads/fargate/instrument.mdx
```diff
@@ -0,0 +1,426 @@
+---
+title: 'Instrument Fargate with CloudFormation'
+description: 'Self-managed ECS Fargate instrumentation for Java, Python, Node.js, .NET, PHP, and Ruby using hand-written CloudFormation, or the Odigos AWS CDK construct.'
```
rayray221: nit, this is only CloudFormation, we should remove the AWS CDK contract from the description.

docs/cloud-connectors/aws/workloads/fargate/instrument.mdx
```diff
@@ -0,0 +1,426 @@
...
+and reports its status once your changes are in place; see
+[Odigos-managed vs. self-managed instrumentation](/cloud-connectors/overview#odigos-managed-vs-self-managed-instrumentation)
+for how that works. Want the connector to do this automatically instead? Enable the instrumentation
+capability for `aws.fargate-task` on the connector.
+
+<Info>
+  Prefer the Odigos CDK construct? See
+  [CDK: Fargate](/cloud-connectors/aws/cdk/fargate). This page is the
+  hand-written CloudFormation / raw ECS form of the same change.
+</Info>
```
rayray221: This is redundant.  I would suggest:


```suggestion
This page shows how to instrument an ECS **service** running on Fargate yourself via CloudFormation, by editing its task
definition directly — no changes made by the connector on your behalf. Odigos still discovers the service
and reports its status once your changes are in place; see
[Odigos-managed vs. self-managed instrumentation](/cloud-connectors/overview#odigos-managed-vs-self-managed-instrumentation)
for how that works. 

<Info>
  Prefer the Odigos CDK construct? See
  [CDK: Fargate](/cloud-connectors/aws/cdk/fargate). This page is the
  hand-written CloudFormation / raw ECS form of the same change.
</Info>
```

docs/cloud-connectors/aws/workloads/fargate/instrument.mdx
```diff
@@ -0,0 +1,426 @@
...
+  must be reachable from the task's network. See
+  [Telemetry export](/cloud-connectors/aws/overview#telemetry-export).
+</Info>
+
+## CloudFormation
+
+Merge these properties into your task definition. The four moves from [How it works](#how-it-works) above
+look like this in CloudFormation — add a volume, add the `odigos-agent` container, mount that volume
+read-only on your application container with a `SUCCESS` dependency, and add the environment variables.
+Only the exact agent path, activation variable, and network protocol change per language. Pick yours below:
```
rayray221: I would suggest removing the reference to the previous steps, it is not needed.


```suggestion
Merge these properties into your task definition. Only the exact agent path, activation variable, and network protocol change per language. Pick yours below:
```

docs/cloud-connectors/aws/workloads/fargate/instrument.mdx
```diff
@@ -0,0 +1,426 @@
...
+## AWS CDK
+
+Prefer a construct over hand-written task-definition changes? Use
+[`@odigos/aws-cdk`](/cloud-connectors/aws/cdk/fargate) — it adds the staging container, volume, mount,
+dependency and environment in one call. The full option list and per-language activation details live on that
+page.
+
+If you need to stay on plain `aws-cdk-lib/aws-ecs`, apply the same four moves shown in the CloudFormation tabs
+above (`addVolume`, non-essential stager container, read-only mount + `SUCCESS` dependency, and the
+environment block passed to `addContainer`).
```
rayray221: nit, this is not needed here it is redundant

docs/cloud-connectors/aws/cdk/overview.mdx
rayray221, line 1: I think Choosing a destination and Choosing which signals to send should be there own page called Destinations.

docs/cloud-connectors/aws/cdk/lambda.mdx
```diff
@@ -0,0 +1,239 @@
...
+## Basic usage
+
+```typescript
+import { OdigosLambdaInstrumentation } from '@odigos/aws-cdk';
+
+new OdigosLambdaInstrumentation(this, 'Odigos', {
+  destination: { endpoint: 'otlp.example.com:4317' },
+  functions: [apiFn, workerFn],
+});
+```
```
rayray221, line 25: Do you define a Odigos construct per cdk stack?  I assume this is the case, perhaps we should explicitly state this?

docs/cloud-connectors/aws/cdk/lambda.mdx
```diff
@@ -0,0 +1,239 @@
...
+
+Both options install byte-for-byte identical instrumentation. Use `IN_ACCOUNT` if your organization's policy
+forbids attaching a Lambda layer owned by an outside AWS account — see
+[Preload the Lambda layer](/cloud-connectors/aws/workloads/lambda/preload) for what that actually deploys.
+It needs the principal running `cdk deploy` to hold `serverlessrepo:CreateCloudFormationTemplate` and
+`serverlessrepo:GetApplication`.
+
+<ParamField path="layerVersion" type="lambda.ILayerVersion">
+  If you mirror the layer somewhere yourself, pass it here directly and neither option above is used.
+</ParamField>
```
rayray221, line 130: This is not clear.  Are you saying that if you pass `layerVersion` that the `layerSource` is not used?  What are validated `layerVersion`?

docs/cloud-connectors/aws/cdk/lambda.mdx
```diff
@@ -0,0 +1,239 @@
...
+    CDK can't read back an environment variable your app set elsewhere, so this
+    can't be detected automatically. Omit it and the variable is set to **only**
+    the Odigos value — for `NODE_OPTIONS`, that can silently drop a flag your
+    application needs to start.
+  </Warning>
+</ParamField>
+
+<Info>
+  Ruby has no verified web-server loader, so Ruby functions are rejected in
+  `WEB_SERVER` mode.
```
rayray221: By rejected I believe you mean unsupported?  If so I think this should read:

```suggestion
Ruby functions are not supported in WEB_SERVER mode because a verified web-server loader is currently unavailable.
```

docs/cloud-connectors/aws/cdk/lambda.mdx
```diff
@@ -0,0 +1,239 @@
...
+
+That fallback exists for a structural reason: a function name that CloudFormation generates for you is a
+reference to the function itself, and a function referring to its own generated name inside its own
+configuration is a circular dependency that CloudFormation refuses to deploy.
+
+<Warning>
+  Odigos compares this against the function's real AWS name. **If your function's
+  name is auto-generated, set `serviceName` explicitly** — otherwise Odigos reports
+  it as drifted. Note that `serviceName` applies to *every* function in a single
+  construct, so use one construct per function when you need different names.
```
rayray221, line 238: I think this should also be mentioned at the top where you have the first example.  I would need to define a construct per function if I need the serviceName to be function specific.

docs/cloud-connectors/aws/cdk/lambda.mdx
```diff
@@ -0,0 +1,239 @@
...
+| `OdigosLambdaInstrumentation` | The construct. Methods: `instrument(fn)`, `instrumentAll(fns)`. |
+| `OdigosInstrumentationAspect` | Applies instrumentation to every function in a scope. |
+| `OdigosLayerSource` | `ODIGOS_PUBLIC` \| `IN_ACCOUNT` |
+| `OdigosInstrumentationMode` | `HANDLER` \| `WEB_SERVER` |
+| `OdigosJavaHandlerDistro` | The four Java handler wrappers. |
+| `renderCollectorConfig` | The same collector-config renderer the construct uses, if you want to inspect what it produces. |
+| `LAYER_ARNS`, `ODIGOS_VERSION` | The layer-ARN table baked into the package, and the Odigos release it came from. |
+| `SAR_APPLICATIONS`, `SAR_SEMANTIC_VERSION` | The Serverless Application Repository applications backing `IN_ACCOUNT`, and their pinned version. |
+
+## Next steps
```
rayray221: These aren't really next steps, maybe use:
```suggestion
## Related Topics
```

docs/cloud-connectors/aws/cdk/fargate.mdx
```diff
@@ -0,0 +1,261 @@
...
+  the default public registry, and the pull failure will block your application
+  container from starting.
+</Warning>
+
+<ParamField path="agentsImage" type="string">
+  Point the copy step at your own mirror of the image:
+
+  ```typescript
+  agentsImage: '111122223333.dkr.ecr.eu-west-1.amazonaws.com/odigos-fargate-agents:v0.0.15',
+  ```
```
rayray221, line 220: I assume the default is the public registry?  Should't this be mentioned?  I would make a note that this section is only relevant if they cannot pull from the public registry.

docs/cloud-connectors/aws/cdk/fargate.mdx
```diff
@@ -0,0 +1,261 @@
...
+
+  See [Preload the Fargate agents image](/cloud-connectors/aws/workloads/fargate/preload) for how to build
+  that mirror.
+</ParamField>
+
+A mirror has to be the same bundle — the environment variables name exact paths inside it. The bundle is
+Linux and glibc only: a Windows task definition stops the build, and Alpine/musl images have no .NET
+profiler available.
```
rayray221: Nit, this assumes a lot for the customer to know about bundles. Would be helpful to be more descriptive about what a bundle is and how it is created.  Even just a link to AWS documentation would be helpful..

docs/cloud-connectors/aws/cdk/fargate.mdx
```diff
@@ -0,0 +1,261 @@
...
+
+| Export | What it is |
+| --- | --- |
+| `OdigosFargateInstrumentation` | The construct. Methods: `instrument(target)`, `instrumentAll(targets)`. |
+| `OdigosFargateTarget` | One container to instrument: `container`, `language`, `service` or `serviceName`, `runtimeVersion`, `existingLoaderValue`. |
+| `OdigosFargateLanguage` | `JAVA` \| `PYTHON` \| `NODEJS` \| `DOTNET` \| `PHP` \| `RUBY` |
+| `sidecarCommand`, `evaluateAppendPattern`, `appendSeparator` | The copy-step command and loader-value rendering, if you want to inspect what they produce. |
+| `FARGATE_AGENTS_IMAGE` | The agents image this package version uses by default. |
+| `FARGATE_DISTROS_VERSION` | The Odigos distros release the activation recipes were mirrored from. |
+| `FARGATE_AGENTS_DIR`, `FARGATE_VOLUME_NAME`, `FARGATE_SIDECAR_CONTAINER_NAME` | The mount path, volume name, and copy-step container name. |
```
rayray221: Why are these relevant if they are not mentioned elsewhere in the document?  If we are going to mention them we need more details about where they are and how to use/interpret them.

docs/cloud-connectors/aws/cdk/lambda.mdx
```diff
@@ -0,0 +1,239 @@
...
+| Export | What it is |
+| --- | --- |
+| `OdigosLambdaInstrumentation` | The construct. Methods: `instrument(fn)`, `instrumentAll(fns)`. |
+| `OdigosInstrumentationAspect` | Applies instrumentation to every function in a scope. |
+| `OdigosLayerSource` | `ODIGOS_PUBLIC` \| `IN_ACCOUNT` |
+| `OdigosInstrumentationMode` | `HANDLER` \| `WEB_SERVER` |
+| `OdigosJavaHandlerDistro` | The four Java handler wrappers. |
+| `renderCollectorConfig` | The same collector-config renderer the construct uses, if you want to inspect what it produces. |
+| `LAYER_ARNS`, `ODIGOS_VERSION` | The layer-ARN table baked into the package, and the Odigos release it came from. |
+| `SAR_APPLICATIONS`, `SAR_SEMANTIC_VERSION` | The Serverless Application Repository applications backing `IN_ACCOUNT`, and their pinned version. |
```
rayray221: Why are these relevant if they are not mentioned elsewhere in the document?  If we are going to mention them we need more details about where they are and how to use/interpret them.

docs/cloud-connectors/aws/cdk/fargate.mdx
TheErikNaz: Yeah, it is not talking about support, but if the instrumentation requires that value to instrument

docs/cloud-connectors/aws/workloads/lambda/preload.mdx
TheErikNaz: But it is really about getting OUR artifacts into THEIR account

docs/cloud-connectors/overview.mdx
TheErikNaz: Yeah, it is a possible and valid option
Doesn't cost us anything :)

rayray221 (approved):

PR #5639, opened 2026-08-20
CORE-1520: Support mTLS in OTLP destinations

damemi:
#### What this PR does / why we need it:
This adds support for mTLS fields in OTLP destinations (grpc/http). Currently using mTLS requires a dynamic destination and manual secret config, this puts that all in the ui.

Tested with sample app at https://github.com/damemi/mtls-test-app
added to a shared repo https://github.com/odigos-io/synthetic-apps/pull/8

Screenshot 2026-08-20 at 11 17 00 AM

cherry-pick:v1.35

#### Changelog entry: Does this PR introduce a user-facing bug fix, feature, dependency update, or breaking change??

```release-note
feat(destinations): support mTLS in OTLP destinations
```

PR #5674, opened 2026-08-25
CORE-1520: Use unique destination secret env vars

damemi:
#5639 added support for more secrets in OTLP destinations, and AI review of this change uncovered that a broader issue is present where the collector config secret env vars are not unique across destinations. This updates them to use unique variables so that secrets don't conflict

cherry-pick:v1.36

```release-note
fix(destinations): Use unique destination secret env vars in collector config
```


common/config/secret_env.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// SecretEnvPlaceholder returns a collector config placeholder that expands from
+// the destination-scoped env var, e.g. ${ODIGOS_DEST_odigos_io_dest_otlp_abc_DATADOG_API_KEY}.
+func SecretEnvPlaceholder(fieldName string, dest ExporterConfigurer) string {
+	return "${" + SecretEnvVarName(fieldName, dest.GetID()) + "}"
+}
+
+// SanitizeEnvIdent maps an arbitrary destination ID to a Kubernetes C_IDENTIFIER
+// fragment (letters, digits, underscore).
```
RonFed, line 32: is this correct with https://www.kubernetes.dev/resources/keps/4369/ ?

RonFed (approved):

BenElferink (approved):
Question:
Does this work also for Destinations configured before the update?

TheErikNaz (approved):

common/config/secret_env.go
damemi: I actually don't think I needed to sanitize the env var names at all since they're just underscores/letters. It's the secret names that are important, so I updated this to use the same santitization function as writing the secrets where relevant

PR #5686, opened 2026-08-26
CORE-1545: (insights): GraphQL support for open guardrail allowlists

damemi:
Expose insights.serviceNames for caller/callee pickers and plumb it through the insights client and webapp adapter so guardrails can combine cluster-wide service names with per-service observed profile data.

For guardrails, we currently only allow a guardrail to be created with a relation to services that have already be seen. Meaning, you can only set allowed/denied caller/callee/egress for callers, callees, and egresses that are already connected to the target service.

This means if you want to define a "violation", you essentially have to wait for that violation to be seen already. This opens up the guardrail page to allow rules to be created for any already seen service (not just services that are already seen connected to the target service). It also adds a free-form text box alongside allowed egress.

other PRs:
https://github.com/odigos-io/ui-kit/pull/1254
https://github.com/odigos-io/odigos-enterprise/pull/3559

```release-note
NONE
```


frontend/graph/insights.graphqls
```diff
@@ -950,6 +843,11 @@ type InsightsStorageHealth {
 
 type Insights {
   services: [InsightsServiceStat!]!
+  """
+  Every distinct service name Insights has seen. Cluster-wide candidate set for
+  guardrail caller/callee allowlists — fetch once and combine with a service profile.
+  """
+  serviceNames: [String!]!
```
BenElferink, line 957: I'm wondering if we should just use the service names coming back from the "get sources" query, instead of adding bloat onto this one 🤔

frontend/graph/insights.resolvers.go
```diff
@@ -704,3 +641,26 @@ func (r *queryResolver) Insights(ctx context.Context) (*model.Insights, error) {
 func (r *Resolver) Insights() InsightsResolver { return &insightsResolver{r} }
 
 type insightsResolver struct{ *Resolver }
+
+// !!! WARNING !!!
+// The code below was going to be deleted when updating resolvers. It has been copied here so you have
```
alonkeyval: :point_up:

alonkeyval:
need to fix the resolver + support the query on the webapp

frontend/graph/insights.graphqls
damemi: So we could, but there is a difference because "get sources" gets things we are instrumenting in odigos, while getting the services gets things that have been discovered through insights based on the spans we've seen. We actually might want to combine both for the broadest prefill coverage

BenElferink: Ahh I see, like those "virtual" nodes seen in the blast radius... Alright sounds good to me :)

BenElferink (approved):

PR #5719, opened 2026-09-01
CORE-1334 | docs: Add and update VM agent documentations to comply with the new non-root version changes and explanations

liranm-odigos:

Added documentations to show how the new non-root VM agent is deployed and managed.
Also updated existing docs that will reflect the new changes.

```release-note
NONE
```


docs/vmagent/setup/configuration/odictl.mdx
```diff
@@ -26,6 +26,17 @@ You will then see a screen similar to the one below:
   alt="Initial odictl TUI screen"
 />
 
+## Install a license
+
+Apply the Odigos Pro token the support team issued. `odictl` validates the JWT before writing `/var/lib/odigos-agent/config/license.jwt`, then you restart the agent:
```
tamirdavid1, line 31: user can either write it directly to the license.jwt file or use odigos license right? worth to mention?

docs/vmagent/setup/configuration/odictl.mdx
```diff
@@ -26,6 +26,17 @@ You will then see a screen similar to the one below:
...
+
+```shell
+sudo odictl -odigos-license '<JWT>'
```
tamirdavid1, line 38: I think the dash is not in the right format, try to copy and use it

docs/vmagent/setup/configuration/add-sources.mdx
```diff
@@ -14,6 +14,10 @@ There are two ways to add [sources](/vmagent/overview#key-concepts) to the Odigo
   **The only exception is Go and Java processes running directly on the host (not in a container).** These are instrumented in place and are **not** restarted.
 </Warning>
 
+<Note>
+  A **non-root** agent needs extra host permissions for some source kinds. Add `odigos` to the `docker` group to instrument and restart containers; add it to `adm` or `systemd-journal` to export `LOGS` from systemd units; and install a polkit rule if the agent should `daemon-reload` and restart systemd services. See [Linux Permissions](/vmagent/setup/permissions).
```
rayray221: Nit, I would not specify the specific's here, I would just say that a non-root agent requires additional permissions to restart services, containers and processes.  See Linux Permissions.  For the Linux Permissions, I would group them into a parent; something like "Control Permissions" so this link can be to that specific section.

docs/vmagent/setup/configuration/overview.mdx
```diff
@@ -48,7 +48,11 @@ When a [destination](/vmagent/setup/configuration/add-destinations) is configure
 |---|---|
 | `process` | Scans the process PID file descriptors for paths containing `*.log` |
 | `docker` | Streams Docker logs for the instrumented container |
-| `systemd` | Streams logs via `journalctl` for the service |
+| `systemd` | Streams logs via the journald receiver for the service |
+
+<Note>
+  A non-root agent can only read the systemd journal if the `odigos` user is in `adm` and/or `systemd-journal`. See [Linux Permissions](/vmagent/setup/permissions#systemd-journal-group-logs).
+</Note>
```
rayray221: I think the whole log collection should be before the Configuration tasks.  To be fair, my original PR was the one to put it after, but I think it would be better to move it before if you would.

docs/vmagent/setup/installation.mdx
```diff
@@ -65,50 +65,145 @@ To install the Odigos VM Agent, configure your package manager or obtain the RPM
...
+
+| Mode | When to use |
+|---|---|
+| **Non-root** (default) | Typical hosts. The agent runs as `odigos` with a fixed set of Linux capabilities. |
+| **Root** | You need the agent to restart systemd units without a [polkit rule](/vmagent/setup/permissions#polkit-restart-systemd-sources), or the host cannot grant ambient capabilities (systemd older than 229). |
+
+**Precedence** (first match wins):
+
+1. `ODIGOS_AGENT_USER_MODE=root` or `nonroot` in the environment of the package script (`apt` / `dnf` / `rpm`)
+2. Sticky file `/var/lib/odigos-agent/user-mode` (written automatically when the env is set)
```
rayray221: I assume that the value in the file should be either `root` or `nonroot`?  We should explicitly state that.  For customers who want to continue using root or need to use root for fresh installs, this is probably the best way for them to ensure this without worrying about the env var.

docs/vmagent/setup/installation.mdx
```diff
@@ -65,50 +65,145 @@ To install the Odigos VM Agent, configure your package manager or obtain the RPM
...
+3. Heuristic: non-root — including an upgrade of a host that is still on root
```
rayray221: I'm not sure what you mean by this.  Is it just that we always default to `nonroot` if neither of the above is set to `root`? If that is the case we should just say 
If neither condition applies, default to `nonroot` for both new installations and upgrades.

docs/vmagent/setup/installation.mdx
```diff
@@ -65,50 +65,145 @@ To install the Odigos VM Agent, configure your package manager or obtain the RPM
...
+
+Hosts whose systemd is older than 229 (RHEL 7, CentOS 7, default Amazon Linux 2) cannot grant ambient capabilities. The installer **forces root** on those hosts, even if you request non-root. Check with `/usr/lib/odigos/bin/odigos-vmagent-caps --explain`.
```
rayray221: This is technically part of the precedence and should be in the list above.

docs/vmagent/setup/installation.mdx
rayray221: I see you have the specifics below, is this case can you add:

```suggestion
2. Sticky file `/var/lib/odigos-agent/user-mode` (written automatically when the env is set; see below)
```

docs/vmagent/setup/installation.mdx
```diff
@@ -65,50 +65,145 @@ To install the Odigos VM Agent, configure your package manager or obtain the RPM
...
+echo root | sudo tee /var/lib/odigos-agent/user-mode
+# or
+echo nonroot | sudo tee /var/lib/odigos-agent/user-mode
+```
+
+To migrate an existing root host without waiting for the next upgrade:
+
+```shell
+sudo /usr/lib/odigos/bin/odigos-migrate-to-nonroot --dry-run
+sudo /usr/lib/odigos/bin/odigos-migrate-to-nonroot
```
rayray221: Nit, this should probably be it's own section and maybe under upgrade instead of install since you would have to have it installed to perform this.

docs/vmagent/setup/installation.mdx
```diff
@@ -65,50 +65,145 @@ To install the Odigos VM Agent, configure your package manager or obtain the RPM
...
+
+`odictl` validates the token **before** writing it, so an invalid JWT cannot replace a working license. The file is stored at `/var/lib/odigos-agent/config/license.jwt` (the same path in root and non-root mode).
 
-Write the license token received from the Odigos support team to the file `/etc/odigos-vmagent/license.jwt`
 ```shell
-echo "<license-token>" | sudo tee /etc/odigos-vmagent/license.jwt
+sudo odictl -odigos-license '<JWT>'
 ```
 
+Wrap the token in quotes so the shell does not alter it. Do not write the JWT into the file by hand — use `odictl` so it is validated first.
```
rayray221: Nit
```suggestion
Wrap the token in single quotes so the shell does not alter it. Do not write the JWT into the file by hand — use `odictl` so it is validated first.
```

docs/vmagent/setup/installation.mdx
```diff
@@ -65,50 +65,145 @@ To install the Odigos VM Agent, configure your package manager or obtain the RPM
...
+
+<Note>
+  The config directory is owned by the agent user (`odigos` in non-root mode, `root` in root mode) and is not world-writable. Use `sudo` for `odictl -odigos-license` and for any YAML you place under `/var/lib/odigos-agent/config`.
```
rayray221, line 157: Wouldn't sudo be unnecessary in `nonroot` mode?

docs/vmagent/setup/installation.mdx
```diff
@@ -65,50 +65,145 @@ To install the Odigos VM Agent, configure your package manager or obtain the RPM
...
 
 If you run Odigos Central, follow [Connecting VM Agent](/central/adding-connections/vmagent) to link this host to Central.
+
+## Layout
+
+Config, license, sources, destinations, actions, and rules live under one tree for both root and non-root:
+
+| Path | Purpose |
+|---|---|
+| `/var/lib/odigos-agent/config` | Agent configuration and `license.jwt` |
```
rayray221: Nit, to make sure users know it is a directory:

```suggestion
| `/var/lib/odigos-agent/config/` | Agent configuration and `license.jwt` |
```

docs/vmagent/setup/installation.mdx
```diff
@@ -65,50 +65,145 @@ To install the Odigos VM Agent, configure your package manager or obtain the RPM
...
+| `/var/lib/odigos-agent/user-mode` | Sticky `root` / `nonroot` preference |
+| `/var/lib/odigos-agent/log` | Agent logs in non-root mode |
+| `/var/odigos` | Instrumentation files (both modes; do not relocate) |
+| `/etc/odigos-otelcol` | Collector configuration |
```
rayray221: Same as above:

```suggestion
| `/var/lib/odigos-agent/log/` | Agent logs in non-root mode |
| `/var/odigos/` | Instrumentation files (both modes; do not relocate) |
| `/etc/odigos-otelcol/` | Collector configuration |
```

docs/vmagent/setup/installation.mdx
```diff
@@ -65,50 +65,145 @@ To install the Odigos VM Agent, configure your package manager or obtain the RPM
...
+| `/etc/systemd/system/odigos-vmagent.service.d/nonroot.conf` | Non-root user, capabilities, and environment |
+
+Older installs used `/etc/odigos-vmagent`. Upgrades copy that tree into `/var/lib/odigos-agent/config` and then delete `/etc/odigos-vmagent`.
```
rayray221: Nit, I do not think this is necessary in the Install section since we have an Update section.

docs/vmagent/setup/system-requirements.mdx
```diff
@@ -23,3 +19,24 @@ import KernelVersion from '../../../snippets/shared/ebpf-kernel-version-note.mdx
...
+| RHEL 7, CentOS 7, Amazon Linux 2 | 3.10 / 4.14 | 219 | Root only (`AmbientCapabilities` unavailable) |
+| RHEL 8, Rocky 8 | 4.18 | 239+ | Non-root |
+| RHEL 9, Rocky 9, Amazon Linux 2023 | 5.14+ | 250+ | Non-root |
+| Debian 11–13, Ubuntu 20.04+ | 5.4+ | 245+ | Non-root |
+
+Confirm what this host can grant:
+
+```shell
+/usr/lib/odigos/bin/odigos-vmagent-caps --explain
+```
```
rayray221, line 40: Nit, an example of the output would be nice

docs/vmagent/setup/upgrade.mdx
```diff
@@ -33,7 +33,31 @@ To upgrade the Odigos VM Agent, run the following based on your Linux package ma
...
+    sudo rpm -Uvh odigos-otelcol-<otelcol-version>.rpm odigos-<agent-version>.rpm
     ```
   </Tab>
-</Tabs>
\ No newline at end of file
+</Tabs>
+
+## What an upgrade changes
+
+Every upgrade defaults to **non-root**, including a host that has been running as root. The upgrade performs the same migration as `/usr/lib/odigos/bin/odigos-migrate-to-nonroot`:
```
rayray221: We need the VM Agent version here.  
```suggestion
Starting with VM Agent v0.x.x, the installation defaults to **non-root**, including a host that was previously installed running as root. The upgrade performs the same migration as `/usr/lib/odigos/bin/odigos-migrate-to-nonroot`:
```

Also, I'm not sure I understand what you mean but "the same migration as `/usr/lib/odigos/bin/odigos-migrate-to-nonroot`:"  What is in this file?  I do not see it anywhere else in the documentation.

docs/vmagent/setup/upgrade.mdx
```diff
@@ -33,7 +33,31 @@ To upgrade the Odigos VM Agent, run the following based on your Linux package ma
...
+
+- Moves leftover `/etc/odigos-vmagent/*` into `/var/lib/odigos-agent/config/`, then deletes `/etc/odigos-vmagent`
+- Moves agent logs from `/var/log` into `/var/lib/odigos-agent/log/`
+- Chowns `/var/odigos` and `/etc/odigos-otelcol` to the `odigos` user
+- Adds `odigos` to the `docker` group when that group exists
+- Disables `odigos-otelcol.service` (the collector becomes a child of the agent)
+- Re-renders `/etc/systemd/system/odigos-vmagent.service.d/nonroot.conf` unless you replaced its `# Managed by the odigos-vmagent package` header
```
rayray221, line 50: I believe we still migrate the data even if we stay on root, this needs to be stated here as well.  Maybe a note/hint?

docs/vmagent/setup/permissions.mdx
```diff
@@ -0,0 +1,304 @@
...
+
+By default the VM Agent runs as the unprivileged `odigos` user, not as root. It still needs a small, fixed set of Linux capabilities, and — depending on which sources you instrument — membership in a few host groups. This page covers what is granted out of the box, why, and how to change it.
+
+<Info>
+  Root-mode installs (`ODIGOS_AGENT_USER_MODE=root`) already have every privilege. Skip this page unless you are tightening a non-root install or switching a host to non-root. See [Installation](/vmagent/setup/installation#root-and-non-root).
+</Info>
+
+## Linux capabilities
+
+The packaged drop-in `/etc/systemd/system/odigos-vmagent.service.d/nonroot.conf` grants the same capability set on every distribution that can run non-root. The installer does **not** pick capabilities from `kernel.perf_event_paranoid`, and it never assigns `CAP_BPF`, `CAP_PERFMON`, or `CAP_CHECKPOINT_RESTORE`.
```
rayray221: Nit, I think stating:
"The installer does **not** pick capabilities from `kernel.perf_event_paranoid`, and it never assigns `CAP_BPF`, `CAP_PERFMON`, or `CAP_CHECKPOINT_RESTORE`."

Here is a bit confusing since we have not mentioned anything about the specific capabilities yet.  I would put this in a Info box after the list of capabilities if at all.

docs/vmagent/setup/permissions.mdx
```diff
@@ -0,0 +1,304 @@
...
+| `CAP_CHOWN` | Hand `odigos-otelcol` secret and log files to the `odigos` user. |
+| `CAP_DAC_OVERRIDE` | Write root-owned host files such as `/etc/ld.so.preload` (dynamic-loader instrumentation). |
+| `CAP_FOWNER` | Replace root-owned scratch left by an earlier root-mode agent in sticky (`1777`) directories: `/tmp/keeplist` and each container's `/tmp/ebpf-java-*-dynamic.json`. Only the owner may unlink or rename there; `CAP_DAC_OVERRIDE` does not lift that rule. |
+| `CAP_NET_RAW` | Open the `AF_PACKET` socket the eBPF instrumenter binds its socket filter to. Dropping this is not a partial loss: the instrumenter's pipelines share one error group, so it exits at startup and **no** eBPF traces are produced. `CAP_SYS_ADMIN` does not substitute for this. |
+| `CAP_NET_ADMIN` | Attach the TC programs that carry trace context across hops (TCP-option propagation). |
+| `CAP_SETUID` + `CAP_SETGID` | Match a target JVM's credentials for HotSpot dynamic attach, which the JVM accepts only from an identical euid/egid. Dropping them leaves JVMs owned by other users (including all containerized ones) instrumentable only by restarting them — also set `use_dynamic_attach=false` if you do. |
```
rayray221: Nit,

```suggestion
| `CAP_SETUID` + `CAP_SETGID` | Match a target JVM's credentials for HotSpot dynamic attach, which the JVM accepts only from an identical euid/egid. Dropping them leaves JVMs owned by other users (including all containerized ones) instrumentable only by restarting them.  If you choose to remove this capability, also set `use_dynamic_attach=false` in the VM Agent configuration (see [Configuration Sections] (/vmagent/setup/configuration/odictl#configuration-sections.) |
```

docs/vmagent/setup/permissions.mdx
```diff
@@ -0,0 +1,304 @@
...
+    |---|---|---|
+    | RHEL 7, CentOS 7, Amazon Linux 2 | 3.10 / 4.14 | **Non-root is not available.** systemd is older than 229, so `AmbientCapabilities=` does not exist. The installer forces root even if you request non-root. Confirm with `/usr/lib/odigos/bin/odigos-vmagent-caps --explain`. |
+    | RHEL 8, Rocky 8 | 4.18 | `CAP_BPF` and `CAP_PERFMON` do not exist until Linux 5.8. eBPF load and `perf_event_open` still require `CAP_SYS_ADMIN`. |
+    | RHEL 9, Rocky 9, Amazon Linux 2023 | 5.14+ | The finer bits exist, but `capabilities(7)` still attributes those operations to `CAP_SYS_ADMIN`, and `perf_event_open` often still needs it. The package therefore grants `CAP_SYS_ADMIN` rather than `CAP_BPF` / `CAP_PERFMON`. |
+  </Tab>
+  <Tab title="Debian / Ubuntu">
+    | Distro | Typical kernel | Why `CAP_SYS_ADMIN` |
+    |---|---|---|
+    | Debian 11–13, Ubuntu 20.04 and later | 5.4+ | The default `kernel.perf_event_paranoid` is `3` or `4`. That sysctl requires `CAP_SYS_ADMIN` for `perf_event_open` (uprobes and profiling) even when `CAP_PERFMON` is present. Debian/Ubuntu also set `kptr_restrict=1`, which is why `CAP_SYSLOG` is in the set. |
```
rayray221, line 60: Nit, If a user could also drop `kernel.perf_event_paranoid` to 1 or 2, we should mention this as an option if it avoids `CAP_SYS_ADMIN`

docs/vmagent/setup/permissions.mdx
```diff
@@ -0,0 +1,304 @@
...
+
+Inspect what the helper decided on this host:
+
+```shell
+/usr/lib/odigos/bin/odigos-vmagent-caps --explain
+```
+
+### Override, add, or remove capabilities
+
+systemd **merges** `AmbientCapabilities=` and `CapabilityBoundingSet=` across drop-in files. A later drop-in can **add** capabilities; it cannot remove one that `nonroot.conf` already listed. To drop or replace capabilities, edit the packaged drop-in itself.
```
rayray221, line 76: I'm not sure we should include this but if we need to, then we should also include a warning:
Modifying Linux Capabilities may impair or break VM Agent functionality. Alter these settings at your own risk or under the guidance of Odigos Support.

docs/vmagent/setup/permissions.mdx
```diff
@@ -0,0 +1,304 @@
...
+  ```shell
+  sudo /usr/lib/odigos/bin/odigos-vmagent-caps --refresh
+  sudo systemctl daemon-reload
+  sudo systemctl restart odigos-vmagent
+  ```
+</Note>
+
+Dropping a capability does not prevent the agent from starting. The matching feature is disabled or degraded, and the agent logs which capability was missing at startup.
+
+## Docker group
```
rayray221: As I mentioned in the other comment, the three of these settings should fall under a parent called "Additional Permissions" or something similar.  

Then we should probably rename the headers of the three types of processes.  Docker group should be just Docker containers.

docs/vmagent/setup/permissions.mdx
```diff
@@ -0,0 +1,304 @@
...
+# uid=... gid=... groups=... docker
+```
+
+Without this group, Docker discovery fails with `EACCES` on the socket: the agent cannot list containers, instrument them, or restart them.
+
+## systemd journal group (logs)
+
+When a destination includes the `LOGS` signal, the collector reads **systemd** source logs from the journal (a shared `journald` receiver). A non-root `odigos` user cannot read the journal unless it is in a group that journald allows.
+
+Add `odigos` to **one** of the groups below (both is also fine), then restart the agent:
```
rayray221: Nit,

```suggestion
Add `odigos` to an appropriate group below based on your Linux Distribution, then restart the agent:
```

docs/vmagent/setup/permissions.mdx
rayray221: Do we support podman too?

docs/vmagent/setup/permissions.mdx
```diff
@@ -0,0 +1,304 @@
...
+Confirm:
+
+```shell
+id odigos
+# uid=... gid=... groups=... docker
+```
+
+Without this group, Docker discovery fails with `EACCES` on the socket: the agent cannot list containers, instrument them, or restart them.
+
+## systemd journal group (logs)
```
rayray221: Like above:
```suggestion
## journald permissions
```

docs/vmagent/setup/permissions.mdx
```diff
@@ -0,0 +1,304 @@
...
+  </Tab>
+</Tabs>
+
+This is only required for `LOGS` on **systemd** sources. Process log files and Docker container logs do not use the journal.
+
+<Note>
+  The odictl **Logs** panel shows logs for the agent and collector units themselves. That is separate from application logs exported to your destination, and does not require these groups.
+</Note>
+
+## polkit (restart systemd sources)
```
rayray221: Like above:
```suggestion
## systemd permissions (polkit)
```

rayray221 (changes requested):

docs/vmagent/setup/configuration/odictl.mdx
liranm-odigos: It's working here

docs/vmagent/setup/permissions.mdx
liranm-odigos: Adjusted. and we don't support it

docs/vmagent/setup/installation.mdx
liranm-odigos: The directory is still owned by user odigos.
So if non root or non user odigos wants to modify the files in that directory it either needs to run as root (sudo) or use user odigos (FYI, user odigos does not have shell, so it's not really possible to switch to that user).

docs/vmagent/setup/permissions.mdx
liranm-odigos: Not always, even if it's 1 or 2 we still need SYS_ADMIN

rayray221 (approved):
Awesome!

PR #5728, opened 2026-09-03
RUN-1115 | feat(instrumentation): set telemetry.distro.name for eBPF distributions

daemon1024:
Each eBPF agent decides on its own whether to report `telemetry.distro.name`, and the Java agent does not report it at all. This sets it in the instrumentation manager from the process's distribution, so every eBPF distribution reports the Odigos distribution name. Generic factories do not get it. An existing value, for example from `OTEL_RESOURCE_ATTRIBUTES`, is kept.

```release-note
eBPF instrumentations now report `telemetry.distro.name` with the Odigos distribution name (for example `java-ebpf-instrumentations`, `golang-enterprise`).
```

cherry-pick: v1.36

### DIFF
--- a/instrumentation/manager.go
+++ b/instrumentation/manager.go
@@ -446,6 +446,19 @@ func (m *manager[ProcessGroup, ConfigGroup, ProcessDetails]) metricsAttributeSet
 	)
 }
 
+// withTelemetryDistroName returns attrs with telemetry.distro.name set to distroName.
+// An existing value (for example from OTEL_RESOURCE_ATTRIBUTES) is kept. attrs is not modified.
+func withTelemetryDistroName(attrs []attribute.KeyValue, distroName string) []attribute.KeyValue {
+	for _, attr := range attrs {
+		if attr.Key == semconv.TelemetryDistroNameKey {
+			return attrs
+		}
+	}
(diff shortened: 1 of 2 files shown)


instrumentation/manager.go
```diff
@@ -604,6 +617,9 @@ func (m *manager[ProcessGroup, ConfigGroup, ProcessDetails]) tryInstrument(ctx c
 	// Distro factory: the process's own instrumentation and the main instrumentation path - report
 	// init/load, track even on failure (so the reporter is notified on exit and a failed distro can
 	// be retried), and run.
+	// The distribution is reported as telemetry.distro.name. Generic factories do not get it, as they
+	// are not the distribution instrumenting the process.
+	settings.ResourceAttributes = withTelemetryDistroName(settings.ResourceAttributes, otelDistro.Name)
```
RonFed, line 622: I think otelDistro can be nil here for cases where there is no ebpf instrumentation but there is a "generic" wrapper instrumentation, should we add a nil check for the distro?

instrumentation/manager.go
daemon1024, line 555: @RonFed we already have a nilcheck here, we would return early incase otelDistro.Name is invalid

RonFed (approved):

PR #5735, opened 2026-09-03
CORE-1634: fix(init-container): Copy enterprise pull secret to user ns for k8s 1.35+

damemi:
We set it up so that the `init-container` mount method pre-pulled the agents image onto the odiglet, so that we didn't need to inject private registry pull secrets into every app that used the init container (since it was already cached on the node).

However, in k8s 1.35+, the kubelet setting `KubeletEnsureSecretPulledImages` went beta and on by default, which *re-verifies* pull secrets from private repos even for already-cached images. This breaks customer mirroring from private registries, and it also breaks our new enterprise pull secret use case.

Our only option now is to copy any imagePullSecrets defined at startup for odigos images (either customer mirror registry or our own enterprise registry) to any namespace where an app is instrumented and inject that secret.

This adds that functionality, and also removes the odiglet pre-pull workaround.

```release-note
fix(init-container): copy odigos pull secrets to instrumented application namespaces to support KubeletEnsureSecretPulledImages on k8s 1.35+
```

### DIFF
--- a/cli/cmd/uninstall-dep.go
+++ b/cli/cmd/uninstall-dep.go
@@ -90,6 +90,11 @@ and rollback any metadata changes made to your objects.`,
 				fmt.Printf("\033[31mERROR\033[0m Failed to remove all sources: %s\n", err)
 				os.Exit(1)
 			}
+			err = removeCopiedImagePullSecrets(ctx, client, ns)
+			if err != nil {
+				fmt.Printf("\033[31mERROR\033[0m Failed to remove copied image pull secrets: %s\n", err)
+				os.Exit(1)
+			}
 			if autoRolloutDisabled {
 				fmt.Println("Odigos is configured to NOT rollout workloads automatically; existing pods will remain instrumented until a manual rollout is triggered.")
 			} else if !cmd.Flag("no-wait").Changed {
@@ -583,6 +588,34 @@ func cleanupNodeOdigosLabels(ctx context.Context, client *kube.Client, ns, _ str
 		}
 	}
 
+	return nil
+}
+
+func removeCopiedImagePullSecrets(ctx context.Context, client *kube.Client, odigosNs string) error {
+	l := log.Print("Removing copied image pull secrets...")
+	list, err := client.CoreV1().Secrets("").List(ctx, metav1.ListOptions{
+		LabelSelector: metav1.FormatLabelSelector(&metav1.LabelSelector{
+			MatchLabels: labels.OdigosSystem,
+		}),
+	})
+	if err != nil {
+		return err
+	}
+
+	var deleteErr error
+	for _, secret := range list.Items {
+		if secret.Namespace == odigosNs {
+			continue
+		}
+		err = client.CoreV1().Secrets(secret.Namespace).Delete(ctx, secret.Name, metav1.DeleteOptions{})
+		if err != nil && !apierrors.IsNotFound(err) {
+			deleteErr = errors.Join(deleteErr, err)
+		}
+	}
+	if deleteErr != nil {
+		return deleteErr
+	}
+	l.Success()
 	return nil
 }
 
--- a/common/odigos_config.go
+++ b/common/odigos_config.go
@@ -685,7 +685,8 @@ type OdigosConfiguration struct {
 	// configuration for odigos own metrics store in the cluster.
 	OdigosOwnTelemetryStore *OdigosOwnTelemetryConfiguration `json:"odigosOwnTelemetryStore,omitempty" yaml:"odigosOwnTelemetryStore"`
 
-	// ImagePullSecrets to use for collectors and init container
+	// ImagePullSecrets to use for collectors and the agents init container.
+	// For k8s-init-container, these secrets are copied into instrumented namespaces.
 	ImagePullSecrets []string `json:"imagePullSecrets,omitempty" yaml:"imagePullSecrets"`
 
 	// global configurations for sampling.
--- a/common/odigos_config.go
+++ b/common/odigos_config.go
@@ -685,7 +685,7 @@ type OdigosConfiguration struct {
 	// configuration for odigos own metrics store in the cluster.
 	OdigosOwnTelemetryStore *OdigosOwnTelemetryConfiguration `json:"odigosOwnTelemetryStore,omitempty" yaml:"odigosOwnTelemetryStore"`
 
-	// ImagePullSecrets to use for collectors and the agents init container.
+	// ImagePullSecrets to use for Odigos components and the agents init container.
 	// For k8s-init-container, these secrets are copied into instrumented namespaces.
 	ImagePullSecrets []string `json:"imagePullSecrets,omitempty" yaml:"imagePullSecrets"`
 
--- a/helm/odigos/templates/cleanup/cleanup-clusterrole.yaml
+++ b/helm/odigos/templates/cleanup/cleanup-clusterrole.yaml
@@ -14,4 +14,7 @@ rules:
 - apiGroups: [""]
   resources: ["nodes"]
   verbs: ["list", "patch"]
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["list", "delete"]
 
--- a/helm/odigos/templates/cleanup/cleanup-clusterrole.yaml
+++ b/helm/odigos/templates/cleanup/cleanup-clusterrole.yaml
@@ -14,7 +14,9 @@ rules:
 - apiGroups: [""]
   resources: ["nodes"]
   verbs: ["list", "patch"]
+{{- if include "odigos.needsImagePullSecretCopy" . }}
 - apiGroups: [""]
   resources: ["secrets"]
   verbs: ["list", "delete"]
+{{- end }}
 
--- a/helm/odigos/templates/instrumentor/clusterrole.yaml
+++ b/helm/odigos/templates/instrumentor/clusterrole.yaml
@@ -33,6 +33,7 @@ rules:
       - list
       - watch
       - get
+{{- if include "odigos.needsImagePullSecretCopy" . }}
   # create cannot be scoped with resourceNames; used to copy image pull secrets into instrumented namespaces.
   - apiGroups:
       - ''
@@ -40,16 +41,7 @@ rules:
       - secrets
     verbs:
       - create
-  - apiGroups:
-      - ''
-    resources:
-      - secrets
-    resourceNames:
-      - odigos-enterprise-registry
-    verbs:
-      - get
-      - update
-      - patch
+{{- end }}
   - apiGroups:
       - 'batch'
     resources:
--- a/helm/odigos/templates/instrumentor/clusterrole.yaml
+++ b/helm/odigos/templates/instrumentor/clusterrole.yaml
@@ -33,6 +33,23 @@ rules:
       - list
       - watch
       - get
+  # create cannot be scoped with resourceNames; used to copy image pull secrets into instrumented namespaces.
+  - apiGroups:
+      - ''
+    resources:
+      - secrets
+    verbs:
+      - create
+  - apiGroups:
+      - ''
+    resources:
+      - secrets
+    resourceNames:
+      - odigos-enterprise-registry
+    verbs:
+      - get
+      - update
+      - patch
   - apiGroups:
       - 'batch'
     resources:
--- a/instrumentor/controllers/agentenabled/pods_webhook.go
+++ b/instrumentor/controllers/agentenabled/pods_webhook.go
@@ -29,7 +29,6 @@ import (
 	podutils "github.com/odigos-io/odigos/instrumentor/internal/pod"
 	webhookenvinjector "github.com/odigos-io/odigos/instrumentor/internal/webhook_env_injector"
 	"github.com/odigos-io/odigos/k8sutils/pkg/env"
-	"github.com/odigos-io/odigos/k8sutils/pkg/pro"
 	"github.com/odigos-io/odigos/k8sutils/pkg/service"
 	k8sutils "github.com/odigos-io/odigos/k8sutils/pkg/utils"
 	"github.com/odigos-io/odigos/k8sutils/pkg/workload"
@@ -217,8 +216,8 @@ func (p *PodsWebhook) injectOdigos(ctx context.Context, pod *corev1.Pod, req adm
 		if len(dirsToCopy) > 0 {
 			// Create the init container that will copy the directories to the empty dir based on dirsToCopy
 			createInitContainer(pod, dirsToCopy, odigosConfiguration)
-			if err := ensureImagePullSecretsForPod(ctx, p.Client, odigosNamespace, pw.Namespace, pod, odigosConfiguration.ImagePullSecrets); err != nil {
-				logger.Error(err, "failed to ensure image pull secrets for init container")
+			for _, name := range odigosConfiguration.ImagePullSecrets {
+				injectImagePullSecret(pod, name)
 			}
 		}
 	}
@@ -550,18 +549,10 @@ func createInitContainer(pod *corev1.Pod, dirsToCopy map[string]struct{}, config
 	pod.Spec.InitContainers = append(pod.Spec.InitContainers, agentInitContainer)
 }
 
-func ensureImagePullSecretsForPod(ctx context.Context, c client.Client, odigosNamespace, podNamespace string, pod *corev1.Pod, configuredSecrets []string) error {
-	present, err := pro.CopyImagePullSecretsIfMissing(ctx, c, c, odigosNamespace, podNamespace, configuredSecrets)
-	if err != nil {
-		return err
-	}
-	for _, name := range present {
-		injectImagePullSecret(pod, name)
-	}
-	return nil
-}
-
 func injectImagePullSecret(pod *corev1.Pod, secretName string) {
+	if secretName == "" {
+		return
+	}
 	for _, existing := range pod.Spec.ImagePullSecrets {
 		if existing.Name == secretName {
 			return
--- a/instrumentor/controllers/agentenabled/pods_webhook.go
+++ b/instrumentor/controllers/agentenabled/pods_webhook.go
@@ -29,6 +29,7 @@ import (
 	podutils "github.com/odigos-io/odigos/instrumentor/internal/pod"
 	webhookenvinjector "github.com/odigos-io/odigos/instrumentor/internal/webhook_env_injector"
 	"github.com/odigos-io/odigos/k8sutils/pkg/env"
+	"github.com/odigos-io/odigos/k8sutils/pkg/pro"
 	"github.com/odigos-io/odigos/k8sutils/pkg/service"
 	k8sutils "github.com/odigos-io/odigos/k8sutils/pkg/utils"
 	"github.com/odigos-io/odigos/k8sutils/pkg/workload"
@@ -216,6 +217,9 @@ func (p *PodsWebhook) injectOdigos(ctx context.Context, pod *corev1.Pod, req adm
 		if len(dirsToCopy) > 0 {
 			// Create the init container that will copy the directories to the empty dir based on dirsToCopy
 			createInitContainer(pod, dirsToCopy, odigosConfiguration)
+			if err := ensureImagePullSecretsForPod(ctx, p.Client, odigosNamespace, pw.Namespace, pod, odigosConfiguration.ImagePullSecrets); err != nil {
+				logger.Error(err, "failed to ensure image pull secrets for init container")
+			}
 		}
 	}
 
@@ -544,6 +548,26 @@ func createInitContainer(pod *corev1.Pod, dirsToCopy map[string]struct{}, config
 		}
 	}
 	pod.Spec.InitContainers = append(pod.Spec.InitContainers, agentInitContainer)
+}
+
+func ensureImagePullSecretsForPod(ctx context.Context, c client.Client, odigosNamespace, podNamespace string, pod *corev1.Pod, configuredSecrets []string) error {
+	present, err := pro.CopyImagePullSecretsIfMissing(ctx, c, c, odigosNamespace, podNamespace, configuredSecrets)
+	if err != nil {
+		return err
+	}
+	for _, name := range present {
+		injectImagePullSecret(pod, name)
+	}
+	return nil
+}
+
+func injectImagePullSecret(pod *corev1.Pod, secretName string) {
+	for _, existing := range pod.Spec.ImagePullSecrets {
+		if existing.Name == secretName {
+			return
+		}
+	}
+	pod.Spec.ImagePullSecrets = append(pod.Spec.ImagePullSecrets, corev1.LocalObjectReference{Name: secretName})
 }
 
 func getInitContainerImage(config common.OdigosConfiguration) string {
--- a/k8sutils/pkg/pro/pullsecret.go
+++ b/k8sutils/pkg/pro/pullsecret.go
@@ -1,12 +1,16 @@
 package pro
 
 import (
+	"context"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/odigos-io/odigos/api/k8sconsts"
 )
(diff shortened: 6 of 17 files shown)


helm/odigos/templates/cleanup/cleanup-clusterrole.yaml
```diff
@@ -14,4 +14,7 @@ rules:
 - apiGroups: [""]
   resources: ["nodes"]
   verbs: ["list", "patch"]
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["list", "delete"]
```
blumamir, line 20: I think this can scare a lot of people and raise lot's of questions.

we should at minimum have a way to opt in/out of this feature so we can continue to operate like now without this very dangerous RBAC.

also worth to gate it only for init-container mount method and not globally

common/odigos_config.go
```diff
@@ -686,7 +686,8 @@ type OdigosConfiguration struct {
 	// configuration for odigos own metrics store in the cluster.
 	OdigosOwnTelemetryStore *OdigosOwnTelemetryConfiguration `json:"odigosOwnTelemetryStore,omitempty" yaml:"odigosOwnTelemetryStore"`
 
-	// ImagePullSecrets to use for collectors and init container
+	// ImagePullSecrets to use for collectors and the agents init container.
```
blumamir: we can take the opportunity to update the comment (not just collectors, all odigos components)

helm/odigos/templates/instrumentor/clusterrole.yaml
```diff
@@ -29,6 +29,23 @@ rules:
...
+  - apiGroups:
+      - ''
+    resources:
+      - secrets
+    resourceNames:
+      - odigos-enterprise-registry
+    verbs:
+      - get
+      - update
+      - patch
```
blumamir: can this be wrapped so that it's only rendered if the pull secret is used in the values?

So we assume this RBAC only when actually needed and avoid for others

k8sutils/pkg/pro/pullsecret.go
```diff
@@ -45,3 +49,89 @@ func EnterpriseRegistryPullSecretLabels() map[string]string {
...
+	dest := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretName,
+			Namespace: destNs,
+			Labels:    EnterpriseRegistryPullSecretLabels(),
+		},
+		Type: src.Type,
+		Data: data,
+	}
+	err = w.Create(ctx, dest)
```
blumamir, line 103: this is problematic to do in webhook, since it can block, fail (and then it's not trivial to retry) and other issues.

k8sutils/pkg/pro/pullsecret.go
```diff
@@ -45,3 +49,89 @@ func EnterpriseRegistryPullSecretLabels() map[string]string {
...
+			continue
+		}
+		if ok {
+			present = append(present, name)
+		}
+	}
+	return present, errs
+}
+
+func copyImagePullSecretIfMissing(ctx context.Context, r client.Reader, w client.Writer, sourceNs, destNs, secretName string) (bool, error) {
```
blumamir: these functions are named so similar 😅

k8sutils/pkg/pro/pullsecret.go
```diff
@@ -45,3 +49,89 @@ func EnterpriseRegistryPullSecretLabels() map[string]string {
...
+		}
+		seen[name] = struct{}{}
+		names = append(names, name)
+	}
+	for _, name := range configured {
+		add(name)
+	}
+	add(k8sconsts.OdigosEnterpriseRegistryPullSecretName)
+	return names
+}
```
blumamir: you can do this once and store in effective config, this way:
- we don't need to run this every time (it does not change)
- we have visibility into the effective list
- keep all config sanitisation in one place

k8sutils/pkg/pro/pullsecret.go
```diff
@@ -45,3 +49,89 @@ func EnterpriseRegistryPullSecretLabels() map[string]string {
...
+
+// CopyImagePullSecretsIfMissing copies each named pull secret from sourceNs into destNs
+// when it is not already present. Missing source secrets are skipped.
+// Returns the secret names that destNs should reference on pods.
+func CopyImagePullSecretsIfMissing(
```
blumamir, line 56: Can we do this once when we reconcile the source into IC, instead of per pod?

cli/cmd/uninstall-dep.go
```diff
@@ -590,6 +595,34 @@ func cleanupNodeOdigosLabels(ctx context.Context, client *kube.Client, ns, _ str
...
+
+	var deleteErr error
+	for _, secret := range list.Items {
+		if secret.Namespace == odigosNs {
+			continue
+		}
+		err = client.CoreV1().Secrets(secret.Namespace).Delete(ctx, secret.Name, metav1.DeleteOptions{})
+		if err != nil && !apierrors.IsNotFound(err) {
+			deleteErr = errors.Join(deleteErr, err)
+		}
```
blumamir, line 613: consider to use "DeleteAllOf" which does it in one go and also avoid the "Get" of sensitive resources into our pod

instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -222,6 +223,9 @@ func (p *PodsWebhook) injectOdigos(ctx context.Context, pod *corev1.Pod, req adm
 		if len(dirsToCopy) > 0 {
 			// Create the init container that will copy the directories to the empty dir based on dirsToCopy
 			createInitContainer(pod, dirsToCopy, odigosConfiguration)
+			if err := ensureImagePullSecretsForPod(ctx, p.Client, odigosNamespace, pw.Namespace, pod, odigosConfiguration.ImagePullSecrets); err != nil {
```
blumamir: we should have a flag to opt-out of this behavior if user manages their own secrets or mirror the images to internal registry with no auth

helm/odigos/templates/cleanup/cleanup-clusterrole.yaml
damemi: if a user wants to use initContainer mount method + a custom, private registry, to me that's the opt-in. That setup won't work without being able to sync secrets across namespaces. so the best we could do is document that more

blumamir: Is private registry always means it needs a pull seecret?

damemi: Yes, a "private" registry always needs a pull secret. Just using a custom registry (`imagePrefix`) does not on its own, but that doesn't hit the issue here at all.

"private registry" is a combination of 2 settings: `imagePrefix` (mirrored somewhere besides registry.odigos.io) + `imagePullSecrets` (access that mirror)

the only reason someone will set `imagePullSecrets` is if they're already using `imagePrefix` (since the default enterprise pull secret is created for them)

`imagePullSecrets` is the issue: these currently only need to exist in the `odigos-system` namespace to pull mirrored odigos components. That's where the secrets rbac requirement comes from

helm/odigos/templates/instrumentor/clusterrole.yaml
damemi: yes updated

k8sutils/pkg/pro/pullsecret.go
damemi: removed from the webhook and now just done in reconcile

k8sutils/pkg/pro/pullsecret.go
damemi: similar to https://github.com/odigos-io/odigos/pull/5735#discussion_r3982636337, removed from the webhook

k8sutils/pkg/pro/pullsecret.go
damemi: the imagepullsecrets are already in effective config, so updated to just use that list

instrumentor/controllers/agentenabled/pods_webhook.go
damemi: removed it from the webhook anyway, but even in sync the attempt to copy is a no-op if the user isn't using 1. initcontainer method and 2. doesn't have any pull secrets to copy

cli/cmd/uninstall-dep.go
damemi: true that would get rid of needing a Get permission, but we'd need a different label specifically for these copied image pull secrets otherwise we would just delete all system secrets at once (like webhook certs which could break uninstall)

helm/odigos/templates/cleanup/cleanup-clusterrole.yaml
blumamir: can't someone use `imagePrefix` without `imagePullSecrets`? it's there own private registry so they don't have to implement authentication

cli/cmd/uninstall-dep.go
blumamir: ok so let's have a lable - sounds good anyway

k8sutils/pkg/pro/pullsecret.go
```diff
@@ -45,3 +49,63 @@ func EnterpriseRegistryPullSecretLabels() map[string]string {
...
+	dest := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretName,
+			Namespace: destNs,
+			Labels:    EnterpriseRegistryPullSecretLabels(),
+		},
+		Type: src.Type,
+		Data: data,
+	}
+	err = w.Create(ctx, dest)
```
damemi: we need a better reconcile than a one time create because if the pull secrets rotate they won't get propagated

damemi:
@blumamir updated to make the auto-syncing an opt in feature so users can manage the secrets themselves. enterprise secret +any secrets listed in imagePullSecrets are still injected into instrumented pods by default. updated docs to explain the new behavior


Config / Helm (opt-in)
syncOdigosPullSecrets defaults to false. Helm writes it into effective config and only then grants instrumentor secrets create plus named get/update/patch on imagePullSecrets + odigos-enterprise-registry. Cleanup secrets list/delete is also gated on that flag.

Copy behavior
CopyImagePullSecrets creates dest secrets labeled odigos.io/copied-image-pull-secret. If dest already has that label, it updates data (rotation). If dest exists without the label, it fails and does not overwrite. Webhook still always injects the names from effective config; the flag only controls copying.

Re-trigger
New PullSecretSyncReconciler watches source secrets in the Odigos namespace. Changing odigos.io/sync-odigos-pull-secrets copies/updates that secret into every IC namespace.

Uninstall
Copied secrets are removed with DeleteCollection on the copy label (no list of all system-object secrets, no skip-odigos-ns loop).

cli/cmd/uninstall-dep.go
```diff
@@ -586,6 +591,20 @@ func cleanupNodeOdigosLabels(ctx context.Context, client *kube.Client, ns, _ str
 	return nil
 }
 
+func removeCopiedImagePullSecrets(ctx context.Context, client *kube.Client, _ string) error {
```
tamirdavid1, line 594: we should do it only if syncOdigosPullSecrets is enable right?
if not we still need the extended rbac to the cleanup job

tests/e2e/mount-method/odiglet-ready.yaml
```diff
@@ -11,9 +11,6 @@ metadata:
       kind: DaemonSet
       name: odiglet
 spec:
-  initContainers:
```
tamirdavid1, line 14: Should we remove it only from tests running on Kubernetes v1.35.0 and above, and only when `syncOdigosPullSecrets` is enabled?

I think the mount method is the right place to test this behavior, once with the flag enabled and once with it disabled.

helm/odigos/values.yaml
```diff
@@ -10,6 +10,22 @@ image:
   tag: ''
 imagePullSecrets: []
 
+# @schema
+# description: |-
+#   Copy Odigos image pull secrets (imagePullSecrets and odigos-enterprise-registry)
+#   into instrumented application namespaces. Off by default.
+#   Enable for k8s-init-container with a private registry (enterprise registry or a
```
blumamir: if it's enterprise registry, then it's not private, right?

helm/odigos/values.yaml
```diff
@@ -10,6 +10,22 @@ image:
...
+#   mirrored registry). Kubernetes 1.35+ verifies pull credentials even for cached
+#   images (KubeletEnsureSecretPulledImages), so the agents init container needs the
+#   secret in the application namespace. The webhook always injects the secret names
+#   onto those pods; this flag only controls copying the Secret objects.
+#   Copies are labeled odigos.io/copied-image-pull-secret=true and are updated on
+#   re-sync. A same-named secret without that label is not overwritten. After
+#   rotating credentials, annotate a source secret in the Odigos namespace with
```
blumamir: this looks like documentation for odigos engineers and maintainers, but I wonder if a normal user will understand or care about all these details.

the values file is user facing and we should mention just the things that are relevant for them IMO

common/odigos_config.go
```diff
@@ -685,9 +685,16 @@ type OdigosConfiguration struct {
...
-	// ImagePullSecrets to use for collectors and init container
+	// ImagePullSecrets to use for Odigos components and the agents init container.
+	// The pod webhook always injects these names onto init-container pods.
+	// Secrets are copied into instrumented namespaces only when SyncOdigosPullSecrets is set.
 	ImagePullSecrets []string `json:"imagePullSecrets,omitempty" yaml:"imagePullSecrets"`
 
+	// SyncOdigosPullSecrets copies ImagePullSecrets from the Odigos namespace into
+	// instrumented namespaces. Opt-in; intended for k8s-init-container with a private
+	// registry (enterprise registry or a mirrored registry). Off by default.
+	SyncOdigosPullSecrets bool `json:"syncOdigosPullSecrets,omitempty" yaml:"syncOdigosPullSecrets"`
```
blumamir: please use `*bool` and differentiate between `null` (unset) and `false` (explicitly set).

if we later want to support setting it from central or ui, it is required in order to know if something is set or not. also to align with other properties in our config

cli/cmd/uninstall-dep.go
damemi: yup good catch

tests/e2e/mount-method/odiglet-ready.yaml
damemi: good catch here too, keeping the image pre-pull in all scenarios by default (worst case it does still have the benefit of warming the node cache so first instrumentation goes quicker). 

I also updated it so the pull secrets are only injected into the pods on k8s 1.35+, aligning with the requirement to have them in the same namespace (it's harmless to inject them even if they're not present, but it can make a lot of noise about not finding them)

helm/odigos/templates/cleanup/cleanup-clusterrole.yaml
```diff
@@ -14,4 +14,9 @@ rules:
 - apiGroups: [""]
   resources: ["nodes"]
   verbs: ["list", "patch"]
+{{- if include "odigos.needsImagePullSecretCopy" . }}
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["list", "delete"]
+{{- end }}
```
blumamir: don't we need

```suggestion
{{- if include "odigos.needsImagePullSecretCopy" . }}
- apiGroups: [""]
  resources: ["secrets"]
  verbs: ["deletecollection"]
{{- end }}
```

rbac permission here for the `DeleteCollection` call?

instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -546,6 +555,22 @@ func createInitContainer(pod *corev1.Pod, dirsToCopy map[string]struct{}, config
 	pod.Spec.InitContainers = append(pod.Spec.InitContainers, agentInitContainer)
 }
 
+func shouldInjectInitContainerPullSecrets(k8sVersion *version.Version) bool {
+	return k8sVersion != nil && k8sVersion.AtLeast(k8sconsts.MinK8SVersionForInitContainerPullSecrets)
+}
```
blumamir, line 560: what is the reason to keep it for k8s versions <1.35 ?

would it be simpler to just do it for all versions instead?

instrumentor/controllers/agentenabled/pods_webhook.go
```diff
@@ -546,6 +555,22 @@ func createInitContainer(pod *corev1.Pod, dirsToCopy map[string]struct{}, config
 	pod.Spec.InitContainers = append(pod.Spec.InitContainers, agentInitContainer)
 }
 
+func shouldInjectInitContainerPullSecrets(k8sVersion *version.Version) bool {
+	return k8sVersion != nil && k8sVersion.AtLeast(k8sconsts.MinK8SVersionForInitContainerPullSecrets)
```
blumamir, line 559: if `k8sVersion` is nil, this will not inject the pull secret which I think it should

instrumentor/controllers/agentenabled/manager.go
```diff
@@ -112,5 +112,17 @@ func SetupWithManager(mgr ctrl.Manager, dp *distros.Provider) error {
...
+		ControllerManagedBy(mgr).
+		Named("agentenabled-pullsecretsync").
+		For(&corev1.Secret{}).
+		WithEventFilter(&syncOdigosPullSecretsAnnotationPredicate{}).
+		Complete(&PullSecretSyncReconciler{
+			Client: mgr.GetClient(),
+		})
+	if err != nil {
+		return err
+	}
```
blumamir: wondering if this controller is logically more appropriate in the `sourceinstrumentation` directory?

k8sutils/pkg/pro/pullsecret.go
```diff
@@ -45,3 +54,127 @@ func EnterpriseRegistryPullSecretLabels() map[string]string {
...
+		return fmt.Errorf("failed to read image pull secret %q in namespace %q: %w", secretName, sourceNs, err)
+	}
+
+	data := make(map[string][]byte, len(src.Data))
+	for k, v := range src.Data {
+		data[k] = append([]byte(nil), v...)
+	}
+
+	var dest corev1.Secret
+	err = r.Get(ctx, client.ObjectKey{Namespace: destNs, Name: secretName}, &dest)
```
blumamir: if we do it with patch, we can avoid reading the secret from the dest namespace and the rbac permissions it needs. also will simplify the flow (no need to branch create vs update

k8sutils/pkg/pro/pullsecret.go
```diff
@@ -45,3 +54,127 @@ func EnterpriseRegistryPullSecretLabels() map[string]string {
...
+		return false
+	}
+	for k, av := range a {
+		bv, ok := b[k]
+		if !ok || !bytes.Equal(av, bv) {
+			return false
+		}
+	}
+	return true
+}
```
blumamir: what is the reason to do that? 
are we doing it anywhere else in odigos?

instrumentor/controllers/agentenabled/pullsecret_controller.go
```diff
@@ -0,0 +1,97 @@
...
+	for i := range ics.Items {
+		ns := ics.Items[i].Namespace
+		if ns == "" || ns == sourceNs {
+			continue
+		}
+		if _, ok := seen[ns]; ok {
+			continue
+		}
+		seen[ns] = struct{}{}
+		copyErr := pro.CopyImagePullSecrets(ctx, r.Client, r.Client, sourceNs, ns, []string{req.Name})
```
blumamir: we are reconciling one single secret here, but sync the full list of image pull secrets. 
This looks redundant and not idomatic for controllers

api/k8sconsts/resources.go
```diff
@@ -36,6 +36,12 @@ const (
 	// RollbackRecoveryAtAnnotation on the same IC, the recovery has been handled.
 	RollbackRecoveryProcessedAtAnnotation = "odigos.io/rollback-recovery-processed"
 
+	// SyncOdigosPullSecretsAnnotation on a source pull secret in the Odigos namespace
```
blumamir: is it on a source? or on a secret?

docs/snippets/shared/instrumentations/configuration/mount-method.mdx
```diff
@@ -109,6 +109,41 @@ This approach is ideal for users who prioritize avoiding hostPath or virtual dev
 
 Works with [Karpenter](../../setup/odigos-with-karpenter) out of the box. No additional Karpenter configuration is required.
 
+#### Init container image pull secrets
+
+Kubernetes 1.35+ enables [`KubeletEnsureSecretPulledImages`](https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/) by default. The kubelet [re-verifies pull credentials](https://kubernetes.io/docs/concepts/containers/images/#ensure-image-pull-credential-verification) even when the agents image is already cached on the node. The injected `odigos-agents` init container therefore needs the same pull secret in the **application** namespace, not only in the Odigos namespace.
```
blumamir, line 114: same here, looks like info not relevant to end users.
I think we should give instructions and not provide internal details and implementation details in our docs

docs/snippets/shared/instrumentations/configuration/mount-method.mdx
```diff
@@ -109,6 +109,41 @@ This approach is ideal for users who prioritize avoiding hostPath or virtual dev
...
+
+This applies when you use `k8s-init-container` with a private registry:
```
blumamir, line 116: to me private registry is a registry maintained by the organization. our `registry.odigos.io` is not private registry.

this should be read from the perspective of the user/organization and not odigos maintainers

docs/snippets/shared/instrumentations/configuration/mount-method.mdx
```diff
@@ -109,6 +109,41 @@ This approach is ideal for users who prioritize avoiding hostPath or virtual dev
...
+**Option A — let Odigos copy the secrets**
+
+```bash
+helm upgrade --install odigos odigos/odigos \
+  --namespace odigos-system \
+  --set instrumentor.mountMethod=k8s-init-container \
+  --set syncOdigosPullSecrets=true
+```
+
+Odigos labels copies with `odigos.io/copied-image-pull-secret=true`. Those copies are updated when the source changes. A secret with the same name that Odigos did not create is left alone.
```
blumamir: give users just the instructions they need, and avoid overwhelming them with esoteric info they don't need to be aware of

docs/snippets/shared/instrumentations/configuration/mount-method.mdx
```diff
@@ -109,6 +109,41 @@ This approach is ideal for users who prioritize avoiding hostPath or virtual dev
...
+
+After rotating credentials in the Odigos namespace, re-sync the labeled copies by annotating the source secret:
+
+```bash
+kubectl annotate secret odigos-enterprise-registry -n odigos-system \
+  odigos.io/sync-odigos-pull-secrets="$(date +%s)" --overwrite
+```
```
blumamir: why do we need this step? can't we just sync it always with the reconciler?

helm/odigos/values.yaml
damemi: enterprise registry is a private registry (you need the pull secret copied to pull from enterprise for init container)

instrumentor/controllers/agentenabled/pods_webhook.go
damemi: if you set a pull secret on a pod and that pull secret doesn't exist, it makes a bunch of noise about not finding the pull secret. it's harmless and the pod still works, but it's unnecessary warnings and we don't need to do this <1.35

instrumentor/controllers/agentenabled/pullsecret_controller.go
damemi: no that's idiomatic, this is level based. ideal controller design doesn't matter *what* item you're reconciling, the point is to reconcile the entire state. reconcile events are just triggers

api/k8sconsts/resources.go
damemi: the "source pull secret" (as in source/destination for a copy)

docs/snippets/shared/instrumentations/configuration/mount-method.mdx
damemi: I think this is very relevant to end users and especially SE who will have to explain why we have this painful flow now

docs/snippets/shared/instrumentations/configuration/mount-method.mdx
damemi: this is what kicks off a sync, otherwise we need to watch all secrets in all namespaces

PR #5755, opened 2026-09-06
CORE-1637: schedule odiglet only on nodes with instrumented pods

blumamir:
Adds an opt-in Helm setting `odiglet.scheduleOnlyOnInstrumentedNodes` so odiglet DaemonSet pods are scheduled only on nodes that already run instrumented workloads. This is intended for the `k8s-init-container` mount method, where agents do not need odiglet present on every node.

When enabled:
- the instrumentor maintains the `odigos.io/instrumented-pods` node label from instrumented pods / InstrumentationConfigs
- the odiglet DaemonSet gets a matching `nodeSelector`
- toggling the flag sets/clears an instrumentor env var so the controllers (or a one-shot label cleanup runnable) are registered on restart

When disabled, leftover node labels are cleaned up by a leader-elected runnable.

```release-note
Add odiglet.scheduleOnlyOnInstrumentedNodes to schedule odiglet only on nodes that already have instrumented pods (requires instrumentor.mountMethod=k8s-init-container).
```

Linear: https://linear.app/odigos/issue/CORE-1637

### DIFF
--- a/helm/odigos/templates/odiglet/daemonset.yaml
+++ b/helm/odigos/templates/odiglet/daemonset.yaml
@@ -29,8 +29,8 @@
 {{- if and $networkMetricsEnabled .Values.odiglet.noHostNetwork }}
 {{ fail "odiglet.noHostNetwork cannot be true when metricsSources.networkMetrics.enabled is true, since OBI network metrics require hostNetwork" }}
 {{- end }}
-{{- if and .Values.odiglet.scheduleOnlyOnInstrumentedNodes (ne .Values.instrumentor.mountMethod "k8s-init-container") }}
-{{ fail "odiglet.scheduleOnlyOnInstrumentedNodes requires instrumentor.mountMethod=k8s-init-container" }}
+{{- if and (dig "scheduleOnlyOnInstrumentedNodes" "enabled" false (.Values.odiglet | default dict)) (ne .Values.instrumentor.mountMethod "k8s-init-container") }}
+{{ fail "odiglet.scheduleOnlyOnInstrumentedNodes.enabled requires instrumentor.mountMethod=k8s-init-container" }}
 {{- end }}
 apiVersion: apps/v1
 kind: DaemonSet
@@ -72,9 +72,6 @@ spec:
       {{- range $k, $v := .Values.nodeSelector }}
       {{- $_ := set $nodeSelector $k $v }}
       {{- end }}
-      {{- end }}
-      {{- if .Values.odiglet.scheduleOnlyOnInstrumentedNodes }}
-      {{- $_ := set $nodeSelector "odigos.io/instrumented-pods" "true" }}
       {{- end }}
       {{- if $nodeSelector }}
       nodeSelector:
@@ -613,9 +610,21 @@ spec:
   {{- if .tolerations }}
       tolerations: {{ toYaml .tolerations | nindent 8 }}
   {{- end }}
-  {{- if .affinity }}
-      affinity: {{ toYaml .affinity | nindent 8 }}
-  {{- end }}
+{{- end }}
+{{- /* Exists match: label value is a discovery timestamp, not a fixed "true". */}}
+{{- if or (and .Values.odiglet.affinity (not (empty .Values.odiglet.affinity))) (dig "scheduleOnlyOnInstrumentedNodes" "enabled" false (.Values.odiglet | default dict)) }}
+      affinity:
+{{- if and .Values.odiglet.affinity (not (empty .Values.odiglet.affinity)) }}
+{{ toYaml .Values.odiglet.affinity | nindent 8 }}
+{{- end }}
+{{- if dig "scheduleOnlyOnInstrumentedNodes" "enabled" false (.Values.odiglet | default dict) }}
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: odigos.io/first-instrumented-pod-at
+                    operator: Exists
+{{- end }}
 {{- end }}
       # Force UID 0 unless runAsNonRoot is opted in; otherwise image USER (non-zero) is used.
       securityContext:
--- a/helm/odigos/templates/odiglet/daemonset.yaml
+++ b/helm/odigos/templates/odiglet/daemonset.yaml
@@ -29,6 +29,9 @@
 {{- if and $networkMetricsEnabled .Values.odiglet.noHostNetwork }}
 {{ fail "odiglet.noHostNetwork cannot be true when metricsSources.networkMetrics.enabled is true, since OBI network metrics require hostNetwork" }}
 {{- end }}
+{{- if and .Values.odiglet.scheduleOnlyOnInstrumentedNodes (ne .Values.instrumentor.mountMethod "k8s-init-container") }}
+{{ fail "odiglet.scheduleOnlyOnInstrumentedNodes requires instrumentor.mountMethod=k8s-init-container" }}
+{{- end }}
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
@@ -60,14 +63,22 @@ spec:
         {{- end }}
         kubectl.kubernetes.io/default-container: odiglet
     spec:
+      {{- $nodeSelector := dict }}
       {{- if .Values.odiglet.nodeSelector }}
-      # Component-specific nodeSelector (overrides global values.yaml nodeSelector)
-      nodeSelector:
-      {{- toYaml .Values.odiglet.nodeSelector | nindent 8 }}
+      {{- range $k, $v := .Values.odiglet.nodeSelector }}
+      {{- $_ := set $nodeSelector $k $v }}
+      {{- end }}
       {{- else if .Values.nodeSelector }}
-      # Global nodeSelector from values.yaml
+      {{- range $k, $v := .Values.nodeSelector }}
+      {{- $_ := set $nodeSelector $k $v }}
+      {{- end }}
+      {{- end }}
+      {{- if .Values.odiglet.scheduleOnlyOnInstrumentedNodes }}
+      {{- $_ := set $nodeSelector "odigos.io/instrumented-pods" "true" }}
+      {{- end }}
+      {{- if $nodeSelector }}
       nodeSelector:
-      {{- toYaml .Values.nodeSelector | nindent 8 }}
+      {{- toYaml $nodeSelector | nindent 8 }}
       {{- end }}
       # we will always have at least one init container
       # if the mountMethod is k8s-init-container, we will pre-pull the agents image here as an init container
--- a/helm/odigos/values.yaml
+++ b/helm/odigos/values.yaml
@@ -1020,6 +1020,10 @@ odiglet:
     #   When true, odiglet runs only on nodes that already have instrumented pods.
     #   By default (false), odiglet runs on all nodes (subject to nodeSelector,
     #   affinity, and tolerations).
+    #   When enabled together with a non-default nodeSelector, affinity, or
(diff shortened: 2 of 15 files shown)


helm/odigos/values.yaml
```diff
@@ -1004,6 +989,17 @@ odiglet:
...
+  # @schema
+  # description: |-
+  #   When true, odiglet DaemonSet pods are only scheduled on nodes that already
+  #   have at least one instrumented pod. By default (false), odiglet runs on all
+  #   nodes (subject to nodeSelector, affinity, and tolerations).
+  #   Requires instrumentor.mountMethod to be set to k8s-init-container.
+  #   Also set as an env var on the instrumentor so toggling this value restarts
+  #   it and enables/disables the node-label controllers.
+  # @schema
+  scheduleOnlyOnInstrumentedNodes: false
```
RonFed: maybe we should have something like
```suggestion
  scheduleOnlyOnInstrumentedNodes:
      enabled: false
```

this will allow us more flexibility in the future if we want to add more configurations to this feature like rate limiting etc'

helm/odigos/templates/odiglet/daemonset.yaml
```diff
@@ -29,6 +29,9 @@
 {{- if and $networkMetricsEnabled .Values.odiglet.noHostNetwork }}
 {{ fail "odiglet.noHostNetwork cannot be true when metricsSources.networkMetrics.enabled is true, since OBI network metrics require hostNetwork" }}
 {{- end }}
+{{- if and .Values.odiglet.scheduleOnlyOnInstrumentedNodes (ne .Values.instrumentor.mountMethod "k8s-init-container") }}
+{{ fail "odiglet.scheduleOnlyOnInstrumentedNodes requires instrumentor.mountMethod=k8s-init-container" }}
```
RonFed: 👍

instrumentor/controllers/manager.go
```diff
@@ -169,7 +171,7 @@ func durationPointer(d time.Duration) *time.Duration {
 	return &d
 }
 
-func SetupWithManager(mgr manager.Manager, dp *distros.Provider, k8sVersion *version.Version) error {
+func SetupWithManager(ctx context.Context, mgr manager.Manager, dp *distros.Provider, k8sVersion *version.Version, scheduleOdigletOnlyOnInstrumentedNodes bool) error {
```
RonFed: nit: we can just read the env var here instead of passing another argument to the function

instrumentor/controllers/instrumentednodes/nodes_controller.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		return err
+	}
+
+	for i := range nodes.Items {
+		if err := r.Client.Patch(ctx, &nodes.Items[i], client.RawPatch(types.MergePatchType, patch)); err != nil {
+			return err
+		}
+	}
+	return nil
```
RonFed: since this is a runnable, should we wrap this in a retry-backoff loop?

instrumentor/controllers/instrumentednodes/nodes_controller.go
```diff
@@ -0,0 +1,149 @@
...
+	if node.Labels == nil {
+		node.Labels = map[string]string{}
+	}
+	if hasInstrumented {
+		node.Labels[k8sconsts.InstrumentedPodsNodeLabel] = k8sconsts.InstrumentedPodsNodeLabelValue
+	} else {
+		delete(node.Labels, k8sconsts.InstrumentedPodsNodeLabel)
+	}
+
+	return r.Patch(ctx, node, client.MergeFrom(original))
```
RonFed: nit: not sure if it is possible to do this with "strategic" patch (if that's what its called) - where we only pass the json patch of the operation we want to do and avoid the deep copy

instrumentor/controllers/instrumentednodes/manager.go
```diff
@@ -0,0 +1,79 @@
...
+	err = builder.
+		ControllerManagedBy(mgr).
+		Named("instrumentednodes-instrumentationconfig").
+		For(&odigosv1.InstrumentationConfig{}).
+		WithEventFilter(odigospredicate.ExistencePredicate{}).
+		Complete(&InstrumentationConfigReconciler{
+			Client: mgr.GetClient(),
+		})
+	if err != nil {
+		return err
```
RonFed, line 79: we don't need pods reconciler?

instrumentor/controllers/instrumentednodes/instrumentationconfig_controller.go
```diff
@@ -0,0 +1,108 @@
...
+	err = r.List(ctx, &pods, client.InNamespace(pw.Namespace), client.MatchingLabels(labelSelector.MatchLabels))
+	if err != nil {
+		return nil, err
+	}
+	for i := range pods.Items {
+		if pods.Items[i].Spec.NodeName != "" {
+			nodes[pods.Items[i].Spec.NodeName] = struct{}{}
+		}
+	}
+	return nodes, nil
```
RonFed, line 116: nit: put this in a default case in the switch case above for better readability

instrumentor/controllers/instrumentednodes/instrumentationconfig_controller.go
```diff
@@ -0,0 +1,108 @@
...
+			syncErrs = append(syncErrs, err)
+		}
+	}
+
+	logger.Info("synced instrumented pods node labels for instrumentation config",
+		"instrumentationConfig", req.NamespacedName, "nodes", len(nodeNames))
+	return utils.K8SUpdateErrorHandler(errors.Join(syncErrs...))
+}
+
+func (r *InstrumentationConfigReconciler) nodeNamesForWorkload(ctx context.Context, pw k8sconsts.PodWorkload) (map[string]struct{}, error) {
```
RonFed, line 58: not sure if we have a common helper we can create for this, since I'm pretty sure the rollback/rollout logic have very similar helpers for getting all the pods of a workload

instrumentor/controllers/instrumentednodes/manager.go
blumamir: pod trigger the node above:

```
		Watches(
			&corev1.Pod{},
			handler.EnqueueRequestsFromMapFunc(mapPodToNode),
			builder.WithPredicates(odigospredicate.ExistencePredicate{}),
		).
```

but I might change it to pod reconciler

instrumentor/controllers/manager.go
blumamir: I want to avoid side effects and be explicit about the inputs

instrumentor/controllers/instrumentednodes/instrumentationconfig_controller.go
blumamir: will be addressed in a refactor. I don't want to change current rollout code

helm/odigos/values.yaml
```diff
@@ -1004,6 +1004,35 @@ odiglet:
...
+  #   Cons: the first time an instrumented pod starts on a node there is a period
+  #   without telemetry until odiglet starts on that node; assumes additional RBAC
+  #   permissions to patch and update node labels.
+  # @schema
+  scheduleOnlyOnInstrumentedNodes:
+    # @schema
+    # description: |-
+    #   When true, odiglet runs only on nodes that already have instrumented pods.
+    #   By default (false), odiglet runs on all nodes (subject to nodeSelector,
+    #   affinity, and tolerations).
```
RonFed, line 1022: nit maybe add a sentence how this integrates with nodeSelector/affininty/tolerations when its enabled

instrumentor/controllers/instrumentednodes/sync.go
```diff
@@ -0,0 +1,151 @@
...
+			return requeueAfter, nil
+		}
+		// remove the label
+		err := patchFirstInstrumentedPodAtNodeLabel(ctx, c, node, nil)
+		if err != nil {
+			return 0, err
+		}
+		return 0, nil
+	}
+}
```
RonFed, line 107:
```suggestion
func syncFirstInstrumentedPodAtNodeLabel(ctx context.Context, c client.Client, node *corev1.Node, hasInstrumented bool, nodeLabelRetention time.Duration) (time.Duration, error) {
	labelValue, hasLabel := node.Labels[k8sconsts.FirstInstrumentedPodAtNodeLabel]

	// label already reflects the desired state
	if hasInstrumented == hasLabel {
		return 0, nil
	}

	// node just got its first instrumented pod: stamp it with the current time
	if hasInstrumented {
		now := time.Now().UTC().Format(k8sconsts.FirstInstrumentedPodAtNodeLabelTimeFormat)
		return 0, patchFirstInstrumentedPodAtNodeLabel(ctx, c, node, now)
	}

	// no instrumented pods left: remove the label, but only after the retention period
	if requeueAfter, delayRemoval := retentionRemaining(labelValue, nodeLabelRetention); delayRemoval {
		return requeueAfter, nil
	}
	return 0, patchFirstInstrumentedPodAtNodeLabel(ctx, c, node, nil)
}
```

instrumentor/controllers/instrumentednodes/pods_controller.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/odigos-io/odigos/k8sutils/pkg/utils"
+)
+
+const maxPodNodeTrackerSize = 10000
+
+// podNodeNamePredicate triggers on pod create/delete and when Spec.NodeName changes.
+type podNodeNamePredicate struct{}
+
+func (p podNodeNamePredicate) Create(e event.CreateEvent) bool {
+	return true
```
RonFed: only if node name is not empty

instrumentor/controllers/instrumentednodes/pods_controller.go
```diff
@@ -0,0 +1,190 @@
...
+func (p podNodeNamePredicate) Generic(e event.GenericEvent) bool {
+	return false
+}
+
+// podNodeTracker maps pod name+namespace to its node so deletes can still sync the node.
+// when a pod is deleted, we cannot "Get" it and discover the node it was running on,
+// thus we need to maintain our track to be able to sync the node when a pod is deleted
+type podNodeTracker struct {
+	mu       sync.Mutex
+	podNodes map[ctrl.Request]string
```
RonFed:
```suggestion
	nodeNames map[ctrl.Request]string
```

instrumentor/controllers/instrumentednodes/pods_controller.go
```diff
@@ -0,0 +1,190 @@
...
+		}
+		if ra > 0 && (requeueAfter == 0 || ra < requeueAfter) {
+			requeueAfter = ra
+		}
+	}
+
+	if requeueAfter > 0 {
+		return ctrl.Result{RequeueAfter: requeueAfter}, nil
+	}
+	return ctrl.Result{}, nil
```
RonFed:
```suggestion
     return ctrl.Result{RequeueAfter: requeueAfter}, nil
```

instrumentor/controllers/instrumentednodes/nodes_controller.go
```diff
@@ -0,0 +1,60 @@
...
+		return utils.K8SUpdateErrorHandler(err)
+	}
+	// Periodically re-sync so the label is cleared after instrumented pods are
+	// deleted. The pods controller cannot handle deletes (node name is unknown
+	// once the pod is gone), so this is the path that eventually removes the label.
+	// it means new pods are handled right away, but pods being deleted are handled at most 1 minute after they are deleted.
+	const periodicRequeue = time.Minute
+	if requeueAfter == 0 || requeueAfter > periodicRequeue {
+		requeueAfter = periodicRequeue
+	}
```
RonFed, line 58:
```suggestion
	requeueAfter = min(requeueAfter, periodicRequeue)
```

RonFed (approved):

PR #5759, opened 2026-09-06
RUN-1121 | feat: run odiglet and data-collection as non-privileged by default

RonFed:

The odiglet daemonset ran its containers as privileged by default, and there was no explicit flag controlling the data-collection container at all — its privilege level was implied by the collected signals. Running it non-privileged required `--set signals="{traces}"`, which coupled an unrelated setting to a security decision and made logs/metrics collection
impossible without privileged mode.

This PR:
- Runs the odiglet and data-collection containers **non-privileged by default**, using the capabilities, appArmorProfile and seLinuxOptions already defined per container.
- Replaces `odiglet.unPrivileged` with per-container `odiglet.odiglet.privileged` and `odiglet.dataCollection.privileged`, both defaulting to `false`. The old key sat at the pod level but only ever controlled the odiglet and init containers, which was misleading.
- Removes the signals-derived privilege decision, so data-collection can run non-privileged with metrics and logs enabled.
- Adds `odiglet.dataCollection.seLinuxOptions` (default `spc_t`), the SELinux domain the container already receives when running privileged, so dropping privileged does not change its domain on SELinux-enforcing nodes.
- Adds `SYSLOG` and `SYS_RESOURCE` to data-collection when `profiling.enabled` is true, only when not already present in the user-supplied capabilities list. The eBPF profiler reads `/proc/kallsyms` addresses (`CAP_SYSLOG`) and raises `RLIMIT_MEMLOCK` (`CAP_SYS_RESOURCE`).
- for OBI `NET_ADMIN` and `NET_RAW` are added to the default capability set of the odiglet container - these can be removed if OBI is not used.
- For community go instrumentation we add the `KILL` capability since the upstream go openteleemtry instrumentation use signal(0) to check for existance of a PID.

```release-note
feat: [BREAKING CHANGE] the odiglet and data-collection containers now run as non-privileged by default. To restore the previous behavior set `odiglet.odiglet.privileged=true` and `odiglet.dataCollection.privileged=true`.
feat: [BREAKING CHANGE] `odiglet.unPrivileged` is replaced by the per-container `odiglet.odiglet.privileged` and `odiglet.dataCollection.privileged` flags, to make clear they apply to individual containers in the odiglet daemonset and not to the whole pod.
feat: [BREAKING CHANGE] the `k8s-csi-driver` mount method now requires `odiglet.odiglet.privileged=true`.
feat: data-collection can now run non-privileged with metrics and logs enabled; it is no longer implied by `signals`.
```

### DIFF
--- a/helm/odigos/templates/odiglet/daemonset.yaml
+++ b/helm/odigos/templates/odiglet/daemonset.yaml
@@ -182,9 +182,14 @@ spec:
             privileged: false
             capabilities:
               add:
-              {{- with .Values.odiglet.odiglet.capabilities }}
-              {{- toYaml . | nindent 14 }}
+              {{- $caps := .Values.odiglet.odiglet.capabilities | default list }}
+              {{- if not (include "odigos.secretExists" .) }}
+              {{- /* the community go instrumentation checks that the target process is alive with
+                     kill(pid, 0), which needs CAP_KILL when the target runs as a different user.
+                     the enterprise distro stats /proc/<pid> instead and needs no capability. */}}
+              {{- if not (has "KILL" $caps) }}{{- $caps = append $caps "KILL" }}{{- end }}
               {{- end }}
+              {{- toYaml $caps | nindent 14 }}
               drop:
                 - ALL
             # for Kubernetes >= 1.30 we can specify AppArmor profile in securityContext
--- a/helm/odigos/templates/odiglet/daemonset.yaml
+++ b/helm/odigos/templates/odiglet/daemonset.yaml
@@ -21,11 +21,7 @@
     {{- end -}}
   {{- end -}}
 {{- end -}}
-{{- $privilegedRequired := or $collectLogs $collectMetrics -}}
 {{- $k8sVersion := include "utils.cleanKubeVersion" . }}
-{{- if and $privilegedRequired .Values.odiglet.unPrivileged }}
-{{ fail "privileged mode is required when collecting logs or metrics, this is currently not supported in unprivileged mode" }}
-{{- end }}
 {{- if and $networkMetricsEnabled .Values.odiglet.noHostNetwork }}
 {{ fail "odiglet.noHostNetwork cannot be true when metricsSources.networkMetrics.enabled is true, since OBI network metrics require hostNetwork" }}
 {{- end }}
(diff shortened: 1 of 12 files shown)


edeNFed (approved):

helm/odigos/values.yaml
```diff
@@ -1102,22 +1126,14 @@ odiglet:
   # @schema
   customContainerRuntimeSocketPath: ''
 
-  # @schema
-  # description: |-
-  #   run the odiglet pod without privileged containers.
-  #   when set to true, odiglet will use the set of capabilities required for eBPF operations.
-  #   by default, odiglet runs as a privileged container.
-  # @schema
-  unPrivileged: false
```
blumamir, line 1111: I wonder if it's really necessary to maintain this in the container level and not in the ds level. do you think users would want to set one and not the other?

helm/odigos/values.yaml
```diff
@@ -890,6 +890,12 @@ odiglet:
         port: 55683
       periodSeconds: 10
 
+    # @schema
+    # description: |-
+    #   Run the odiglet container, and its init container, as non-privileged containers.
+    # @schema
+    unPrivileged: false
```
blumamir, line 897: I guess if it's possible then all people would prefer to use it right, but then we should mention when it would not work and under which conditions it needs to be set?

helm/odigos/templates/odiglet/daemonset.yaml
```diff
@@ -99,7 +95,7 @@ spec:
           {{- end }}
           {{- end }}
           securityContext:
-          {{- if not .Values.odiglet.unPrivileged}}
+          {{- if not .Values.odiglet.odiglet.unPrivileged}}
```
blumamir, line 98: will it work for oss? should we check and fail the deployment if it's used where not supported?

helm/odigos/values.yaml
```diff
@@ -963,12 +969,30 @@ odiglet:
...
+
+    # @schema
+    # description: |-
+    #   seLinuxOptions for the data-collection container when running as unPrivileged.
+    #   "spc_t" is the SELinux domain the container already gets when running privileged.
+    # @schema
+    seLinuxOptions:
+      type: "spc_t"
```
blumamir, line 995: When someone will need to use it, how can he know what to write here in order to make it work?
Should we mention when to use it and how?

helm/odigos/values.yaml
```diff
@@ -963,12 +969,30 @@ odiglet:
...
     # @schema
     capabilities:
       - SYS_ADMIN
       - BPF
       - PERFMON
       - IPC_LOCK
+      # uncomment when profiling.enabled is true - the eBPF profiler reads
+      # /proc/kallsyms addresses and raises RLIMIT_MEMLOCK
+      # - SYSLOG
+      # - SYS_RESOURCE
```
blumamir, line 982: we can write this into the template, right? it will be easy to miss. if we choose to keep it, should we also add a comment next to profiling.enabled to make it more discoverable?

blumamir (commented):
added few nit comments

blumamir (approved):

edeNFed (approved):

blumamir (approved):

PR #5786, opened 2026-09-09
fix: create empty URLTemplatization config when action fields are omitted

alonkeyval:
Recommendation CreateAction with type URLTemplatization and empty fields left `spec.urlTemplatization` unset, so the admission webhook rejected the request.

When converting GraphQL Action input to a CR, InferDbAttributes/DbQueryTemplatization already key off `action.type` and create an empty config. URLTemplatization only looked at field groups and returned nil when both were omitted. Align the converter with that pattern so type URLTemplatization always produces `spec.urlTemplatization` (including `{}`), matching catalog YAML and fixing recommendation enable from the UI.

```release-note
Fix creating URL Templatization actions with empty fields (e.g. from Recommendations).
```

### DIFF
--- a/api/k8sconsts/odiglet.go
+++ b/api/k8sconsts/odiglet.go
@@ -35,7 +35,6 @@ const (
 
 	OdigletLocalTrafficServiceName = "odiglet-local"
 	OdigletMetricsServerPort       = 8080
-	OdigletWaspServicePort         = 4040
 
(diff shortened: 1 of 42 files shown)


alonkeyval:
block untill uikit upgrade

BenElferink (approved):

PR #5803, opened 2026-09-11
RUN-1126 | fix(openshift): run odiglet unprivileged by default

RonFed:

`openshift.enabled=true` with the unprivileged odiglet default is broken today (after #5759 ) — the API server rejects the DaemonSet, because mountPropagation: Bidirectional is only allowed on privileged containers.
The privileged existed so the init container could chroot("/host") and run the host's semanage + restorecon, making the agents readable by instrumented containers.
`ApplyOpenShiftSELinuxSettings` now sets the SELinux type on the agent files directly, with lsetxattr. No chroot, no host tooling. It runs before the copy, so rsync's output inherits the directory's type and only files predating the run need fixing.

The chart drops the /host and /etc/selinux mounts, the Bidirectional propagation and both volumes. The init container gains seLinuxOptions from the existing odiglet.odiglet.seLinuxOptions value, without it the container lands in container_t and cannot relabel anything.
Relabeling is gated on ODIGOS_OPENSHIFT_ENABLED to make sure we won't label the files on platform with SELinux that don't require this. Note that the SELinux flow in the init container was also gated to run only on openshift in the old code - but that was done implicitly - the mount was done in openshift, and that caused the exec to succeed there.

While testing, discovered we are missing RBAC for corn-jobs, jobs, argo-rollouts and static pods workloads - so these are updated in the chart under the openshift flag.

Verified on OpenShift 4.22.7 / RHCOS (CRC).

```release-note
fix(openshift): run odiglet unprivileged by default
fix(openshift): support cornjobs, jobs, argo rollouts and static pods as workloads for instrumentation
```


damemi (commented):
overall I like this approach a lot better than the manual chrooting. have some notes though (ai reviewer helped):

values.yaml documents `odiglet.odiglet.seLinuxOptions: null `as supported. With:

```
openshift.enabled=true
odiglet.odiglet.privileged=false
odiglet.odiglet.seLinuxOptions=null
```

the chart renders the init container without an SELinux domain but still sets `ODIGOS_OPENSHIFT_ENABLED=true`. The init then tries Lsetxattr from container_t, gets EACCES/EPERM, and exits before copying agents. We should either give the init container a dedicated required relabel-capable SELinux context independent of the main odiglet override, or fail Helm rendering with a clear validation error for this combination. Continuing after relabel failure would leave agents unreadable, so I would not silently degrade.

common/fs/agents.go:87-88 silently returns success for unexpected root labels; odiglet.go:304 then logs “Labeled agent files” with zero changes. Return a distinct result or warning containing the observed type.

Fresh installs no longer create a persistent semanage fcontext rule. A later host restorecon can revert /var/odigos to var_t until odiglet restarts. Existing upgraded nodes also retain the old unmanaged rule. This tradeoff should be documented and explicitly tested.

The last one is interesting because using the selinux exec directly did create these persistent rules, and if they got reverted somehow it would be a huge pain to figure out why things aren't working

RonFed:
> values.yaml documents odiglet.odiglet.seLinuxOptions: null as supported. With:
openshift.enabled=true
odiglet.odiglet.privileged=false
odiglet.odiglet.seLinuxOptions=null
the chart renders the init container without an SELinux domain but still sets ODIGOS_OPENSHIFT_ENABLED=true. The init then tries Lsetxattr from container_t, gets EACCES/EPERM, and exits before copying agents. We should either give the init container a dedicated required relabel-capable SELinux context independent of the main odiglet override, or fail Helm rendering with a clear validation error for this combination. Continuing after relabel failure would leave agents unreadable, so I would not silently degrade.

Good point, added a fail assertion in the chart for that.

> common/fs/agents.go:87-88 silently returns success for unexpected root labels; odiglet.go:304 then logs “Labeled agent files” with zero changes. Return a distinct result or warning containing the observed type.

Added to the logging the root dir context type.

> Fresh installs no longer create a persistent semanage fcontext rule. A later host restorecon can revert /var/odigos to var_t until odiglet restarts. Existing upgraded nodes also retain the old unmanaged rule. This tradeoff should be documented and explicitly tested.

The last one is interesting because using the selinux exec directly did create these persistent rules, and if they got reverted somehow it would be a huge pain to figure out why things aren't working

That's a good point and it is really a behavior change and a tradeoff. Basically, in theory while the odiglet is running, the host dir `/var/odigos` might get relabeled - and we don't have a watch/"reconciliation" for that - we only align it on odiglet init. I'm not sure if that is a realistic case to be happening - i.e having some tool/manual command change the SELinux in runtime - and if we should take the effort to defend against that - which adds permissions to the init container and I'm not sure if we can do non-privileged.

damemi (approved):

PR #5805, opened 2026-09-12
RUN-834, RUN-1063 | chore: stop wiring odigosebpf into metrics pipelines and drop the JVM metrics map

daemon1024:

```release-note
```


autoscaler/controllers/nodecollector/collectorconfig/metrics.go
```diff
@@ -10,6 +13,16 @@ const (
...
 	odigosMetricsPipelineName = "metrics"
+
+	// JVM runtime metrics from the eBPF Java agent arrive over OTLP on otlp/in,
+	// which joins the metrics pipeline only with agentsTelemetry. They are
+	// reported whatever that setting says, so without it they get a route of
+	// their own.
+	jvmRuntimeMetricsPipelineName = "metrics/jvm-runtime"
+	jvmRuntimeMetricsFilterName   = "filter/jvm-runtime"
+	// Must equal ebpf-java-instrumentation's jvmmetrics.ScopeName: renaming the
+	// scope there silently filters these metrics out.
```
Tomerkan, line 24: Can you confirm the producer actually stamps jvm-ebpf-metrics? It's not in ebpf-java-instrumentation (meter name there is jvm-metrics, and there's no jvmmetrics.ScopeName), and the real emitter is odigos-enterprise's odiglet — if it mismatches, this filter silently drops all JVM metrics.

daemon1024:
Confirmed, and it needed fixing. `jvmmetrics.ScopeName` now exists in ebpf-java-instrumentation — it landed with odigos-io/ebpf-java-instrumentation#589, which replaced the old OTel meter with a pdata pipeline. The scope was also renamed in review from `jvm-ebpf-metrics` to the producing Go package, `github.com/odigos-io/ebpf-java-instrumentation/jvmmetrics`, so this filter was about to drop every JVM metric exactly as you describe. Updated the constant and the test to match.

autoscaler/controllers/nodecollector/collectorconfig/metrics.go
daemon1024: Verified end to end, the scope name was updated after the PR was raised to include audios. This matches that now.

damemi (approved):
approving for time sensitive request

RonFed (changes requested):
Blocking for the dedicated JVM metrics pipeline which I don't understand.

RonFed (approved):

PR #5806, opened 2026-09-12
RUN-834, RUN-1063 | chore: stop wiring odigosebpf into metrics pipelines and drop the JVM metrics map

daemon1024:

```release-note
```


damemi (approved):
approving for time sensitive request

daemon1024:
Superseded: #5805 is being closed (its only content was the dedicated JVM metrics pipeline, which we are dropping), and GitHub will not retarget a PR that is part of a stack. Reopened against main from the same branch, same commits.

PR #5807, opened 2026-09-12
RUN-834, RUN-1063 | chore: remove the JVM metrics map from instrumentation settings

daemon1024:
Nothing writes the JVM metrics eBPF map now that the Java agent reports over OTLP, so the settings and manager options that carried it are removed.

```release-note
Removed the JVM metrics eBPF map from instrumentation Settings and manager options.
```


daemon1024:
Folded into #5806 — the commit now sits on that branch.

PR #5818, opened 2026-09-14
Devops 87 migrate oss into using docker bake definition

TheErikNaz:

Migrate OSS builds to use docker bake

```release-note
None
```


docker-bake.hcl
```diff
@@ -0,0 +1,321 @@
...
+variable "IMG_SUFFIX" {
+  default = ""
+}
+
+# Comma-separated platform list, e.g. "linux/amd64,linux/arm64". A single
+# entry produces a normal image; more than one requires --push (or a
+# containerd image store) since a multi-platform result can't be `--load`ed.
+variable "PLATFORMS" {
+  default = "linux/amd64"
+}
```
RonFed, line 55: why is this the default? is it keeping the same behavior we have today? for running locally on a mac, linux/arm64 is used

Makefile
```diff
@@ -20,30 +20,36 @@ LINT_CMD = $(GOLANGCI_LINT) run -c ../.golangci.yml
...
 RHEL?=false
-BUILD_DIR=.
-
-include cli.mk
-
+LD_FLAGS?=
+# Platform(s) passed to docker-bake.hcl. A single platform (the default) can be
+# --load'ed into the local docker engine; more than one (e.g. for multi-arch)
+# requires --push. push-% always builds for PUSH_PLATFORMS regardless of this.
+PLATFORMS?=linux/amd64
```
RonFed: why is the default defined both here and in the bake file? can we have one source of truth?

Makefile
```diff
@@ -20,30 +20,36 @@ LINT_CMD = $(GOLANGCI_LINT) run -c ../.golangci.yml
...
+PUSH_PLATFORMS?=linux/amd64,linux/arm64
```
RonFed, line 33: this should come from CI ideally explictly

docker-bake.hcl
TheErikNaz: Done, Default now is local arch

Makefile
TheErikNaz: I prefer it being hardened as both archs (I don't see a case where we build and push only a single arch) in the CI
If someone wants to build a single arch and push to ecr, or whatever, he can save some runtime

Makefile
TheErikNaz: Done :)

docker-bake.hcl
```diff
@@ -0,0 +1,324 @@
...
+#
+# Groups:
+#   images / images-rhel - the 7 components deployed to a dev/e2e cluster
+#                           (matches `make build-images` / `build-images-rhel`)
+#   default (default group), rhel - aliases for images / images-rhel
+#   oss / oss-rhel        - every OSS component, including operator and cli
+#   all                   - everything: oss + oss-rhel
+#
+# Usage:
+#   docker buildx bake                                   # images group, linux/amd64
```
RonFed, line 22:
```suggestion
#   docker buildx bake                                   # images group, native arch
```

Makefile
```diff
@@ -121,126 +129,56 @@ $(HELM_SCHEMA_BIN):
...
-	--build-arg SUMMARY="$(SUMMARY)" \
-	--build-arg DESCRIPTION="$(DESCRIPTION)" \
-	--build-arg LD_FLAGS="$(LD_FLAGS)" \
-	--build-arg RHEL="$(RHEL)"
+# Build a single OSS component image via docker-bake.hcl (Dockerfile, target
+# stage, build args, tags, summary/description all live there).
+# Works for any bake target: autoscaler, scheduler, instrumentor, odiglet,
+# agents, collector, ui, operator, cli. RHEL=true builds that component's
+# -rhel variant instead (see docker-bake.hcl for what that changes).
+# IMAGE=registry/name:tag overrides the default tag (used by deploy-%).
```
RonFed, line 137: nit: think the comment can be trimmed and the components list can get out dated here - we have enough comments in the bake file IMO

RonFed (approved):

PR #5826, opened 2026-09-15
CORE-1667: Add Instrumentor finalizer

damemi:
If the instrumentor is accidentally deleted, or deleted before sources, the source objects can hang forever due to their finalizer and prevent uninstall from completing. This adds a finalizer to the Instrumentor itself which is cleaned up by proper uninstall methods to remove it to prevent instrumentor deletion.

```release-note
fix(instrumentor): Add finalizer to instrumentor pods to prevent hung uninstalls on sources
```


RonFed:
Not sure what the goal here is.
If a user is running
`kubectl delete ns odigos-system` - This might leave the odigos components in an "un-defined" state, more specifically - I think it might delete the CRDs and leave the CRs (e.g the Sources object instances) - if the Source CRD definition is deleted, I think even if the Instrumentor is kept, we will be stuck?

cli/cmd/uninstall-dep.go
```diff
@@ -287,6 +292,100 @@ func uninstallDeployments(ctx context.Context, client *kube.Client, ns, _ string
...
+			newFinalizers = append(newFinalizers, f)
+		}
+		if !updated {
+			continue
+		}
+		dep.Finalizers = newFinalizers
+		_, err = client.AppsV1().Deployments(ns).Update(ctx, dep, metav1.UpdateOptions{})
+		if err != nil {
+			return err
+		}
```
blumamir, line 326: update can return conflict here. will we retry it? what will happen to the uninstall?

cli/cmd/uninstall-dep.go
```diff
@@ -287,6 +292,100 @@ func uninstallDeployments(ctx context.Context, client *kube.Client, ns, _ string
...
+			if f == k8sconsts.InstrumentorDeploymentFinalizer {
+				updated = true
+				continue
+			}
+			newFinalizers = append(newFinalizers, f)
+		}
+		if !updated {
+			continue
+		}
+		dep.Finalizers = newFinalizers
```
blumamir, line 322: nit: this code is 3 times in this file, consider a util

helm/odigos/templates/instrumentor/deployment.yaml
```diff
@@ -6,6 +6,8 @@ metadata:
     odigos.io/system-object: "true"
   name: {{ .Values.instrumentor.deploymentName | default "odigos-instrumentor" }}
   namespace: {{ .Release.Namespace }}
+  finalizers:
+    - odigos.io/instrumentor-finalizer
```
blumamir, line 10: if the instrumentor is "deleted" for any reason, and then odigos is re-deployed with helm, it will still have the `deletionTimestamp` set which is incorrect and can cause controllers to treat it as deleting.

we have this bug in sources as well which can manifest in some very hard to track sequences.

blumamir (commented):
I think the finalizer is a powerful tool, but also one that can create some very hard-to-find and annoying bugs, dead-locks, incomplete states etc. these were all observed with the finalizers for sources.

could be helpful to map and track all the places where errors can occur, and make sure we can recover from them.
also, there are timeouts and flags to uninstall which might affect the flow of the code.

PR #5835, opened 2026-09-16
CORE-1670: Unify RHEL publish into a wrapper job

damemi:
This changes how we publish rhel images. The old way was a total recompile of the checked out branch for each image, all to just drop it in a ubi9 base image (with license generation). This unifies and simplifies it: Copy from the existing standard image into a ubi9 dockerfile. It also adds a "license image" that can be built once with the rhel publish flow and copied from to satisfy the license requirements.

Removes the Enterprise flow: https://github.com/odigos-io/odigos-enterprise/pull/3759

```release-note
NONE
```


rhel/Dockerfile.wrap
```diff
@@ -0,0 +1,115 @@
...
+FROM base AS wrap-cli
+WORKDIR /
+COPY --from=source /ko-app /ko-app
+# ko names the binary after the main package directory ("cli")
+ENTRYPOINT ["/ko-app/cli"]
+
+# Central UI is a Node app on node:alpine. Rebase onto the published image so
+# node/yarn stay available (they are musl-linked and cannot run on ubi-micro),
+# then overlay Red Hat labels and the shared licenses tree.
+FROM source AS wrap-central-ui
```
RonFed, line 98: should this be in this repo?

rhel/Dockerfile.wrap
```diff
@@ -0,0 +1,115 @@
+# Wrap an already-published Odigos component image onto ubi-micro with Red Hat
+# labels and the shared odigos-licenses image. No recompilation.
```
RonFed, line 2:
```suggestion
# labels and the shared odigos-licenses image.
```

rhel/Dockerfile.wrap
```diff
@@ -0,0 +1,115 @@
...
+# Target (one of): wrap-app, wrap-ui, wrap-collector, wrap-enterprise-collector,
+#   wrap-odiglet, wrap-agents, wrap-operator, wrap-cli, wrap-central-ui
+
+ARG SOURCE_IMAGE
+ARG LICENSES_IMAGE
+
+FROM ${SOURCE_IMAGE} AS source
+FROM ${LICENSES_IMAGE} AS licenses
+
+FROM registry.access.redhat.com/ubi9/ubi-micro:latest AS base
```
RonFed, line 18: I saw we use `latest` also today, but I think we should prefer using a pinned version to avoid "surprises" in latest, wdyt?

rhel/Dockerfile.wrap
RonFed, line 1: I think it might be more clear to place each component wrapper Dockerfile next to its original one to make this clearer and more consistent with what we have today.

rhel/Dockerfile.wrap
```diff
@@ -0,0 +1,115 @@
...
+WORKDIR /instrumentations/
+CMD ["/root/odiglet"]
+
+# Agents init container: published image is busybox with sh/cp for
+# `sh -c 'cp …'` chaining. Copy those tools onto ubi-micro and keep uid 1000.
+FROM base AS wrap-agents
+COPY --from=source /bin/busybox /bin/busybox
+COPY --from=source /bin/sh /bin/sh
+COPY --from=source /bin/cp /bin/cp
+COPY --from=source /instrumentations/ /instrumentations/
```
RonFed, line 78: the original rhel flavor of the agents image didn't copy these - does that mean using the init container mount method with the rhel image would fail one "cp not found"

.github/workflows/publish-modules-rhel.yml
```diff
@@ -1,298 +1,254 @@
...
+          secrets: |
+            github_token=${{ secrets.RELEASE_BOT_TOKEN }}
+          tags: |
+            ${{ env.AR_REGISTRY }}/${{ env.LICENSES_IMAGE_NAME }}:${{ inputs.tag }}
+            keyval/${{ env.LICENSES_IMAGE_NAME }}:${{ inputs.tag }}
+            ghcr.io/odigos-io/${{ env.LICENSES_IMAGE_NAME }}:${{ inputs.tag }}
+            quay.io/odigos/${{ env.LICENSES_IMAGE_NAME }}:${{ inputs.tag }}
+
+  publish-images-rhel:
+    needs: publish-licenses
```
RonFed, line 101: we need the regular images publish as a dependency for this, right? I think currently we don't have this wired in this workflow? if I understand correctly this is triggered in the same tag as the regular images publish via a tag push?

.github/workflows/publish-modules-rhel.yml
```diff
@@ -1,298 +1,254 @@
...
-              client-payload: '{"tag": "${{ inputs.tag }}"}'
+  publish-licenses:
+    runs-on: depot-ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+      packages: write
+    steps:
+      - name: Checkout OSS repository
+        uses: actions/checkout@v5
```
blumamir, line 36: use latest version

.github/workflows/publish-modules-rhel.yml
```diff
@@ -1,298 +1,254 @@
...
+
+      - name: Login to Artifact Registry
+        uses: docker/login-action@v3
+        with:
+          registry: us-central1-docker.pkg.dev
+          username: oauth2accesstoken
+          password: ${{ steps.gcp-auth.outputs.access_token }}
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
```
blumamir, line 57:
```suggestion
        uses: docker/login-action@v4
```

.github/workflows/publish-modules-rhel.yml
```diff
@@ -1,298 +1,254 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
```
blumamir, line 67: I think we can avoid the ghcr, unless something uses it?

.github/workflows/publish-modules-rhel.yml
```diff
@@ -1,298 +1,254 @@
...
+          password: ${{ secrets.QUAY_TOKEN }}
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v4
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+
+      - name: Build and push odigos-licenses image
+        uses: docker/build-push-action@v6
```
blumamir, line 83:
```suggestion
        uses: docker/build-push-action@v7
```

rhel/Dockerfile.licenses
```diff
@@ -0,0 +1,53 @@
...
+    if [ "${ENTERPRISE_REF}" = "skip" ]; then \
+      echo "Skipping enterprise clone (ENTERPRISE_REF=skip)"; \
+      mkdir -p /src/enterprise; \
+    else \
+      TOKEN="$(cat /run/secrets/github_token)"; \
+      test -n "${TOKEN}" || { echo "github_token secret required unless ENTERPRISE_REF=skip" >&2; exit 1; }; \
+      git clone --depth 1 --branch "${ENTERPRISE_REF}" \
+        "https://x-access-token:${TOKEN}@github.com/odigos-io/odigos-enterprise.git" \
+        /src/enterprise; \
+    fi
```
blumamir, line 37: I think it's unhealthy to have this code here.
should we have something similar in enterprise to keep good sepration?

rhel/Dockerfile.wrap
blumamir: I support pinning

rhel/Dockerfile.wrap
```diff
@@ -0,0 +1,115 @@
...
+#   NAME / VERSION / RELEASE / SUMMARY / DESCRIPTION - Red Hat image labels
+#
+# Target (one of): wrap-app, wrap-ui, wrap-collector, wrap-enterprise-collector,
+#   wrap-odiglet, wrap-agents, wrap-operator, wrap-cli, wrap-central-ui
+
+ARG SOURCE_IMAGE
+ARG LICENSES_IMAGE
+
+FROM ${SOURCE_IMAGE} AS source
+FROM ${LICENSES_IMAGE} AS licenses
```
blumamir, line 16: this is more appropriate as build context and not build args

rhel/Dockerfile.wrap
```diff
@@ -0,0 +1,115 @@
...
+CMD ["/odigosotelcol"]
+
+# Enterprise collector (extra supervisor + managed config; dirs must be writable
+# by the non-root user when a volume is mounted on the supervisor path)
+FROM base AS wrap-enterprise-collector
+COPY --from=source /odigosotelcol /odigosotelcol
+COPY --from=source /odigosopampsupervisor /odigosopampsupervisor
+COPY --from=source --chown=65532:65532 /etc/odigos/managed-collector/ /etc/odigos/managed-collector/
+COPY --from=source --chown=65532:65532 /var/lib/odigos/collector-supervisor/ /var/lib/odigos/collector-supervisor/
+CMD ["/odigosotelcol"]
```
blumamir, line 59: I think it's bad practice to do this enterprise build in oss

rhel/Dockerfile.wrap
blumamir: I support

rhel/Dockerfile.wrap
damemi: red hat scans the version for "freshness" so pinning makes it fall from A->F, and since we don't release rhel that often that itself is a surprise when it happens. we're not going to find anything on our own before we push a release anyway

.github/workflows/publish-modules-rhel.yml
damemi: no, the rhel release is a manual trigger. it's intentionally not done for every release

rhel/Dockerfile.wrap
damemi: I want to put the rhel publish in a single place, whether that's all in OSS or all in enterprise, I don't like this current repo split between 2 jobs. 

same for comments in https://github.com/odigos-io/odigos/pull/5835#discussion_r4034103547
https://github.com/odigos-io/odigos/pull/5835#discussion_r4034555990
https://github.com/odigos-io/odigos/pull/5835#discussion_r4034672085

rhel/Dockerfile.wrap
damemi: trying to simplify rhel into a single repo, see https://github.com/odigos-io/odigos/pull/5835#discussion_r4036846253

rhel/Dockerfile.licenses
damemi: see https://github.com/odigos-io/odigos/pull/5835#discussion_r4036846253, don't want to split this across 2 action triggers in different repos

rhel/Dockerfile.wrap
damemi: not actually a build, it's a copy from the already published image. see https://github.com/odigos-io/odigos/pull/5835#discussion_r4036846253

if we don't want to do the enterprise copy in oss, I'll move everything in this PR to the enterprise repo (including oss copies). but for what we use rhel for and how this approach is working now there's no reason to have 2 separate jobs

rhel/Dockerfile.wrap
RonFed: I think having all the different components in a single Dockerfile is not clear and prone to get out of date when one of the components needs a change in the wrapper Dockerfile. 
Consider a new component being added, I think the split version is much more clear on the fact that a rhel wrapper needs to be added. 

From a release/build point, it a matter of replacing the target flag with the file flag. 

I think from a maintenance point, it's easier to understand if it's split.

damemi: I do agree with the single dockerfile being messy... I'm not against doing separate dockerfiles (actually like that) but what I really want is for this release process to be as minimal as possible with the smallest footprint and least impact on the other releases and people's workflows. I will think about this a little more no rush on this right now

PR #5841, opened 2026-09-17
Fix: RuntimeDetailsChangedPredicate compared containers by list position, not identity

kushalShukla-web:
### Commit message

fix: match containers by name instead of position in runtime details predicate

### PR description

**Problem**

`RuntimeDetailsChangedPredicate.Update()` compared `old.Status.RuntimeDetailsByContainer[i]` to `new[i]` by index. This assumed the list stays in the same order across updates. Since runtime details are currently only ever set once, this held by luck — but the code's own TODO flagged it as unsafe once runtime details can be re-detected/updated more than once, since nothing guarantees list order stays stable.

Concretely: if the list order changes with no real per-container change, the predicate can report a false change (unnecessary reconcile + rollout). Conversely, comparing the wrong pair by position can also mask a real change.

**Fix**

In `runtime_details_changed.go`, `Update()` now indexes the old details by `ContainerName` into a map and looks up each new container by name, so comparison is order-independent and tied to the actual container identity.

```


kushalShukla-web:
@damemi  let me know if any changes required here , happy to fix them :)

PR #5851, opened 2026-09-19
CORE-1675 | feat: replace Source finalizers with IC owner references

blumamir:
Source finalizers block deletion until InstrumentationConfigs are cleaned up, which complicates Odigos uninstall (Sources can get stuck terminating after the instrumentor is gone).

This change stops adding Source finalizers and instead sets non-controller owner references from enabling Source(s) onto InstrumentationConfigs. When no enabled Source still marks a workload for instrumentation, ICs are cleaned up via reconcile (including on owner-ref changes), without finalizer orchestration.

Also explicitly deletes the IC when the workload is NotFound, since multi-owner refs can prevent GC from removing it when only the workload is deleted.

Linear: https://linear.app/odigos/issue/CORE-1675/replace-source-finalizers-with-owner-references-for

```release-note
NONE
```


instrumentor/controllers/sourceinstrumentation/common.go
```diff
@@ -511,3 +519,41 @@ func updateRecoveredFromRollbackAt(ic *odigosv1.InstrumentationConfig, sourceRol
...
+func activeEnablingSources(sources *odigosv1.WorkloadSources) map[types.UID]*odigosv1.Source {
+	result := make(map[types.UID]*odigosv1.Source, 2)
+	if sources.Workload != nil && !odigosv1.IsDisabledSource(sources.Workload) {
+		result[sources.Workload.UID] = sources.Workload
+	}
+	if sources.Namespace != nil && !odigosv1.IsDisabledSource(sources.Namespace) {
+		result[sources.Namespace.UID] = sources.Namespace
+	}
+	return result
+}
```
damemi, line 532: what happens when you delete a disabled source? Currently, if you have an active namespace source + an active exclusion workload source, and you delete the exclusion (disabled) source, that should trigger a re-instrumentation of that workload. but since disabled sources aren't included I think it will be a no op

instrumentor/controllers/sourceinstrumentation/common.go
```diff
@@ -272,10 +273,11 @@ func syncWorkload(ctx context.Context, k8sClient client.Client, scheme *runtime.
...
 	if err != nil {
-		// if err is not nil it means obj is invalid, so we must return.
-		// instrumentation config has the workload as owner, so it will be deleted automatically by k8s,
-		// thus NotFound is expected and we can return without error.
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		if apierrors.IsNotFound(err) {
+			// Workload is gone; delete any InstrumentationConfig explicitly.
+			return ctrl.Result{}, deleteWorkloadInstrumentationConfig(ctx, k8sClient, pw)
+		}
+		return ctrl.Result{}, err
```
damemi, line 280: the instrumentationconfig controller (what we're relying on for syncworkload here) only calls syncworkload when the IC is already deleted (see https://github.com/odigos-io/odigos/blob/416fa3d03203da0a130ecd1dd0c64dfb6768ae11/instrumentor/controllers/sourceinstrumentation/instrumentationconfig_controller.go#L44-L58)

so in this flow:
1. Deployment deleted
2. GC removes deployment from ownerref
3. Ownerref predicate fires
4. but that controller only syncs on `if IC !found`

So the IC won't get cleaned up with the current approach

damemi (commented):
found 2 critical path notes that we can probably update to work around. but overall I'm still unsure if this tradeoff is worth the change, or if we're jumping out of the pan into the fire

Finalizers on sources give us problems for sure, but they're at least idiomatic problems. This isn't an idiomatic approach to ownerrefs, and while I do like the core idea of leaning on ownerrefs instead of finalizers (ownerrefs tend to fail safer and we shift a lot of responsibility from our controllers to native k8s GC), we are also doing additional handling that's outside the intent for ownerrefs and I think could give us some new unanticipated problems

that said I do think this is progress toward a better architecture, but perhaps not something that's critical to rush into right now if we don't need to. but if it's not too much pain to fix these 2 comments I'm still open to seeing how that looks

instrumentor/controllers/sourceinstrumentation/common.go
blumamir: good catch, fixed by always reconciling, and also adding delete events on the workload object to the events filter

PR #5859, opened 2026-09-22
RUN-834, RUN-1063 | chore: stop wiring odigosebpf into metrics pipelines

daemon1024:

```release-note
```


daemon1024:
Superseded by #5805, which now carries these commits.

